PR #24, opened 2016-05-03
Fix wrong rule name in recommended.js

boris-petrov:
lint-self-closing-void-elements is actually called self-closing-void-elements

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -6,7 +6,7 @@ module.exports = {
     'block-indentation': 2,
     'html-comments': true,
     'nested-interactive': true,
-    'lint-self-closing-void-elements': true,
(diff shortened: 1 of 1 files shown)


rwjblue:
Awesome, thank you for catching this!!!

rwjblue:
I will add some tests for this type of thing, sorry it bit you.

rwjblue:
First step here is done in https://github.com/rwjblue/ember-template-lint/pull/25.  I still need to add an explicit test for the `recommended` config that errors loudly if I make this mistake again.

rwjblue:
OK, now fully tested in https://github.com/rwjblue/ember-template-lint/pull/26.  We can not add invalid rule names, or rule configuration to default configs any more.

Thanks again for reporting and fixing!

PR #27, opened 2016-05-03
Remove `bare-strings` from recommended configuration.

rwjblue:
I believe that the majority of Ember applications should deal with localization properly, and this was why I originally configured `bare-strings` as a default. However, @mmun suggested that this is very much a concern that should be controlled outside of ember-template-lint (likely via the presence of an addon like `ember-i18n` or `ember-intl`).

I agree with @mmun, and would like to work with ember-intl/ember-i18n to get this to "just work" when those addons are present, but in the default case we should not include the `bare-strings` rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,17 @@ You can turn on specific rules by toggling them in a
 
 ```javascript
 module.exports = {
-  extends: 'recommended'
+  extends: 'recommended',
+
+  rules: {
+    'bare-strings': true
+  }
 }
 ```
 
-This extends from the builtin recommended configuration ([lib/config/recommended.js](https://github.com/rwjblue/ember-template-lint/blob/master/lib/config/recommended.js)).
+This extends from the builtin recommended configuration ([lib/config/recommended.js](https://github.com/rwjblue/ember-template-lint/blob/master/lib/config/recommended.js)),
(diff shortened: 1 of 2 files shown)


jamesarosen:
I'd be quite happy to change ember-i18n to declare _add this JSCS setting when installing_ if there's an easy hook to do so.

rwjblue:
@jamesarosen - I think I'm looking for some flag on the addon itself to indicate that it is intended as a "translations library" or whatever the right term is.  At that point, I could detect if a library is present, and enable the "no bare strings" rule by default...

rwjblue:
Something like:

```
// index.js of ember-i18n
module.exports = {
  name: 'ember-i18n',
  
  isLocalizationFramework: true
};
```

Or something like that...

rwjblue:
As long as the major players in the space provide the same flag, I can use it...

jasonmit:
I am onboard or can write to the config to flip bare strings on in the afterInstall hook

mmun:
I like that!

jamesarosen:
I hereby solemnly swear to support @rwjblue's `isLocalizationFramework` API. Or really any API he suggests in earnest. That guy's sharp.

PR #42, opened 2016-05-12
rule Quotes (WIP)

oriSomething:
[the original PR from ember-cli-template-lint](https://github.com/rwjblue/ember-cli-template-lint/pull/88)
#### config:
- `false` - disable
- `true` - only require qoutes for HTML attributes 
- `"single"` - always single quotes
- `"double"` - always double quotes
- `["single", "avoidEscape"]` - prefer single quotes unless there is an escape needed only to single quotes
- `["double", "avoidEscape"]` - prefer double quotes unless there is an escape needed only to double quotes
#### todos:
- [x] adding support for HTML attributes (TextNode)
- ~~adding support for HTMLBars helper attributes~~
- [x] attribute MustacheStatement
- [x] attribute ConcatStatement
- [x] MustacheStatement not inside attribute
- ~~ConcatStatement not inside attribute~~
- [x] adding support for only requiring qoutes
- [ ] adding README / and to blueprints
- [ ] adding avoidEscape supports
- [ ] adding avoidEscape to README

if i miss something, please tell me

### DIFF
--- a/lib/rules/lint-quotes.js
+++ b/lib/rules/lint-quotes.js
@@ -13,7 +13,7 @@
 
  The following values are valid configuration:
 
-   * boolean -- `false` for disabled. `true` to enforce any qoutes
+   * boolean -- `false` for disabled. `true` to enforce any quotes
    * string -- `single` of `double` for enforce style
  */
 
@@ -60,51 +60,8 @@ module.exports = function(addonContext) {
     );
   };
 
-  LogQuotes.prototype._processAttributeTextNode = function(attribute) {
-    if (attribute.value.chars.length === 0) {
-      return;
-    }
-
-    var source = this.sourceForNode(attribute);
-    var qouteCharacter = source[source.length - 1];
-
-    if (!this.config.shouldTestQuoteType) {
-      if (qouteCharacter === '"' || qouteCharacter === '\'') {
-        return;
-      } else {
-        this.log({
-          message: 'Quotes: you should use quotes for attributes',
-          line: attribute.loc.start.line,
-          column: attribute.loc.start.column,
-          source: source
-        });
-        return;
-      }
-    }
-
-    if (this.config.quoteCharacter === qouteCharacter) {
-      return;
-    }
-
-    var undesiredQouteStyle = (
-      this.config.quotesType === SINGLE_QUOTES_NAME ?
-        DOUBLE_QUOTES_NAME :
-        SINGLE_QUOTES_NAME
-    );
-
-    this.log({
-      message: 'Quotes: you got ' + undesiredQouteStyle +
-        ' quotes when you set quotes style to ' + this.config.quotesType +
-        ' quotes',
-      line: attribute.loc.start.line,
-      column: attribute.loc.start.column,
-      source: source
-    });
-    return;
-  };
-
-  LogQuotes.prototype._processNode = function(node) {
-    var source = this.sourceForNode(node);
+  LogQuotes.prototype._processNode = function(node, source) {
+    source = _.isString(source) ? source : this.sourceForNode(node);
     var qouteCharacter = source[source.length - 1];
 
     if (!this.config.shouldTestQuoteType) {
@@ -161,12 +118,16 @@ module.exports = function(addonContext) {
   LogQuotes.prototype._processAttribute = function(attribute) {
     switch(attribute.value.type) {
     case 'TextNode':
-      return this._processAttributeTextNode(attribute);
+      if (attribute.value.chars.length > 0) {
+        this._processNode(attribute, this.sourceForNode(attribute));
+      }
+      break;
     case 'MustacheStatement':
-      return this._processAttributeMustacheStatement(attribute.value);
-    // case 'ConcatStatement':
-    default:
-      return null;
+      this._processAttributeMustacheStatement(attribute.value);
+      break;
+    case 'ConcatStatement':
+      /** @todo */
+      break;
     }
   };
 
--- a/lib/rules/lint-quotes.js
+++ b/lib/rules/lint-quotes.js
@@ -53,7 +53,7 @@ module.exports = function(addonContext) {
     throw new Error(errorMessage);
   };
 
-  LogQuotes.prototype._processStringLiteralNode = function(node, source) {
+  LogQuotes.prototype._processStringLiteral = function(node, source) {
     source = _.isString(source) ? source : this.sourceForNode(node);
     var qouteCharacter = source[source.length - 1];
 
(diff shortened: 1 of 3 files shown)


oriSomething:
@rwjblue hey, currently `TextNode` attributes have no `loc` property which make this rule making much more complicated. is there a change it can be added to `HTMLBars`?

rwjblue:
TextNode's in attributes (there values) have no loc (I will look into fixing), but the AttrNode's do (and should contain the quotes used for that specific attribute). The bare-strings rule has an example of the AttrNode processing.

Also, note I am going to try and land the traverse API change that is open in the other PR, I believe this will help here...

oriSomething:
this change is more than welcome, since API ESLint was prove itself to me a few times in the past

rwjblue:
Ok, https://github.com/rwjblue/ember-template-lint/pull/39 has landed so this will need some tweaks to work there. 

I think it would be easiest to use the 'AttrNode' visitor, then you can remove some of the custom iteration you were forced to do.

oriSomething:
i have some issue with `ConcatStatement` because to check it i feel i make a little hack. do you know a better way to check it. because `ConcatStatement` has only `pairs` and `loc`, not something that i can get the quotes in a non hacky way.

[the commit](https://github.com/rwjblue/ember-template-lint/pull/42/commits/054cc3786dd636bca15fb3b7bf37501371af4a24#diff-c56719b7c6de7e1aa67db5f152052e9bR126)

rwjblue:
As of https://github.com/tildeio/htmlbars/pull/464, we should now be getting proper `loc` info for `node.value` in `AttrNode` visitor.

I am hoping that this unlocks what you are looking for:

```
SomeThing.prototype.visitors = function() {
  return {
    AttrNode: function(node) {
      var valueSource = this.sourceForNode(node.value);
      var valueFirstChar = valueSource.slice(0,1);
      var doubleQuoted = valueFirstChar === '"';
      var singleQuoted = valueFirstChar === "'";
      var quoteUsed = doubleQuoted || singleQuoted;
    }
  };
};
```

rwjblue:
That update to HTMLBars broke some of the bare-strings tests (because we were assuming that attribute value text nodes would be ignored since they had no loc info).  I fixed that on master, and locked down HTMLBars version (we can rely on greenkeeper.io to keep us up to date).

lib/rules/lint-quotes.js
```diff
@@ -0,0 +1,148 @@
...
+ <img alt='tomster' src="tomster.jpg">
+ ```
+
+ The following values are valid configuration:
+
+   * boolean -- `false` for disabled. `true` to enforce any quotes
+   * string -- `single` of `double` for enforce style
+ */
+
+var _ = require('lodash');
```
rwjblue: I know it is more verbose, but I prefer separate vars for each util method. Would you mind doing that too?

lib/rules/lint-quotes.js
```diff
@@ -0,0 +1,148 @@
...
+      case 'StringLiteral':
+        this._processStringLiteral(node);
+        break;
+      case 'SubExpression':
+        this._processMustacheStatement(node);
+        break;
+      }
+    }, this);
+
+    _([])
```
rwjblue: I don't think that `_` is needed here, these are normal array methods...

lib/rules/lint-quotes.js
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  LogQuotes.prototype._processConcatStatement = function(node) {
+    // NOTE: check if there is no more reasonable solution
+    this._processStringLiteral(node, this.sourceForNode(node));
+  };
+
+  LogQuotes.prototype._processAttribute = function(node) {
+    switch(node.value.type) {
+    case 'TextNode':
```
rwjblue: Can these be moved into the `visitors` instead?  You can see how I updated the `bare-strings` to track `this._currentAttrNode` (via `enter` / `exit` visitors) and then in `TextNode` visitor you can check for `this._currentAttrNode`.

test/unit/plugins/lint-quotes-test.js
```diff
@@ -0,0 +1,129 @@
...
+      template: '<img alt="tomster">',
+      result: {
+        message: 'Quotes: you got double quotes when you set quotes style to be single quotes',
+        source: 'alt="tomster"',
+        line: 1,
+        column: 5
+      }
+    }, {
+      config: 'single',
+      template: '<img alt={{url "tomster"}}>',
```
rwjblue: This is fine (and is required for things like `<div style={{someSafeString}}></div>`).  We should disambiguate between `ConcatStatement` (which is in quotes) and `MustacheStatement` (which is not in quotes).

rwjblue: Whoops! Sorry, I misread the message here!  You are testing that the quotes inside the mustache are correct.

lib/rules/lint-quotes.js
oriSomething: [it's possible](https://github.com/oriSomething/lodash-modules-sublime) ;-)
but it means there is no `chain` and lazy evaluation of streams

rwjblue:
This is looking really good! I left some inline comments / suggestions.

I want to make sure we are on the same page with `avoidEscape`. Basically, given `['double', 'avoidEscape'], I would like the the following good/bad:

```hbs
{{! good }}
<div data-foo="blah blah"></div>
<div data-foo="{{some-helper}}"></div>
<div data-foo="{{some-helper 'string args inside use single quotes'}}"></div>

{{! bad }}
<div data-foo='blah blah'></div>
<div data-foo='{{some-helper-in-single-quotes}}'></div>
<div data-foo="{{some-helper "string args inside double quotes cannot also use double quotes"}}"></div>
```

I want ^^ configuration to be the default when a config value of `true` is used (I think this is the right default for ember-cli apps).

rwjblue:
I'm happy to help push this over the finish line (and stop you from having to do so much extra work / rebases / etc). Let me know how I can help...

lib/rules/lint-quotes.js
oriSomething: it's a partly question of style guide. if you prefer to prefer always native array functions so be it. but since there is lodash anyway i don't mind to bring the performance gains it brings. since native array functions are still slower.

just tell me what are the project prefernces

oriSomething: @rwjblue can you please answer?

rwjblue: Ya, seems ok to me, I just didn't realize you were using it really.

oriSomething:
@rwjblue since i barely have the time, and i fill like i the one who responsible for the delay in this rule, tell me if you want to continue or you fine with the slowness of my coding in this PR

derekdowling:
Hey @oriSomething are you still wanting to complete this? I'd love to see feature make it in.

oriSomething:
@derekdowling As I've told before, I barely have time to complete this feature currently (and I guess for at least two months more) 😕. If you want to continue from here, you are invited. Just let everyone know

PR #49, opened 2016-05-15
Add isInteractiveElement helper function.

rwjblue:
Beginning part of #49.


rwjblue:
Working on migrating the `nested-interactive` rule to use this. Will PR after this lands...

PR #50, opened 2016-05-15
Update nested-interactive rule to use `isInteractiveElement` helper.

rwjblue:
Removes support for `nested-interactive` config to be an array. The
underlying functionality will be brought back before releasing.

I intend to change the configuration to use an object so that we can make it clearer what is being whitelisted. My current thoughts are something like:

```js
rules: {
  'nested-interactive': {
    ignoredTags: ['a', 'button'], // list of tag names to ignore
    ignoreTabindex: true, // ignore the tabindex check
    ignoreUsemapAttribute: ['img', 'object'], // ignore `usemap` check for specific tag names
    additionalInteractiveTags: ['some-custom-tag'], // not sure this is needed, but it seams neat :P
  }
}
```

I believe that we can convert the currently allowed array into the above object and issue a deprecation so we don't need a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,20 +191,11 @@ This rule forbids the following:
 The following values are valid configuration:
 
   * boolean -- `true` indicates all whitelist test will run, `false` indicates that the rule is disabled.
-  * array -- an array of whitelisted tests as the following
-
-Whitelist of option in the configuration (are tags name or element attributes):
-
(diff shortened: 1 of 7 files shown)


rwjblue:
Implemented all changes and is now backwards compatible (with a warning).  Will land once CI passes...

PR #55, opened 2016-05-19
Add tests for indentation calculation for escaped curlies and raw blocks

### LINKED ISSUE #54: Escaped curlies cause incorrect indentation error
The following template:
```handlebars
<div>
  \{{example}}
</div>
```
causes the following error:
```
block-indentation: Incorrect indentation for `{{example}}
` beginning at L2:C3. Expected `{{example}}
` to be at an indentation of 2 but was found at 0.(dummy/pods/examples/computed-properties/any/template @ L2:C3): 
`<div>
  \{{example}}
</div>`
```
It seems like the indentation calculation is thrown off by the line starting with the sequence `\{{`.

mriska:
Fixes #54.

The following template:
```handlebars
<div>
  \{{example}}
</div>
```
caused the following error:
```
block-indentation: Incorrect indentation for `{{example}}
` beginning at L2:C3. Expected `{{example}}
` to be at an indentation of 2 but was found at 0.(dummy/pods/examples/computed-properties/any/template @ L2:C3): 
`<div>
  \{{example}}
</div>`
```

It seems like the indentation calculation is thrown off by the line starting with the sequence `\{{`.

The problem seemed to be due to the template compiler compiling `\n  ` and the following `{{example}}` into separate textNodes.

### DIFF
--- a/test/unit/rules/lint-block-indentation-test.js
+++ b/test/unit/rules/lint-block-indentation-test.js
@@ -8,6 +8,16 @@ generateRuleTests({
   config: 2,
 
   good: [
+    [
+      '<div>',
+      '  \\{{example}}',
+      '</div>'
+    ].join('\n'),
+    [
+      '{{{{if isMorning}}}}',
+      '  Good Morning',
+      '{{{{/if}}}}'
+    ].join('\n'),
     '\n  {{#each cats as |dog|}}\n  {{/each}}',
     '<div><p>Stuff</p></div>',
     '<div>\n  <p>Stuff Here</p>\n</div>',


rwjblue:
Thank you for digging in and figuring out what is happening.  I'm not certain that this fix is quite correct though.  I suspect (I'll confirm shortly) that there is actually a bug in HTMLBars that is reporting the location information for `\{{foo}}` starts at column `1` instead of `0`.  We should fix that bug, and then I don't think any changes will be needed here.

rwjblue:
Demo: http://rwjblue.jsbin.com/cecume/edit?html,js,output

Given a template of `\{{foo}}`, the text node indicates that it starts at column 1 when it starts at 0.

mriska:
Thanks for the swift reply, and a big thanks for the template linting addon in general. 

I actually found another bug also that might be related:
```handlebars

{{{{if isMorning}}}}
  Good Morning
{{{{/if}}}}
```
Gives the following error message:
```
Incorrect indentation for `if` beginning at L1:C0. Expected `{{/if}}` ending at L3:C11 to be at an indentation of 0 but was found at 4."
```
I have added that test to this PR and removed my incorrect fix from it.

mriska:
I will take a look at the HTMLBars repo and see if I understand it well enough to generate a test and fix for the underlying issue.

rwjblue:
@mriska - Awesome!  You can add the test in this file (this is where I have been adding most tests for this stuff): https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-syntax/tests/loc-node-test.js.  A failing test would help a ton there (its always the first thing I have to do when fixing these issues anyways). I'm not 100% sure where the fix would go for the `\{{foo}}` case, but I suspect it has to do with how handlebars itself parses the `\`.  The first thing to check would be in the loc info in the `ContentStatement` [here](https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-syntax/lib/parser/handlebars-node-visitors.js#L105) when the `TextNode` is being processed for `\{{foo}}`.

test/unit/plugins/lint-block-indentation-test.js
```diff
@@ -8,6 +8,12 @@ generateRuleTests({
   config: 2,
 
   good: [
+    '<div>\n' +
```
rwjblue: I've been slowly working on changing from string concat, to using arrays to do the concatting via `.join('\n')`.  Would you mind tweaking these to be like this:

```js
  [
    '<div>',
    '  \\{{example}}',
    '</div>'
  ].join('\n'),
```

rwjblue:
Also, I think that newer example (`{{{{if foo}}}}`) is a bug in ember-template-lint and not HTMLBars, can you file a separate issue so we can track that independently?

mriska:
Perfect, thanks for the additional info. I will update the PR and file a separate issue.

mriska:
Opened new issue #65 for raw block issue.

mriska:
@rwjblue - Tests are now updated for style and rebased. Have not had the time to look at a possible fix for the raw blocks.

mriska:
@rwjblue, Thanks for fixing these.

PR #64, opened 2016-05-28
Rule: img tags must have alt attributes

### LINKED ISSUE #45: Rule: img tags must have alt attributes
Hello! Please let me know if this is the way I should do it. If this issue is cool, I'll add more rule proposals.

Bad:

```
<img src="hello.jpg">
<img src="//cdn.verylongurl.com/1234/bucket/hello.jpg">
```

Good:

```
<img src="hello.jpg" alt="">
<img src="logo.jpg" alt="Your Website Name">
<img src="save-icon.jpg" alt="Save draft">
<img src={{userGeneratedUrl}} alt={{userGeneratedCaption}}>
```

Screen readers read images as text. Without `alt` attributes, screen readers fallback to reading the full `src` for an image out loud. With CDNs, that will get very annoying :-) To mitigate, always provide `alt` attributes for images. The value of the attribute should present the content and function of the image. Blank alt attributes are preferred if the image is purely presentational, but if it displays information (e.g. an image of text), transcribe the text; or if it's tied to functionality, write what it does as if it were a button. Ref: http://webaim.org/techniques/alttext/

binhums:
Requires `img` tags to have an `alt` attribute defined. See #45 for more information.

Fixes #45

### DIFF
--- a/test/unit/helpers/ast-node-info-test.js
+++ b/test/unit/helpers/ast-node-info-test.js
@@ -1,0 +1,13 @@
+var assert = require('power-assert');
+var parse = require('htmlbars/dist/cjs/htmlbars-syntax').parse;
+var AstNodeInfo = require('../../../lib/helpers/ast-node-info');
+
+describe('isImgElement', function() {
+  it('can detect an image tag', function() {
+    var tableAst = parse('<table></table>');
+    assert(AstNodeInfo.isImgElement(tableAst.body[0]) === false);
+
+    var imgAst = parse('<img />');
+    assert(AstNodeInfo.isImgElement(imgAst.body[0]) === true);
+  });
+});
--- a/test/unit/rules/lint-img-alt-attributes-test.js
+++ b/test/unit/rules/lint-img-alt-attributes-test.js
@@ -1,0 +1,28 @@
+'use strict';
+
+var generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'img-alt-attributes',
+
+  config: true,
+
+  good: [
+    '<img alt="hullo">'
+  ],
+
+  bad: [
+    {
+      template: '<img>',
+
+      result: {
+        rule: 'img-alt-attributes',
+        message: 'img tags must have an alt attribute',
+        moduleId: 'layout.hbs',
+        source: '<img>',
+        line: 1,
+        column: 0
+      }
+    }
+  ]
+});
--- a/test/unit/rules/lint-img-alt-attributes-test.js
+++ b/test/unit/rules/lint-img-alt-attributes-test.js
@@ -8,7 +8,8 @@ generateRuleTests({
   config: true,
 
   good: [
-    '<img alt="hullo">'
+    '<img alt="hullo">',
+    '<img src="hello.jpg" alt="">'
   ],
 
   bad: [
--- a/test/unit/rules/lint-img-alt-attributes-test.js
+++ b/test/unit/rules/lint-img-alt-attributes-test.js
@@ -8,8 +8,7 @@ generateRuleTests({
   config: true,
 
   good: [
-    '<img alt="hullo">',
-    '<img src="hello.jpg" alt="">'
+    '<img alt="hullo">'
   ],
 
   bad: [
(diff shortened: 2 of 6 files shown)


binhums:
Pushing this up for initial feedback, does this seem sensible?

Notably does not implement more nuanced tests described in https://github.com/rwjblue/ember-template-lint/issues/45#issuecomment-219041310. I would suggest that we can iterate to add those in if they are requested :shipit:

binhums:
@rwjblue any guidance on adding the fix section. I had a go using the htmlbars print and attr builders but that seemed heavy handed.

binhums:
Also what are your thoughts on adding a link in the rule message to a W3C description or something.

binhums:
Removing WIP flag, @rwjblue any comments on implementation?

test/unit/rules/lint-img-alt-attributes-test.js
```diff
@@ -0,0 +1,29 @@
...
+var generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'img-alt-attributes',
+
+  config: true,
+
+  good: [
+    '<img alt="hullo">',
+    '<img src="hello.jpg" alt="">'
```
Robdel12: I know we've had some back and forth on this but I don't think having an empty alt tag is any different than not having one at all.

While we're creating rules for `img` tags, should we enforce an `src` attribute? The spec says it's required. https://www.w3.org/TR/html-markup/img.html

test/unit/helpers/ast-node-info-test.js
```diff
@@ -0,0 +1,13 @@
+var assert = require('power-assert');
+var parse = require('htmlbars/dist/cjs/htmlbars-syntax').parse;
+var AstNodeInfo = require('../../../lib/helpers/ast-node-info');
+
+describe('isImgElement', function() {
+  it('can detect an image tag', function() {
+    var tableAst = parse('<table></table>');
+    assert(AstNodeInfo.isImgElement(tableAst.body[0]) === false);
+
+    var imgAst = parse('<img />');
```
rwjblue: I don't think we need the closing `/` here (and if added in an app template we would error for self closing void elements).

rwjblue:
@binhums - This looks good, can you add another bad test for `` :trollface:?

test/unit/helpers/ast-node-info-test.js
Robdel12: Interesting. Does HTMLBars throw an error for a closing `/`? It's optional and that's how I've always written my `img` tags. 

> Optionally, a "/" character, which may be present only if the element is a void element.

https://www.w3.org/TR/html-markup/syntax.html#void-element

(also just learned all about void elements, woo! 🎉 )

test/unit/rules/lint-img-alt-attributes-test.js
binhums: @Robdel12 I imagine a framework for required attributes for certain elements will probably be an eventual solution. E.g. like in axe-core:

https://github.com/dequelabs/axe-core/blob/4ef74435ac92a88c602690ac1e243c404f750d00/lib/commons/aria/index.js#L159

binhums: On `alt=""`:

> @binhums: i think screenreaders read out the whole URL in the 2nd case, and skips over in the former case

https://embercommunity.slack.com/archives/topic-a11y/p1464808554001560

binhums: I'll double check.

Robdel12: That sounds like a terrible experience to me :/

test/unit/helpers/ast-node-info-test.js
rwjblue: @Robdel12 - Nope, definitely doesn't throw.  But then if that were the only reason to change behaviors this linter wouldn't exist 😉 .

binhums:
Turns out the AST still reads in `value.chars` as an empty string for `` vs ``, added a helper to distinguish these cases.

test/unit/rules/lint-img-alt-attributes-test.js
rwjblue: Personally, I'd prefer to require at least one char in `alt`.  Which would make this actually a `bad` example.

Sorry to be so wishy washy on it...

binhums: NP, I'm in favour of being strong here.

binhums:
Updated to disallow all kinds of empty `alt` attribute.

Robdel12:
🙏🏼🙏🏼

rwjblue:
Thank you!

sarenji:
With this rule now disallowing blank `alt` tags, how do we denote an `` as purely presentational?

rwjblue:
Good point. I am far from an expert, but I suspect we might make an exception in the rule for "role=presentation". Does that seem correct?

Robdel12:
I would say `aria-hidden="true"`. If it's truly presentational and you're dead set on not giving an alt tag then make the screen reader ignore the image.

rwjblue:
@Robdel12 / @sarenji - Created https://github.com/rwjblue/ember-template-lint/issues/72 to track that.

PR #81, opened 2016-06-05
Allow child elements of non-interactive elements

### LINKED ISSUE #80: Invalid interactive detected on parent nodes
Currently the invalid-interactive rule is triggered when a non-interactive element has a child with an interaction - regardless of whether the child is an interactive element. E.g.

```
<li>
  <button {{action 'foo'}}>
</li>
```

I have a fix incoming.

binhums:
The invalid interactive rule only keeps track of entering and exiting non-interactive elements. This means that the rule will be triggered on a valid interactive element that has any non-interactive parents.

This PR resets the state of the current element for every `ElementNode` that is visited.

Fixes #80.

### DIFF
--- a/lib/rules/lint-invalid-interactive.js
+++ b/lib/rules/lint-invalid-interactive.js
@@ -16,9 +16,7 @@ module.exports = function(addonContext) {
 
     var visitor = {
       enter: function(node) {
-        if (!isInteractiveElement(node)) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Good catch, thank you!

rwjblue:
Released in v0.5.13.

Issue #57, opened 2016-05-20
Indentation for empty elements


amk221:
For elements with a lot of attributes, I like to put them on their own line:

```htmlbars
<div
  id="my-button"
  class="button"
  style={{escape "border: 1px solid red"}}
  onclick={{action 'foo'}}
  onmouseover={{action 'bar'}}
  onmouseout={{action 'baz'}}
  tabindex=0
  title="the title"></div>
```

With the above example, I get:
`Expected '</div>'  to be at an indentation of 0 but was found at 20`

I can't suppress the linting error by doing this:

```htmlbars
<div
  ...
  class="button">
  {{~how to suck up whitespace?~}}
</div>
```

amk221:
In fact, I also get this when the element is not empty (unsure what to title this issue?)

```htmlbars
<div
  ...
  class="button">&times;</div>
```

rwjblue:
I'm going to close this one in favor of #63, but I'm thinking/working on a fix...

PR #82, opened 2016-06-05
Don't check end block indentation of empty ElementNodes

### LINKED ISSUE #63: Textareas display whitespace from block indentation
Since textareas are block elements, the linter will instruct the closing tag to match the indentation of the opening tag.

e.g. to change

    <textarea></textarea>

to

    <textarea>
    </textarea>

When this rule is obeyed the textarea will include the added whitespace in the input box.

image

This took a while to debug :-)

Heres the rule message:

```
block-indentation: Incorrect indentation for `textarea` beginning at L31:C6. Expected `</textarea>` ending at L36:C88 to be at an indentation of 6 but was found at 77.(live-ops/templates/components/new-last-minute-job-form @ L36:C88): 

`<textarea class="form-control"
                id="job-instructions"
                rows="3"
                placeholder="Do it well"
                value={{job.instructions}}
                oninput={{action 'updateInstructions' value='target.value'}}></textarea>`
```

Any ideas for a potential fix? In the mean time I'll disable the rule.

binhums:
The block-indentation rule is raised if the start tag spans multiple lines - even if the start tag is empty. E.g.

```
<div id="multi"
       class="line"></div>
```

This PR skips checking the indentation of an end tag if the element does not contain any children. Note that whitespace will generate a TextNode child if it exists.

Fixes #63 which is a duplicate of #57.

### DIFF
--- a/lib/helpers/ast-node-info.js
+++ b/lib/helpers/ast-node-info.js
@@ -12,12 +12,20 @@ function isTextNode(node) {
   return node.type === 'TextNode';
 }
 
+function isCommentStatement(node) {
+  return node.type === 'CommentStatement';
+}
+
 function isElementNode(node) {
   return node && node.type === 'ElementNode';
 }
 
 function isComponentNode(node) {
   return node && node.type === 'ComponentNode';
+}
+
+function isMustacheStatement(node) {
+  return node.type === 'MustacheStatement';
 }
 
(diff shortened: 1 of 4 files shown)


rwjblue:
Seems good to me, thank you for all the cleanup here too!

PR #94, opened 2016-06-28
Do not validate if source doesn't match node

btecu:
Currently is possible for nodes to be matched as closing parent nodes that are of different type.
`</iframe>` should not be considered closing for `</if>` or `{{/if}}`.


rwjblue:
Awesome, thank you!!

PR #98, opened 2016-07-08
Tilde fix

btecu:


lib/rules/lint-block-indentation.js
```diff
@@ -224,6 +224,8 @@ module.exports = function(addonContext) {
         } else {
           childStartColumn = firstNonWhitespace;
         }
+      } else if (this.sourceForNode(child) === '') {
```
rwjblue: What type of node is `child` in this case?

btecu: `ElementNode`

kmiyashiro:
Are there any blockers for this fix?

rwjblue:
I believe this has been addressed on master already (and tests added [here](https://github.com/rwjblue/ember-template-lint/blob/a48191c1e3820f7cf304eba15af1cbcf87dfeeef/test/unit/rules/lint-block-indentation-test.js#L146-L152)).

Closing, but please open an issue if I missed something...

PR #103, opened 2016-07-28
Add `link-rel-noopener` to recommended rules

cibernox:
Also improve the wording of the rule in the readme to better explain why `rel="noopener"`
solves problems.


cibernox:
@rwjblue any thoughts on this?

rwjblue:
Sorry, yes, I completely agree with the change, but I need to queue it up for 0.6.0 since it may be considered a breaking change.

Issue #60, opened 2016-05-23
No inline {{link-to}}


nathanhammond:
I'm not sure we can actually detect this anymore because of the behind-the-scenes transformation, but:

Good:
```handlebars
{{#link-to "foo"}}Foo{{/link-to}}
```

Bad:
```handlebars
{{link-to "Foo" "foo"}}
```

rwjblue:
Our plugins run unrelated to Embers own (Embers plugins are not ran during linting).

PR #110, opened 2016-08-21
Add inline-link-to rule

mitchlloyd:
I wanted to get my feet wet with this project and I don't like the inline form of link to so I took a shot at implementing a rule for #60.

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,32 @@ The following values are valid configuration (same as the `nested-interactive` r
     * `ignoreUsemapAttribute` - When `true` ignores the `usemap` attribute on `img` and `object` elements. Defaults `false`.
     * `additionalInteractiveTags` - An array of element tag names that should also be considered as interactive. Defaults to `[]`.'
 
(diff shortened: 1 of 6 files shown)


rwjblue:
This looks great! @nathanhammond and I share your hatred of inline link-to...

My only question is RE: the rule name.  I realize that they are all over the map so far and I should come up with a "rule for the rule names". 

Would this rule may be better as `inline-link-to` (which follows the patter of `triple-curlies` / `bare-strings` / `nested-interactive` / `self-closing-void-elements` and forbids the thing it describes)?

mitchlloyd:
"inline-link-to" works for me. Going forward maybe this project could adopt
naming similar to ESLint where rules are generally affirmative "eqeqeq" and
forbidding rules begin with "no".

rwjblue:
Yeah, I agree

mitchlloyd:
Updated. Seems like a good one to squash if this looks good.

PR #111, opened 2016-09-20
Check for unsafe style concatenation

ef4:
Ember has a runtime warning that says "Binding style attributes may introduce cross-site scripting vulnerabilities." It can only be avoided by always marking the bound value with `Ember.String.htmlSafe`. While we can't detect statically if you're always providing a safe string, we can detect cases common where it's impossible that you're doing so. 

Forbidden examples:

```hbs
<div style="{{make-background url}}">
<div style="background-style: url({{url}})">
```

Allowed examples:

```hbs
<div style={{make-background url}}>
<div style="background-image: url(/pretty.png)">
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,26 @@ This rule is configured with one boolean value:
 
   * boolean -- `true` for enabled / `false` for disabled
 
+#### style-concatentation
+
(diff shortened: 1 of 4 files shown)


rwjblue:
Awesome, thanks @ef4!

PR #113, opened 2016-09-29
[BUGFIX] Honor `additionalInteractiveTags` option in `invalid-interactive` rule

### LINKED ISSUE #101: invalid-interactive additionalInteractiveTags not working?
I have this:

```<form {{action 'foo' on="submit"}}>```

and its failing the invalid-interactive check.  We don't have buttons in the form, but we want to be able to submit the form with the enter key for accessibility.  It seems like it should be okay for forms to submit actions...

I still think its odd to default forms to fail this check.  I'm trying to change this behavior with this rule and it doesn't seem to be working:

        'invalid-interactive': {additionalInteractiveTags:['form']},

when looking at 'lint-invalid-interactive.js' it doesn't appear to be using additionalInteractiveTags like 'lint-nested-interactive.js' is...

cibernox:
Closes #101

### DIFF
--- a/lib/rules/lint-invalid-interactive.js
+++ b/lib/rules/lint-invalid-interactive.js
@@ -11,12 +11,23 @@ module.exports = function(addonContext) {
     return parseConfig(this.ruleName, config);
   };
 
+  InvalidInteractive.prototype.isCustomInteractiveElement = function(node) {
+    var additionalInteractiveTags = this.config.additionalInteractiveTags || [];
+
+    if (additionalInteractiveTags.indexOf(node.tag) > -1) {
+      return true;
+    } else {
+      return false;
+    }
+  };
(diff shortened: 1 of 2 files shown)


rwjblue:
This looks good, but please don't tell me you are actually using `<div>` as an interactive element 😝

cibernox:
@rwjblue No, but `<form {{action "foo" on="submit"}}>` shouldn't trigger this rule and it does. Specifying `form` in the list of interactive tags is the simplest way to fix this. 

Thechnically the best fix would be to only consider a bad practice to bind to the `click` event (via `onclick={{action "foo"}}` or via `{{action "foo"}}`), but have `onmousenter={{}}` on a `<div>` is not a violation of this rule.

PR #115, opened 2016-09-29
Adding More Whitelisted Characters

derekdowling:
These are sometimes used in their bare format without the need for translations. It would be nice if they were included by default.


rwjblue:
Thank you!

PR #122, opened 2016-10-28
Add deprecation rule about inline View helper

duggiefresh:
Good: `{{foo}}`
Bad: `{{view 'foo'}}`, `{{view.some.path}}`

For https://github.com/rwjblue/ember-cli-template-lint/issues/28

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,38 @@ Instead, you should write the template as:
 
 More information is available at the [Deprecation Guide](http://emberjs.com/deprecations/v1.x/#toc_code-in-code-syntax-for-code-each-code).
 
+#### deprecated-inline-view-helper
+
+In Ember 1.12, support for invoking the inline View helper was deprecated.
+
+For example, this rule forbids the following:
+
+```hbs
+{{view 'this-is-bad'}}
+
+{{view.also-bad}}
+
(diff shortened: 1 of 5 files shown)


rwjblue:
Awesome, thank you for working on this! 

Can you also add "bad" tests for these scenarios:

```hbs
{{foo-bar baz=view.derp}}
```

```hbs
<div data-foo={{view.hallo}}></div>
```

This is looking really good...

duggiefresh:
Sure thing, anything for you!

duggiefresh:
Updated per request, @rwjblue 

It should be noted that determining values for `TextNode` is tricky, since it seems we can't differentiate between Strings and Booleans being passed in. Hence, why I omitted the other attrs being passed through here: https://github.com/rwjblue/ember-template-lint/pull/122/files#diff-d20b3fb270c2429d9b0638c6a423f1c9R88

Thanks!

rwjblue:
More run tests (sorry!):

```hbs
{{#foo-bar derp=view.dumb}}{{/foo-bar}}
```

I think you'll have to implement `BlockStatement` (like `MustacheStatement`) for that one

duggiefresh:
Updated per request! 🍨

rwjblue:
Thanks Doug!

PR #134, opened 2016-11-09
Handlebars comment instructions

### LINKED ISSUE #79: Add mechanism to disable rules inline within a given block.
The current implementation of `<!-- template-lint disabled=true -->` or `<!-- template-lint bare-strings=false -->` disables the rule for the entire "rest" of the template they are in. We should add a better control/configuration system so that it is possible to disable all or a single rule within a given block.

The principle behind this (i.e. how we think of it applying to the DOM tree) is that we're providing two ways of configuring the linter in a template:

1. Add an instruction as an in element comment (`<div {{! template-lint`) that applies to it and (optionally) to its descendants.
2. Add an instruction as a child-less node (`{{! template-lint`) that applies to all of its later siblings nodes and (mandatorily) to their descendants.

```hbs
<div>
  {{! template-lint disabled=true }}
  {{! Things are disabled here }}
</div>
{{! But not disabled here }}
```

```hbs
{{#foo-bar as |baz|}}
  {{! template-lint invalid-interactive=false }}
  {{! invalid-interactive is disabled here }}
{{/foo-bar}}
{{! invalid-interactive is reset to its original / global configuration here }}
```

```hbs
<div>
  <div {{! template-lint rule='invalid-interactive' disabled=true}}>
    {{! invalid-interactive *not* disabled }}
  </div>
</div>
```

```hbs
<div>
  <div {{! template-lint rule='invalid-interactive' disabled=true recursive=true}}>
    {{! invalid-interactive *is* disabled }}
  </div>
</div>
```

<details>
  <summary>Original Proposal</summary>

Off the top of my head, I would like to add something like:

``` hbs
{{#template-lint disabled=true}}
  {{! Things are disabled here }}
{{/template-lint}}

{{! But not disabled here }}
```

These `{{template-lint` component invocations would be completely removed at build time (just like the HTML comments are today). This would also allow us to slowly phase out the current usage of HTML comments, which is somewhat odd anyways. 

Here is an early mockup of what this might look like: http://rwjblue.jsbin.com/vayamom/edit?html,js,output
</details>

bendemboski:
Fixes #79.

This implements a new way of providing in-template instructions for configuring the rules. Rather than using HTML comments that modify a rule for the rest of the file, we can now use handlebars comments to modify a rule in some DOM-related scope.

This also allows full inline rule configuration rather than just enabling/disabling since it's easy to embed JSON strings inside handlebars comments.

I ended up solving the recursive/non-recursive issue in a fairly blunt, but I think effective way. It's pretty simple -- if we think of instructions as having names and arguments, then the name plus where the comment appears in the DOM tells you what the scope of the instruction is, and the arguments tell you what rules it affects and how. So it's all pretty much like we discussed in #79, except you can append `-tree` to the instruction name to make it recursive:

```handlebars
<div {{! template-lint-disable "my-rule" }}>
  {{! my-rule *not* disabled }}
</div>
<div {{! template-lint-disable-tree "my-rule" }}>
  {{! my-rule *is* disabled }}
</div>
```

and similarly for `template-lint-enable` and `template-lint-configure`.

Still TODO:

- [x] Deprecate old-style HTML comment instructions
- [x] Update unit tests to use new-style comment instructions
- [x] Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,19 +87,19 @@ It is also possible to disable specific rules (or all rules) in a template itsel
 <!-- disable all rules -->
 {{! template-lint-disable }}
 
-<!-- disable only bare-strings -->
+<!-- disable bare-strings -->
 {{! template-lint-disable bare-strings }}
 
-<!-- disable only bare-strings and triple-curlies -->
+<!-- disable bare-strings and triple-curlies -->
 {{! template-lint-disable bare-strings triple-curlies }}
 
 <!-- enable all rules -->
 {{! template-lint-enable }}
 
-<!-- enable only bare-strings -->
+<!-- enable bare-strings -->
 {{! template-lint-enable bare-strings }}
 
-<!-- enable only bare-strings and triple-curlies -->
+<!-- enable bare-strings and triple-curlies -->
 {{! template-lint-enable bare-strings triple-curlies }}
 ```
 
@@ -151,6 +151,8 @@ An in-element instruction with the `-tree` suffix will apply to that element and
   </p>
 </div>
 ```
+
+Note that enabling a rule (`{{! template-lint-enable }}`) that has been configured in-template (`{{! template-lint-configure }}`), will restore it to its default configuration rather than the modified in-template configuration for the scope of the `{{! template-lint-enable }}` instruction.
 
 ### Configuration Keys
 
--- a/README.md
+++ b/README.md
@@ -84,14 +84,73 @@ module.exports = {
 It is also possible to disable specific rules (or all rules) in a template itself:
 
 ```hbs
-{{! disable all rules for this template }}
-<!-- template-lint disable=true -->
+<!-- disable all rules -->
+{{! template-lint-disable }}
 
-{{! disable specific rules for this template }}
-<!-- template-lint bare-strings=false -->
+<!-- disable only bare-strings -->
+{{! template-lint-disable bare-strings }}
+
+<!-- disable only bare-strings and triple-curlies -->
+{{! template-lint-disable bare-strings triple-curlies }}
+
+<!-- enable all rules -->
+{{! template-lint-enable }}
+
+<!-- enable only bare-strings -->
+{{! template-lint-enable bare-strings }}
+
+<!-- enable only bare-strings and triple-curlies -->
+{{! template-lint-enable bare-strings triple-curlies }}
 ```
 
-It is not currently possible to change rule configuration in the template.
+and to configure rules in the template:
+
+```hbs
+{{! template-lint-configure bare-strings ["ZOMG THIS IS ALLOWED!!!!"] }}
+
+{{! template-lint-configure bare-strings {"whitelist": "(),.", "globalAttributes": ["title"]} }}
+```
+
+The configure instruction can only configure a single rule, and the configuration value must be valid JSON that parses into a configuration for that rule.
+
+These configuration instructions do not modify the rule for the rest of the template, but instead modify it within whatever DOM scope the comment instruction appears.
+
+An instruction will apply to all later siblings and their descendants:
+
+```hbs
+<!-- disable for <p> and <span> and their contents, but not for <div> or <hr> -->
+<div>
+  <hr>
+  {{! template-lint-disable }}
+  <p>
+    <span>Hello!</span>
+  </p>
+</div>
+```
+
+An in-element instruction will apply to only that element:
+
+```hbs
+<!-- enable for <p>, but not for <div>, <hr> or <span> -->
+<div>
+  <hr>
+  <p {{! template-lint-enable }}>
+    <span>Hello!</span>
+  </p>
+</div>
+```
+
+An in-element instruction with the `-tree` suffix will apply to that element and all its descendants:
+
+```hbs
+<!-- configure for <p>, <span> and their contents, but not for <div> or <hr> -->
+<div>
+  <hr>
+  <p {{! template-lint-configure-tree block-indentation "tab" }}>
+    <span>Hello!</span>
+  </p>
+</div>
+```
 
 ### Configuration Keys
 
@@ -173,15 +232,6 @@ but allows the following:
 ```hbs
 {{!-- comment goes here --}}
 ```
-
-Html comments containing linting instructions such as:
-
-```hbs
-<!-- template-lint bare-strings=false -->
-```
-
-are of course allowed (and since the linter strips them during processing, they will not get compiled and rendered into the DOM regardless of this rule).
-
 
 #### triple-curlies
 
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -1,18 +1,7 @@
 'use strict';
 
-var assign = require('lodash').assign;
-
-function invokeVisitorEnter(visitor, args, thisContext) {
-  var func;
-
-  if (typeof visitor === 'function') {
-    func = visitor;
-  } else if (visitor && visitor.enter) {
-    func = visitor.enter;
-  }
-
-  return func.apply(thisContext, args);
-}
+var _ = require('lodash');
+var assign = _.assign;
 
 module.exports = function(options) {
   var log = options.log;
@@ -30,7 +19,13 @@ module.exports = function(options) {
     this._log = log;
     this.ruleName = ruleName;
     this.severity = defaultSeverity;
+    // To support DOM-scoped configuration instructions, we need to maintain
+    // a stack of our configurations so we can restore the previous one when
+    // the current one goes out of scope. The current one is duplicated in
+    // this.config so the rule implementations don't need to worry about the
+    // fact that there is a stack.
     this.config = this.parseConfig(config);
+    this.configStack = [this.config];
   }
 
   BasePlugin.prototype.parseConfig = function(config) {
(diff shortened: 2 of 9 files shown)


lib/rules/base.js
```diff
@@ -30,7 +19,13 @@ module.exports = function(options) {
     this._log = log;
     this.ruleName = ruleName;
     this.severity = defaultSeverity;
+    // To support DOM-scoped configuration instructions, we need to maintain
+    // a stack of our configurations so we can restore the previous one when
+    // the current one goes out of scope. The current one is duplicated in
+    // this.config so the rule implementations don't need to worry about the
+    // fact that there is a stack.
     this.config = this.parseConfig(config);
+    this.configStack = [this.config];
```
rwjblue: Do we intend for this to be "public API" for the rules?  I haven't thought deeply about it, but it seems like it might be ok to make it public.  

If we don't want it to be public, we should underscore it.

lib/rules/base.js
```diff
@@ -48,57 +43,300 @@ module.exports = function(options) {
...
+    visitors.CommentStatement = visitors.CommentStatement || this._wrapVisitor(null, astHandlers.CommentStatement);
+    visitors.MustacheCommentStatement = visitors.MustacheCommentStatement || this._wrapVisitor(null, astHandlers.MustacheCommentStatement);
+    visitors.ElementNode = visitors.ElementNode || this._wrapVisitor(null, astHandlers.ElementNode);
+
+    //
+    // Set up our handlers
+    //
+
+    // Manage elementStack
+    astHandlers.ElementNode.keys.children.enter.before.push(function(elementNode) {
```
rwjblue: Since these are essentially static, should they just be included in the definition of `astHandlers` above? 

I'm not sold that moving the declaration above makes sense, but I am curious to your thoughts on it...

lib/rules/base.js
```diff
@@ -48,57 +43,300 @@ module.exports = function(options) {
...
+          // Applies to only this element, so pop the config when we enter the
+          // element's children, re-push it when we exit the element's
+          // children, and then re-pop it when we exit the element entirely.
+          astHandlers.ElementNode.keys.children.enter.before.unshift(function(node) {
+            if (node === elementNode) {
+              pluginContext._popConfig(config.value);
+              return false;
+            }
+          });
+          astHandlers.ElementNode.keys.children.exit.after.push(function(node) {
```
rwjblue: This slightly confuses me.  Why do we need to push this config back after the children exit? What can be after an elements children?

lib/rules/base.js
```diff
@@ -48,57 +43,300 @@ module.exports = function(options) {
...
     return visitors;
   };
 
-  BasePlugin.prototype.processVisitor = function(ruleVisitor) {
-    var pluginContext = this;
-    var visitor = {
-      enter: function(node) {
-        if (!pluginContext.isDisabled()) {
-          invokeVisitorEnter(ruleVisitor, [node], pluginContext);
+  BasePlugin.prototype.visitors = function() { };
```
rwjblue: Complete aside (for a future issue/PR): we should probably throw if `this.visitors()` returns a non-object...

lib/rules/base.js
```diff
@@ -48,57 +43,300 @@ module.exports = function(options) {
...
+    internalHandlers = internalHandlers || {};
+    var beforeHandlers = internalHandlers.before || [];
+    var afterHandlers = internalHandlers.after || [];
+    var plugin = this;
+    var i, ret;
+
+    return function() {
+      for (i = 0; i < beforeHandlers.length; i++) {
+        ret = beforeHandlers[i].apply(plugin, arguments);
+        if (ret === false) {
```
rwjblue: This is **really** clever, and I love it!  Can we change the "special" value to be not `false` though?  Something like a module scoped `var REMOVE_HANDLER = {};` (and the handlers can just do `return REMOVE_HANDLER;`)...

lib/rules/base.js
bendemboski: I was thinking that we don't want it to be public -- rules shouldn't be concerned about *how* their configuration is specified, just *what* the configuration is. My goal was to make that all a black box that resolves to a single `this.config` at any given time. So, unless you disagree I'll underscore `configStack`.

lib/rules/base.js
bendemboski: Hmm...yeah, I suppose the outer handlers are static (although they dynamically push additional handlers). Conceptually I was thinking of it more as an observable mixin, where declaring the `astHandlers` is the equivalent of the mixin setting up its observable framework, and then these registrations are the equivalent of the subclass registering its handlers in a constructor.

I was actually going to set up `astHandlers` in the plugin constructor before realizing that they if the plugin instance were used for multiple traverses (which I don't think they are?) that the astHandlers would need to be reset for each traverse.

So anyway, I'm not that's actually an answer, but I think to my eye the code ends up more readable in this imperative form, rather than being declarative. But I could easily be swayed...

bendemboski:
> Should we allow {{! template-lint-enable "fake" to be the same as {{! template-lint-enable fake?

Yeah, I suppose that makes sense -- it looks like eslint supports quoted rule names as well, so that would be in line with them. However, to keep the implementation sane I would opt to simply detect and remove leading/trailing quotes *after* space-splitting the string, rather than doing full quote parsing -- i.e. we would *not* support spaces inside quotes or escaped quotes. So if anybody tries to name a rule to contain spaces or quotes, after slapping them several times we'd need to tell them that our comment syntax doesn't support their rule.

lib/rules/base.js
rwjblue: Nope, seems good. The code structure alone seems better at this point.

lib/rules/base.js
rwjblue: Gotcha, and it makes sense. Let's leave a comment so we have a fighting chance to remember what we were thinking...

rwjblue:
> However, to keep the implementation sane I would opt to simply detect and remove leading/trailing quotes after space-splitting the string, rather than doing full quote parsing 

Seems reasonable. I think the easy heuristic is to throw an error if the rule name provided doesn't match one of our known rules.

bendemboski:
> Seems reasonable. I think the easy heuristic is to throw an error if the rule name provided doesn't match one of our known rules.

Sounds good.

> We should throw if `{{! template-lint-configure` is passed multiple rule names.

Right now I just throw everything after the (first) rule name at `JSON.parse` so I think we'll throw a JSON parsing error in that case regardless, since I can't think of any rule name that (quoted or unquoted) could be prepended with a space to a valid JSON string to produce another valid JSON string.

I suppose *if* we fail to parse the JSON, I could split off the first word from it and if it matches a known rule name, throw a different error than the generic 'bad JSON' error? Do you think that provides enough value to be worthwhile? I don't think it would be hard to implement, I'm just not sure it's that much more valuable than something like `failed to parse JSON: 'second-rule false'`.

rwjblue:
> Right now I just throw everything after the (first) rule name at JSON.parse so I think we'll throw a JSON parsing error in that case regardless, since I can't think of any rule name that (quoted or unquoted) could be prepended with a space to a valid JSON string to produce another valid JSON string.

This seems OK to me, but we should make sure our error messages are fairly well tailored here.  They should include the fact that the JSON failed to parse, the source for the comment node, the template name, line, and column.

bendemboski:
Yeah, since we're using the same mechanism used to report rule errors, we get all that for free 😄

bendemboski:
Removed [WIP] because all the items are done -- I think all that's left is review/feedback.

README.md
```diff
@@ -84,14 +84,73 @@ module.exports = {
...
 
-{{! disable specific rules for this template }}
-<!-- template-lint bare-strings=false -->
+<!-- disable only bare-strings -->
+{{! template-lint-disable bare-strings }}
+
+<!-- disable only bare-strings and triple-curlies -->
+{{! template-lint-disable bare-strings triple-curlies }}
+
+<!-- enable all rules -->
```
rwjblue: Just to confirm: this enables the rules as configured in the prior config stack, right?

```hbs
<div>
  {{! template-lint-disable bare-string }}
  <div>
    {{! template-lint-disable block-indentation }}
    <div>
      {{! template-lint-enable }} <b>what is the rule set now?</b>
    </div>
  </div>
</div>
```

README.md
```diff
@@ -84,14 +84,73 @@ module.exports = {
...
+{{! template-lint-enable }}
+
+<!-- enable only bare-strings -->
```
rwjblue: Does this enable only `bare-strings` or just enable bare-strings in addition to whatever else was enabled (I believe it does the latter?)?

bendemboski:
> How do we handle HTML Comments / disabling and reenabling?

Ah, right, I should probably write a unit test for that case to verify. The answer is that we are preserving the previous behavior of HTML comment instructions, which is that when we encounter one, its config applies for the rest of the file until/unless modified by another instruction.

As you can see [here](https://github.com/rwjblue/ember-template-lint/pull/134/files#diff-821ff4b48b6a2b56ee6a5aeded2a611cR484), when we encounter an HTML comment instruction we overwrite all existing config stack frames so the comment's configuration will persist until another HTML comment overwrites it or a handlebars comment pushes another config stack frame.

Seems like the most intuitive, er, least un-intuitive behavior. I was planning to not include this in the README as it's kinda confusing and not really a case we want to put much effort into supporting, but I can explain it there if you think it's necessary. I could also issue a warning if we detect HTML comments and handlebars comments in the same file, but I would hope that anybody encountering unexpected behavior would immediately realize that mixing the two is kinda weird.

bendemboski:
Oh, I guess ^^ didn't answer your question, and made me realize something I totally didn't until just now. Before this PR, there's no way to re-enable disabled rules -- see [here](https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/base.js#L119).

So all the HTML comments can do is disable rules for the rest of the template unless subsequent handlebars comments re-enable them.

rwjblue:
>  I was planning to not include this in the README as it's kinda confusing and not really a case we want to put much effort into supporting, but I can explain it there if you think it's necessary.

Agreed. I think its totally fine to leave it out of the README. Frankly, we should completely remove the `<!-- template-lint` documentation (which I think you did already?).

> So all the HTML comments can do is disable rules for the rest of the template unless subsequent handlebars comments re-enable them.

Right, but maybe that is fine?  I really don't care to deal with this at *all* now that we have a better system.  Using Handlebars comments this will all be possible (and much more logical)...

README.md
bendemboski: Yeah, the latter. The usage of the word 'only' is confusing -- I'll fix it.

README.md
rwjblue: Yes, that suggestion sounds perfect.  👍

bendemboski:
@rwjblue I believe this is ready for another review.

rwjblue:
👏

nathanhammond:
This. Is. Fantastic. ❤️

PR #138, opened 2016-11-12
Support for Plugins

ppcano:
@rwjblue Basic implementation for #136. 

- [X] Add a new `plugins `.
- [X] Expand `extends` to support external configurations. 
- [X] `extends` supports an array of configurations.
- [ ] Expose public API.

I don't know which APIs will become public. It may be better if `exposing the public API` is done by somebody else.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -1,28 +1,41 @@
 'use strict';
 
 var path = require('path');
+var resolve = require('resolve');
 var assign = require('lodash').assign;
-var existsSync = require('exists-sync');
 var chalk = require('chalk');
-var rules = require('./rules');
+var defaultRules = require('./rules');
+var defaultConfigurations = require('./config/index');
 
-var KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore'];
+var KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore', 'plugins'];
 
-function readConfigFromDisk(options) {
-  var providedConfigPath = options.configPath;
-  var configPath = providedConfigPath || path.join(process.cwd(), '.template-lintrc');
 
-  if(existsSync(configPath) || existsSync(configPath + '.js') || existsSync(configPath + '.json')) {
-    return require(configPath);
-  } else {
+function resolveConfigPath(configPath, pluginPath) {
+
+  configPath = configPath || path.join(process.cwd(), '.template-lintrc');
+
+
+  if (pluginPath) {
+
+    //throws exception if not found
+    pluginPath = resolve.sync(pluginPath, { basedir: path.dirname(configPath) });
+
+  }
+
+  try {
+    var filePath = pluginPath || configPath;
+    return require(filePath);
+  } catch(e) {
     return {};
   }
+
 }
 
 function ensureRootProperties(config) {
   config.rules = config.rules || {};
   config.pending = config.pending || [];
   config.ignore = config.ignore || [];
+  config.plugins = config.plugins || [];
 }
 
 function migrateRulesFromRoot(config, options) {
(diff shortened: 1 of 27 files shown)


ppcano:
@rwjblue Included latest changes. 

Let me know any suggested changes or improvements, thanks!

lib/get-config.js
```diff
@@ -4,9 +4,25 @@ var path = require('path');
...
+var KNOWN_EXTENDED_CONFIGURATION = ['recommended'];
+
+function readPluginFromDisk(options, pluginPath) {
+  var providedConfigPath = options.configPath;
+  var configPath = providedConfigPath || path.join(process.cwd(), '.template-lintrc');
+
+  var configDir = path.dirname(configPath);
+  pluginPath = path.join(configDir, pluginPath);
+
+  if(existsSync(pluginPath) || existsSync(pluginPath + '.js') || existsSync(pluginPath + '.json')) {
```
rwjblue: Lets use `resolve` package here instead.  

```js
var resolve = require('resolve');
// ...snip...
var pluginPath = resolve.sync(pluginPath, { basedir: configDir });
if (!pluginPath) {
  throw new Error('Invalid plugin provided, could not find `' + pluginPath + '`.');
}
return require(pluginPath);
```

lib/get-config.js
```diff
@@ -4,9 +4,25 @@ var path = require('path');
...
 var chalk = require('chalk');
-var rules = require('./rules');
+var defaultRules = require('./rules');
+
+var KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore', 'plugins'];
+var KNOWN_EXTENDED_CONFIGURATION = ['recommended'];
+
+function readPluginFromDisk(options, pluginPath) {
+  var providedConfigPath = options.configPath;
+  var configPath = providedConfigPath || path.join(process.cwd(), '.template-lintrc');
```
rwjblue: Now that we need this in two places (here and in `readConfigFromDisk`), lets extract it to a shared function:

```js
function resolveConfigPath(options) {
  var configPath = options.providedConfigPath || path.join(process.cwd(), '.template-lintrc');

  try {
    return require.resolve(configPath);
  } catch(e) {
    return null;
  }
}
```

lib/get-config.js
```diff
@@ -50,12 +67,124 @@ function migrateRulesFromRoot(config, options) {
...
+    if (typeof plugin === 'string') {
+      plugin = readPluginFromDisk(options, plugin);
+    } else if (typeof plugin !== 'object') {
+      logger.log(chalk.yellow('Plugin should be string to the plugin file or inline plugin object'));
+    }
+
+    if (plugin && plugin.name) {
+      pluginsHash[plugin.name] = plugin;
+    } else {
+      logger.log(chalk.yellow('Plugin name property is empty'));
```
rwjblue: We should include the info we have in this warning/error.  I _think_ the only info we have is `configPlugins[i]` and whatever `plugin` was resolved to?

lib/get-config.js
```diff
@@ -78,13 +207,17 @@ function validateRules(config, options) {
...
+
+
 module.exports = function(options) {
   var config = options.config || readConfigFromDisk(options);
 
   ensureRootProperties(config);
   migrateRulesFromRoot(config, options);
 
-  config = processExtends(config);
+  config = processPlugins(config, options);
```
rwjblue: Since we are mutating the `config` object in these functions, we shouldn't have to return and keep reassigning, right?

lib/get-config.js
```diff
@@ -50,12 +67,124 @@ function migrateRulesFromRoot(config, options) {
...
+  var loadedRules = assign({}, defaultRules);
+
+  //load plugin rules
+  for (var name in config.plugins) {
+
+    var pluginRules = config.plugins[name].rules;
+    if (pluginRules) {
+      loadedRules = assign(loadedRules, pluginRules);
+    } else {
+      logger.log(chalk.yellow('Plugin ' + name + ' does not define rules'));
```
rwjblue: I think that now that we allowed plugins to provide configuration and/or rules, that we shouldn't log a warning when no rules are detected.  Since it seems totally legit to me for a plugin to simply provide a set of configurations.

lib/get-config.js
```diff
@@ -50,12 +67,124 @@ function migrateRulesFromRoot(config, options) {
...
+
+  return config;
+
+}
+
+function processExtends(config, options) {
+
+  var logger = options.console || console;
+
+  function _getExtendedRules(extendsName) {
```
rwjblue: Can this be extracted to be a stand alone function (i.e. not nested)?  I can't tell if it requires closure scope access to things for it to function...

lib/get-config.js
```diff
@@ -4,9 +4,25 @@ var path = require('path');
 var assign = require('lodash').assign;
 var existsSync = require('exists-sync');
 var chalk = require('chalk');
-var rules = require('./rules');
+var defaultRules = require('./rules');
+
+var KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore', 'plugins'];
+var KNOWN_EXTENDED_CONFIGURATION = ['recommended'];
```
rwjblue: Instead of this, lets create a `lib/config/index.js` with:

```js
"use strict";

module.exports = {
  'recommended': require('./recommended')
};
```

This makes it much more similar to how `rules` are done, and how plugins will work...

Then we can use `name in KNOWN_CONFIGURATIONS` below in `_getExtendedRules`...

ppcano:
@rwjblue Applied the changes.

1) Let me know when I should rebase into one commit

2) Node 4 and 0.12 fails because using  [resolve](https://github.com/rwjblue/ember-template-lint/pull/138/commits/1e58d95b3e4a35e94aeb20defd35dc845bcbb036). Should I rollback to use `existSync`?

rwjblue:
Do we have any tests for a plugin loading a configuration that extends from another plugins configuration?

ppcano:
@rwjblue 

> Do we have any tests for a plugin loading a configuration that extends from another plugins configuration?

No. Is this the only test you are missing?

ppcano:
@rwjblue Complete latest changes. Waiting your final approval to rebase into one commit.

rwjblue:
Looks good. Let's rebase / squash and get it out in a beta version for testing...

ppcano:
@rwjblue DONE

rwjblue:
Thanks for working so hard on this @ppcano! I'll try to get it released in a few hours (on mobile ATM).

PR #140, opened 2016-11-17
Enable empty alt attributes for img-alt-attributes test

mydea:
See https://github.com/rwjblue/ember-cli-template-lint/issues/174

I really like the new "img-alt-attributes" rule. However, it currently also fails on e.g. ``, 
which is valid. Consider this use case for icons:

```html
<a href='users.html'>
  <img src='users.png' alt=''>
  Users
</a>
```

An empty alt tag has a distinctively different semantic meaning from a missing alt tag - an image with an empty alt tag will simply be ignored by screen readers, while an image with a missing alt tag will have its file name read (or something like this). In the above use case, by adding an alt tag "users" to the icon, a screen reader would read out "Users Users", which wouldn't be helpful at all. Further information on the topic can be found here: https://www.w3.org/TR/WCAG20-TECHS/H67.html

Considering this, the rule should allow empty alt tags and only error on missing ones.

### DIFF
--- a/lib/rules/lint-img-alt-attributes.js
+++ b/lib/rules/lint-img-alt-attributes.js
@@ -49,13 +49,5 @@ module.exports = function(addonContext) {
 function isAltPresent(node) {
   var altAttribute = AstNodeInfo.findAttribute(node, 'alt');
 
-  if (!altAttribute) { return false; }
-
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

PR #141, opened 2016-11-17
Enable submit actions on form elements for invalid-interactive test

mydea:
When using forms, for keyboard users it is generally better to have:

```hbs
<form {{action "submitForm" on="submit"}}>
  <input>
  <button>Submit</button>
</form>
```

instead of

```hbs
<form>
  <input>
  <button {{action "submitForm"}}>Submit</button>
</form>
```

However, this is currently reported as an error when using the invalid-interactive test.
Of course, you could whitelist `form`, but that would also allow `<form {{action "doSomethingOnClick"}}></form>`, something that shouldn't be allowed.

This PR adds a special case handling for this and allows an action on a form element, but only if `on="submit"` is set on it.

### DIFF
--- a/lib/rules/lint-invalid-interactive.js
+++ b/lib/rules/lint-invalid-interactive.js
@@ -44,6 +44,11 @@ module.exports = function(addonContext) {
         var modifierName = node.path.original;
 
         if (modifierName === 'action') {
+          // Allow {{action "foo" on="submit"}} on form tags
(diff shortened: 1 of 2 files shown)


rwjblue:
This seems good to me. 

@nathanhammond - Can you confirm from your side that this should be allowed?

MelSumner:
@rwjblue this is reasonable.

rwjblue:
Thanks @MelSumner!

PR #149, opened 2016-12-03
lint inline styles rule and tests

### LINKED ISSUE #97: Disallow inline 'style' attribute in html or hbs
Im most projects, we recommend that using css to style the dom elements. But some ones may prefer to write inline style rules, because it's just quick! I would be wonderful if we have a rule that totally disallow inline style rules.

``` css
<div class="navbar" style="margin-bottom: 10px; border: 0;">
...
</div>
```

And it would be even more wonderful if we can add inline directive comment to disable a rule for some rare reasons. I'm not sure if currently we support that.

``` hbs
{{!-- template-lint bare-strings:false --}}
```

jedrula:
resolves #97

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -13,6 +13,7 @@ module.exports = {
     'img-alt-attributes': true,
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
-    'unused-block-params': true
+    'unused-block-params': true,
+    'inline-styles': true
   }
 };
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -13,7 +13,6 @@ module.exports = {
     'img-alt-attributes': true,
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
-    'unused-block-params': true,
-    'inline-styles': true
+    'unused-block-params': true
   }
 };
--- a/README.md
+++ b/README.md
@@ -397,6 +397,26 @@ The block form is a little longer but has advantages over the inline form:
 This rule is configured with one boolean value:
 
   * boolean -- `true` for enabled / `false` for disabled
+  
+#### inline-styles
+
+Inline styles are not the best practice because they are hard to maintain and usually make the overall size of the project bigger. This rule forbids the inline styles.
+
+Forbidden:
+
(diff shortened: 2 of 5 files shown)


rwjblue:
Awesome!  Thanks for picking this up!

Can you add tests for the following scenarios (I suspect they all already will pass):

* `<div style={{foo}}></div>`
* `<div style="{{foo}} {{bar}}"></div>`
* `<div style></div>` (I don't even know if this should be "wrong", but its definitely weird)


Any other scenarios that you can think of?

jedrula:
I was thinking of adding some dynamic values for the styles, so sure, I will add that.

I aslo thought that maybe I should do:  

`<div style=""></div>`

that is probably very rare and does no harm - but again - weird and might as well be tested

maybe also one with multiple styles like: 

`<h1 style="color:blue;margin-left:30px;">This is a heading.</h1>`

rwjblue:
Sounds good, ping me when you update and I'll take another look.

Thanks again for working on it!

rwjblue:
Looks great!  I think this only needs some additions to the README, and we can ship..

jedrula:
@rwjblue I have added a few lines to the README (have a look if that will do) and added the rule to the recommended set.

lib/config/recommended.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     'img-alt-attributes': true,
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
-    'unused-block-params': true
+    'unused-block-params': true,
+    'inline-styles': true
```
rwjblue: We have to wait on changing the recommended set for 0.7.0 (as it is a "breaking" change, since it could fail apps that were previously passing using recommended).

Can you back out this change?

rwjblue:
The README looks good, only need to remove the recommended rule set changes for now and this should be good to go...

jedrula:
should we remove that or set it to false ?

rwjblue:
> @rwjblue should we remove that or set it to false ?

Please remove

jedrula:
@rwjblue done

rwjblue:
Thank you!

Issue #108, opened 2016-08-19
run linter from command line


RafLeszczynski:
Is there a way to run linter form the command line?

RafLeszczynski:
if not from what I see in this PR https://github.com/rwjblue/ember-template-lint/pull/96/files it should be relatively easy to write a script that could be executed from command line.

If anybody is currently working on this please let me know so we won't duplicate the effort.

rwjblue:
ember-cli-template-lint includes a command that runs against a project (`ember template-lint:print-failing` or something like that). I think it should be possible to add another command for `ember template-lint` that exits with non-zero exit code (and emits normal console output).

bardzusny:
Hello @rwjblue - I wrote a simple script that could perhaps be put to use here. It's simple `bin/ember-template-lint` executable, accepting path of `.hbs` file to check and returning formatted JSON with all detected errors.

I'm actually already using it daily with (a-bit-hacked version of) https://github.com/vim-syntastic/syntastic to provide me with in-editor linting, which I actually find the main use-case here (as opposed to analysing `ember-cli` output / awaiting test pipeline results).

I will happily clean it up a bit, write a spec or two, and contribute back upstream - could you perhaps confirm that simple executable script to be put into this repository is indeed the correct idea, and creating PR with it won't be wasted effort? Thanks!

rwjblue:
I would definitely love to land support for this with a bin script, just haven't had time to land one myself yet.

shuriu:
I see CLI support has landed as per #152, and the README is updated.

Unfortunately after trying to install via `yarn` initially, I've found no executable.
I've tried installing using the `npm` command from the README but still no executable.

Am I missing something?

Thanks, and great work on this btw! 👍

shuriu:
Sorry, just saw that this feature hasn't been released yet.

sukima:
When will this be released?

jasnross-bt:
Would love to see this released. Just ran into the same issue as noted above after following the README: installed using npm, but no bin script was available.

rwjblue:
This is now released in 0.7.0-beta.3.  Closing....

erichonkanen:
@bardzusny do you have the template linting working inside vim? mind sharing how to set that up?

bardzusny:
@erichonkanen see [ALE](https://github.com/w0rp/ale) and PRs: https://github.com/w0rp/ale/pull/452 , https://github.com/w0rp/ale/pull/589 .

[...]

PR #152, opened 2016-12-11
CLI executable

bardzusny:
Related issue: #108 .

Node.JS script making it possible to run `ember-template-lint` check from command line. Clear use-cases for it include:

* Pre-commit/push hooks
* Editor integrations

The idea laid in my backlog for quite a while, finally decided to push it even in initial, very much WIP form.

As it stands, I'm already using this script with `--json` flag together with hacked-up version of [Syntastic](https://github.com/vim-syntastic/syntastic) to provide me with in-editor template linting.

## Current functionality

```bash
# Prints all errors/warnings found in "app/template/application.hbs", in same format as currently provided by "ember-cli-template-lint" integration addon
$ ember-template-lint app/templates/application.hbs

# Prints all errors/warnings as JSON - for editor integrations
$ ember-template-lint --json app/templates/application.hbs
```

## Things to consider/do pre-merge:

- [x] **Node.JS version support.** I mercilessly used ESNext syntax, which could be a problem if we're to support anything older than current LTS version of Node.JS (6). No Babel here. We'd definitely want to confirm that it runs OK with 4, I think.
- [x] **Parsing command-line arguments.** Currently very primitive search of `process.argv` for: `--json` flag and path of `.hbs` file to lint. On one hand I'm not too comfortable with leaving it as-is, on the other including something like [yargs](http://yargs.js.org/) seems like definite overkill.
- [x] **Allow linting on more than one `.hbs` file at once.** Partially related to the above.
- [x] ~~**Handle erroneous arguments**. Anything more informative than `TypeError`s when someone passes-in path to non-existent `.hbs` file.~~ N/A anymore - if no valid file paths are passed-in, linting doesn't happen at all.
- [x] **Code duplication with [ember-cli-template-lint](https://github.com/rwjblue/ember-cli-template-lint).** Mostly limited to several lines in `errorsToMessages` function (see https://github.com/rwjblue/ember-cli-template-lint/blob/master/broccoli-template-linter.js#L74 ), still worth mentioning. Didn't really consider `ember-cli-template-lint` in the bigger image, will be happy to hear opinions on where exactly we want to divide the two (general idea being that we want `ember-cli-template-lint` to be just consumer of `ember-template-lint`).
- [x] **Tests.** Once everything else is out of the way, we'll definitely want to make sure command output is correctly formatted/exit code matches expectations/etc.
- [x] **README.** Make sure it includes information about provided executable.


bardzusny:
All of the steps listed above are up for discussion/additions are welcome, too. TL;DR is that this (executable script for linter) is not something I recall coding before - I'd love to hear all pointers/suggestions and iterate within this PR until we arrive at basic, but functional, sensible and properly tested version - that can be merged with pride 🚀.

rwjblue:
I'm very sorry for the delay here!  Answers below:

> Node.JS version support.

We need to support all supported Nodes (based on [the maintenance schedule](https://cdn.rawgit.com/nodejs/LTS/a7b6ce43/schedule.png)), which ATM is Node 4, 6, and 7.  I submitted #189 to get our test matrix in order.

> Parsing command-line arguments.

I think what we are doing now is perfectly fine.  Its super simple, and I would rather not add a dependency here when our needs are so simple.

> Allow linting on more than one .hbs file at once.

Yes, we need to do a few things here:

* Allow many arguments and treat all of them that do not start with `--` (to handle the `--json` case) as globs.
* Add support for `glob`s, iterating each matched file and running the linting against it.

> Handle erroneous arguments.

Agreed, but I do not consider this a blocker. It just makes the ergonomics a bit better.

> Code duplication with ember-cli-template-lint.

Lets solve this by exposing the shared function as a static method on `Linter` in `lib/index.js`.  This allows `ember-cli-template-lint` to use the function via `let errorsToMessages = require('ember-template-lint').errorsToMessage;`.

> Tests.

Yes, please. 😸

rwjblue:
@bardzusny - Oh, also one other item to add to the checklist is a blurb about this in the README.

rwjblue:
@bardzusny - Friendly ping!

bardzusny:
Hey @rwjblue - TL;DR here is that your response is exactly what I wished for originally, and clarifies everything I need to push this PR forward. Will do my best to do this soon, thanks! 🎊

bin/ember-template-lint
```diff
@@ -0,0 +1,82 @@
...
+      let globPath;
+      let isDirectory;
+
+      try {
+        isDirectory = fs.statSync(fileArg).isDirectory();
+      } catch(e) {
+        isDirectory = false;
+      }
+
+      if (isDirectory) globPath = `${fileArg}/**/*.hbs`;
```
rwjblue: I have a general preference for requiring a block with `if` / `else`, would you mind adding `{` / `}` around these?

bin/ember-template-lint
```diff
@@ -0,0 +1,82 @@
...
+    const filePath = path.resolve(relativeFilePath);
+    const fileErrors = lintFile(filePath, relativeFilePath.slice(0, -4));
+
+    if (fileErrors.some((err) => err.severity > 1).length) exitCode = 1;
+
+    errors[filePath] = fileErrors;
+    return errors;
+  }, {});
+
+  printErrors(errors);
```
rwjblue: Should we print the errors for each file that we lint (instead of waiting for all the linting to be finished before priting any output)?  I'm thinking of the non `--json` case (but I think its probably fine in both cases).  Specifically, if I lint a large directory that takes a while (e.g. a few thousand templates) I'd like to see any linting errors that have been discovered as quickly as possible.

bardzusny: Interesting point - definitely makes sense for non-`--json` case, when printing JSON though: what we're printing is just one object, with arrays of errors under keys matching appropriate file paths.

bardzusny:
Thanks for review @rwjblue - still very much WIP, as you can see. One thing that interests me is what you think about first point in the checklist:

>Node.JS version support.

Right now script happily runs under v4, despite some ES6 constructs being used.

We can either leave it as-is, which is:
* less hassle (d'uh ;) ).

Or still re-write it to ES5, which would mean:
* consistency with basically all of currently existing code.
* no complications when running linter against added code (related to the first point - didn't look at that one yet to be honest, but we definitely don't want separate `.eslintrc` just for `bin/` directory, I think).

rwjblue:
No need to support < Node 4 (I've already dropped Node 0.12 from .travis.yml).

bin/ember-template-lint
rwjblue: Ya, confirm. --json has not benefit from emitting early, but not also should not suffer (and having one code path is nice...)

test/unit/bin/ember-template-lint-test.js
```diff
@@ -0,0 +1,149 @@
...
+          assert.equal(stderr, '', 'doesn\'t print anything to error output');
+          done();
+        });
+      });
+    });
+
+    describe('given path to non-existing file', function() {
+      it('should exit without error and any console output', function(done) {
+        exec('../../../bin/ember-template-lint app/templates/application-1.hbs', {
+          cwd: './test/fixtures/with-errors'
```
bardzusny: These repetitive, relative paths can look a bit edgy - I don't really have the mindset of striving for DRY-ness/sparkling code in tests, too much. They're here, they work, they're simple and readable. Comments/suggestions welcome.

test/unit/bin/ember-template-lint-test.js
```diff
@@ -0,0 +1,149 @@
...
+        });
+      });
+    });
+
+    describe('given path to single file with errors', function() {
+      it('should print errors', function(done) {
+        exec('../../../bin/ember-template-lint app/templates/application.hbs', {
+          cwd: './test/fixtures/with-errors'
+        }, function(err, stdout, stderr) {
+          assert.ok(err, 'exits with error');
```
bardzusny: Could be described as `exits with non-zero exit code`, and we could check this exit code as well - do you think it's worth the hassle?

test/unit/bin/ember-template-lint-test.js
```diff
@@ -0,0 +1,149 @@
...
+          assert.ok(stdout, 'prints errors to standard output');
+          assert.equal(stderr, '', 'doesn\'t print anything to error output');
+          done();
+        });
+      });
+    });
+
+    describe('given wildcard path resolving to single file', function() {
+      it('should print errors', function(done) {
+        exec('../../../bin/ember-template-lint app/templates/*', {
```
bardzusny: We still have only one file there. Also, we're always giving just one full/wildcard path - no test cases for more than this given at once. Should we cover that, too? I opted to try to re-use current fixtures, giving us just one template file each to operate on.

test/unit/bin/ember-template-lint-test.js
```diff
@@ -0,0 +1,149 @@
...
+      });
+    });
+
+    describe('given directory path', function() {
+      it('should print errors', function(done) {
+        exec('../../../bin/ember-template-lint app', {
+          cwd: './test/fixtures/with-errors'
+        }, function(err, stdout, stderr) {
+          assert.ok(err, 'exits with error');
+          assert.ok(stdout, 'print linting errors to standard output');
```
bardzusny: Or maybe just `prints errors to standard output` - I'll try to pass these descriptions through some grammar nazi before merging. 😆

bardzusny:
Hello @rwjblue - thinking that  we're close to finishing here 🎊 . Feel invited to take a look, most importantly whether you see anything left to do above minor fixes/README mention.

Re. https://github.com/rwjblue/ember-template-lint/pull/152#discussion_r97345249 - I had a go at quick implementation. Quick sum-up, let me know if you see any holes in my reasoning:
* `--json` case - early-emitting single object while ensuring that _valid_, nicely formatted JSON string is outputted is something that I found both problematic (indendation/trailing commas/etc.) and not very practical (because if we stop linting process mid-way we won't get parse-able JSON string, anyway - it won't have closing object brace, to start with).
* non-`--json` case: Got functional version, happy to push it.  Holding myself back, because it seems like a non-blocker that would still mean a bit of divergence/inconsistency in usage of `errorsToMessages` static method (different paths for `--json`/no-`--json`).

rwjblue:
This looks great, thanks for keeping at it! 

I think the only thing left is some README treatment...

rwjblue:
ZOMG, this is looking very good!!!!

@bardzusny - I'm happy with where this is, let me know when you are ready and we will land it...

bardzusny:
@rwjblue happy that we waited a little bit, some missing things only now came to my mind :) quite confident that we're good to go now - let's 🚀 .

rwjblue:
👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #154, opened 2016-12-12
Added CLI support

Dejavs:
It provides CLI support. By executing CLI command:

./node_modules/ember-template-lint/bin/lint
Above command finds all *.hbs files and checking them for linting errors. Output is similar to ember-cli-template-lint. It is possible to adjust searching directory by adding -d option.

I attached output log file which shows how this command works on real project.
[out.txt](https://github.com/rwjblue/ember-template-lint/files/644936/out.txt)


Dejavs:
The advantege of putting the whole code into lib directory is that it's under eslint control ;)

rwjblue:
Thank you for working on this! We landed an alternate implementation in https://github.com/rwjblue/ember-template-lint/pull/152.

Would you mind taking a look and opening issues for anything that you spot that doesn't meet your needs?

PR #157, opened 2017-01-06
bare string config failed for rupee symbol html entitiy

MatrixZ:


test/unit/rules/lint-bare-strings-test.js
```diff
@@ -51,6 +51,11 @@ generateRuleTests({
     },
 
     {
+      config: ['&#8377;'],
```
rwjblue: This is invalid config I believe, please change the config to be `['₹']` instead...

rwjblue:
Thank you! More tests are always good. 😃

PR #159, opened 2017-01-10
Fix indentation check on text nodes after a whitespace eating expression

fhrbek:
This is a proposal how to (partially) fix https://github.com/rwjblue/ember-template-lint/issues/91 concerning text nodes after expressions with a tilde.

I added a test that would fail without the changes in this PR. At the same time, however, one of the existing tests got broken:

The following structure was in the "good" samples.
```
{{#if isMorning}}  Good morning
{{else if isAfternoon}}  Good afternoon
{{else}}
  Good night
{{/if}}
```

Is that really correct? It looks weird to me. I removed this test from the good samples and moved a variation of it to the bad samples.

I don't insist on my proposal being correct, please take it as an inspiration for fixing the problem.

Note: I also tried to investigate other cases when there's something else than a text node in this position, however, it looks like the parser eats the white space and reports the element being on a wrong line which is something the linter can hardly do anything about - the fix needs to be done elsewhere. I also noticed that even the ```ember build``` command reports wrong line numbers if there are syntax errors at similar positions - completely unrelated to linting.

### DIFF
--- a/lib/rules/lint-block-indentation.js
+++ b/lib/rules/lint-block-indentation.js
@@ -206,7 +206,8 @@ module.exports = function(addonContext) {
         continue;
       }
 
-      var childStartColumn;
+      var childStartColumn = child.loc.start.column;
+
       // sanitize text node starting column info
       if (AstNodeInfo.isTextNode(child)) {
         // TextNode's include leading newlines, but those newlines do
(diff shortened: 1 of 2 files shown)


fhrbek:
@rwjblue Could you please take a look at this? It would be so nice to get all the "tilde" related problems fixed.

rwjblue:
Yep, sorry for the delay!  This change looks good to me, and makes the logic a bit simpler to grok too...

rwjblue:
I'd love your help continuing down the list of whitespace control issues if you have the time / energy for it.

Not sure if this helps, but I have found the recently added support for glimmer AST in `astexplorer.net` to be super helpful for tracking things down in the AST.  Checkout https://astexplorer.net/#/nQKEYXxIIi for an example...

fhrbek:
@rwjblue Thanks for merging this! Yes, I'd like to help burning the list of whitespace control issues because now we have a lot of non-existent linting problems reported in our code base.

The AST explorer with glimmer looks useful. There I can easily demonstrate the problem we have in the linter. Here are three code snippets to compare:

1. https://astexplorer.net/#/5aw8u0mBd8
2. https://astexplorer.net/#/1xrnXWf6Gm/1
3. https://astexplorer.net/#/cWuWTh9X0J

If you pick the start of the ```<p>``` element, you will see the difference there: with whitespace eaten (1), the element is reported as starting at line 3, column 18. This is IMHO wrong. Without the "tilde" (2), the reported location is line 4, column 2 which is correct. I believe that this is the root cause of the linter problem we observe.
The last snippet (3) is a "manual transformation" of (1) and it is parsed correctly. It looks like there's a problem parsing ```{{else if condition~}}```.

Questions:
1. What can we do about glimmer reporting wrong locations? Where to start?
2. Where should we continue this discussion? (this PR is closed so I don't think it's the right place)

PR #164, opened 2017-01-18
Add tests for 'else if' with whitespace eaten

fhrbek:
This PR adds tests for ```if-else if``` structures with whitespace eaten after ```else if``` statement.
For the tests to pass, glimmer with https://github.com/tildeio/glimmer/pull/390 is required.

Note: #163 is a prerequisite for this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "dependencies": {
     "chalk": "^1.1.3",
     "resolve": "^1.1.3",
-    "glimmer-engine": "^0.19.1",
+    "glimmer-engine": "^0.21.2",
     "lodash": "^4.11.1"
   },
   "bugs": {
--- a/test/unit/helpers/ast-node-info-test.js
+++ b/test/unit/helpers/ast-node-info-test.js
@@ -1,5 +1,5 @@
 var assert = require('power-assert');
-var preprocess = require('glimmer-engine/dist/node_modules/glimmer-syntax').preprocess;
+var preprocess = require('glimmer-engine/dist/node_modules/@glimmer/syntax').preprocess;
 var AstNodeInfo = require('../../../lib/helpers/ast-node-info');
 
 describe('isImgElement', function() {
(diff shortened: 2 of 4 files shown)


rwjblue:
Just to confirm: you have tested with that PR and all tests here (including the new one) pass now?

rwjblue:
Because based on https://github.com/rwjblue/ember-template-lint/pull/163, I was thinking that we had some changes that were needed before we could test against glimmer-engine@0.20+

fhrbek:
@rwjblue Yes, in the environment I set up, all tests (both existing and new ones) passed. All I had to do was link the new glimmer and change paths to glimmer-syntax in ```test/unit/helpers/ast-node-info-test.js``` and ```test/unit/helpers/is-interactive-element-test.js``` (not part of this PR).

fhrbek:
@rwjblue Yes, checking #163, the tests fail on the path to glimmer-syntax. That's exactly what I had to solve to be able to run my (and other) tests and the whole linter on my real project.

rwjblue:
@fhrbek - Since you have already fixed, would you mind adding those commits to this branch so I can just merge?

fhrbek:
@rwjblue Do you mean adding the glimmer upgrade & related path changes here? If so, to which version of glimmer?

rwjblue:
Yes, to glimmer-engine@0.20.2 (which I just published and includes the fix to the `{{else if foo~}}` case).

fhrbek:
@rwjblue gonna do that right away...

Just to verify, while I'm preparing the commit: you mean to this PR, and you mean 0.21.2, not 0.20.2 (see https://github.com/tildeio/glimmer/commit/a426bfebf950ba569f56e2fb10751f01cc86ca08), right?

rwjblue:
Whoops, yes, sorry about that

fhrbek:
Done, all tests pass.

PR #169, opened 2017-01-24
Add missing syntax highlighting

ryanponce:
Syntax highlighting was missing for the `unused-block-params` rule examples.


rwjblue:
Thank you!

PR #175, opened 2017-02-09
Fix config with plugins

fhrbek:
This PR adds two commits:
- tests that fail with certain plugin configuration, or after repeated loading of the config
- fixes for the newly added tests

Some more background:

The first problem is more or less cosmetic - with a plugin configuration, there may be some non-default rules added (implemented in the plugin source). Those rules are evaluated as invalid as they are not part of the default set of rules. In this PR, rules are validated against loaded rules (which consist of the default rules and additional rules declared in plugins) rather than the default rules, which fixes the problem. Without the fix, a meaningless warning is reported.

The second problem is more serious - if the configuration is loaded more than once within a single runtime, it can lose important settings, making the whole configuration unusable. For example, in the first pass, ```loadedRules``` and ```loadedConfigurations``` keys are added as root configuration keys. Those keys contain important information, especially when there are some plugins declared. During the second pass, however, those keys are moved under ```rules``` as they are considered as invalid in the configuration root, and then they are replaced with newly constructed keys. At the same time, the ```plugins``` key is restructured during the first pass (from an array to a hash), and totally destroyed in the second pass (the hash is processed like an array, which means no plugins are found). As a result of combining the two problems described above, the newly constructed keys lose essential information (```plugins``` is empty and ```loadedRules``` does not contain rules coming from the plugins). The final consequence of this is that none of the declared plugins are effective.

Note: Even in a situation when the configuration is loaded with the ```require``` function from a file, multiple attempts to load such a configuration results in a broken configuration due to caching of the required files (the cached structures are changed by the configuration loader and then served again).

The second problem is fixed by adding a boolean flag to the configuration (```._processed```), saying that the configuration has already been processed and should simply be returned on all subsequent attempts to get the configuration.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -203,7 +203,7 @@ function validateRules(config, options) {
   var invalidKeysFound = [];
 
   for (var key in config.rules) {
-    if (!defaultRules[key]) {
+    if (!config.loadedRules[key]) {
       invalidKeysFound.push(key);
     }
   }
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome job tracking this down! Thank you!!

PR #179, opened 2017-02-24
Loading config: fix order of extending configurations

### LINKED ISSUE #178: Wrong order of loading extended/custom rules configurations
Currently such `.template-lintrc.js` file:
```javascript
module.exports = {
  extends: 'recommended',

  rules: {
    'invalid-interactive': false
  }
};
```

Will result in `invalid-interactive` rule being turned on, due to its presence within `recommended` config that we're extending: https://github.com/rwjblue/ember-template-lint/blob/master/lib/config/recommended.js#L12 .

This seems to be due to wrong order of loading configurations: it's extended configurations that are assigned to custom rules setup, thus overriding them. [Seemingly faulty line in `lib/get-config.js` file.](https://github.com/rwjblue/ember-template-lint/blob/master/lib/get-config.js#L179).

bardzusny:
- [x] Fix `with-plugins-overwriting` fixture to have correct order of extended configurations, corresponding to expected result within acceptance spec test case (https://github.com/rwjblue/ember-template-lint/blob/master/test/acceptance-test.js#L469).
- [x] Fix clear typo/mistake in `get-config` unit test: don't try to override `bare-strings` `false` rule config value with another `false`.
- [x] `processExtends` function within `lib/get-config.js`: extend extended configurations instead of extending custom config _with_ them (:mindblown: - fixes two test cases described above).

Fixes #178 .

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -176,7 +176,7 @@ function processExtends(config, options) {
 
         if (configuration.rules) {
 
-          config.rules = assign(config.rules, configuration.rules);
+          config.rules = assign(configuration.rules, config.rules);
 
         } else {
 
--- a/test/fixtures/with-plugins-overwriting/.template-lintrc.js
+++ b/test/fixtures/with-plugins-overwriting/.template-lintrc.js
@@ -4,8 +4,8 @@ module.exports = {
     './plugins/plugin2'
   ],
   extends: [
-    'plugin1:enable-inline-component',
     'plugin2:disable-inline-component',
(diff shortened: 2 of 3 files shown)


bardzusny:
Note: this PR is rather hastily done, with me just wanting to use latest `master` revision in the project I'm currently working on. Let's confirm twice that I didn't misunderstand anything before proceeding (i.e. merging).

rwjblue:
Awesome, thank you!

Issue #167, opened 2017-01-23
Document plugins


rwjblue:
Plugin support landed in https://github.com/rwjblue/ember-template-lint/pull/138 (with some additional discussion in https://github.com/rwjblue/ember-template-lint/issues/136). 

We need to document this new functionality in the README.

fhrbek:
@ppcano would you have some time to provide docs for the plugins? We'd appreciate a new release with #137, and the plugins look promising to us too 👍. I'd like to help but I have too little knowledge of the plugin support (all I could do would be an attempt to reverse engineer #138 which would probably be less than ideal).

ppcano:
@fhrbek We left pending to make public a subset of APIs in order people could easier build their plugins.  

@rwjblue I think the documentation should describe both aspects: How the plugin feature works and what public APIs are available.

fhrbek:
@rwjblue @ppcano I tried out the plugin support on our project. It looks really good, however, I encountered issues using it until I found and fixed the cause - see #175. Feel free to review the PR and ask questions if something is unclear. All I can say is that only with this PR I'm able to use my first plugin in our project successfully.

mchat:
We'd also like to use plugin support in our project. It looks like the main blocker at this point is documentation. I don't have a lot of context myself, but would be more than willing to draft an initial version of documentation and open a PR if that helps unblock the release!

From what I understand, the documentation would need to include:

- [ ] how to set up a plugin in your config file (based on proposal in #136)
  - [ ] inline plugins
  - [ ] strings that resolve to plugin files
  - [ ] options for plugins
    - [ ] `name`
    - [ ] `rules`
    - [ ] `configurations`
  - [ ] extending plugins
  - [ ] using rules defined in plugins
- [ ] public APIs that you can use in your plugin rules
  - [ ] `buildPlugin` to set up base rule
  - [ ] `visitors` for each node
  - [ ] logging errors from visitors

It sounds like there's still some discussion to be had around exactly what APIs should be exposed for plugins to use. I just added what seems like the most basic ones to this list, but I'm very new to this project and may be completely off base here.

If this outline seems reasonable, I can draft some documentation and open a PR.

ppcano:
@mchat Thanks for moving this forward. Unfortunately now, I don't have much time, but I will try to help.

Most of the built-in plugins are using the following APIs. 

```
var AstNodeInfo = require('../helpers/ast-node-info');
var buildPlugin = require('./base');
```

[...]

PR #180, opened 2017-02-25
Document plugin support

mchat:
For #167. First draft of documentation for the plugin support system. I'm still very new to this project, so would appreciate a careful review to ensure that I've correctly understood / communicated how the plugin system and rules APIs work.

- [x] how to set up a plugin in your config file (based on proposal in #136)
  - [x] inline plugins
  - [x] strings that resolve to plugin files
  - [x] options for plugins
    - [x] `name`
    - [x] `rules`
    - [x] `configurations`
  - [x] extending plugins
  - [x] using rules defined in plugins
- [x] public APIs that you can use in your plugin rules
  - [x] `buildPlugin` to set up base rule
  - [x] AST node helpers 

I also updated the usage documentation in the README — it looks like the API has changed for that method since the documentation was first written.

cc: @ppcano 

@rwjblue the documentation for plugins was a bit long, so I created a `docs` directory. Would you prefer to keep all documentation in the main README instead?

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -16,7 +16,7 @@ Each plugin object can include these properties.
 
   Object that defines new rules.
   Each key represents the name of the rule that is defined.
-  Each value should be a string that represents a relative file path to the rule definition. See [Rule APIs](#rule-apis) for more detail.
+  Each value should be a Rule object. See [Rule APIs](#rule-apis) for more detail.
 
 * `configurations` -- `Object`
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,0 +1,267 @@
+## Plugin Support
+
+You can customize the linter with rules that are more specific to your use case with the `ember-template-lint` plugin system.
+
+Plugins can define new rules and set up default configurations that can be extended.
+
+### Adding Plugins to Your Configuration
+
+In order to enable a plugin, you must add it to the `plugins` key in your configuration file.
+
+`plugins` accepts an array of either objects, or strings that resolve to files that export plugin objects.
+
+#### Define a plugin inline
+
+```javascript
+// .template-lintrc.js
+
+module.exports = {
+  plugins: [
+    {
+      // Define rules for this plugin. Each path should map to a plugin rule
+      rules: [
+        'disallow-inline-components': require('./lib/template-lint-rules/disallow-inline-components'),
+        'another-custom-rule': require('.lib/template-lint-rules/another-custom-rule')
+      ],
+
+      // Define configurations for this plugin that can be extended by the base configuration
+      configurations: [
+        'no-inline': {
+            rules: {
+              'disallow-inline-components': true
+            }
+        }
+      ]
+    },
+  ],
+
+  extends: [
+    // Extend configuration defined in the inline plugin
+    'no-inline'
+  ],
+
+  rules: {
+    'bare-strings': true,
+
+    // Use a rule defined in the inline plugin
+    'another-custom-rule': true
+  }
+}
+```
+
+#### Define a plugin based on a relative path
+```javascript
+// .template-lintrc.js
+
+module.exports = {
+  plugins: [
+    // Define a plugin that is exported elsewhere (i.e. a third-party plugin)
+    './plugins/some-local-plugin',
+    'some-npm-package/third-party-plugin',
+  ],
+
+  extends: [
+    'recommended',
+
+    // Extend configuration defined in plugins
+    'some-local-plugin:recommended',
+    'third-party-plugin:strict'
+  ],
+
+  rules: {
+    'bare-strings': true,
+
+    // Use a rule defined in a plugin
+    'rule-defined-in-some-local-plugin': true
+  }
+}
+
+```
+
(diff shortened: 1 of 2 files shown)


docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+
+You can customize the linter with rules that are more specific to your use case with the `ember-template-lint` plugin system.
+
+Plugins can define new rules and set up default configurations that can be extended.
+
+### Adding Plugins to Your Configuration
+
+In order to enable a plugin, you must add it to the `plugins` key in your configuration file.
+
+`plugins` accepts an array of either objects, or strings that resolve to files that export plugin objects.
```
rwjblue: Small language tweak suggestion:

> `plugins` accepts an array of either plugin objects or strings to be passed to `require` which resolve to plugin objects.

docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### Define a plugin inline
+
+```javascript
+// .template-lintrc.js
+
+module.exports = {
+  plugins: [
+    {
```
rwjblue: Lets define the `Plugin` interface because its the same interface in both the inline definition and relative path definition examples.

docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+      // Define rules for this plugin. Each path should map to a plugin rule
+      rules: [
```
rwjblue: I think this should be `{` (since `rules` is not an array its an object).

docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+        'disallow-inline-components': require('./lib/template-lint-rules/disallow-inline-components'),
+        'another-custom-rule': require('.lib/template-lint-rules/another-custom-rule')
+      ],
+
+      // Define configurations for this plugin that can be extended by the base configuration
+      configurations: [
```
rwjblue: s/[/{/

docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+
+  Object that defines new configurations that can be extended.
+  Each key represents the name of the configuration object.
+  Each value should be a configuration object, that can include the [same properties as the base config object](../README.md#configuration-keys) in any `.template-lintrc.js` -- i.e. `rules`, `extends`, `ignore`, etc.
+
+### Rule APIs
+
+Every rule defined by a plugin can use these public APIs defined by `ember-template-lint`.
+
+### Building a rule object
```
rwjblue: Should probably be an h4 (####)

docs/plugins.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### Define a plugin inline
+
+```javascript
+// .template-lintrc.js
+
+module.exports = {
+  plugins: [
+    {
+      // Define rules for this plugin. Each path should map to a plugin rule
```
fhrbek: I think that even an inline plugin has to define the ```name``` attribute.

docs/plugins.md
mchat: Ah, good catch! Bad copy/paste on my part.

docs/plugins.md
mchat: What do you mean by `define the Plugin interface`? Do you mean add an extra section of documentation? Something more like a docstring?

docs/plugins.md
mchat: Sounds good! I'll also make `AST Node Helpers` an h4 to match.

docs/plugins.md
rwjblue: The doc currently explains inline plugins differently from the types of plugins that you would require, and they are not really any different.  Each entry in the plugins list must conform to a general "interface" (in a general hand wavy sense). A "Plugin" is a specific thing, that has an expected set of properties.

E.g. it has a `name`, `rules` (Object whose keys are the rule names and the values are the rules), `configurations` (Object whose keys are the config names, and the values are the config values), etc.

If we can define what a "Plugin" is, then the sections for `Define a plugin inline` and `Define a plugin based on a relative path` are really no different.

mchat: That makes sense! Thank you for clarifying.

I just pushed a commit that reorders / clarifies the documentation. Let me know if that's what you're looking for, or if you have any other feedback on how to make the plugin interface more understandable.

docs/plugins.md
```diff
@@ -0,0 +1,229 @@
...
+
+* `name` -- `string` (required)
+
+  The name of the plugin. Will be used to namespace any configuration objects.
+
+* `rules` -- `Object`
+
+  Object that defines new rules.
+  Each key represents the name of the rule that is defined.
+  Each value should be a string that represents a relative file path to the rule definition. See [Rule APIs](#rule-apis) for more detail.
```
rwjblue: The values are actual "Rules" objects (not a string).  Commonly it is done as in the example below:

```js
rules: {
  'some-rule-name': require('./rules/some-rule-name')
}
```

mchat: Ah, good catch! Just updated.

rwjblue:
Thank you!!!!!

eventualbuddha:
Anything holding up a release including this? Let us know if we can help at all!

Issue #120, opened 2016-10-20
Incorrect indentation on line starting with &nbsp;


reidab:
Linting a template with a segment like the following incorrectly flags the indentation:

``` handlebars
{{#if foo}}
  &nbsp;bar
{{/if}}
```

with

```
block-indentation: Incorrect indentation for `   bar
` beginning at L2:C0. Expected `   bar
` to be at an indentation of 2 but was found at 3. (thedyrt-ui/templates/components/at-trip):
`{{#if foo}}
  &nbsp;bar
{{/if}}`
```

duggiefresh:
What would be the expectation here? Thanks!

reidab:
It's looking like ember-template-lint is treating the `&nbsp;` string as a space for the purpose of calculating source code indentation, instead of treating the `&` as the first character of that line's source.

I'd expect both of the following to pass linting, but currently only the second one does:

``` handlebars
{{#if foo}}
  &nbsp;bar
{{/if}}
```
``` handlebars
{{#if foo}}
  &rarr;bar
{{/if}}
```

rwjblue:
Agreed.  I am unsure why `&nbsp;` is being treated differently than `&rarr;` off the top of my head.

rwjblue:
@reidab - Can you double check if this is still an issue, and open a failing test case for this (basically like [this one](https://github.com/rwjblue/ember-template-lint/blob/master/test/unit/rules/lint-block-indentation-test.js#L89-L92)) if it is still an issue?

reidab:
Confirmed that it's still happening on the latest version and submitted failing tests in #184.

rwjblue:
We can fix it now! We need to update our usage of @glimmer/compiler to use the new `mode: 'codemod'` and it will opt-out of converting HTML entities to their underlying thing. 

See https://github.com/glimmerjs/glimmer-vm/pull/938 for background on that new parsing mode.

PR #184, opened 2017-02-27
Add failing indentation tests for lines starting with &nbsp;

reidab:
For #120:

- `&nbsp;` followed by text (`&nbsp;Hello`) seems to interpret the `&nbsp;` as a space and fails: `Expected '   Hello\n' to be at an indentation of 2 but was found at 3."`
- `&nbsp;` followed by HTML (`&nbsp;<div></div>`) seems to interpret the `&nbsp;` as six separate whitespace characters and fails: `Expected '<div>' to be at an indentation of 2 but was found at 8.`

### DIFF
--- a/test/unit/rules/lint-block-indentation-test.js
+++ b/test/unit/rules/lint-block-indentation-test.js
@@ -74,6 +74,8 @@ generateRuleTests({
       '{{/if}}',
     ].join('\n'),
     ['<div class="multi"', '     id="lines"></div>'].join('\n'),
+    ['{{#if foo}}', '  &nbsp;Hello', '{{/if}}'].join('\n'),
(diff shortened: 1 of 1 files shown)


rwjblue:
Oh, no. This is tricky indeed...

MelSumner:
@rwjblue folks shouldn't be indenting paragraph text with whitespace, they should be using CSS (attach a class to the element and put padding in it). It's the whole point of the rule.

rwjblue:
@MelSumner - That is **definitely** not the point of the `block-indentation` rule 😉, it _may_ be the point of the other rule you recently worked on (`no-whitespace-for-layout` landed in #819) but I actually don't think it is. Using spaces in a template to format them so that they can be read by a human is a perfectly valid thing to do. Either way, it is important to keep in mind that each rule should work independently and that any number of combinations of rules will be used together.

Issue #187, opened 2017-03-06
configure link-rel to validate both noopener and noreferrer


ingride:
Firefox [does not currently support "noopener"](https://bugzilla.mozilla.org/show_bug.cgi?id=1222516) and this will be addressed on version 52, that will be released end of March '17.  

To prevent this [vulnerability](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/), best practice would be to validate both `noopener` and `noreferrer` .

From a conversation with @rwjblue we have decided to make this rule configurable to be able to validate `both` while maintaining the backward compatibility and allow current users to address this at their own pace.

rwjblue:
👋 @ingride 😃 

I think this is done now (and the recommended config uses the newer `strict` config by default)

PR #190, opened 2017-03-23
validate both noopener and noreferrer by extending the rule config

ingride:
Fix for #187 .

This PR extends the configuration options for `link-rel-noopener` to allow enforcing validation for both `noopener` **and** `noreferrer` on  rule.

Backward compatibility is preserved by maintaining the boolean values `true` to enable and `false` to disable. 

To configure the new validation

``` javascript
module.exports = {
  rules: {
    'link-rel-noopener': [true, 'strict']
  }
}; 
```

cc @paddyobrien  and @GavinJoyce

### DIFF
--- a/lib/rules/lint-link-rel-noopener.js
+++ b/lib/rules/lint-link-rel-noopener.js
@@ -37,7 +37,7 @@ module.exports = function(addonContext) {
 
     var errorMessage = 'The bare-strings rule accepts one of the following values.\n ' +
       '  * boolean - `true` to enable / `false` to disable\n' +
-      '  * array -- [`true`, `strict`] to enable validation for both noopener AND noreferrer\n' +
+      '  * string -- `strict` to enable validation for both noopener AND noreferrer\n' +
       '\nYou specified `' + JSON.stringify(config) + '`';
 
     switch (configType) {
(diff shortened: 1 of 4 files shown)


lib/rules/lint-link-rel-noopener.js
```diff
@@ -18,9 +18,45 @@ var buildPlugin = require('./base');
...
+
 module.exports = function(addonContext) {
   var LinkRelNoopener = buildPlugin(addonContext, 'link-rel-noopener');
 
+  LinkRelNoopener.prototype.parseConfig = function(config) {
+    var configType = typeof config;
+
+    var errorMessage = 'The bare-strings rule accepts one of the following values.\n ' +
+      '  * boolean - `true` to enable / `false` to disable\n' +
+      '  * array -- [`true`, `strict`] to enable validation for both noopener AND noreferrer\n' +
```
rwjblue: Can we swap to `true`, `false`, and `'strict'` (e.g. use a single string for `"strict"` mode).

ingride: done

rwjblue: Implementation looks good, but this error message needs to be updated I believe.

rwjblue:
Thank you!!

PR #192, opened 2017-04-19
block rule missing error

kellyselden:
test case


rwjblue:
@kellyselden - Just to confirm, you are expecting that `<span>` to be required to be indented but it isn't, is that right?

kellyselden:
@rwjblue Yes.

kellyselden:
This can be merged now, it was fixed via a previous PR.

rwjblue:
Thanks @kellyselden!

PR #193, opened 2017-04-19
block rule missing error

kellyselden:
test case

### DIFF
--- a/test/unit/rules/lint-block-indentation-test.js
+++ b/test/unit/rules/lint-block-indentation-test.js
@@ -520,6 +520,21 @@ generateRuleTests({
     },
     {
       template: [
+        '{{#foo-bar}}',
+        '{{/foo-bar}} content'
+      ].join('\n'),
+
+      result: {
+        rule: 'block-indentation',
+        message: 'Incorrect indentation for `{{! bad comment }}` beginning at L1:C6. Expected `{{! bad comment }}` to be at an indentation of 2 but was found at 6.',
+        moduleId: 'layout.hbs',
+        source: '<div> {{! bad comment }}\n  {{foo-bar}}\n</div>',
+        line: 1,
+        column: 6
+      }
+    },
+    {
+      template: [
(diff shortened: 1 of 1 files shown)


rwjblue:
@kellyselden - Just to confirm, you are saying that in `{{/foo-bar}} content` that ` content` is incorrectly indented?

kellyselden:
@rwjblue Yes.

rwjblue:
This behavior is intentional. We only validate that the first child on a given line is intended properly.

If we forbid this behavior (as you are suggesting by this failing test) the following common cases would now become failures:

```hbs
{{#if something}}
  {{foo}} - {{bar}}
{{/if}}
```

```hbs
<p>
  <span>{{name}}</span>
</p>
```

kellyselden:
I'm sorry but I don't follow. Your examples don't seem related to me. A deeper example would be

```hbs
{{#a-b}}
  {{#b-c}}
  {{/b-c}} x
{{/a-b}}

{{#a-b}}
  {{#b-c}}
  {{/b-c}}
  x
{{/a-b}}
```

It seems to me that that can be linted without failing your above examples. Maybe it could even fall under a different rule `exclusive-opening-and-closing-lines` or something.

rwjblue:
Hmm. I was responding to the generic concept of “trailing content”, but you make a good point. There is something conceptually different between extra content in the same line as a block end vs “having multiple items in the same line”....

Not sure if it should be the same rule or a new one. 🤔

rwjblue:
Hmm. I was responding to the generic concept of “trailing content”, but you make a good point. There is something conceptually different between extra content in the same line as a block end vs “having multiple items in the same line”....

Not sure if it should be the same rule or a new one. 🤔

dcyriller:
Do you mind if I close this PR?

It looks like it highlights an existing issue. But some time has passed since it was opened. And it is now clear that this issue would be better solved on Prettier side?

PR #195, opened 2017-04-22
Adding deep link to rule.

alvincrespo:
When I was reading the docs, I wanted to know where this rule was coming from. Adding the link helps clarify this is a rule part of the library.


rwjblue:
Thank you!

PR #201, opened 2017-05-05
Replace "glimmer-engine" with "@glimmer/syntax" and "@glimmer/compiler"

Turbo87:
/cc @rwjblue

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const precompile = require('glimmer-engine').precompile;
+const precompile = require('@glimmer/compiler').precompile;
 const getConfig = require('./get-config');
 
 const WARNING_SEVERITY = 1;
--- a/lib/rules/deprecations/lint-deprecated-inline-view-helper.js
+++ b/lib/rules/deprecations/lint-deprecated-inline-view-helper.js
@@ -72,6 +72,9 @@ module.exports = function(addonContext) {
         },
 
         MustacheStatement(node) {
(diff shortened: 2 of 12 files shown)


Turbo87:
looks like `power-assert` is doing some strange stuff... I'm working on replacing it with Chai instead

rwjblue:
hehe, `power-assert` was a fun experiment, but its definitely finnicky...

Turbo87:
yeah, I'm used to this style of assertions from https://docs.pytest.org/ but it certainly has its downsides...

Turbo87:
looks like we have two options here:

- v0.21.x and v0.22.x of the `@glimmer` packages are not compatible with Node 4 because the use destructuring

- v0.23.0-beta is apparently breaking a lot of tests here 😞

rwjblue:
We can fix 0.22 and publish an update

Turbo87:
> We can fix 0.22 and publish an update

It looks like we would need to touch `emberjs-build` for that. I've tried to build `glimmer-vm@0.22.0` but it just failed and I have no idea how to make it work since it seems to throw several issues somewhere in lerna... I give up

Turbo87:
I've had another look at this and it seems that `node.loc.start` is often wrong which seems to be causing most of the test failures...

rwjblue:
😭

rwjblue:
Just rebased and fixed things up.  Waiting for CI to agree, but will land once finished.

Turbo87:
🎉  thanks!

PR #211, opened 2017-05-26
Rule disallow `{{unless}}`

### LINKED ISSUE #30: Rule: Disallow `{{unless}}`
Good:

``` hbs
{{#unless somePredicate}}
{{/unless}}

<div class="{{unless foo 'no-foo'}}"></div>
```

Bad:

``` hbs
{{#unless (and foo bar)}}
{{/unless}}
```

``` hbs
{{#unless foo}}
{{else}}
{{/unless}}
```

``` hbs
{{#if foo}}
{{else unless bar}}
{{/if}}
```

Mostly because it is impossible to reason about....

duggiefresh:
Fixes #30

### DIFF
--- a/lib/rules/lint-unless-helper.js
+++ b/lib/rules/lint-unless-helper.js
@@ -1,0 +1,84 @@
+'use strict';
+
+const buildPlugin = require('./base');
+
+const messages = {
+  followingElseBlock: 'Using an {{else}} block with {{unless}} should be avoided.',
+  asElseUnlessBlock: 'Using an `{{else unless}}` block should be avoided.',
+  withHelper: 'Using {{unless}} in combination with other helpers should be avoided.'
+};
+
+module.exports = function(addonContext) {
+  return class Plugin extends buildPlugin(addonContext, 'unless-helper') {
+    visitors() {
+      return {
+        MustacheStatement(node) {
+          if (node.params[0].path) {
+            this._withHelper(node);
+          }
+        },
+
+        BlockStatement(node) {
+          let nodeInverse = node.inverse;
+          let nodePathOriginal = node.path.original;
+
+          if (nodeInverse) {
+            if (nodePathOriginal === 'unless') {
+              if (nodeInverse.body[0].path) {
+                this._followingElseIfBlock(nodeInverse);
+              } else {
+                this._followingElseBlock(nodeInverse);
+              }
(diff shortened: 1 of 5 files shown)


duggiefresh:
I'll update the README as well. Just wanted to get some 👀 on this beforehand. Thanks!

lib/rules/lint-unless-helper.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const buildPlugin = require('./base');
+
+const messages = {
+  followingElseBlock: 'Invoking an `{{else}}` helper with the `unless` helper is not recommended',
```
rwjblue: Small verbiage tweak here:

> Using an `{{else}}` block with `{{unless}}` should be avoided.

lib/rules/lint-unless-helper.js
```diff
@@ -0,0 +1,85 @@
...
+  followingElseIfBlock: 'Invoking an `{{else if}}` helper with the `unless` helper is not recommended',
```
rwjblue: I think we can use the `followingElseBlock` for this one (e.g. we don't need two different messages in this case).

lib/rules/lint-unless-helper.js
```diff
@@ -0,0 +1,85 @@
...
+  asElseUnlessBlock: 'Invoking an `{{else unless}}` helper is not recommended',
+  withHelper: 'Invoking an `{{unless}}` helper with another template helper is not recommended'
```
rwjblue: Small verbiage tweak here:

> Using `{{unless}}` in combination with other helpers should be avoided.

lib/rules/lint-unless-helper.js
```diff
@@ -0,0 +1,85 @@
...
+};
+
+function logMessage(context, message, loc, actual) {
```
rwjblue: If you put this on your `class` below, you can call it via `this.logMessage()` and not pass in `context`.  I don't have a significant preference, just curious if you do 😄...

duggiefresh:
Thanks for the review @rwjblue! Updated per request, and I added a basic entry in the README

rwjblue:
OK, I rebased over the bonkers changes that I have been doing on master, and updated this to account for those tweaks.

Thank you @duggiefresh!!!

Issue #215, opened 2017-06-12
Issue with block-indentation rule when used with UTF8-BOM encoding


initram:
For better or worse we are stuck with having byte order mark (BOM) in our template files. 

This causes issues with the block-indentation rule, as it thinks the BOM is a 1 char indent, resulting in errors reported for the closing tag of the first tag in the file. I would hope that it is not a big issue ignoring the BOM or at least stop treating it as whitespace.

rwjblue:
I'll admit to being completely ignorant of BOM and its associated issues, can you submit a failing test PR with an example of how a template would look? You can add one to [this array](https://github.com/rwjblue/ember-template-lint/blob/master/test/unit/rules/lint-block-indentation-test.js#L11-L15), which should fail (in the same way your app templates fail) so we can try to take a look at fixing...

rwjblue:
Resolved by #216

PR #216, opened 2017-06-12
Added "byte order mark" test

initram:
Added failing test to unit test suite for block-indentation rule. The test input starts with a "byte order mark".

See issue #215

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -96,7 +96,7 @@ class Linter {
     try {
       precompile(stripBom(options.source), {
         moduleName: options.moduleId,
-        rawSource: options.source,
+        rawSource: stripBom(options.source),
         plugins: {
           ast: this.buildASTPlugins(pluginConfig)
         }
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,6 +2,7 @@
 
 const precompile = require('glimmer-engine').precompile;
 const getConfig = require('./get-config');
+const stripBom = require('strip-bom');
 
 const WARNING_SEVERITY = 1;
 const ERROR_SEVERITY = 2;
@@ -93,7 +94,7 @@ class Linter {
     };
 
     try {
-      precompile(options.source, {
+      precompile(stripBom(options.source), {
         moduleName: options.moduleId,
         rawSource: options.source,
         plugins: {
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "mocha-eslint": "^2.0.2",
     "mocha-only-detector": "^0.1.0",
     "rimraf": "^2.5.2",
+    "strip-bom": "^3.0.0",
     "testem": "^1.7.0"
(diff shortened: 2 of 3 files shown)


initram:
It is all a bit difficult to get setup correctly within javascript, but the committed changes causes the error that i saw when using a real file with the UTF8-BOM (which is different from the char I have inserted in the test). You can read more here: https://stackoverflow.com/questions/17879198/adding-utf-8-bom-to-string-blob/27975629

If you want a real file with UTF8-BOM, then I have attached one here.
[bom.txt](https://github.com/rwjblue/ember-template-lint/files/1068600/bom.txt)

rwjblue:
OK, turns out that I think this should be "easy" (famous last words 😝).  Support was added to `ember-cli-htmlbars` to strip the BOM in https://github.com/ember-cli/ember-cli-htmlbars/commit/99e25764322f32a69ef2dc1964eac8330c57a3b8.  We just need to do roughly the same here before we pass the template through the rest of the system.

Specifically:

* Add `strip-bom` to dependencies in `package.json`
* Use `strip-bom` [here](https://github.com/rwjblue/ember-template-lint/blob/aa4084331b4ea59963254dd3ac599ea72f58489b/lib/index.js#L96) before passing `options.source` into `precompile`.

initram:
Tried to implement the changes you suggested, but I not sure if I also have tro strip the BOM from the "rawSource". I did not when making the changes, but looking at the code, it looks like it may be needed to provide the correct source code snippet for errors. Let me know what you think.

rwjblue:
Aha! Yes, you are correct! 

To prove this, let's add a  "bad example" to the tests by putting a space after the BOM and expecting the line/column info is correct in the error message (I suspect that it will not be).

initram:
Not sure if you get notified when I add commits to the pull request, but I now also strip the BOM from the rawSource and added a test to ensure that it works.

rwjblue:
The case that concerns me is:

```hbs
\uFEFF {{if foo}}
{{/if}}
```

In this case, I believe the error would be reported as being incorrect indentation for `{{if foo}}` starting at col 1, but in reality (in the template) its actually col 2.

---

Fixing this to ensure the error locations when on the same line as the BOM marker are correct, would require modifying the `block-indentation` rule (instead of just stripping BOM out before passing it in).

rwjblue:
Perhaps, we should land this, then follow-up with a failing test PR with my snippet above. That way we can get you unblocked, and continue iterating...

initram:
The BOM should not be considered a character, so it is correct to say that the position is in column 1 in that example. The BOM is more like meta-data that describes how the file should be interpreted, and not a real character, which is also why we should make sure it does not show up in the source text after reading the file. I see no further work required for this case.

rwjblue:
Thank you @initram !

Issue #214, opened 2017-06-12
Option to ignore template files within a folder


akashdsouza:
Instead of matching the moduleId to the exact string, it would be helpful if all templates within a specific directory could be ignored. Something like `app/templates/directory/*`

This could be useful for engines wherein the option is available to ignore all template files within the engine

rwjblue:
Seems like a reasonable capability to add, mind taking a shot at it?

akashdsouza:
Sure

rwjblue:
This was fixed in #218.

PR #218, opened 2017-06-14
Ignore using minimatch expressions

josex2r:
This pull-request adds the ability to use minimatch expressions to ignore files.
Example:

```javascript
// .template-lintrc
{
  ignore: ['some/path/**/*.hbs']
}
```

I think it solves the issue #214

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -10,7 +10,6 @@ const defaultConfigurations = require('./config/index');
 
 const KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore', 'plugins'];
 
-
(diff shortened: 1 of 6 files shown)


lib/get-config.js
```diff
@@ -198,6 +199,10 @@ function processExtends(config, options) {
 
 }
 
+function processIgnores(config) {
+  config.ignore = config.ignore.map((pattern) => Minimatch(pattern));
```
rwjblue: Can you update this to call `new Minimatch(pattern)`?

rwjblue:
OK, I rebased and fixed that nit-picky comment I left.  Will land once green.

josex2r:
Thanks :-)

PR #245, opened 2017-07-14
Feature eslint style formatter

Exelord:
Hi!
I've just redesigned the formatter to look like Eslint. It also shows the count of all errors. Now fixing and debugging is even more simple are more readable.

@rwjblue What do you think about it?
screen shot 2017-07-15 at 16 13 52

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -13,7 +13,7 @@ function printErrors(errors) {
     console.log(JSON.stringify(errors, null, 2));
   } else {
     Object.keys(errors).forEach(filePath => {
-      console.log(Linter.errorsToMessages(errors[filePath]));
+      console.log(Linter.errorsToMessages(filePath, errors[filePath]));
     });
   }
 }
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -16,9 +16,15 @@ function printErrors(errors) {
     let count = 0;
 
     Object.keys(errors).forEach(filePath => {
-      let fileErrors = errors[filePath];
-      console.log(Linter.errorsToMessages(filePath, fileErrors));
+      let options = {};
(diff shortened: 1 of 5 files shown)


bin/ember-template-lint.js
```diff
@@ -7,14 +7,27 @@ var path = require('path');
...
   } else {
+    let count = 0;
+
     Object.keys(errors).forEach(filePath => {
-      console.log(Linter.errorsToMessages(errors[filePath]));
+      let options = {};
+      let fileErrors = errors[filePath] || [];
+      count += fileErrors.length;
+
+      if (process.argv.includes('--desc')) {
```
rwjblue: Can we tweak the flag? Perhaps `--verbose`?

Exelord:
@rwjblue Thank you :) I applied the feedback but I can't figure out why the tests are failing under node 4 :/

Exelord:
Do you think about integrating this plugin with ESLint itself?

rwjblue:
> Do you think about integrating this plugin with ESLint itself?

No, I haven't really considered that.

bin/ember-template-lint.js
```diff
@@ -7,14 +7,27 @@ var path = require('path');
...
   } else {
+    let count = 0;
+
     Object.keys(errors).forEach(filePath => {
-      console.log(Linter.errorsToMessages(errors[filePath]));
+      let options = {};
+      let fileErrors = errors[filePath] || [];
+      count += fileErrors.length;
+
+      if (process.argv.includes('--verbose')) {
```
rwjblue: Per http://node.green/#ES2016-features-Array-prototype-includes-Array-prototype-includes, Node 4 does not have a `Array.prototype.includes` method.

Exelord:
My bad :/ Thank you for pointing out :) Fixed.

rwjblue:
Thank you!!!

PR #249, opened 2017-07-18
Move strip-bom to dependencies (un-break atom plugin)

bendemboski:
It's used from lib/index.js, so it needs to be a dependency, not a devDependency.

I suspect this hasn't come up before since `strip-bom` is a dependency of `ember-cli-htmlbars`, so Ember projects that do any template linting would already have `strip-bom` installed for other reasons. But I just tried updating the [linter-ember-template](https://github.com/binhums/linter-ember-template) plugin to the latest `ember-template-lint` and it breaks because of this missing dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "lodash": "^4.11.1",
     "minimatch": "^3.0.4",
     "node-glob": "^1.2.0",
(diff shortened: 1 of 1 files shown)


rwjblue:
Dang. Sorry about this.

PR #252, opened 2017-08-05
[Patch] New rule to lint duplicate attributes

siva-sundar:
Throws warning when a duplicate attribute is detected in any of the following

- ElementNode
- MustacheStatement
- BlockStatement
- SubExpression


lib/rules/index.js
```diff
@@ -18,5 +18,6 @@ module.exports = {
   'invalid-interactive': require('./lint-invalid-interactive'),
   'style-concatenation': require('./lint-style-concatenation'),
   'simple-unless': require('./lint-simple-unless'),
-  'unused-block-params': require('./lint-unused-block-params')
+  'unused-block-params': require('./lint-unused-block-params'),
+  'duplicate-attributes': require('./lint-duplicate-attributes')
```
rwjblue: Let's call this "no-duplicate-attributes"?

lib/rules/lint-duplicate-attributes.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const Rule = require('./base');
+
+const logDuplicateAttributes = function ({ attributes = [], indentifier, type }) {
```
rwjblue: Can't use destructuring (we still support node 4).

lib/rules/lint-duplicate-attributes.js
```diff
@@ -0,0 +1,48 @@
...
+
+  attributes.forEach((attribute, index) => {
+    for (currentIndex = index; currentIndex < length; currentIndex++) {
+      currentAttribute = attributes[currentIndex];
+      if (attribute[indentifier] === currentAttribute[indentifier] && index !== currentIndex) {
+        this.log({
+          message: `Duplicate attributes found in the ${type}.`,
+          line: currentAttribute.loc && currentAttribute.loc.start.line,
+          column: currentAttribute.loc && currentAttribute.loc.start.column,
+          source: this.sourceForNode(currentAttribute)
```
rwjblue: I think the source should be the node containing the attribute (not the attribute itself), what do you think?

lib/rules/index.js
siva-sundar: Sure 👍

lib/rules/lint-duplicate-attributes.js
siva-sundar: Yeah, that will enable us to locate the code much quicker. Can we log as I have mentioned below ?

`Duplicate attribute 'name' found in the SubExpression. `
`(hash name=name age=age name=name) `

siva-sundar:
Should this be`recommended` rule ?

rwjblue:
Yes, it is a good idea to make it recommended but doing so would be a breaking change. 

Let's land and release the new rule, then create an issue to update the recommended config for our next major version bump.

lib/rules/lint-duplicate-attributes.js
siva-sundar: I have added logs as mentioned above. Is it ok?

siva-sundar:
👍

siva-sundar:
I have done the changes you have mentioned. Let me know if anything else needs to be done

lib/rules/lint-duplicate-attributes.js
rwjblue: Yes, this looks good. 😸

rwjblue:
Thank you!

PR #255, opened 2017-08-10
block-indentation: allow single-line blocks with inverses

bendemboski:
This rule previous allowed

```handlebars
{{#if foo}}foo{{/if}}
```

but not

```handlebars
{{#if foo}}foo{{else}}bar{{/if}}
```

Now it allows both (plus uglier forms with else ifs and whatnot).

Personally I'm not a big fan of a newly-allowed syntax that this enables:

```handlebars
{{#if man}}man{{else if this}}this{{else if is}}is{{else}}ugly{{/if}}
```

But I don't think that has any _indentation_ issues, so if we don't like that we should write another `no-super-fugly-if-blocks` rule.

### DIFF
--- a/lib/rules/lint-block-indentation.js
+++ b/lib/rules/lint-block-indentation.js
@@ -113,8 +113,14 @@ module.exports = class BlockIndentation extends Rule {
     };
   }
 
-  /*eslint no-unused-expressions: 0*/
   process(node) {
+    // Nodes that start and end on the same line cannot have any indentation
+    // issues (since the column of the start block was already checked in the
+    // parent's validateBlockChildren())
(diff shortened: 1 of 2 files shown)


lib/rules/lint-block-indentation.js
```diff
@@ -113,8 +113,14 @@ module.exports = class BlockIndentation extends Rule {
     };
   }
 
-  /*eslint no-unused-expressions: 0*/
```
bendemboski: This doesn't appear to be needed anymore, so I removed it.

rwjblue:
> we should write another no-super-fugly-if-blocks rule

LOL, agreed

rwjblue:
Thank you!

PR #259, opened 2017-08-19
Patch lint attribute spacing

siva-sundar:
Implements https://github.com/rwjblue/ember-template-lint/issues/28

### DIFF
--- a/lib/rules/lint-attribute-indentation.js
+++ b/lib/rules/lint-attribute-indentation.js
@@ -1,0 +1,301 @@
+'use strict';
+
+const Rule = require('./base');
+
+const getWhiteSpaceLength = function(statement) {
+  let whiteSpace = statement.match(/^\s+/) || [];
+  return (whiteSpace[0] || '').length;
+};
+
+module.exports = class AttributeSpacing extends Rule {
+
+  getBlockParamStartLoc(node) {
+    let actual, expected;
+    let programStartLoc = node.program.loc.start;
+    let nodeStart = node.loc.start;
+    if (node.params.length === 0 && node.hash.pairs.length === 0) {
+      expected = {
+        line: nodeStart.line + 1,
+        column: nodeStart.column
+      };
+      if (nodeStart.line === programStartLoc.line) {
+        let displayName = `{{#${node.path.original}`;
+        /*
+        {{#employee-details as |employee|}}
+          {{employee.name}}
+        {{/employee-details}}
+        */
+        actual = {
+          line: nodeStart.line,
+          column: displayName.length
+        };
+      } else {
+        /*
+        {{#employee-details
+          as |employee|}}
+          {{employee.name}}
+        {{/employee-details}}
+        */
+        let source = this.sourceForNode({
+          loc: {
+            start: {
+              line: programStartLoc.line,
+              /*
+                Setting column as 0, to get the entire line for calculating the start column.
+                For instance, the below will result in `                    as |employee|}}` => 19
+                  `{{#employee-details
+                                      as |employee|}}
+                    {{employee.name}}
+                  {{/employee-details}}
+                  `
+              */
+              column: 0
+            },
+            end: programStartLoc
+          }
+        });
+        //
+        /*
+          Determining the actual column by calculating the whitespace length
+            `{{#employee-details
+                                as |employee|}} => 19
+              {{employee.name}}
+            {{/employee-details}}`
+        */
+        actual = {
+          line: programStartLoc.line,
+          column: getWhiteSpaceLength(source)
+        };
+      }
+
+    } else {
+      let paramOrHashPairEndLoc;
+
+      if (node.params.length) {
+        /*
+        The block form may contain only positional params as below
+        {{#employee-details
+          firstName
+          lastName
+          age
+        as |employee|}}
+          {{employee.fullName}}
+        {{/employee-details}}
+        */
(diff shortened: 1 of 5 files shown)


siva-sundar:
Implements https://github.com/rwjblue/ember-template-lint/issues/28

lib/rules/lint-attribute-indentation.js
```diff
@@ -0,0 +1,301 @@
...
+
+const getWhiteSpaceLength = function(statement) {
+  let whiteSpace = statement.match(/^\s+/) || [];
+  return (whiteSpace[0] || '').length;
+};
+
+module.exports = class AttributeSpacing extends Rule {
+
+  getBlockParamStartLoc(node) {
+    let actual, expected;
```
siva-sundar: Is there a direct way to get the location of block params?

lib/rules/lint-attribute-indentation.js
```diff
@@ -0,0 +1,301 @@
...
+        {{contact-details
+          age
+          firstName=firstName
+          fullName=fullName
+        }}
+    */
+    let expectedColumnStart = node.loc.start.column + 2; //params should be after 2 positions from component start node
+    let expectedLineStart = node.loc.start.line + 1;
+
+    this.iterateParams(node.params, 'positional', expectedLineStart, expectedColumnStart, node);
```
siva-sundar: Current implementation enforces the `positional params` to follow the same indentation rule.

rwjblue: Seems good to me (I think)....

test/unit/rules/lint-attribute-indentation-test.js
```diff
@@ -0,0 +1,165 @@
...
+    //Block form with no params
+    '{{#contact-details' + '\n' +
+    'as |contact|}}' + '\n' +
+    '  {{contact.fullName}}' + '\n' +
+    '{{/contact-details}}'
+  ],
+
+  bad: [{
+      //MustacheStatement
+    template: '{{contact-details firstName=firstName lastName=lastName}}',
```
rwjblue: I think the rule needs to be a bit nuanced here.  I think this should be a "good" example (non-block single line invocation with less than 80 chars).

test/unit/rules/lint-attribute-indentation-test.js
```diff
@@ -0,0 +1,165 @@
...
+      'column': 55,
+      'line': 1,
+      'message': 'Incorrect indentation of close curly braces \'}}\' for the component \'{{contact-details}}\' beginning at L1:C55. Expected to be indentation at L2:C0 with an of 0 but was found at 55',
+      'moduleId': 'layout.hbs',
+      'rule': 'attribute-indentation',
+      'source': '{{contact-details firstName=firstName lastName=lastName}}'
+    }]
+  }, {
+      //MustacheStatement with a positional param
+    template: '{{contact-details firstName lastName=lastName}}',
```
rwjblue: Same as above, I think this should be a good example.

siva-sundar:
Thank you. I have added an option to configure the length of the open invocation statement. Also made the changes you have mentioned.

test/unit/rules/lint-attribute-indentation-test.js
```diff
@@ -0,0 +1,182 @@
...
+    '{{contact-details' + '\n' +
+    '  firstName=firstName' + '\n' +
+    '  lastName=lastName' + '\n' +
+    '}}',
+    //Non Block form with no params
+    '{{contact-details}}',
+    //Non Block form within 70 characters
+    {
+      config: {
+        'open-invocation-max-len': 70
```
rwjblue: Just to be clear, this one would pass even without custom config right?

siva-sundar: Yes, I'll change the example to have more than the default length(80 chars).

siva-sundar:
I have added few more test cases. Can you take a look at it?

rwjblue:
Awesome, thank you very much for all of your hard work on this!

test/unit/rules/lint-attribute-indentation-test.js
```diff
@@ -0,0 +1,183 @@
...
+      config: {
+        'open-invocation-max-len': 120
+      },
+      template: '{{contact-details firstName=firstName lastName=lastName avatarUrl=avatarUrl age=age address=address phoneNo=phoneNo}}'
+    },
+    //Open-invocation with multiple lines.
+    '{{contact-details' + '\n' +
+    '  firstName=firstName' + '\n' +
+    '  lastName=lastName' + '\n' +
+    '}}',
```
amk221: Will attribute-indentation support not having to have }} on its own line. I like to indent over multiple lines, but retain the }} at the end

siva-sundar: We don't have an option to specify this style of indentation. Perhaps, we could create a sub rule?

cc: @rwjblue

amk221: Worth a though, as I think this should be supported to:
```handlebars
{{my-component
  foo='bar'
  baz='qux'
  my-attr=(component 'my-other-component'
    foo='bar'
    baz='qux')}}
```

rwjblue: for single line invocation, closing at the end of the last line seems ok (and we can update this rule to allow it).

@amk221 - Can you create an issue to track this?

PR #262, opened 2017-09-06
Adding rule to check for template-length

psbanka:
Hi there, I'd love to be able to add a linting rule (ideally as a plugin) to have the linter fail when a template is over a certain size (assuming it should be broken up into pieces) or BELOW a certain size (assuming it should be inlined into a component).

~For some reason, disabling tests through instruction comments does
not seem to work and unit tests failing. If anyone has any idea what might be wrong here, I'd certainly appreciate the assistance. I wonder if we are not getting the configuration change because I'm using the `Program` hook to look for template-length?~

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -15,9 +15,6 @@ module.exports = {
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
     'simple-unless': true,
-    'unused-block-params': true,
-    'template-length': {
-      max: 200
-    },
+    'unused-block-params': true
   }
 };
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -15,6 +15,9 @@ module.exports = {
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
     'simple-unless': true,
-    'unused-block-params': true
+    'unused-block-params': true,
+    'template-length': {
+      max: 200
+    },
   }
 };
--- a/lib/rules/lint-template-length.js
+++ b/lib/rules/lint-template-length.js
@@ -5,12 +5,12 @@
 
  The following values are valid configuration:
 
-   * boolean -- `true` for enabled (same as `unix`) / `false` for disabled
+   * boolean -- `true` for enabled (defaults of max: 200 / min: 5) / `false` for disabling this rule
    * object --
      * max {number} - the longest a template should be without failing a test (assuming the
-       right thing to do would be to split the template up into pieces.
+       right thing to do would be to split the template up into pieces).
      * min {number} - the shortest a template should be without failing a test (assuming the
-       right thing to do would be to inline the template.
+       right thing to do would be to inline the template).
  */
 
 const Rule = require('./base');
--- a/lib/rules/lint-template-length.js
+++ b/lib/rules/lint-template-length.js
@@ -1,0 +1,93 @@
+'use strict';
+
+/*
+ Enforce template size constraints
+
+ The following values are valid configuration:
+
+   * boolean -- `true` for enabled (same as `unix`) / `false` for disabled
+   * object --
+     * max {number} - the longest a template should be without failing a test (assuming the
+       right thing to do would be to split the template up into pieces.
+     * min {number} - the shortest a template should be without failing a test (assuming the
+       right thing to do would be to inline the template.
+ */
+
+const Rule = require('./base');
+
+const MAX_LENGTH = 100;
+const MIN_LENGTH = 5;
+
+const DEFAULT_CONFIG = {
+  max: MAX_LENGTH,
+  min: MIN_LENGTH,
+};
+
+function isValidConfigObjectFormat(config) {
+  for (let key in config) {
+    let value = config[key];
+    let valueIsInteger = Number.isInteger(value);
+
+    if (key === 'min' && !valueIsInteger) {
+      return false;
+    } else if (key === 'max' && !valueIsInteger) {
+      return false;
+    }
+  }
+  return true;
+}
+
+module.exports = class LengthValidation extends Rule {
+  parseConfig(config) {
+    let configType = typeof config;
+
+    let errorMessage = 'The template-length rule accepts one of the following values.\n ' +
+      '  * boolean - `true` to enable / `false` to disable\n' +
+      '  * object -- An object with the following keys:' +
+      '    * `min` -- Minimum length of a template ' +
+      '    * `max` -- Maximum length of a template. ' +
+      '\nYou specified `' + JSON.stringify(config) + '`';
+
+    switch (configType) {
+    case 'boolean':
+      // if `true` use `DEFAULT_CONFIG`
+      return config ? DEFAULT_CONFIG : {};
+    case 'object':
+      if (isValidConfigObjectFormat(config)) {
+        return config;
+      } else {
+        throw new Error(errorMessage);
+      }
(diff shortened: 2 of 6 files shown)


lib/rules/lint-template-length.js
```diff
@@ -0,0 +1,93 @@
...
+      return {};
+    default:
+      throw new Error(errorMessage);
+    }
+  }
+
+  visitor() {
+    return {
+      Program(node) {
+        if (node.loc.start.line <= 1 && node.loc.start.column === 0 && node.loc.end.column === 0) {
```
rwjblue: What is this guarding?

lib/rules/lint-template-length.js
```diff
@@ -0,0 +1,93 @@
...
+          const lines = node.body.map((chunk) => {
```
rwjblue: In general, you can use `this.source` in the rule itself to get an array of all the lines (e.g. you shouldn't need to map over `node.body` at all)

lib/rules/lint-template-length.js
```diff
@@ -0,0 +1,93 @@
...
+            const source = this.sourceForNode(chunk);
+            return source.split('\n').length;
+          }).reduce((sum, value) => sum + value, 0);
+          if (this.config.max && lines > this.config.max) {
+            this.log({
+              message: `Template length of ${lines} exceeds ${this.config.max}`,
+              line: node.loc.start.line,
+              column: node.loc.start.column
+            });
+          } else if (this.config.min && lines < this.config.min) {
```
rwjblue: I'm not sure how you envision this part working.  Specifically, with the eslint-plugin-hbs you created even the inlined templates would be linted and there would be no way to avoid this from logging...

lib/config/recommended.js
```diff
@@ -15,6 +15,9 @@ module.exports = {
     'style-concatenation': true,
     'deprecated-inline-view-helper': true,
     'simple-unless': true,
-    'unused-block-params': true
+    'unused-block-params': true,
+    'template-length': {
```
rwjblue: In general, we don't update recommended ruleset when adding new rules. Adding new rules (or changing config) for recommended is often times a breaking change, and I don't think we are quite ready for another major version bump.  

Can you remove the changes to `lib/config/recommended.js`?

lib/rules/lint-template-length.js
```diff
@@ -0,0 +1,93 @@
...
+    case 'undefined':
+      return {};
+    default:
+      throw new Error(errorMessage);
+    }
+  }
+
+  visitor() {
+    return {
+      Program(node) {
```
rwjblue: The way the comment disabling system works is that when it visits a comment node which contains an template lint directive, it updates the current configuration such that the rule is disabled. When the rule is disabled, none of the other visitors are called.

The "rule can be disabled via comment" tests are failing because of the way this visitor is structured. Specifically, it is not possible for the visitor system to "see" a comment node that disables the rule before the `Program` visitor here is invoked (which would immediately call `this.log`).  Thankfully, the fix is relatively straightforward though:

```js
visitor() {
  return {
    Program: {
      exit(node) {
        // move your implementation here, since it is in the exit handler it will no longer
        // be called if the rule has been disabled by any inline comments within the file
      }
    }
  }
}
```

rwjblue:
Personal opinion time:

I totally agree that we _should_ strive to use logical separation of concerns which in many cases will naturally lead to this rule being satisfied, but I am not personally in favor of _using_ this rule in projects that I work on. In my experience the idea that a given file is "bad" because it is long and to satisfy linting you have to split things up leads to very hard to reason about codebases. In my experience, splitting up a single logical unit into multiple files simply because it is "long" roughly means that to find where some HTML snippet is you have to solve a murder mystery. 😝 

---

That personal thought aside, I am totally in favor of _having_ this rule in the codebase.  Thank you for working on it!

lib/rules/lint-template-length.js
psbanka: This is a hack that is no longer necessary when using `this.source`. I seemed that this callback would get called multiple times per template-file, and I wanted to ensure that it was being called for the entire file and not for a snippet. This if guards that we are starting on the first line and column. Also guards that we are ending on the first column on the last line, which only works if every template in the system ends in a newline.

lib/config/recommended.js
psbanka: Yes, makes sense. Will do!

lib/rules/lint-template-length.js
psbanka: Hmm. You make a good point. I wanted to be able to throw a lint error when an `.hbs` template file was too short to cause a developer to inline their template instead. I hadn't thought that the same rules are going to apply to inline templates since `eslint-plugin-hbs` uses the identical configuration.

jamesarosen: Could `eslint-plugin-hbs` support `ifInline` and `unlessInline` options for each rule?

lib/rules/lint-template-length.js
psbanka: Actually, this is still needed, otherwise you get stuff like this:
```
/Users/psbanka/dev/copy-tango/app/pods/onboarding/step1/template.hbs
  5:19  error  Template length of 129 exceeds 100  template-length
  30:86  error  Template length of 129 exceeds 100  template-length
  18:106  error  Template length of 129 exceeds 100  template-length
  41:101  error  Template length of 129 exceeds 100  template-length
  9:30  error  Template length of 129 exceeds 100  template-length
... etc
  -:-  error  Template length of 129 exceeds 100  template-length
```

So I left the `node.loc.start.line <= 1 && node.loc.start.column === 0 ` guard in, but checking for last column == 0 seemed to be unnecessary.

psbanka:
@rwjblue – you're welcome, and thanks for all the help 😃 ! I think I caught all the items you mentioned in review and I added documentation (including adding a note as to why you might *not* want to use this check, drawing heavily from your "personal opinion." 

I hope this looks good ow; glad to fix anything else you see.

lib/rules/lint-template-length.js
rwjblue: FWIW, the reason for this is that every block statement also has a `Program`.  What you are doing with that conditional is roughly ensuring that we only check the _first_ `Program` node.

A simpler (and easier to grok for our future selves) check might be:

```js
visitor() {
  let seenOuterProgram = false;
  return {
    Program: {
      exit(node) {
        if (seenOuterProgram === false) { return; }
        // current logic goes here  
        seenOuterProgram = true;
      }
    }
  }
}
```

lib/rules/lint-template-length.js
```diff
@@ -0,0 +1,95 @@
...
+
+  visitor() {
+    return {
+      Program: {
+
+        // implementation goes here in exit(): in the exit handler, the rule will not
+        // be called if it has been disabled by any inline comments within the file.
+
+        exit(node) {
+          if (node.loc.start.line > 1 || node.loc.start.column > 0) { return; }
```
rwjblue: I commented in the thread over here https://github.com/rwjblue/ember-template-lint/pull/262#discussion_r137914213, and suggested another possible work around.

Would you mind trying that alternate guard instead?  If it doesn't work, we can land it with the guard you have in place now (but we should add a comment to explain why we need the specific guard)...

lib/rules/lint-template-length.js
psbanka: Nice. Thanks!

PR #268, opened 2017-09-14
rule lint-simple-unless throws an error when template contains an emp…

JoelWAnna:
…ty inverse block

/template.hbs
  -:-  error  Cannot read property 'path' of undefined  undefined


JoelWAnna:
Obviously I don't need to have the empty inverse block, but it was not obvious from the error where the problem was

rwjblue:
Thank you!

PR #273, opened 2017-10-01
Check indentation of else block too

### LINKED ISSUE #272: indentation is not tested correctly for the else block
The following code complains about the "baz" text not being indented correctly, but it does not complain that the `{{bar}}` block is indented incorrectly.
```
{{#if foo}}
  foo
{{else}}
    {{#bar}}
    baz
    {{/bar}}
{{/if}}
```

initram:
Make sure that the else-block is also returned from the childrenFor, so that they can be tested for indentation.

Fixes #272

### DIFF
--- a/lib/helpers/ast-node-info.js
+++ b/lib/helpers/ast-node-info.js
@@ -66,6 +66,9 @@ function childrenFor(node) {
     return node.body;
   }
   if (node.type === 'BlockStatement') {
+    if(node.inverse){
+      return node.program.body.concat(node.inverse.body);
+    }
     return node.program.body;
   }
   if (node.type === 'ElementNode') {
--- a/lib/rules/lint-block-indentation.js
+++ b/lib/rules/lint-block-indentation.js
@@ -158,7 +158,7 @@ module.exports = class BlockIndentation extends Rule {
   }
 
   validateBlockChildren(node) {
-    let children = AstNodeInfo.childrenFor(node);
+    let children = AstNodeInfo.childrenFor(node).filter(x => !x._isElseIfBlock);
 
     if (!AstNodeInfo.hasChildren(node)) {
       return;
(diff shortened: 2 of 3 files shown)


lib/helpers/ast-node-info.js
```diff
@@ -66,6 +66,9 @@ function childrenFor(node) {
     return node.body;
   }
   if (node.type === 'BlockStatement') {
+    if(node.inverse){
+      return node.program.body.concat(node.inverse.body);
```
rwjblue: I believe that its possible to only have an inverse (and therefore `node.program.body` is null/undefined).

Perhaps something like this might work:

```js
return [].concat(
  node.program && node.program.body,
  node.inverse && node.inverse.body
).filter(Boolean);
```

initram: Do you have an example of this? then I would be able to add that to the test suite.

rwjblue: I think this will do it:

```hbs
{{#unless}}
{{/unless}}
```

initram: Okay, I will try that tomorrow. 
It looks like it should only be inverse that is nullable here: https://github.com/glimmerjs/glimmer-vm/blob/e8eb555378f2699676cf39a11fc6dc96af847b58/packages/%40glimmer/syntax/lib/types/nodes.ts#L47

rwjblue: ya, agreed

I just did a quick test on astexplorer.net and it seems that program is populated there also, so maybe it just isn't possible?

initram: I have not been able to produce a BlockStatement without a `program`. I do not this it is possible, which would also make sense when looking at the typescript definition.
But if you still prefer a null check for the possibility that it may sneak in later, then just let me know.

rwjblue: Nope, seems fine to me. Thank you for digging into this!

rwjblue:
Will merge and release shortly....

PR #279, opened 2017-10-03
Make unit ember-template-lint-tes run on windows

initram:
Previously the unit tests that ran ember-template-lint using node failed on windows, due to using forward slashes instead of backslashes as path seperator. But also it is not possible to just execute a .js file in windows. You have to provide a "real" executable to run, in this case node.

So I changed the uses of exec to use execFile with node as the first argument. This was not strictly needed, but execFile has better support for passing arguments and is a bit more efficient, so I saw it as an opportunity to change that as well.

This requires node to be in the PATH, so that it can be executed without an absolute path. Not sure if that is an issue for running it on OSX and Linux.

### DIFF
--- a/test/unit/bin/ember-template-lint-test.js
+++ b/test/unit/bin/ember-template-lint-test.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const exec = require('child_process').exec;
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you!

PR #286, opened 2017-10-06
Allow colgroup and caption in tables

### LINKED ISSUE #285: Changes for table-group rule.
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table) the allowed contents of a `<table>` are:

> In this order:
> 1. an optional `<caption>` element,
> 2. zero or more `<colgroup>` elements,
> 3. an optional `<thead>` element,
> 4. either one of the following:
>   i) zero or more `<tbody>` elements
>   ii) one or more `<tr>` elements
> 5. an optional `<tfoot>` element

We are intentionally disallowing the option in 4 above (e.g. we require `<tr>` to be in `<tbody>`), but we still need to account for:

- [x] `<caption>`
- [x] `<colgroup>`

We also need to enforce the ordering that is mentioned in the spec...

t-sauer:
Fixes #285 

Basically only `thead`, `tbody`, `tfoot`, `caption` and `colgroup` are allowed children for tables.

### DIFF
--- a/lib/rules/lint-table-groups.js
+++ b/lib/rules/lint-table-groups.js
@@ -3,22 +3,14 @@
 const Rule = require('./base');
 const message = 'Tables must have a table group (thead, tbody or tfoot).';
 
-const TABLE_GROUPS = ['thead', 'tbody', 'tfoot'];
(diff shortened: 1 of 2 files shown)


rwjblue:
👍 - Thank you for adding these!

PR #287, opened 2017-10-06
assert Program node type called

kellyselden:
@rwjblue Shouldn't this be called? It seems it would be needed to implement a final eol assertion.


kellyselden:
Looks like it is being called...

rwjblue:
Thank you! Free tests!! 🎉

PR #288, opened 2017-10-06
add eol-last rule

kellyselden:
It's interesting though that the inline comments to disable rules won't work for this, because the `Program` node type runs before any inline comment processing to disable rules. Is there a way around this?

### DIFF
--- a/lib/rules/lint-eol-last.js
+++ b/lib/rules/lint-eol-last.js
@@ -1,0 +1,90 @@
+'use strict';
+
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+
+module.exports = class EolLast extends Rule {
+  parseConfig(config) {
+    let configType = typeof config;
+
+    switch (configType) {
+    case 'boolean':
+      if (!config) {
+        return false;
+      }
+      break;
+    case 'string':
+      if (['always', 'never'].indexOf(config) > -1) {
+        return config;
+      }
+      break;
+    case 'undefined':
+      return false;
+    }
+
+    let errorMessage = createErrorMessage(this.ruleName, [
+      '  * "always" - enforces that files end with a newline',
+      '  * "never" - enforces that files do not end with a newline'
+    ], config);
+
+    throw new Error(errorMessage);
+  }
+
+  visitor() {
+    return {
+      Program: {
+
+        // implementation goes here in exit(): in the exit handler, the rule will not
+        // be called if it has been disabled by any inline comments within the file.
+
+        exit(node) {
+          let line = node.loc.start.line;
+          let column = node.loc.start.column;
+
+          // yielded content makes it in here too
+          //
+          // `{{#my-component}}
+          //   test
+          // {{/my-component}}`
+          //
+          // becomes
+          //
+          // `
+          //   test
+          // `
+          //
+          // check for that case
+          if (line !== 1 || column !== 0) {
+            return;
+          }
+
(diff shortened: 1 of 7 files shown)


kellyselden:
Nvm, it turns out it works with inline comments.

rwjblue:
Take a look at https://github.com/rwjblue/ember-template-lint/pull/262#discussion_r137540375 for details.

kellyselden:
@rwjblue updated.

lib/rules/lint-eol-last.js
```diff
@@ -0,0 +1,59 @@
...
+          // {{/my-component}}`
+          //
+          // becomes
+          //
+          // `
+          //   test
+          // `
+          //
+          // check for that case
+          if (line !== 1 || column !== 0) {
```
rwjblue: I'd rather keep a closure scope variable that says we have processed and noop.

Basically like https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/lint-template-length.js#L77.

lib/rules/lint-eol-last.js
```diff
@@ -0,0 +1,59 @@
...
+            return;
+          }
+
+          let source = this.sourceForNode(node);
+
+          let lastChar = source[source.length - 1];
+          let message;
+          switch (this.config) {
+          case 'always':
+            if (lastChar !== '\n') {
```
rwjblue: Shouldn't we allow the EOL char(s) to be up to the user?  Some folks want `\r\n` (two chars, I know), and others want `\n`.  We _could_ leave this up to the `linebreak-style` rule maybe?

I ultimately want the default line ending to be auto-detected from `.editorconfig` (see https://github.com/rwjblue/ember-template-lint/issues/284), but that doesn't need to block this rule...

lib/rules/lint-eol-last.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+const Rule = require('./base');
+
+module.exports = class EolLast extends Rule {
```
rwjblue: This should implement a `parseConfig` so that we can enforce valid config before beginning linting any templates.

lib/rules/lint-eol-last.js
kellyselden: I tried that but it won't work. The inner content goes through the `exit` first, so it's the reverse order of what I need.

lib/rules/lint-eol-last.js
kellyselden: This is already agnostic, whether the EOL is '\r\n' or '\n', the last char of the string will always be '\n'. (except old mac where it is only '\r'.......)

lib/rules/lint-eol-last.js
kellyselden: done with tests

test/unit/rules/lint-eol-last-test.js
```diff
@@ -0,0 +1,75 @@
...
+        moduleId: 'layout.hbs',
+        message: 'template cannot end with newline',
+        line: 1,
+        column: 0,
+        source: 'test\n'
+      }
+    }
+  ],
+
+  error: [
```
kellyselden: new category for testing hard errors in configs

kellyselden:
This is ready for review/merge again. I wasn't able to get https://github.com/rwjblue/ember-template-lint/pull/288#discussion_r143318707 working.

rwjblue:
Thanks @kellyselden!

PR #299, opened 2017-10-09
block rule missing error part 3

kellyselden:
top-level blocks are not checked for indentation.


duggiefresh:
@kellyselden Hola, thanks for the test case. Do you plan on working on this? Otherwise, I can take ownership of this. Thanks!

kellyselden:
No plans on my end. Thanks!

PR #302, opened 2017-10-13
Top level block indentation

duggiefresh:
For top level blocks, the starting block helper should be located
at line 1 and column 0.

Closes #299


rwjblue:
Thank you @duggiefresh!!!

PR #304, opened 2017-10-16
add no-trailing-spaces rule

kellyselden:


rwjblue:
@kellyselden - Looks like this needs a rebase...

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+- Add new `no-trailing-spaces` rule.
```
rwjblue: I like to give an example of good and bad, but it seems hard for this one (since whitespace would not be displayed anyways...

lib/rules/lint-no-trailing-spaces.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const Rule = require('./base');
+
+module.exports = class NoTrailingSpaces extends Rule {
+  visitor() {
+    return {
+
+      // implementation goes here in exit(): in the exit handler, the rule will not
```
rwjblue: This comment seems to be out of line with the actual implementation (which is using `enter` of `Program`, not `exit`).

lib/rules/lint-no-trailing-spaces.js
```diff
@@ -0,0 +1,49 @@
...
+        // `
+        //
+        // check for that case
+        if (line !== 1 || column !== 0) {
+          return;
+        }
+
+        let source = this.sourceForNode(node);
+
+        source.split('\n').forEach((line, i) => {
```
rwjblue: The lines are already available in `this.source` as an array of lines for the whole file.

kellyselden: it includes the newlines `[ '{{#my-component}}\n', '  test \n', '{{/my-component}}', '' ]`, which means it wouldn't be as clean as a simple forEach. Would you still prefer I switch?

CHANGELOG.md
kellyselden: added examples best I could

lib/rules/lint-no-trailing-spaces.js
kellyselden: I forgot to put it in `exit`. This simplified the tests too.

PR #316, opened 2017-11-15
add quotes rule

kellyselden:
Closes #42.

This a MVP of the quotes rule that can be built upon if necessary.

### DIFF
--- a/docs/rule/quotes.md
+++ b/docs/rule/quotes.md
@@ -1,0 +1,21 @@
+#### quotes
+
+Enforce the consistent use of either double or single quotes. Similar to https://eslint.org/docs/rules/quotes.
+
+Enforce either:
+
+```hbs
+<div class="my-class">test</div>
+{{my-helper "hello there"}}
+```
+
+or:
+
+```hbs
+<div class='my-class'>test</div>
+{{my-helper 'hello there'}}
+```
+
+The following values are valid configuration:
+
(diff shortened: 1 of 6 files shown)


docs/rule/quotes.md
```diff
@@ -0,0 +1,21 @@
+#### eol-last
```
rwjblue: Seems like a copy/paste error...

lib/rules/lint-quotes.js
```diff
@@ -0,0 +1,68 @@
...
+
+    throw new Error(errorMessage);
+  }
+
+  visitor() {
+    let badChar;
+    let message;
+    switch (this.config) {
+    case 'double':
+      badChar = '\'';
```
rwjblue: Thoughts on using a template string here instead to avoid the escaping?

kellyselden: I think I got a linting error when I tried that.

kellyselden: yep "Strings must use singlequote."

docs/rule/quotes.md
kellyselden: fixed.

btecu:
https://github.com/ember-template-lint/ember-template-lint/pull/42#issuecomment-218965107 would be nicer. It's nice to be able to have:
```
<div class="some-component" {{action 'doSomething'}}>
  ...
</div>
```

I believe right now it enforces one type of quotes and defaults to double quotes.

rwjblue:
Yep, confirm

PR #322, opened 2017-12-06
allow chaining of plugins and extends

### LINKED ISSUE #321: support plugins extending other plugins
The main use case is to have a plugin extend base `recommended` so you don't have to do `extends: ['recommended', 'plugin1:recommended'],` yourself. I tried inside of a plugin:

```js
configurations: {
  recommended: {
    extends: 'recommended',
    rules: {
    }
  }
}
```

but it didn't work.

kellyselden:
Closes #321

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -100,7 +100,7 @@ function migrateRulesFromRoot(config, source, options) {
   }
 }
 
-function processPlugins(config, options) {
+function processPlugins(config, options, checkForCircularReference) {
 
   let logger = options.console || console;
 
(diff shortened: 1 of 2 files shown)


kellyselden:
Blocked by #323 (the reason for the test failures).

kellyselden:
now working and no longer affected by the mutated config.

rwjblue:
Will need a rebase...

kellyselden:
> Can you add a test that shows what happens to a circular reference?

What would you like to happen? Throw if another plugin with the same name is found?

rwjblue:
Well, it isn't 100% obvious what should happen.  I think we have (at least) the following options:

* Throw if we hit some amount of recursion
* Attempt to process a certain number of loops

I _think_ the safe bet is to throw if we detect recursion for now, that way if we figure out a real solution we can remove the error without a breaking change...

What do you think?

kellyselden:
I went Above and Beyond™ and added loop detection code. Let me know what you think.

kellyselden:
The Travis gods are now pleased.

kellyselden:
@rwjblue ping

PR #323, opened 2017-12-06
no longer mutate the config

kellyselden:
This issue was uncovered in #322. The config was being mutated (even if `require`d). This was causing terrible issues, with the set of recommended rules getting overwritten among others. While changing this, it also uncovered a bunch of incorrect or brittle tests.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -31,23 +31,39 @@ function resolveConfigPath(configPath, pluginPath) {
 
 }
 
-function ensureRootProperties(config) {
-  config.rules = config.rules || {};
-  config.pending = config.pending || [];
-  config.ignore = config.ignore || [];
-  config.plugins = config.plugins || [];
+function ensureRootProperties(config, source, options) {
+  let logger = options.console || console;
+
+  config.rules = assign({}, source.rules || {});
+  config.pending = (source.pending || []).slice();
+  config.ignore = (source.ignore || []).slice();
+
+  let plugins = source.plugins || [];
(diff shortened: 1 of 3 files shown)


lib/get-config.js
```diff
@@ -31,23 +31,39 @@ function resolveConfigPath(configPath, pluginPath) {
...
+
+  let plugins = source.plugins || [];
+  config.plugins = Array.isArray(plugins) ? plugins.slice() : assign({}, plugins);
+
+
+  let extendedList = [];
+  if (source.extends) {
+    if (typeof source.extends === 'string') {
+      extendedList = [source.extends];
+    } else if (source.extends instanceof Array) {
```
rwjblue: Should use `Array.isArray` here instead of `instanceof`.

kellyselden: fixed

kellyselden:
Added a mutation check that fails on master. `deep-freeze` didn't make sense to me in this context, so I accomplished using a simple clone. I hope you like it.

test/unit/get-config-test.js
```diff
@@ -240,4 +241,18 @@ describe('get-config', function() {
     expect(secondMessage).to.not.be.ok;
   });
 
+  it('does no mutate the config', function() {
```
rwjblue: @kellyselden - This test looks good, can you confirm it failed prior to the other changes?

kellyselden: confirmed it failed terribly on master!

rwjblue: awesome, thank you!

rwjblue:
` ember-template-lint 0.8.14 published 🎉`

PR #328, opened 2017-12-18
remove usused plugin object fallback

kellyselden:
If `plugins` was an object, it would fail on this line https://github.com/rwjblue/ember-template-lint/blob/f568dca61c3a87a6dcf364cdb0cd5b0347fa198a/lib/get-config.js#L89, so I think it is safe to remove this code.


rwjblue:
Thanks!

PR #329, opened 2017-12-20
add failing test for block-indentation: attributes

kellyselden:
Trying to enforce the attribute indentation. Trying to prevent:

```hbs
{{foo-bar
baz=true
}}
```

in favor of

```hbs
{{foo-bar
  baz=true
}}
```


rwjblue:
The `block-indentation` rule is only about enforcing indentation of _children_ of the block, not arguments/attributes passed to it.  I've rebased this and moved the tests into the right file, to show that they do fail as appropriate.

kellyselden:
Ah maybe it's a new rule since this was created. Nevertheless, thanks!

Issue #332, opened 2018-01-11
Strange Warning [Expected 'title' to be indentation at an of 2 but was found at 2]


FalkF:
For this HBS:
```hbs
{{print-debug
  (hash
    client=environment.APP.version
    server=environment.apiVersion
    commitHash=commitHash
  )
  title="Version"
}}
```
I get following output:
```bash
  9:2  error  Incorrect indentation of attribute 'title' beginning at L9:C2. Expected 'title' to be indentation at an of 2 but was found at 2  attribute-indentation
```

Edit:
ember-template-lint does also seem to have a hard time detecting t.body or t.head from ember-light-table:
```hbs
...

  {{#t.body
    canExpand=true
    multiRowExpansion=false
  as |body|}}

...
```
```bash
  4:5  error  Incorrect indentation of positional param 't.body' beginning at L4:C5. Expected 't.body' to be indentation at an of 4 but was found at 5  attribute-indentation

```

rwjblue:
Very strange indeed!  Would you mind submitting a failing test case (around [here](https://github.com/rwjblue/ember-template-lint/blob/master/test/unit/rules/lint-attribute-indentation-test.js#L51-L53)) to make it easier to track down?

rwjblue:
Closed by #334

PR #334, opened 2018-01-11
fix(lint-attribute-indentation)

FalkF:
#332

### DIFF
--- a/lib/rules/lint-attribute-indentation.js
+++ b/lib/rules/lint-attribute-indentation.js
@@ -234,6 +234,11 @@ module.exports = class AttributeSpacing extends Rule {
     if (type === 'positional') {
       paramType = 'positional param';
       namePath = 'original';
+      if (node.path.original === 'component'
+          && node.params.length > 0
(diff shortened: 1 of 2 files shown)


lib/rules/lint-attribute-indentation.js
```diff
@@ -275,8 +280,8 @@ module.exports = class AttributeSpacing extends Rule {
       expectedLineStart = node.loc.start.line + 1;
     }
 
-    this.iterateParams(node.params, 'positional', expectedLineStart, expectedColumnStart, node);
-    this.iterateParams(node.hash.pairs, 'attribute', expectedLineStart + node.params.length, expectedColumnStart, node);
+    expectedLineStart = this.iterateParams(node.params, 'positional', expectedLineStart, expectedColumnStart, node);
```
FalkF: My supervisor did mention that it is probably a bad choice to make ```iterateParams``` return ```expectedLineStart```, since most other function have no return values and it needlessly entangles them.
Do you see another way to make this work?

rwjblue:
restarted

PR #335, opened 2018-01-11
table-groups textnode bug

FalkF:
table-groups gave false positives, since text nodes where not supported.


rwjblue:
Generally looks good, can you also add a "bad" test case for using text nodes (to ensure that the regexp for non-whitespace is working right)?

Something like:

```hbs
<table> Stuff<thead></thead></table>
```

Can you also add a "good" test for the following (to confirm the text node regexp matches multiple lines properly)?

```hbs
<table>
  <thead></thead>
</table>
```

rwjblue:
Thank you!

PR #340, opened 2018-01-19
Allow indentation to be configurable and fix multi-line attributes linting

DanielCMS:
See https://github.com/rwjblue/ember-template-lint/issues/339.

As a drive-by fix, also added a patch for long attribute linting


rwjblue:
Oh no, I'm very sorry to have let this sit 😢. Would you mind rebasing so we can try to get it landed?

DanielCMS:
@rwjblue Thanks for fixing the tests! I thought I would do it next week or so but always better to have the author putting additional eyes on :+1:

PR #342, opened 2018-01-21
Don't allow trailing dots in a PathExpression

siva-sundar:
Implements https://github.com/rwjblue/ember-template-lint/issues/341

### DIFF
--- a/docs/rule/no-trailing-dot-in-path-expression.md
+++ b/docs/rule/no-trailing-dot-in-path-expression.md
@@ -1,6 +1,30 @@
 ## no-trailing-dot-in-path-expression
 
-This rule doesn't allow trailing dot(s) on a path expression.
+This rule doesn't allow trailing dot(s) on a path expression. Since the trailing dot is treated as a separate path expression which represents the context associated to the template.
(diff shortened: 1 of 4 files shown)


siva-sundar:
I have added more info about the rule. Please let me know if anything else is required.

siva-sundar:
@rwjblue is there anything that needs to be updated in this PR?

siva-sundar:
I have 2 opinions:

1) This leads to another issue too, consider the following code
```js
<div class={{if some.path. 'a'}}>Sample Text</div>
```
Where some.path is a controller property, this renders as follows
```js
<div class="<twiddle@controller:application::ember232>">Sample Text</div>
```


2) Can we have this until this error is thrown from Glimmer?

siva-sundar:
Hey @rwjblue any updates?

Turbo87:
@rwjblue @siva-sundar it looks like this PR is restricting us from updating to the latest `@glimmer/compiler` release. with that release the parser will fail and no longer be able to parse the template at all. the only option I see is removing the rule again 🤔

PR #344, opened 2018-01-29
Add failing test case

KillaPL:
I believe I found a bug. Whenever there is a component in a block form, but it has no content inside, template lint will believe it violates the `eol-last` rule.

Fix I provided solves the issue.

Please let me know if I understand what's going on here correctly, and if there is anything I can do more to help fix this issue.

### DIFF
--- a/lib/rules/lint-eol-last.js
+++ b/lib/rules/lint-eol-last.js
@@ -58,6 +58,14 @@ module.exports = class EolLast extends Rule {
             return;
           }
 
+          let bodyLength = node.body.length;
+          // if there is a block component invocation without a body
+          // it will make it here too
+          // check for that case
+          if (bodyLength === 0) {
+            return;
+          }
+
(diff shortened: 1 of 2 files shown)


FalkF:
As far as I know block form should not be used without content, but there could be exceptions.
And I think ```{{/my-component}}'``` belongs to the next line:
```js
'{{#my-component}}' + '\n' + 
'{{/my-component}}'
```

For the test above no error is thrown, if I am right however there should be one telling us to use non Blockform. Maybe adjust it, so that instead of returning you throw a new error plus you can try to throw it when it is not on the same line. But I am neither a pro or maintainer, so take this with a grain of salt :wink:

KillaPL:
There might be additional rule to *discourage* block form without context, but it is indeed correct syntax and should not fail with linter.

rwjblue:
Thank you@

KillaPL:
Happy to help ^^

PR #345, opened 2018-01-30
Failing test case for attribute-indentation

KillaPL:
I'm providing failing test case for attribute-indentation rule, as discussed in https://github.com/rwjblue/ember-template-lint/issues/331

I have no clue what the expected format is in this case, I don't think it's handled. If handling this case isn't supposed to be handled by this rule, maybe it can ignore this case?

Please let me know if there is anything more I can do to help nailing this down.


FalkF:
I solved this one this in #334. :)

KillaPL:
I'm very glad I could help! I will check if we can safely use the attribute-indentation rule as soon, as new version is out! : )

rwjblue:
Replaced by #334

PR #355, opened 2018-03-04
Allow form elements to have reset actions

andybluntish:
An exception to the invalid-interactive rule already exists for allowing `onsubmit` attributes and `{{action "foo" on="submit"}}` on form elements. This makes that same exception for `reset` events. 

- allow `onreset={{action "foo"}}` on form tags
- allow `{{action "foo" on="reset"}}` on form tags


rwjblue:
Thank you!

PR #356, opened 2018-03-12
feat(lint-simple-unless): add whitelist

### LINKED ISSUE #353: feat(lint-simple-unless): add ability to whitelist helpers
I would like to be able to allow some specific helpers when using unless.

FalkF:
This PR adds two configuration options to the ```simple-unless``` rule:
- ```whitelist``` - an array of helpers to whitelist, e.g. ```'simple-unless': ['not', 'or']```.
- ```maxHelpers``` - number of maximum nested Helpers, defaults to 1 when at least one Helper is whitelisted. Setting maxHelpers without whitelisting at least one Helper will allow all Helpers and only apply the limit.
 
closes #353

### DIFF
--- a/lib/rules/lint-simple-unless.js
+++ b/lib/rules/lint-simple-unless.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
 
 const messages = {
   followingElseBlock: 'Using an {{else}} block with {{unless}} should be avoided.',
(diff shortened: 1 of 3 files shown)


lib/rules/lint-simple-unless.js
```diff
@@ -1,14 +1,71 @@
...
+      if (Array.isArray(config)) {
+        return {
+          whitelist: config,
+          maxHelpers: 1
+        };
+      } else if (isValidConfigObjectFormat(config)) {
+        // default any missing keys to empty values
+        return {
+          whitelist: config.whitelist || [],
+          maxHelpers: config.maxHelpers || 1
```
rwjblue: Why would this default to 1?

lib/rules/lint-simple-unless.js
```diff
@@ -65,18 +122,58 @@ module.exports = class LintSimpleUnless extends Rule {
   }
 
   _withHelper(node) {
-    let loc = node.params[0].loc.start;
-    let actual = `{{unless (${node.params[0].path.original} ...`;
+    let whitelist = this.config.whitelist; // let { whitelist, maxHelpers } = this.config;
+    let maxHelpers = this.config.maxHelpers;
+
+    if (typeof maxHelpers !== 'undefined' || typeof whitelist !== 'undefined') {
```
rwjblue: We should ensure that maxHelpers is always initialized up in the config parsing code. We shouldn’t need to compare against undefined....

rwjblue: Basically, I think we should remove this conditional completely. As long as `maxHelpers` is initialized to 0 and `whitelist` is initialized to `[]` we can always do the loop / check and still have the same result as before...

lib/rules/lint-simple-unless.js
FalkF: when somebody uses and object without maxHelpers, instead of an array.
e.g. :
```js
'simple-unless': { whitelist: ['not', 'or'] }
```

FalkF:
Im currently testing my changes. I will remove the WIP, when I finished this and wrote some more tests.

FalkF:
@rwjblue :tada: im done, can I be part of todays merge party

rwjblue:
I would prefer to reduce the number of config options down to Boolean and the object with `whitelist` and `maxHelpers`. I find it really hard to reason about all the various combinations, and it severely hampers how we teach this rule.

PR #361, opened 2018-03-21
[FEAT] validate-input-helper

runspired:
This PR contains two new rules for the `{{input` helper.

The first validates that it is not used via `{{#input` which would result in a runtime error. See: https://github.com/emberjs/ember.js/pull/15445 for context.

The second validates that `tagName` is not set. When `tagName` is set, and is set to something other than `input`, the helper becomes useless, and sometimes will not render at all.

### DIFF
--- a/test/unit/rules/lint-no-input-block-test.js
+++ b/test/unit/rules/lint-no-input-block-test.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const message = require('../../../lib/rules/lint-no-input-block').message;
+
+generateRuleTests({
+  name: 'no-input-block',
+
+  config: true,
+
+  good: [
+    '{{button}}',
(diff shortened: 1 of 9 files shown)


test/unit/rules/lint-no-input-block-test.js
```diff
@@ -0,0 +1,30 @@
...
+const message = require('../../../lib/rules/lint-no-input-block').message;
+
+generateRuleTests({
+  name: 'no-input-block',
+
+  config: true,
+
+  good: [
+    '{{button}}',
+    '{{#button}}{{/button}}',
```
rwjblue: I’m not sure this is actually valid. Button would have to be a keyword? Can you update to x-button or something?

runspired: 👍

runspired: Addressed

PR #364, opened 2018-03-23
Replace Mocha/Chai with Jest

Turbo87:
because coverage, interactive watch mode, snapshots, etc. :)

/cc @rwjblue


Turbo87:
I give up... no idea why this is running so mega slow :(

Issue #380, opened 2018-04-02
link-rel-noopener false-positive


boris-petrov:
We have a number of places where we use `rel="noreferrer noopener"` which is not OK according to `link-rel-noopener` as it wants it to be `noopener noreferrer`. This means exactly the same, or am I mistaken?

rwjblue:
Nope, I think you are definitely correct. Mind sending in a PR fixing?

boris-petrov:
Done.

PR #381, opened 2018-04-02
Fix link-rel-noopener not working for "noreferrer noopener"

boris-petrov:
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/380.


boris-petrov:
@rwjblue - good catch, I didn't think that there were more options. I've updated the PR.

PR #382, opened 2018-04-04
Fix linter for mustaches containing only string literal

KillaPL:
Mustaches containing only string literal are legal. I found them useful when in, say, addon dummy app I want to render readable html, it's easy to do 

`{{"<div>example text</div>"}}`

When `isMultipartPath` finds template code like this, it receives something like

```
{ type: 'StringLiteral',
  value: '<div>example text</div>',
  original: '<div>example text</div>',
  loc:
   { source: null,
     start: { line: 1, column: 2 },
     end: { line: 1, column: 14 } } }
```

Which contains no `parts`, this generating error. I simply added a guard against this case.

### DIFF
--- a/lib/plugins/transform-dot-component-invocation.js
+++ b/lib/plugins/transform-dot-component-invocation.js
@@ -43,7 +43,7 @@ module.exports = function transformDotComponentInvocation(env) {
 };
 
 function isMultipartPath(path)  {
-  return path.parts.length > 1;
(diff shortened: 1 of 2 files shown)


rwjblue:
Would you mind rebasing to get CI to run? When you submitted the PR originally we were having an issue with CI...

KillaPL:
Is that ok? :)

rwjblue:
Yep, looks great thank you!

Issue #388, opened 2018-04-04
Node 4 EOL


FalkF:
Node 4 will be unsupported after this month.
Can we join the upgrade train and get some new language features?
https://github.com/nodejs/release

rwjblue:
yes, we are queuing up for a 0.9.0 too...

PR #392, opened 2018-04-15
Add no-implicit-this rule

mmun:
This rule disallows paths with an implicit `this` as described in https://github.com/emberjs/rfcs/pull/308.

For example, `{{foo}}` and `{{my-component foo.bar}}` are disallowed.

This lint suggests three possible alternatives:

1. If the root of the path is passed in as an argument to the component then the path should be prefixed with `@`
e.g. `onClick` → `@onClick`. This suggestion is only possbile when using Ember 3.1+.

2. If the root of the path is a property on `this` then prefix the path with `this.`
e.g. `isExpanded` → `this.isExpanded`.

3. If the path is in fact a component or helper invocation that has no arguments, you must add this component to the `allow` whitelist in the rule's configuration.

Unfortunately, without something like module unification it is near impossible to know for certain whether something like `{{foo-bar}}` is a component or helper invocation, but I believe that an app will generally not have many components like this so it's ok to require them to be whitelisted.

### DIFF
--- a/lib/rules/lint-strict-paths.js
+++ b/lib/rules/lint-strict-paths.js
@@ -82,7 +82,7 @@ module.exports = class StrictPaths extends Rule {
         } else {
           let valid = path.data ||
                       path.this ||
-                      scope.isLocal(path.parts[0]) ||
+                      scope.isLocal(path) ||
                       this.config.allow.indexOf(path.original) > -1;
 
 
--- a/lib/rules/lint-strict-paths.js
+++ b/lib/rules/lint-strict-paths.js
@@ -2,6 +2,7 @@
 
 const Rule = require('./base');
 const createErrorMessage = require('../helpers/create-error-message');
+const Scope = require('./internal/scope');
 
 function message(original) {
   return `Ambiguous path '${original}' is not allowed. ` +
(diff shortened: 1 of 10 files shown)


lib/rules/lint-strict-paths.js
```diff
@@ -0,0 +1,146 @@
...
+
+    return {
+      PathExpression(path) {
+        if (nextPathIsCallee) {
+          // All paths are valid callees so there's nothing to check.
+        } else {
+          let valid = path.data ||
+                      path.this ||
+                      scope.isLocal(path.parts[0]) ||
+                      this.config.allow.includes(path.original);
```
mmun: node 4 doesn't have `includes`

FalkF: https://github.com/ember-template-lint/ember-template-lint/issues/388
😄

mmun: :D

rwjblue:
@mmun - I just pushed some changes, mostly just cleanup and rebasing stuff.  I think this rule is good to go once we get some docs. Thoughts on `no-ambiguous-paths` vs `strict-paths`?

rwjblue:
OK, I have updated the rule name to `no-implicit-this-paths` and added documentation.

Anything else needed here?

rwjblue:
Chatted with @mmun about this, and I think we both prefer `no-implicit-this`.  I'll update accordingly...

Issue #402, opened 2018-05-01
Contextual component indentation issue


graemechapman:
I have encountered a weird linting issue when using a component helper in inside contextual components.

```
{{#my-component as |something|}}
  {{#something.content}}
    {{#component "my-component"}}
      Content
    {{/component}}
  {{/something.content}}
{{/my-component}}
```

produces the error
```
6:24  error  Incorrect indentation for `component` beginning at L2:C2. Expected `{{/component}}` ending at L6:C24 to be at an indentation of 2 but was found at 10.  block-indentation
```

It works fine with normal component invocation, only seems to be an issue when using the component helper (and only in block form). 
However if I change the alias of the contextual component to a single character, it passes 😕

rwjblue:
Would you mind putting together a failing test case for this?

graemechapman:
@rwjblue sure, see https://github.com/ember-template-lint/ember-template-lint/pull/403

monovertex:
Is there any progress on this issue? We're encountering this as well in several of our templates. I would be happy to try my hand at a PR, but I'm unfamiliar with the codebase. If anyone has any pointers where should I look, I can give it a shot.

rwjblue:
I haven't dug into it much yet, I think that figuring out why the closing `{{/component}}` is seen as being at column 2 in the failing test case in #403 should be a great place to start.

PR #403, opened 2018-05-01
Failing test case for block component inside contextual component

graemechapman:
Adds a failing test case for the issue discussed in https://github.com/ember-template-lint/ember-template-lint/issues/402.


rwjblue:
Just rebased and pushed a new commit which includes the fix for this bug, thanks again for reporting!

graemechapman:
Awesome, thanks @rwjblue  :slightly_smiling_face:

PR #409, opened 2018-05-17
Remove tabpanel from interactive elements array

eglenfield:
We are having an issue with the template linter and elements with the `role="tabpanel"`. The tabpanel role is not an interactive element as defined on the W3C website, so the nested interactive error should not be fired (the superclass defined [here](https://www.w3.org/WAI/PF/aria/roles#tabpanel) is a region, not a widget). It would be great if this PR could get merged ASAP as the workaround we're using is to disable the linter which isn't ideal.

### DIFF
--- a/lib/helpers/is-interactive-element.js
+++ b/lib/helpers/is-interactive-element.js
@@ -29,7 +29,6 @@ const ARIA_WIDGET_ROLES = [
   'slider',
   'spinbutton',
   'tab',
-  'tabpanel',
   'textbox',
   'tooltip',
   'treeitem'


rwjblue:
Hmm, https://www.w3.org/TR/wai-aria/#widget_roles definitely lists `tabpanel` as a widget, and lists `widget` as an "interactive control" https://www.w3.org/WAI/PF/aria/roles#widget. 🤔

eglenfield:
@rwjblue The docs are contradicting themselves in the sense the first link you supplied does say a `tabpanel` is a widget, but it isn't defines as a subclass in the second link. In instances where I've seen tabs used on other websites, the tabpanel may incorporate radio buttons or search boxes, which would mean that most websites aren't following the W3C standards. I can understand `tab` and `tablist` roles being interactive controls, but in the case of a `tabpanel` I'm not so sure 🤷‍♀️

rwjblue:
I tend to agree with you, its just annoying that the docs are not consistent enough to make this clearer...

rwjblue:
Thank you!

PR #412, opened 2018-05-22
add aria-* properties to bare-strings rule

steveax:
This commit adds aria-label [1], aria-placeholder [2],
aria-roledescription [3], and aria-valuetext [4] to the
GLOBAL_ATTRIBUTES array in the bare-strings rule.

These are 4 of the 5 aria properties that take free-form strings
(aria-keyshortcuts is the other and it's not clear to me that would
be translated)

[1]: https://www.w3.org/TR/wai-aria-1.1/#aria-label
[2]: https://www.w3.org/TR/wai-aria-1.1/#aria-placeholder
[3]: https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription
[3]: https://www.w3.org/TR/wai-aria-1.1/#aria-valuetext

### DIFF
--- a/lib/rules/lint-bare-strings.js
+++ b/lib/rules/lint-bare-strings.js
@@ -26,7 +26,11 @@ const Rule = require('./base');
 const createErrorMessage = require('../helpers/create-error-message');
 
 const GLOBAL_ATTRIBUTES = [
-  'title'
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

Issue #384, opened 2018-04-04
Prevent using `attrs` inside of component templates


Turbo87:
see https://github.com/ember-template-lint/ember-cli-template-lint/issues/23

rwjblue:
Implemented in #416

PR #416, opened 2018-05-28
Add `no-attr-in-components` rule

akashdsouza:
https://github.com/ember-template-lint/ember-template-lint/issues/384

### DIFF
--- a/docs/rule/no-attrs-in-components.md
+++ b/docs/rule/no-attrs-in-components.md
@@ -1,0 +1,25 @@
+## no-attrs-in-components
+
+This rule prevents the usage of `attrs` property to access values passed to the component since all the values can be accessed directly from the from the template.
+
+For example, doing
+
+```components/templates/layout.hbs
+{{attr.foo}}
+```
+is bad. The correct way would be to directly consume `foo` in the template like,
+
+```components/templates/layout.hbs
+{{foo}}
+```
+
+or if you using Ember 3.1 and above,
+
+```components/templates/layout.hbs
+{{@foo}}
+```
+
+
+The following values are valid configuration:
+
+  * boolean -- `true` for enabled / `false` for disabled
--- a/lib/rules/lint-no-attr-in-components.js
+++ b/lib/rules/lint-no-attr-in-components.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const Rule = require('./base');
+const componentTemplateRegex = new RegExp('templates/components|components/.*/template|ui/components|-components/');
+
+module.exports = class LinkNoAttrComponent extends Rule {
+  isComponentTemplate() {
+    return componentTemplateRegex.test(this.templateEnvironmentData.moduleName);
+  }
+  visitor() {
+    return {
+      PathExpression(node) {
+        if (!this.isComponentTemplate()) {
+          return;
+        }
+        if (node.parts && node.parts[0] === 'attrs') {
+          this.log({
+            message: 'Component templates should not contain `attrs`.',
+            line: node.loc && node.loc.start.line,
+            column: node.loc && node.loc.start.column,
+            source: this.sourceForNode(node)
+          });
+        }
+      }
+    };
+  }
+};
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -36,4 +36,5 @@ module.exports = {
   'no-unnecessary-concat': require('./lint-no-unnecessary-concat'),
   'no-partial': require('./lint-no-partial'),
   'invocable-blacklist': require('./lint-invocable-blacklist'),
+  'no-attr-in-components': require('./lint-no-attr-in-components'),
 };
--- a/test/unit/rules/lint-no-attr-in-components-test.js
+++ b/test/unit/rules/lint-no-attr-in-components-test.js
@@ -1,0 +1,129 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-attr-in-components',
+
+  config: 'true',
+  good: [
+    '<div></div>',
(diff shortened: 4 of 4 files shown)


docs/rule/no-attrs-in-components.md
```diff
@@ -0,0 +1,18 @@
+## no-attrs-in-components
+
+This rule prevents the usage of `attrs` property to access values passed to the component since all the values can be accessed directly from the from the template.
+
+For example, doing
+
+```components/templates/layout.hbs
+{{attr.foo}}
+```
+is bad. The correct way would be to directly consume `foo` in the template like,
```
rwjblue: Can you also add a "good" example using `@foo` (works in Ember 3.1+)?

lib/config/recommended.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     'deprecated-inline-view-helper': true,
     'deprecated-render-helper': true,
     'simple-unless': true,
-    'unused-block-params': true
+    'unused-block-params': true,
+    'no-attr-in-components': true,
```
rwjblue: Please do not update the default config in this PR (adding a new default rule is a breaking change, so it can't get released within the current minor version). Can you remove this entry?

lib/rules/lint-no-attr-in-components.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const Rule = require('./base');
+
+module.exports = class LinkNoAttrComponent extends Rule {
+  isComponentTemplate() {
+    let moduleName = this.templateEnvironmentData.moduleName;
+    let regex = RegExp('templates/components/|components/.*/template.hbs');
```
rwjblue: Lets store this regexp in module scope (no need to recreate it for each check).

rwjblue: We should consider [module unification](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md) here as well. 

I think the following would work `new RegExp('/templates/components|components/.*/template|ui/components|-components/')`, what do you think?

lib/rules/lint-no-attr-in-components.js
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+    if (attributes && attributes.length || params.length) {
+      this.logAttrsInComponentTemplates(node, attributes, 'value');
+      this.logAttrsInComponentTemplates(node, params, '');
+    }
+  }
+  visitor() {
+    return {
+      SubExpression(node) {
```
rwjblue: I think that you can likely get away with using `PathExpression` as the main hook here (instead of `SubExpression`, `BlockStatement`, `MustacheStatement` and not need to manually traverse `params` and `hash` of each)...

akashdsouza: The reason I did not use `PathExpression` was that I felt this rule should allow components to be named `attrs` or to be present in folders named `attrs` (although these cases aren't ideal)

lib/rules/lint-no-attr-in-components.js
akashdsouza: I don't really have a lot of experience working with module unification. I could take a look. If you feel that the above regex would work, I would gladly change it.

rwjblue: Ya, I think it’s at least pretty close and we can always tweak it if we need to later.

lib/rules/lint-no-attr-in-components.js
rwjblue: Hmm, how would using `PathExpression` prevent using components in folders named attrs?

Also, components already cannot be named exactly `attrs` due to the dash requirement...

akashdsouza: Sorry I first thought of components. But the actual problem would be with helpers named `attrs`

rwjblue: We _could_ handle that while using path expression visitor (by only matching when `node.original.startsWith(‘attrs.’)`), but TBH I don’t think having a helper named attrs is common (or a good idea 😜).

I’d still much prefer to go the path expression route. It also would ensure compliance with angle bracket invocation (those would be `ElementNode`s)...

docs/rule/no-attrs-in-components.md
akashdsouza: Done

lib/rules/lint-no-attr-in-components.js
akashdsouza: Agreed. I guess I got caught up a bit with what the rule should and shouldn't support. I have made all the changes.
I have just one question though. Although this case may never ever come up, should we also guard against simply using `{{attrs}}` ?

rwjblue: >  Although this case may never ever come up, should we also guard against simply using {{attrs}} ?

Hmm, good point. I agree, I don't think we want to allow that...

I think we basically want to log an error whenever `node.parts[0] === 'attrs` (for the `PathExpression` hook), what do you think? That should catch all of these:

* `{{attrs}}`
* `{{foo-bar blah=attrs}}`
* `{{attrs.derp}}`

rwjblue:
I think it will also need a rebase once the main changes are done...

lib/rules/lint-no-attr-in-components.js
akashdsouza: Yes it should work against all the cases and that includes the helpers. I guess we can go ahead with `node.parts[0]` check for now and see if the helper case really is a concern later on. Thoughts?

rwjblue: Yep, totally agreed.

akashdsouza:
Rebased

rwjblue:
Great job, thanks for working on this!

PR #422, opened 2018-06-07
Allow angle bracket components to have actions

cibernox:


lib/helpers/is-angle-bracket-component.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function isDynamicComponent(element) {
+  let open = element.tag.charAt(0);
+
+  let maybeLocal = element.tag.split('.').length > 1;
```
rwjblue: This is fine for now, but can you create an issue to address actually tracking locals and expose `isLocal` to each plugin?

cibernox:
@rwjblue done in #423

Issue #191, opened 2017-04-14
erroneous "deprecated-inline-view-helper" violation


csantero:
With 0.6.4, the `deprecated-inline-view-helper` rule is triggering for the following syntax:

```
{{foo-bar qux="view"}}
```

Yielding the linter message "The inline form of `view` is deprecated. Please use the `Ember.Component` instead."

rwjblue:
Should be fixed by https://github.com/ember-template-lint/ember-template-lint/pull/431 when it lands...

boris-petrov:
Still happening with version 3.9.0. :)

bmish:
Looks like we'll need a volunteer to continue working on the fix that was started in #431...

lifeart:
closing as likely outdated and there is no active champions on it

boris-petrov:
@lifeart why did you close this? It's still an issue in the current version - 7.0.8.

lifeart:
@boris-petrov issue remains as-is for 3 years and related PR is closed, and nobody working on it, I'm happy to reopen it if we found contributor to fix it.

PR #431, opened 2018-06-15
Add and fix a test case for deprecated-inline-view-helper

lstrzebinczyk:
In the application I am working on, we have an icon component, that accepts an icon name as param. One of available icon names is "view", and those usages are flagged by `deprecated-inline-view-helper`.

This change makes sure that the rule does not flag string literals containing "view" string, and adds proper test case.

If the MR handles this update incorrectly in any sort, I am happy to update.

Fixes https://github.com/ember-template-lint/ember-template-lint/issues/191

### DIFF
--- a/lib/rules/deprecations/lint-deprecated-inline-view-helper.js
+++ b/lib/rules/deprecations/lint-deprecated-inline-view-helper.js
@@ -38,8 +38,9 @@ function manageMustacheViewInvocation(context, node, isBlockStatement) {
     for (let i = 0; i < pairs.length; i++) {
       let currentPair = pairs[i];
       let originalValue = currentPair.value.original;
+      let valueType = currentPair.value.type;
 
-      if (typeof originalValue === 'string' && originalValue.split('.')[0] == 'view') {
+      if (valueType !== 'StringLiteral' && typeof originalValue === 'string' && originalValue.split('.')[0] == 'view') {
         if (isBlockStatement) {
           context.processInBlockStatement(node, currentPair);
         } else {
--- a/test/unit/rules/deprecations/lint-deprecated-inline-view-helper-test.js
+++ b/test/unit/rules/deprecations/lint-deprecated-inline-view-helper-test.js
@@ -14,7 +14,8 @@ generateRuleTests({
   good: [
     '{{great-fishsticks}}',
     '{{input placeholder=(t "email") value=email}}',
-    '{{title "CrossCheck Web" prepend=true separator=" | "}}'
+    '{{title "CrossCheck Web" prepend=true separator=" | "}}',
+    '{{x-component foo="view"}}'
   ],
 
   bad: [


lstrzebinczyk:
Tests seem to have failed on node 6, while were successfull on node 4 and 8... Did I do something wrong?

lstrzebinczyk:
So just to make sure I get this right: the extended rule would be that any property key-value pair in which key contains "Binding" and value contains "view" should be disallowed ?

rwjblue:
Yeah, I think something like:

```js
let isInvalidStringLiteral = propertyType === ‘StringLiteral’ 
  && propertyName.endsWith(‘Binding`)
  && propertyValue.startsWith(‘view’)
```

lstrzebinczyk:
Exactly what I had in mind! Already working on it.

lstrzebinczyk:
Using a template like this:

```
'{{foo-bar someBinding=“view.something”}}'
```

generates such mustache data:

```
{ type: 'HashPair',
  key: 'someBinding',
  value: 
   { type: 'PathExpression',
     original: '“view.something”',
     this: false,
     parts: [ '“view', 'something”' ],
     data: false,
     loc: SourceLocation { source: undefined, start: [Object], end: [Object] } },
  loc: 
   { source: null,
     start: { line: 1, column: 10 },
     end: { line: 1, column: 38 } } }
```

Which seems like a bug. This should be a string literal in this case, right?

lstrzebinczyk:
I am more than happy to fix that, but I don't think this happens in this repo. Glimmer compiler maybe?

rwjblue:
What version were you testing on? I did a quick test in the local terminal with the version of @glimmer/syntax that we are using now, and it returned a `StringLiteral`...


```
let syntax = require('@glimmer/syntax');
let ast = syntax.preprocess('{{foo-bar bazBinding="view.qux"}}');
console.log(JSON.stringify(ast, null, 2));
```


```json
{
  "type": "Program",
  "body": [
    {
      "type": "MustacheStatement",
      "path": {
        "type": "PathExpression",
        "original": "foo-bar",
        "this": false,
        "parts": [
          "foo-bar"
        ],
        "data": false,
        "loc": {
          "start": {
            "line": 1,
            "column": 2
          },
          "end": {
            "line": 1,
            "column": 9
          }
        }
      },
      "params": [],
      "hash": {
        "type": "Hash",
        "pairs": [
          {
            "type": "HashPair",
            "key": "bazBinding",
            "value": {
              "type": "StringLiteral",
              "value": "view.qux",
              "original": "view.qux",
              "loc": {
                "source": null,
                "start": {
                  "line": 1,
                  "column": 21
                },
                "end": {
                  "line": 1,
                  "column": 31
                }
              }
            },
            "loc": {
              "source": null,
              "start": {
                "line": 1,
                "column": 10
              },
              "end": {
                "line": 1,
                "column": 31
              }
            }
          }
        ],
        "loc": {
          "source": null,
          "start": {
            "line": 1,
            "column": 10
          },
          "end": {
            "line": 1,
            "column": 31
          }
        }
      },
      "escaped": true,
      "loc": {
        "source": null,
        "start": {
          "line": 1,
          "column": 0
        },
        "end": {
          "line": 1,
          "column": 33
        }
      }
    }
  ],
  "blockParams": [],
  "loc": {
    "source": null,
    "start": {
      "line": 1,
      "column": 0
    },
    "end": {
      "line": 1,
      "column": 33
    }
  }
}
```

lstrzebinczyk:
I did a git clone, yarn install, and there I was. Package.json:

```
{
  "name": "ember-template-lint",
  "version": "0.8.18",
  "description": "Lint your templates.",
  "keywords": [],
  "homepage": "https://github.com/rwjblue/ember-template-lint",
  "bugs": {
    "url": "https://github.com/rwjblue/ember-template-lint/issues"
  },
  "license": "MIT",
  "author": "Robert Jackson <me@rwjblue.com>",
  "main": "lib/index.js",
  "bin": "./bin/ember-template-lint.js",
  "repository": {
    "type": "git",
    "url": "git+ssh://git@github.com/rwjblue/ember-template-lint.git"
  },
  "scripts": {
    "test": "mocha"
  },
  "dependencies": {
    "@glimmer/compiler": "^0.35.5",
    "chalk": "^2.0.0",
    "glob": "^7.0.0",
    "minimatch": "^3.0.4",
    "resolve": "^1.1.3",
    "strip-bom": "^3.0.0"
  },
  "devDependencies": {
    "chai": "^4.0.0",
    "eslint-plugin-node": "^6.0.0",
    "mocha": "^5.0.0",
    "mocha-eslint": "^4.1.0",
    "mocha-only-detector": "^1.0.0",
    "testem": "^2.0.0"
  },
  "engines": {
    "node": ">= 4"
  }
}
```

And, relevantly: 
```
"@glimmer/compiler@^0.35.5":
  version "0.35.5"
```
in yarn.lock

I found my result inside the rule code itself, while writing a test, and trying to get it to run, maybe this has anything to do with it?

lstrzebinczyk:
This seems like an issue that needs to be solved properly, but I am loosing a trail at one point. One step at a time:

I am adding a `bad` case to tests in `test/unit/rules/deprecations/lint-deprecated-inline-view-helper`, with this object:

```
    {
      template: '{{foo-bar someBinding=“view.something”}}',

      result: {
        message,
        moduleId: 'layout.hbs',
        source: '{{foo-bar someBinding=“view.something”}}',
        line: 1,
        column: 11,
        fix: {
          text: '{{foo-bar some=something}}'
        }
      }
    }
```

line, column are copy-pasted from somewhere else now, and fix is my guess. At this point I am trying to get any response.

So, this object is called by this loop: https://github.com/ember-template-lint/ember-template-lint/blob/master/test/helpers/rule-test-harness.js#L59, and we are specifically interested in this call: https://github.com/ember-template-lint/ember-template-lint/blob/master/test/helpers/rule-test-harness.js#L72

The `verify` method this calls is defined here: https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/index.js#L98. This, in turn, calls the `precompile` method from glimmer compiler, more specifically,  `require('@glimmer/compiler').precompile`.

I dug into the node-modules, and this calls the code in `node_modules/@glimmer/compiler/dist/commonjs/es5/lib/compiler`, function `precompile`.

The compiled code itself looks like this:

```
function precompile(string) {
    var options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : defaultOptions;

    var ast = (0, _syntax.preprocess)(string, options);

    var meta = options.meta;

    var _TemplateCompiler$com = _templateCompiler2.default.compile(ast, options),
        block = _TemplateCompiler$com.block;

    var idFn = options.id || defaultId;
    var blockJSON = JSON.stringify(block.toJSON());
    var templateJSONObject = {
        id: idFn(JSON.stringify(meta) + blockJSON),
        block: blockJSON,
        meta: meta
    };
    // JSON is javascript
    return JSON.stringify(templateJSONObject);
}
```

And the ast, if JSON.stringified for when this specific test is run contains this object:

```
{
   "type":"Program",
   "body":[
      {
         "type":"MustacheCommentStatement",
         "value":" template-lint-disable ",
         "loc":{
            "source":null,
            "start":{
               "line":1,
               "column":0
            },
            "end":{
               "line":1,
               "column":28
            }
         }
      },
      {
         "type":"MustacheStatement",
         "path":{
            "type":"PathExpression",
            "original":"foo-bar",
            "this":false,
            "parts":[
               "foo-bar"
            ],
            "data":false,
            "loc":{
               "start":{
                  "line":2,
                  "column":2
               },
               "end":{
                  "line":2,
                  "column":9
               }
            }
         },
         "params":[

         ],
         "hash":{
            "type":"Hash",
            "pairs":[
               {
                  "type":"HashPair",
                  "key":"someBinding",
                  "value":{
                     "type":"PathExpression",
                     "original":"“view.something”",
                     "this":false,
                     "parts":[
                        "“view",
                        "something”"
                     ],
                     "data":false,
                     "loc":{
                        "start":{
                           "line":2,
                           "column":22
                        },
                        "end":{
                           "line":2,
                           "column":38
                        }
                     }
                  },
                  "loc":{
                     "source":null,
                     "start":{
                        "line":2,
                        "column":10
                     },
                     "end":{
                        "line":2,
                        "column":38
                     }
                  }
               }
            ],
            "loc":{
               "source":null,
               "start":{
                  "line":2,
                  "column":10
               },
               "end":{
                  "line":2,
                  "column":38
               }
            }
         },
         "escaped":true,
         "loc":{
            "source":null,
            "start":{
               "line":2,
               "column":0
            },
            "end":{
               "line":2,
               "column":40
            }
         }
      }
   ],
   "blockParams":[

   ],
   "loc":{
      "source":null,
      "start":{
         "line":1,
         "column":0
      },
      "end":{
         "line":2,
         "column":40
      }
   }
}
```

From which the important part is `"type":"PathExpression","original":"“view.something”"`; it contains the incorrect type I mentioned before.

I assume this is the object that goes (or parts of which goes) into the rule definition here: https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/deprecations/lint-deprecated-inline-view-helper.js#L73. I assume, because I can't find a clear call site of the `MustacheStatement` function.

If I understand correctly, in current version if this MR the `bad` case we are talking about will continue to go unchecked.

Forgive the digging into the compiled code inside node modules, experience in writing apps with ember doesn't translate easily into hacking libraries. I tried to create a reproduction app that would show the compiled output of template string, but I can't seem to be able to require glimmer compiler inside an ember app...

I noticed that your example uses `preprocess`, and the glimmer code used to compile sources in the tests is `precompile`. I understand that you used preprocess, because it's seen here: https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/helpers/parse.js, but this is actually only used in some unit tests, not to actually parse the cases in tests. 
However, precompile uses preprocess underneath: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/compiler/lib/compiler.ts#L70, so this seems to be a dead end.

So the real question here is: where and how do I create a reproducible test case for this issue? Maybe this should be a test case somewhere in https://github.com/glimmerjs/glimmer-vm/tree/master/packages/%40glimmer/compiler/test ?

dcyriller:
@lstrzebinczyk this PR's branch has been deleted. Do you want to push it again? 

Now that several `@glimmer/syntax` versions have been released, StringLiteral issue you ran into might have been fixed there.

lstrzebinczyk:
Oh boy, this was so long ago, I kind of forgot about this one. I'm gonna close it, there's a fair chance this is no longer relevant. Sorry for the holdup!

PR #435, opened 2018-06-17
Ignore Warnings / Only Report Errors with "--quiet" option

### LINKED ISSUE #428: Only report errors / hide warnings
Is it possible to only report errors / hide warnings for the CLI? Something like the "--quiet" option in eslint?

I was looking to integrate this in a CI setup with the "pending" config so my team could start using it, but I found out pretty quick that if there are errors, it also shows all other warnings too! So finding the error in the thousands of warnings doesn't really work. 

I played around with the code a bit and was able to make such an option, so I could take a crack at making a PR if you guys are open to it?

srt4rulez:
Resolves #428 

If you pass "--quiet" to the CLI, it will now remove warnings from the console output. 

I also updated the console output to show the number of warnings, like how eslint does:

Eg: ✖ 3 problems (2 errors, 1 warnings)

I understand this may not be desirable in this PR, as each PR should probably implement one feature, so feel free to tell me that. 

I'm also new to nodejs (but not javascript) and tests in general, so constructive feedback is welcome!

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -10,11 +10,27 @@ var linter = new Linter();
 const chalk = require('chalk');
 
 function printErrors(errors) {
+  const quiet = process.argv.indexOf('--quiet') !== -1;
+
   if (process.argv.indexOf('--json') + 1) {
+    let filteredErrors;
+
+    Object.keys(errors).forEach(filePath => {
+      let fileErrors = errors[filePath] || [];
+
+      filteredErrors = fileErrors.filter((error) => {
+        if (quiet && error.severity < Linter.ERROR_SEVERITY) {
+          return false;
+        }
+
+        return true;
+      });
+
+      errors[filePath] = filteredErrors;
+    });
+
     console.log(JSON.stringify(errors, null, 2));
   } else {
-    const quiet = process.argv.indexOf('--quiet') !== -1;
-
     let errorCount   = 0;
     let warningCount = 0;
 
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -13,21 +13,39 @@ function printErrors(errors) {
   if (process.argv.indexOf('--json') + 1) {
     console.log(JSON.stringify(errors, null, 2));
   } else {
-    let count = 0;
+    const quiet = process.argv.indexOf('--quiet') !== -1;
+
+    let errorCount   = 0;
+    let warningCount = 0;
 
     Object.keys(errors).forEach(filePath => {
-      let options = {};
+      let options = {
+        quiet: quiet
+      };
       let fileErrors = errors[filePath] || [];
-      count += fileErrors.length;
+
+      errorCount   += fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY).length;
+      warningCount += fileErrors.filter(error => error.severity === Linter.WARNING_SEVERITY).length;
 
       if (process.argv.indexOf('--verbose') + 1) {
         options.verbose = true;
       }
 
-      console.log(Linter.errorsToMessages(filePath, fileErrors, options));
+      const messages = Linter.errorsToMessages(filePath, fileErrors, options);
+      if (messages !== '') {
+        console.log(messages);
+      }
     });
 
-    console.log(chalk.red(chalk.bold(`✖ ${count} problems`)));
+    if (quiet) {
+      warningCount = 0;
+    }
+
+    const count = errorCount + warningCount;
+
+    if (count > 0) {
+      console.log(chalk.red(chalk.bold(`✖ ${count} problems (${errorCount} errors, ${warningCount} warnings)`)));
+    }
   }
 }
 
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -12,36 +12,27 @@ const chalk = require('chalk');
 function printErrors(errors) {
   const quiet = process.argv.indexOf('--quiet') !== -1;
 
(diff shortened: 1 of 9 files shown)


lib/index.js
```diff
@@ -9,7 +9,7 @@ const chalk = require('chalk');
 const TransformDotComponentInvocation = require('./plugins/transform-dot-component-invocation');
 
 const WARNING_SEVERITY = 1;
-const ERROR_SEVERITY = 2;
+const ERROR_SEVERITY   = 2;
```
rwjblue: Would you mind resetting this? I do aesthetically prefer them to be aligned, but the cost of maintaining is much higher (e.g. following project history ends up being harder)...

lib/index.js
```diff
@@ -158,21 +158,33 @@ class Linter {
...
     };
 
     if (errors.length === 0) {
       return '';
     }
 
-    let errorsMessages = errors.map((error) => this._formatError(error, options)).join('\n');
+    let errorsMessages = errors
+      .map((error) => this._formatError(error, options))
+      .filter(error => error !== '')
```
rwjblue: Hmm, this seems surprising, why would we have empty strings?

lib/index.js
```diff
@@ -158,21 +158,33 @@ class Linter {
...
+      return '';
+    }
 
     return `${chalk.underline(filePath)}\n${errorsMessages}\n`;
   }
 
   static _formatError(error, options) {
     let message = '';
 
+    if (options.quiet && error.severity === WARNING_SEVERITY) {
```
rwjblue: I know that it doesn't matter just yet, but I think conceptually we want to say that quiet means "only show errors". Can we make this conditional reflect that (e.g. `options.quiet && error.severity < ERROR_SEVERITY`)?

lib/index.js
```diff
@@ -158,21 +158,33 @@ class Linter {
...
+      return message;
```
rwjblue: TBH, I'd prefer to return something else here (especially if we are going to filter them out in `Linter.errorsToMessages`).  Perhaps `return null`?

srt4rulez:
Oh! I completely forgot about `--json`. Will do!

lib/index.js
srt4rulez: I do not mind and agree 100%. I was actually wondering if you'd comment on this or not haha. I'll revert the module exports as well.

lib/index.js
srt4rulez: Yes, I wasn't really sure about this. I couldn't think of any other way to remove the warnings before it gets to the `.join('\n')` and create empty new lines everywhere. 

Ah, what about using filter to do the quiet and severity check instead?

```js
.filter((error) => {
    if (options.quiet && error.severity < ERROR_SEVERITY) {
        return false;
    }
    
    return true;
})
```

Then there are no changes to `_formatError` and it will always format an error (as the function suggests)

rwjblue: Ya, sounds good to me...

srt4rulez:
@rwjblue I have made the requested changes. Please review.

bin/ember-template-lint.js
```diff
@@ -10,24 +10,58 @@ var linter = new Linter();
...
 
     Object.keys(errors).forEach(filePath => {
-      let options = {};
+      let options = {
+        quiet: quiet
+      };
       let fileErrors = errors[filePath] || [];
-      count += fileErrors.length;
+
+      errorCount   += fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY).length;
```
rwjblue: Hmm, this update looks nice.  I _think_ with some small tweaks here, we can completely abstract `--quite` into the `bin/` code.

I'm thinking:

```js
let errors = fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY);
let warnings = quiet ? [] : fileErrors.filter(error => error.severity === Linter.WARNING_SEVERITY);

let errorCount = errors.length;
let warningCount = warnings.length;

// ...snip...
const messages = Linter.errorsToMessages(filePath, errors.concat(warnings), options;

// ...snip...
```

This way `Linter.errorsToMessages` _always_ is used for formatting strings from errors... 

What do you think?

srt4rulez: I think that's a great idea.

srt4rulez: Adding onto this, what are your thoughts on moving that logic above the check for the `--json` param and re-assigning errors, so we don't have to do it twice?

```js
Object.keys(errors).forEach(filePath => {
  let fileErrors = errors[filePath] || [];

  let errorsFiltered = fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY);
  let warnings = quiet ? [] : fileErrors.filter(error => error.severity === Linter.WARNING_SEVERITY);

  errorCount += errorsFiltered.length;
  warningCount += warnings.length;

  errors[filePath] = errorsFiltered.concat(warnings);
});

// check for json
// else foreach errors, console log Linter.errorsToMessages as it currently is
```

rwjblue: Yep, makes sense!

srt4rulez:
Absolutely!

srt4rulez:
I've updated the README. 

I think in the future it would be nice to have a section for the arguments you can pass, instead of just throwing more examples to the list 😛 hehe. 

Also, would you like me to rebase onto master?

rwjblue:
Thank you! I’ll try to get this released in the AM (feel free to ping me if I forget 😜)...

srt4rulez:
🎉 You're welcome. And thank-you for the feedback, sir!

PR #442, opened 2018-06-19
update no-bare-strings docs to include aria-*

steveax:
In a previous commit [1], the following aria-* properties
were added to the default no-bare-strings rule:

- aria-label
- aria-placeholder
- aria-roledescription
- aria-valuetext

But the docs for that rule were not updated. This commit
updates the docs to include the additional aria-* properties.

[1]: 4a58a8a2ecc13fec3c42b6303aff4e0e2cf67394

Previous PR: #412


rwjblue:
Sorry about letting the docs get out of sync, thank you very much for fixing them!

steveax:
Well, since it was my previous PR, I think you're off the hook ;-)

Issue #446, opened 2018-06-28
Improve attribute-indentation (or add a new rule) to require proper indenation for HTML


ventuno:
The `attribute-indentation` rule currently applies to handlebar elements/components only.
It would be nice to apply it to HTML as well. Not 100% sure if it's better to:
1. Simply extend the existing rule to cover HTML;
1. Keep this existing rule but add a configuration parameter to include/exclude HTML;
1. Create a new rule.

[1] https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/attribute-indentation.md

rwjblue:
I think 2 is probably the best bet. We are preparing for a major version bump soon, so we can change the default config (and enabling this rule in the recommended config) when we do.

ventuno:
Thanks @rwjblue, I created #447 as an early attempt to implement this.
It's just a very rough implementation for void HTML elements (see [1] for the terminology), I'm creating the pull request just to see if I'm headed in the correct direction.
Most of the naming needs to be figured out, I'm just trying to maximize the amount of existing logic that we can reuse.

A few questions:
1. I'm naming the configuration `parse-html-elements` let me know if that sounds reasonable;
1. Currently we position the "closing brackets" (`}}`) in a newline:
```hbs
{{some-component
  param1
}}
```
What should be preferred for HTML?
```html
<!-- option 1 -->
<input
  type="text"
>
```
```html
<!-- option 2 -->
<input
  type="text">
```
I personally prefer option 2, let me know what you think.

[1] https://en.wikipedia.org/wiki/HTML_element#Syntax

rwjblue:
This has actually been on the docket to fix also, see https://github.com/ember-template-lint/ember-template-lint/issues/267. Basically, I want to allow both by default, and have a config option to force one or the other.

ventuno:
Thanks @rwjblue, what if I work on option 1 right now (e.g. existing option) and take up #267 separately right after this one?
I’d like to keep the pull request’s scope as small as possible.

rwjblue:
Sounds like a great plan 😃

ventuno:
Open for review: https://github.com/ember-template-lint/ember-template-lint/pull/447.

ventuno:
Closed by https://github.com/ember-template-lint/ember-template-lint/pull/447.

PR #447, opened 2018-06-29
Attribute indentation support for HTML

ventuno:
Implements: https://github.com/ember-template-lint/ember-template-lint/issues/446.

HTML attributes parsing can by activated setting `process-elements` to `true`.

### DIFF
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -118,4 +118,4 @@ Block form (HTML)
 #### Configuration:
   * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
   * object - { 'open-invocation-max-len': n characters } - Maximum length of the opening invocation.
-  * object - { 'parse-html-elements': `true` } - Also validate the indentation of HTML elements.
+  * object - { 'process-elements': `true` } - Also validate the indentation of HTML/SVG elements.
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -116,6 +116,6 @@ Block form (HTML)
 ```
 
 #### Configuration:
-  * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
+  * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines. This will only enforce the rule for handlebar elements and not HTML/SVG elements.
   * object - { 'open-invocation-max-len': n characters } - Maximum length of the opening invocation.
   * object - { 'process-elements': `true` } - Also validate the indentation of HTML/SVG elements.
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -18,6 +18,18 @@ Block form
   {{/employee-details}}
 ```
 
+Non-Block form (HTML)
+``` html
+
+  <input disabled id="firstName" value={{firstName}} class="input-field first-name" type="text">
+```
+
+Block form (HTML)
+``` html
+
+  <a href="https://www.emberjs.com" class="emberjs-home link" rel="noopener" target="_blank">Ember JS</a>
+```
+
 #### Allowed:
 
 Non-Block form
@@ -58,6 +70,18 @@ Non-Block form with Helper unfolded
   }}
 ```
 
+Non-Block form (HTML)
+``` html
+
+  <input
+    disabled
+    id="firstName"
+    value={{firstName}}
+    class="input-field first-name"
+    type="text"
+  >
+```
+
 Block form
 ``` hbs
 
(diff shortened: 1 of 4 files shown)


ventuno:
@rwjblue this is probably a good point in which looking at this makes sense.

lib/rules/lint-attribute-indentation.js
```diff
@@ -12,7 +12,7 @@ const canApplyRule = function(node, type, config) {
   let end;
   let start = node.loc.start;
 
-  if (type === 'MustacheStatement') {
+  if (type === 'NonBlockStatement') {
```
rwjblue: I'd generally prefer to always pass `node.type` (instead of making our own custom `NonBlockStatement` string here) into this method. We can invert this conditional to keep the simpler logic though.

Something like:

```js
if (type !== 'BlockStatement') {
  // ...snip...
}
```

lib/rules/lint-attribute-indentation.js
```diff
@@ -285,30 +289,68 @@ module.exports = class AttributeSpacing extends Rule {
     let expectedColumnStart = node.loc.start.column + this.config.indentation; //params should be after proper positions from component start node
     let expectedLineStart = node.loc.start.line + 1;
 
-    expectedLineStart = this.iterateParams(node.params, 'positional', expectedLineStart, expectedColumnStart, node);
```
rwjblue: These two lines seem to have been deleted, but I would think they need to end up in the else clause of `if (node.type === 'ElementNode') {` right?

lib/rules/lint-attribute-indentation.js
```diff
@@ -360,6 +414,9 @@ module.exports = class AttributeSpacing extends Rule {
       if ('indentation' in config) {
         result.indentation = config.indentation;
       }
+      if ('parse-html-elements' in config) {
```
rwjblue: `html-elements` seems somewhat redundant (and we do theoretically support `svg` elements too). Maybe `process-elements`?

lib/rules/lint-attribute-indentation.js
ventuno: Done: removed `NonBlockStatement` and passed `node.type` directly (see `visitor()`).

lib/rules/lint-attribute-indentation.js
ventuno: Yeah, they are in lines 295-296 (below), am I missing something?

lib/rules/lint-attribute-indentation.js
ventuno: Done.

ventuno:
@rwjblue responded to your comments and added the tests as requested (thanks to the new tests I also discovered some latent issues, which is great).
As of now "angle bracket invocation" code will be treated as HTML, does that make sense?

rwjblue:
> As of now "angle bracket invocation" code will be treated as HTML, does that make sense?

Yep! That’s exactly what I would expect...

ventuno:
Thanks @rwjblue, all issues have been addressed and I've just pushed another change to fix an error message. Let me know if you think this needs any further changes or if we're good to go.

docs/rule/attribute-indentation.md
```diff
@@ -79,6 +103,19 @@ Block form (open invocation < 80 characters)
...
+    href="https://www.emberjs.com"
+    class="emberjs-home link"
+    rel="noopener"
+    target="_blank">
+    Ember JS
+  </a>
+```
+
 #### Configuration:
   * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
```
rwjblue: Can you update this to show exactly what `true` expands to (e.g. I assume it does not enable `process-elements`, but we should be explicit)

lib/rules/lint-attribute-indentation.js
```diff
@@ -360,6 +415,9 @@ module.exports = class AttributeSpacing extends Rule {
       if ('indentation' in config) {
         result.indentation = config.indentation;
       }
+      if ('process-elements' in config) {
+        result.parseHtmlElements = true;
```
rwjblue: This seems odd, the way this is written it is impossible to disable `process-elements` explicitly.

I think that you can update this to `result.parseHtmlElements = config['process-elements']`, this way we will use `true` or `false` based on whatever the users configuration has in it.

lib/rules/lint-attribute-indentation.js
```diff
@@ -379,14 +438,26 @@ module.exports = class AttributeSpacing extends Rule {
...
       MustacheStatement(node) {
-        if (canApplyRule(node, 'MustacheStatement', this.config)) {
-          this.validateNonBlockForm(node);
+        if (canApplyRule(node, node.type, this.config)) {
+          return this.validateNonBlockForm(node);
+        }
+        return node.loc.end.line;
+      },
+      ElementNode(node) {
+        if (this.config.parseHtmlElements) {
```
rwjblue: Can you tweak this to match the flag name? `this.config.processElements`

docs/rule/attribute-indentation.md
ventuno: Done, also updated the error message in `parseConfig` to specify that when `true` it will only apply to handlebar components and not HTML/SVG elements.

lib/rules/lint-attribute-indentation.js
ventuno: Makes sense, fixed and added a test in which I explicitly set `process-elements` to `false`.

lib/rules/lint-attribute-indentation.js
ventuno: Done as well, now they all match :-).

ventuno:
Thanks for the quick follow up @rwjblue. I addressed the latest feedback as well.

rwjblue:
Thank you very much!

Issue #267, opened 2017-09-14
attribute-indentation closing brackets on same line


amk221:
To support cases like:

```handlebars
{{my-component
  foo='bar'
  baz='qux'
  my-attr=(component 'my-other-component' data=(hash
    foo='bar'
    baz='qux'))}}
```
```handlebars
{{my-component
  foo='bar'
  baz='qux'
  my-attr=(component 'my-other-component' data=(hash
    foo='bar'
    baz='qux'
  ))
}}
```

rwjblue:
I am in favor of allowing the closing to be at the end of the last line, but I'm unsure if it should be a config option or if we should just allow either format.  If we make it config (horrible strawman: `allowBlocklessMustacheEndOnLastAttribute` ) we can still default to allowing both when the rule is `true`, but folks can lock their config down to force one format or the other (which I think is desirable).

rwjblue:
@siva-sundar - Thoughts?

siva-sundar:
Sounds good. We can allow both formats by default.

rwjblue:
This still seems like a good idea, but I haven't had time to work on it.

I'm definitely happy to help others through the process if someone has the time...

rwjblue:
Closed by #447

ventuno:
@rwjblue https://github.com/ember-template-lint/ember-template-lint/pull/447 closed https://github.com/ember-template-lint/ember-template-lint/issues/446, not this one.

I'm going to work on this next, but I noticed that https://github.com/ember-template-lint/ember-template-lint/pull/271/files is still open. @siva-sundar, are you still working on it or can I take over?

rwjblue:
Eeeck!Sorry about that!

PR #455, opened 2018-07-21
Validate close brace location

ventuno:
Implements: https://github.com/ember-template-lint/ember-template-lint/issues/267.

Validates the location of the close brace for mustache/block statements (`mustache-open-end`) and HTML/SVG elements (`end-element-open`).
This can be configured to be on a new line (`new-line`, the current behavior) or next to the last attribute (`last-attribute`).

### DIFF
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -119,3 +119,5 @@ Block form (HTML)
   * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
   * object - { 'open-invocation-max-len': n characters } - Maximum length of the opening invocation.
   * object - { 'process-elements': `true` } - Also validate the indentation of HTML/SVG elements.
+  * object - { 'end-element-open': `new-line`|`last-attribute` } - Enforce the position of the closing brace `>` to be on a new line or next to the last attribute (defaults to `new-line`).
+  * object - { 'mustache-open-end': `new-line`|`last-attribute` } - Enforce the position of the closing braces `}}` to be on a new line or next to the last attribute (defaults to `new-line`).
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -119,5 +119,5 @@ Block form (HTML)
   * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
   * object - { 'open-invocation-max-len': n characters } - Maximum length of the opening invocation.
   * object - { 'process-elements': `true` } - Also validate the indentation of HTML/SVG elements.
-  * object - { 'end-element-open': `new-line`|`last-attribute` } - Enforce the position of the closing brace `>` to be on a new line or next to the last attribute (defaults to `new-line`).
+  * object - { 'element-open-end': `new-line`|`last-attribute` } - Enforce the position of the closing brace `>` to be on a new line or next to the last attribute (defaults to `new-line`).
   * object - { 'mustache-open-end': `new-line`|`last-attribute` } - Enforce the position of the closing braces `}}` to be on a new line or next to the last attribute (defaults to `new-line`).
--- a/lib/rules/lint-attribute-indentation.js
+++ b/lib/rules/lint-attribute-indentation.js
@@ -465,7 +465,9 @@ module.exports = class AttributeSpacing extends Rule {
       '  * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.',
       '  * { open-invocation-max-len: n characters, indentation: m  } - n : The max length of the opening invocation can be configured',
       '  *                                                            - m : The desired indentation of attribute',
-      '  * { process-elements: `boolean` }                         - `true` : Also parse HTML/SVG attributes'
+      '  * { process-elements: `boolean` } - `true` : Also parse HTML/SVG attributes',
+      '  * { end-element-open: `new-line`|`last-attribute` } - Enforce the position of the closing brace `>` to be on a new line or next to the last attribute (defaults to `new-line`)',
+      '  * { mustache-open-end: `new-line`|`last-attribute` } - Enforce the position of the closing braces `}}` to be on a new line or next to the last attribute (defaults to `new-line`)'
     ], config);
 
     throw new Error(errorMessage);
--- a/lib/rules/lint-attribute-indentation.js
+++ b/lib/rules/lint-attribute-indentation.js
@@ -434,7 +434,7 @@ module.exports = class AttributeSpacing extends Rule {
         result.blocklessMustacheEnd = config['blockless-mustache-end'];
       }
       if ('elements-end' in config) {
-        // if elements-end is set, assument process-elements=true
+        // if elements-end is set, assume process-elements=true
         result.processElements = true;
         result.elementsEnd = config['elements-end'];
       }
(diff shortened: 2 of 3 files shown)


ventuno:
@rwjblue, this is probably a good moment for a first review. I feel I'm done code-wise, but docs are missing. I wanted to agree first on the naming of the configuration parameters before I proceed as I'm not too sure about the naming I picked.
I didn't want to have a super-long configuration name like `allowBlocklessMustacheEndOnLastAttribute` [1], so I thought something like `blockless-mustache-end` (`elements-end` for HTML/SVG attributes) with `new-line` (as the default) and `last-attribute` as possible values would help making the configuration parameters simpler. How does that sound?

Also, if someone sets`elements-end` to a value, I'm automatically setting `process-elements` to `true`. Does that make sense?


[1] https://github.com/ember-template-lint/ember-template-lint/issues/267#issuecomment-329550338

lib/rules/lint-attribute-indentation.js
```diff
@@ -408,6 +424,8 @@ module.exports = class AttributeSpacing extends Rule {
       let result = {
         maxLength: 80,
         indentation: 2,
+        blocklessMustacheEnd: 'new-line',
```
rwjblue: Hmm, why is this only for blockless? Seems like the same concept applies here:

```hbs
{{#some-thing
  whatever=qux
  stuff=goes-here
  onclick=(action 'handleClick')}}
  stuff here
{{/some-thing}}
```

vs

```hbs
{{#some-thing
  whatever=qux
  stuff=goes-here
  onclick=(action 'handleClick')
}}
  stuff here
{{/some-thing}}
```

lib/rules/lint-attribute-indentation.js
```diff
@@ -408,6 +424,8 @@ module.exports = class AttributeSpacing extends Rule {
...
+        elementsEnd: 'new-line',
```
rwjblue: Since this is specifically talking about how to end the _opening_ of the element, how about `endElementOpen`?

lib/rules/lint-attribute-indentation.js
rwjblue: Also, how about `mustacheOpenEnd`?

ventuno: Thanks @rwjblue, my understanding by reading https://github.com/ember-template-lint/ember-template-lint/issues/267 is that this issue is about validating the closing brackets of mustache expressions only. Currently there is no validation for the closing brackets of block elements (is this a bug?).
I'm more than happy to handle the validation of block elements as well. Should we handle this with a separate configuration parameter to support the current behavior?
These parameters would be (trying to summarize all feedback):
* `mustacheEndOpen`: `true|false` (`true` `}}` on a new line, `false` next to the last attribute);
* `blockExpressionEndOpen`: `true|false` (`true` `}}` on a new line, `false` next to the last attribute);
* `elementEndOpen`: `true|false` (`true` `}}` on a new line, `false` next to the last attribute).

How does that sound?

lib/rules/lint-attribute-indentation.js
ventuno: See above, I like the naming, but I would keep the `*EndOpen` for consistency, what do you think?

lib/rules/lint-attribute-indentation.js
rwjblue: > Currently there is no validation for the closing brackets of block elements (is this a bug?).

Ya, I am aware. The main issue is that its fairly tricky. For mustache statements we know exactly where the `}}` is (because the `node.loc.end` represents it), but for blocks we have to find it (likely using the source) since `node.loc.end` is actually referencing the block closing tag.

I do really want to have coverage for the "block end open tag", happy to treat that as another issue after landing this (to keep moving forward with the incremental improvements and whatnot) if you think that works better...

> Should we handle this with a separate configuration parameter to support the current behavior?

I _do_ feel strongly that we should validate block open, but I _do not_ feel strongly around the config. I think my gut feeling would be that an app would want both blockless and blocks to be similarly formatted (and therefore we would only need a single config for those), but again I don't really feel very strongly there...

lib/rules/lint-attribute-indentation.js
rwjblue: ya, seems good (the main point I was making was to make it clear that we are talking about the "end" of the "open" as opposed to the actual closing tag)

rwjblue:
Eeck, looks like we have pulled in a bunch more commits (likely ones I added to master?)...

ventuno:
Yup, my bad, I was trying to rebase with your latest changes, got excited and rebased with master. Should I close this and cherry-pick my changes into a new one?

rwjblue:
Nah, it should be easy enough to fix. 

I would do (assuming `origin` references this remote and not your personal one):

```js
git fetch origin
git rebase origin/master -i
# delete all the lines you see before your own
# I think that means all but the last 4
```

I'm happy to do it for you though (maintainers are allowed to push to branches of PR's by default), but you'd have to do a hard reset locally.

ventuno:
Alright I think I got this. Thanks for the help I had to force push because of the previous merge, but I guess that should be fine.

ventuno:
@rwjblue this should be good to go. This also covers the close brace location of block statements. Let me know what you think.

One open issue which I'm not sure how to handle:
```html
<SiteHeader
  @selected={{this.user.country}} as |Option|>
    {{#each this.availableCountries as |country|}}
      <Option @value={{country}}>{{country.name}}</Option>
    {{/each}}
</SiteHeader>
```
the above case shows an "angle bracket invocation" with a block param and the closing brace `>` aligned with the last attribute.
The problem here is that while I can easily determine the location of the attributes, it's not as easy to determine the location of the block param in this type of invocation (while it's easier for block statements).

docs/rule/attribute-indentation.md
```diff
@@ -119,3 +119,5 @@ Block form (HTML)
   * boolean - `true` - Enables the rule to be enforced when the opening invocation has more than 80 characters or when it spans multiple lines.
   * object - { 'open-invocation-max-len': n characters } - Maximum length of the opening invocation.
   * object - { 'process-elements': `true` } - Also validate the indentation of HTML/SVG elements.
+  * object - { 'end-element-open': `new-line`|`last-attribute` } - Enforce the position of the closing brace `>` to be on a new line or next to the last attribute (defaults to `new-line`).
```
rwjblue: the two option names seem confusing to me, it seems like we want to be consistent. I'd prefer to call it `element-open-end`. What do you think?

ventuno: Makes sense to me. I didn't feel we had total agreement on naming so I went with the most recent suggestions and figured we'd go from there.

Should I maybe validate that the values passed are either `new-line` or `last-attribute`?
As of now, I default the values to `new-line` and throughout the code I only check if `=== 'last-attribute'` so any other value than `last-attribute` will actually behave as `new-line` which I believe could be confusing.

rwjblue: > Should I maybe validate that the values passed are either new-line or last-attribute?

Yes, that makes sense, good idea.

ventuno: Done. Config values are validated now :-).

test/unit/rules/lint-attribute-indentation-test.js
```diff
@@ -270,13 +475,15 @@ generateRuleTests({
     '{{#t.body' + '\n' +
     '  canExpand=(helper help)' + '\n' +
     '  multiRowExpansion=false' + '\n' +
-    'as |body|}}' + '\n' +
+    'as |body|' + '\n' +
+    '}}' + '\n' +
```
rwjblue: I think that this is "fine" (it definitely is correct based on what the rule options are called), but practically speaking this feels _super bad_ to me...

This is a good candidate for some additional follow up after this lands...

rwjblue:
Thank you!

PR #460, opened 2018-07-25
Fix issue with attribute-indentation and ElementNode's without attributes.

rwjblue:
After releasing 1.0.0-beta.1 and updating a few small apps I identified an issue in the logic for validating an element's attributes. Essentially, if there were no attributes we would attempt to call `node.params.length` (and `node.params` is `undefined`).

This fixes the issue by:

* simplify `validateNonBlockForm` and `validateBlockForm` by moving `ElementNode` specific code into the `ElementNode` visitor directly.
* correct issues with detecting the "end" of the open tag for `ElementNode`'s (once the prior issue was fixed this cropped up) by extracting a specific method for `getEndLocationForOpen`.

### DIFF
--- a/lib/rules/lint-attribute-indentation.js
+++ b/lib/rules/lint-attribute-indentation.js
@@ -8,11 +8,8 @@ const getWhiteSpaceLength = function(statement) {
   return (whiteSpace[0] || '').length;
 };
 
-const canApplyRule = function(node, type, config) {
-  let end;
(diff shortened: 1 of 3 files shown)


rwjblue:
@ventuno - Would you mind reviewing this for me? I think you are the most familiar with this code at the moment...

rwjblue:
Thanks for reviewing @ventuno!

PR #469, opened 2018-08-03
Adds missing rules to `rules.md`

duggiefresh:


rwjblue:
Thank you @duggiefresh!

PR #474, opened 2018-08-15
element containing dot should not be considered ambiguous

### LINKED ISSUE #473: no-shadowed-elements should allow component names including a dot to start lower case
Currently this is considered a violation of `no-shadowed-elements` rule:

```
<Foo as |s|>
  <s.bar />
</Foo>
```

If local variable is a hash and therefore component is referred to be dot notation, it can't shadow elements. Therefore the hash should not be enforced to start with an uppercase char IMO.

jelhan:
Before a yielded hash had to start with uppercase char to not violate no-shadowed-elements rule even so it could not shadow any element cause having a dot in path.

Fixes #473

### DIFF
--- a/lib/rules/lint-no-shadowed-elements.js
+++ b/lib/rules/lint-no-shadowed-elements.js
@@ -21,8 +21,9 @@ module.exports = class NoShadowedElements extends Rule {
 
         let firstChar = node.tag.charAt(0);
         let startsWithUpperCase = firstChar === firstChar.toUpperCase() && firstChar !== firstChar.toLowerCase();
+        let containsDot = node.tag.includes('.');
 
-        if (!startsWithUpperCase) {
+        if (!startsWithUpperCase && !containsDot) {
           this.log({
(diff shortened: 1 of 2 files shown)


lib/rules/lint-no-shadowed-elements.js
```diff
@@ -21,8 +21,9 @@ module.exports = class NoShadowedElements extends Rule {
 
         let firstChar = node.tag.charAt(0);
         let startsWithUpperCase = firstChar === firstChar.toUpperCase() && firstChar !== firstChar.toLowerCase();
+        let containsDot = node.tag.includes('.');
 
-        if (!startsWithUpperCase) {
+        if (!startsWithUpperCase && !containsDot) {
```
rwjblue: This should be caught by the `this.isLocal(node)` check above 🤔

rwjblue: Hmm, actually I think I mis-read things a bit. This fix seems good to me, sorry about that...

PR #488, opened 2018-09-06
Add no bare string failing test

cbou:
The rule 'no-bare-string' should not produce an error with `<script>var foo = 'bar';</script>`


Turbo87:
looks like this has been fixed a while ago :)

PR #499, opened 2018-09-10
Fix an error when a route has a name ending in "components".

BnitoBzh:
This test fails when the application (or addon) has a route with a name ending in "components"
It's necessary to specified the parent folder name "templates/"

### DIFF
--- a/test/unit/rules/lint-no-outlet-outside-routes-test.js
+++ b/test/unit/rules/lint-no-outlet-outside-routes-test.js
@@ -25,6 +25,19 @@ generateRuleTests({
       }
     },
     {
+      template: '{{#outlet}}Route with "components" name{{/outlet}}',
+      meta: {
+        moduleId: 'app/templates/components.hbs',
+      },
+      result: {
+        message,
+        moduleId: 'app/templates/components.hbs',
+        source: '{{outlet}}',
+        line: 1,
+        column: 0
+      }
+    },
+    {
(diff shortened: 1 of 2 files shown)


rwjblue:
Can you add a test that demonstrates what this is fixing?

BnitoBzh:
Tests updated

test/unit/rules/lint-no-outlet-outside-routes-test.js
```diff
@@ -24,6 +24,19 @@ generateRuleTests({
         column: 0
       }
     },
+    {
+      template: '{{#outlet}}Route with "components" name{{/outlet}}',
+      meta: {
+        moduleId: 'app/templates/components.hbs',
```
rwjblue: Can you double check and confirm this would have failed previously? AFAICT the original code looked for `components/` which would not match this tests module...

BnitoBzh:
Hum subtle ...

I can update my test to improve it with this module id : 
`moduleId: 'app/templates/routes/components/index.hbs'`

But in practice i completely don't how Ember react when we create a root route named 'components' because it will generates a template name inside components folder for all sub routes ...

So my test only tests for second level route named 'components' ...

rwjblue:
I think this was fixed by https://github.com/ember-template-lint/ember-template-lint/pull/762. Sorry for the long delay here.

Issue #505, opened 2018-09-11
Incorrect indentation of multi-line angle bracket with block params


nightire:
If I understand correctly, with rules set as:

``` javascript
'attribute-indentation': {
  'element-open-end': 'last-attribute',
},
```

the following snippets should be passed:

``` htmlbars
<UiPhotoswipe
  @model={{this.mediaList}}
  @enabled={{this.zoomEnabled}}
as |medium|>
  <img src={{medium.src}} alt={{medium.name}}>
</UiPhotoswipe>
```

but instead it causes:

> Incorrect indentation of close bracket '>' for the element '<UiPhotoswipe>' beginning at L4:C11. Expected '<UiPhotoswipe>' to be at L3:C31.

however it is fine if without block params:

``` htmlbars
<UiPhotoswipe
  @model={{this.mediaList}}
  @enabled={{this.zoomEnabled}}>
  ...
</UiPhotoswipe>
```

Note: one line invocation does not have this issue.

PR #506, opened 2018-09-11
test: add failing case for multi-line angle bracket with block params

nightire:
Add test cases for issue #505


nightire:
@rwjblue done.

Issue #437, opened 2018-06-18
Add documentation for no-shadowed-elements


rwjblue:
I can't believe that I forgot this when I landed the rule 😭

Willibaur:
@rwjblue [PR-515](https://github.com/ember-template-lint/ember-template-lint/pull/515) in place for this tasks.
Thans

PR #515, opened 2018-09-18
[Issue-437] - Add-no-shadowed-elements

Willibaur:
@rwjblue Documentation added for `no-shadowed-elements` mentioned on #437.

I was not 100% sure about the explanation of the rule, pls correct me if I got it wrong, happy to make changes as needed.

Thanks

### DIFF
--- a/docs/rule/no-shadowed-elements.md
+++ b/docs/rule/no-shadowed-elements.md
@@ -1,0 +1,30 @@
+## no-shadowed-elements
+
+This rule prevents situations where a `variable` declared within an iterator has the same `name` than any other primitive element like `<div>, <body>` etc.
+
+The following code will throw an error:
+
+```hbs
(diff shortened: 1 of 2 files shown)


docs/rule/no-shadowed-elements.md
```diff
@@ -0,0 +1,30 @@
+## no-shadowed-elements
+
+This rule prevents situations where a `variable` declared within an iterator has the same `name` than any other primitive element like `<div>, <body>` etc.
```
rwjblue: Suggested tweak/replacement:

> This rule prevents ambiguity in situations where yielded block param which starts with a lower case letter, is also used within the block itself as an element name.

Willibaur:
@rwjblue Feedback implemented.

rwjblue:
Thank you!

PR #518, opened 2018-09-19
Add blacklist support to simple-unless rule

mattbalmer:
A project I'm working on has need of a `blacklist` feature in conjunction with the 'simple-unless' rule. We would like to allow most cases for `{{unless}}`, but prevent certain helpers from being used with them.

Code changes include adding that blacklist support, as well as modifying `maxHelpers` to treat a value of `-1` as 'no limit'. This is useful because otherwise, to use the `blacklist`, one would need to specify an arbitrarily large number for `maxHelpers` (as it would raise errors on helpers _not_ in the `blacklist` when the limit, defaulted to 0, was reached).

### DIFF
--- a/lib/rules/lint-simple-unless.js
+++ b/lib/rules/lint-simple-unless.js
@@ -12,6 +12,7 @@ const messages = {
 
 const DEFAULT_CONFIG = {
   whitelist: [],
+  blacklist: [],
   maxHelpers: 0,
 };
 
@@ -23,6 +24,8 @@ function isValidConfigObjectFormat(config) {
     if (value === undefined) {
       config[key] = DEFAULT_CONFIG[key];
     } else if (key === 'whitelist' && !valueIsArray) {
+      return false;
+    } else if (key === 'blacklist' && !valueIsArray) {
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


bradcypert:
I can see how this is useful. Seems like a good feature to have.

lib/rules/lint-simple-unless.js
```diff
@@ -142,6 +147,14 @@ module.exports = class LintSimpleUnless extends Rule {
             this._logMessage(message, loc.line, loc.column, actual);
           }
 
+          if (blacklist.length > 0 && blacklist.indexOf(param.path.original) > -1) { // blacklist.includes(param.path.original)
```
rwjblue: I think we can use `.includes` now (we dropped Node 4 support a little while back),  mind swapping to that?

rwjblue:
@mattbalmer - 👋 think you might have time to circle back to this one? Again, I'm sorry I let it linger so long after you initially submitted it...

lib/rules/lint-simple-unless.js
mattbalmer: Sounds good! Also updated a similar usage in the whitelist check above.

mattbalmer:
Hey, no worries and sorry for my own slow response! Your suggestion has been included, and the styles updated to match the new linter rules. Let me know if you spot any other changes you'd like!
Thanks

rwjblue:
Thank you!

PR #526, opened 2018-09-24
add failing test for multi-line ember action

NullVoxPopuli:


dcyriller:
@NullVoxPopuli is nesting action really a thing? Can the issue be reproduced with maybe a concat helper instead of nested actions?

NullVoxPopuli:
It in with the queue helper, yeah ;)

Though, probs not with the action helper with octane out.

But the repro just needs to have sub expressions on multiple lines

rwjblue:
I'd definitely accept a fix for this, but I'm going to close the failing test PR for now since in our default setup we actually don't allow any form of `{{action`/`(action`.

Issue #527, opened 2018-09-24
Custom path for Config?


krivaten:
Looking through the source, it seems we are not currently able to define a custom location for the config. I see a few places where `options.config` is checked, but nowhere it is being set. Is this something you would accept a PR for?

If so, I was thinking of checking the `process.argv` for something like "config-path" before the `new Linter()` is invoked within "ember-template-lint.js" unless you had a different approach you would prefer?

rwjblue:
Definitely seems reasonable to have a `—config-path` type command line argument...

PR #528, opened 2018-09-25
[ISSUE-527] Add option for config-path

krivaten:
## Problem:
If someone wants to use or share a single lintrc file across multiple in-repo addons, there is no way to do so without symlinking or creating a separate repo to share configs, which could be overkill for some folks. Most of the necessary code is already in place to enable this feature, we just need a place to define the variable.

----

## Current Behavior:
No way to set a custom `configPath`.

## Solution:
Updated the `./bin/ember-template-lint.js` to allow a user to pass in a `--config-path` argument followed by the relative path to the desired location. Doing so will result in the defining of the `configPath` option on the `Linter` Class.

Note: To do this, I had to move where `Linter` is invoked, which required me to add the `linter` to the list of arguments that gets passed to the `lintFile` function.

Example: `ember-template-lint . --config-path ../../.my-template-lintrc.js`

Fixes https://github.com/ember-template-lint/ember-template-lint/issues/527

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ Example usage:
 
 # ignore warnings / only report errors
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --quiet
+
+# define custom config path
+./node_modules/.bin/ember-template-lint app/templates/application.hbs --config-path .my-template-lintrc.js
 ```
 
 ### ESLint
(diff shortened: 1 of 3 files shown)


krivaten:
Of course, sorry it totally slipped by mind! If you want anything more in depth, just let me know.

PR #541, opened 2018-10-15
Make example pass rule

benmurden:
Example as it is written does not pass the rule `attribute-indentation`. This makes the example passing.

### DIFF
--- a/docs/rule/attribute-indentation.md
+++ b/docs/rule/attribute-indentation.md
@@ -110,7 +110,8 @@ Block form (HTML)
     href="https://www.emberjs.com"
     class="emberjs-home link"
     rel="noopener"
-    target="_blank">
+    target="_blank"
+  >
     Ember JS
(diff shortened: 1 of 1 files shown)


benmurden:
https://github.com/benmurden/ember-template-lint-test

This is a newly initialized ember app with two routes, `passing` and `failing`, each template passing and failing respectively with default options.

`passing.hbs`
```handlebars
<div>
  <a
    href="https://www.emberjs.com"
    class="emberjs-home link"
    rel="noopener"
    target="_blank"
  >
    Ember JS
  </a>
</div>
```

`failing.hbs`
```handlebars
<div>
  <a
    href="https://www.emberjs.com"
    class="emberjs-home link"
    rel="noopener"
    target="_blank">
    Ember JS
  </a>
</div>
```

rwjblue:
Thank you for confirming!

Issue #558, opened 2018-11-05
Audio/Video events break rule no-invalid-interactive


HenryVonfire:
Attaching actions to Audio/Video events break the rule `no-invalid-interactive`. Example:

```
<video 
  controls
  id="myVideo" 
  width="320" 
  height="176" 
  oncanplay={{action "canplay"}}  
>
 ...
</video>

```

> error  Interaction added to non-interactive element  no-invalid-interactive

rwjblue:
Thank you for reporting! Would you mind adding a test and adding (possibly with a refactor) the exceptions to the rule?  The current exceptions are done here:

https://github.com/ember-template-lint/ember-template-lint/blob/9fc0ebe1c3f0aa74b43ffa18f960573e3a490d16/lib/rules/lint-no-invalid-interactive.js#L77-L85

HenryVonfire:
Should I ignore specific media events or just ignore completely the tags `audio` and `video`?

rwjblue:
Good question, I'd like to say that `audio` and `video` should just be considered "interactive elements". If that is the case, then all we'd need to do is add to this list:

https://github.com/ember-template-lint/ember-template-lint/blob/9fc0ebe1c3f0aa74b43ffa18f960573e3a490d16/lib/helpers/is-interactive-element.js#L5-L15

Once added to that list, the rule should automatically allow `audio` / `video` to have `{{action`...

rwjblue:
Lets go in that direction while I try to get some clarification from some of our community A11Y experts...

MelSumner:
Here is the specification reference: 
https://dev.w3.org/html5/pf-summary/interactive-elements.html

Here is a list- and it has qualifiers. The rule should check for the qualifiers. 
- `a`
- `audio` (if the `controls` attribute is present) 
- `bb` 
- `button` 
- `details` 
- `embed` 
- `iframe` 
- `img` (if the `usemap` attribute is present) 
- `input` (if the `type` attribute is not in the hidden state) 
- `keygen` 
- `label` 
- `menu` (if the `type` attribute is in the tool bar state) 
- `object` (if the `usemap` attribute is present) 
- `select` 
- `textarea` 
- `video` (if the `controls` attribute is present)

rwjblue:
Perfect, thank you!

rwjblue:
This was fixed by #559

PR #559, opened 2018-11-05
Added video and audio with control attribute as interactive elements

HenryVonfire:
Fix for #558


lib/helpers/is-interactive-element.js
```diff
@@ -11,7 +11,9 @@ const INTERACTIVE_TAG_NAMES = [
   'keygen',
   'label',
   'select',
-  'textarea'
+  'textarea',
+  'audio',
```
rwjblue: Please remove from `INTERACTIVE_TAG_NAMES` and instead create a helper method like:

```js
function isAudioVideoWithControls(node) {
  return (node.tag === 'audio' || node.tag === 'video') && ast.hasAttribute(node, 'controls');
}
```

Then add a snippet to `reasons` function (just below) like the following:

```js
if (isAudioVideoWithControls(node)) {
  return `an <${node.tag}> element with the \`controls\` attribute`;
}
```

HenryVonfire:
Added requested changes and fixed tests.

PR #571, opened 2018-12-03
Add ability to put RegExp into rules['no-implicit-this'].allow

iki6:


lib/rules/lint-no-implicit-this.js
```diff
@@ -82,8 +90,9 @@ module.exports = class StrictPaths extends Rule {
           let valid = path.data ||
                       path.this ||
                       this.isLocal(path) ||
-                      this.config.allow.indexOf(path.original) > -1;
-
+                      this.config.allow.some((item) => {
+                        return item instanceof RegExp ? item.test(path.original) : item === path.original
```
H1D: You have nice `isRegExp()` function for this. Maybe it is better to use it

H1D:
looks like you forgot to include `test/fixtures/test-selector` directory

PR #573, opened 2018-12-03
Allow RegExp's with no-implicit-this 

iki6:
The rule that permits `rules['no-implicit-this'].allow` array to include only strings prevents efficient work with ember-test-selectors (it is necessary to put all of `data-test-*` attributes in allow array, but this list may be large)


rwjblue:
This test for example:

https://github.com/ember-template-lint/ember-template-lint/blob/4b2cec3eaee6b41d0910ad76256c1275ecc75a1f/test/unit/rules/lint-no-implicit-this-test.js#L24-L27

Issue #572, opened 2018-12-03
Incorrect attribute-indentation for triple curly invocation


hakubo:
in a template with triple curly e.g.
```hbs
<div>
  {{{i18n
    param=true
    otherParam=false
  }}}
</div>
```

I get something like:
`Incorrect indentation of close curly braces '}}' for the component '{{i18n}}' beginning at L11:C5. Expected '{{i18n}}' to be at L11:C4.  attribute-indentation`

The workaround is to delete one space before closing }}} - leading to one space indent.

rwjblue:
Thank you for reporting! Definitely a bug. Would you mind submitting a failing test case PR (so we can confirm when we get it fixed)?

hakubo:
@rwjblue done ^
I think the issue is somewhere here
https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/lint-attribute-indentation.js#L59

but my understanding of it is too shallow to proceed. I can try to fix it but I'd need some guidance.

lifeart:
This fixing test:
```js
validateCloseBrace(node, nextLocation) {
   const isNotEscaped = node.escaped === false;
   if (isNotEscaped) {
     expectedStartLocation.column++;
   }
}
```

PR #578, opened 2018-12-10
Add tripple curly test

hakubo:
A test for this bug https://github.com/ember-template-lint/ember-template-lint/issues/572


Turbo87:
looks like this got fixed in https://github.com/ember-template-lint/ember-template-lint/pull/845

PR #580, opened 2018-12-15
Rejects invalid configPath configuration

xcambar:
This PR ensures that the process breaks (with an informative message) when the user provides a path to a config file that does not exist.

The current (read: before the PR) behaviour is that, because the module fails to load, it is replaced by a default configuration (empty set of rules), which is disconcerting when you're editing a config file and see no result because you typo'ed the path :p

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -72,8 +72,7 @@ function checkConfigPath() {
   var configPathIndex = process.argv.indexOf('--config-path');
   var configPath = null;
   if (configPathIndex > -1) {
-    var configPathValue = process.argv[configPathIndex + 1];
-    configPath = path.join(process.cwd(), configPathValue);
+    configPath = process.argv[configPathIndex + 1];
   }
 
   return configPath;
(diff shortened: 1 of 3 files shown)


rwjblue:
Thank you!

bin/ember-template-lint.js
```diff
@@ -83,7 +82,14 @@ function run() {
   var exitCode = 0;
 
   var configPath = checkConfigPath();
-  var linter = new Linter({ configPath });
+  var linter;
+  try {
+    linter = new Linter({ configPath });
+  } catch (e) {
```
rwjblue: Sorry I missed this in review, I don't understand why we need to `console.error(e.message)` and `process.exit()` here. If `new Linter({ configPath })` throws an error the **default** thing is to log that error to the console and exit the process, right?

xcambar: As far as I remember, I thought the user would be better off with a simpler (more understandable, less technical?) output message than with the raw output from Node's error handler.

That is debatable but maybe then it just raises the question of error management in general. YMMV.

PR #582, opened 2018-12-19
Makes statusForModule compatible with absolute paths

xcambar:
This PR allows to define pending statuses with relative paths (as before)
and unifies the paths when the files to be checked are given with absolute paths.

In other words, this allows to define pending modules as `pending: ['relative/path/to/module']`
whil calling `ember-template-lint /home/user/app/relative/path/to/module` from `/home/user/app`.


rwjblue:
Thank you!

PR #584, opened 2018-12-24
feat: add rule 'no-extra-mut-helper-argument'.

bmish:
Adds new rule to catch the common mistake of passing an extra `value` argument to the `mut(attr)` helper.


bmish:
Anyone want to have a look?

PR #586, opened 2018-12-28
Lints from stdin

xcambar:
Hi!
Similar to ESLint, this PR allows `ember-template-lint` to lint from stdin, considering the argument `--stdin-filename` (as used in ESLint) is present with a value.

This is ver useful for editors that often allow linting to happen during edition, _ie_ before the file is actually saved to disk.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -8,8 +8,17 @@ var globby = require('globby');
 var Linter = require('../lib/index');
 const chalk = require('chalk');
 
+let [templatePatterns, args] = process.argv.slice(2).reduce(function([files, options], arg) {
+  if (options.length || (arg.slice(0, 2) === '--')) {
+    options = options.concat(arg);
+  } else {
+    files = files.concat(arg);
+  }
+  return [files, options];
+}, [[], []]);
+
 function printErrors(errors) {
-  const quiet = process.argv.indexOf('--quiet') !== -1;
+  const quiet = args.indexOf('--quiet') !== -1;
 
   let errorCount = 0;
   let warningCount = 0;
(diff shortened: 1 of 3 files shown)


rwjblue:
Haven't had a chance to review just yet, but mind rebasing so I can (hopefully) merge if all is well?

xcambar:
Done, thanks for your interest!

scottkidder:
Would love to see this get merged to improve integration with https://github.com/w0rp/ale

xcambar:
Tools like `flycheck` parse the output as json when available (here with the `--json` option). They use the `filePath` to reconcile with the current buffer.

That's why it is needed to work properly, because otherwise, the property mentioned above would only contain the name of the temp file created by ember-template-lint.

Note: This options also exists for ESLint. Behaviour differs slightly though.

rwjblue:
> the property mentioned above would only contain the name of the temp file created by ember-template-lint

Why does ember-template-lint need to create a temporary file at all? Wouldn't it just emit the JSON on stdout?

xcambar:
> Why does ember-template-lint need to create a temporary file at all?

It does so for input because the original implementation of `lintFile` takes a file path and I wanted to deliver value without modifying the existing APIs too much. See
https://github.com/ember-template-lint/ember-template-lint/blob/master/bin/ember-template-lint.js#L60-L63

> Wouldn't it just emit the JSON on stdout?

This is what happens.

seanCodes:
Any chance this could get merged soon? I’m trying to create a SublimeLinter plugin and lack of stdin support is an issue.

sukima:
I have a simpler solution (lest code changes) updated to the latest on master at PR #783

PR #591, opened 2019-01-06
feat: update 'style-concatenation' rule to catch an unsafe usage of the 'concat' helper.

bmish:
This change includes:
* Updates existing rule to catch an additional unsafe case
* General rule/doc/test cleanup/improvement


rwjblue:
Thank you!

PR #614, opened 2019-01-17
feat: add new rule 'no-element-event-actions'.

bmish:
## no-element-event-actions

Using HTML element event properties such as `onclick` for Ember actions is not recommended for the following reasons:

* It doesn't work for SVGs (since there is no `onclick` property of `SVGElement`).
* It can lead to confusing and unexpected behavior. For a comprehensive explanation of why, read [Deep Dive on Ember Events].

```hbs
<button onclick={{action "submit"}}>Submit</button>
```

```hbs
<button {{action "submit"}}>Submit</button>
```

```hbs
<button {{action "submit" on="doubleClick"}}>Submit On Double Click</button>
```

### References

* [Deep Dive on Ember Events]
* Ember [Template Action](https://guides.emberjs.com/release/templates/actions/) documentation
* [List of DOM Events](https://developer.mozilla.org/en-US/docs/Web/Events)

[Deep Dive on Ember Events]: https://medium.com/square-corner-blog/deep-dive-on-ember-events-cf684fd3b808

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -14,6 +14,7 @@
 * [no-bare-strings](rule/no-bare-strings.md)
 * [no-debugger](rule/no-debugger.md)
 * [no-duplicate-attributes](rule/no-duplicate-attributes.md)
+* [no-element-event-actions](rule/no-element-event-actions.md)
 * [no-extra-mut-helper-argument](rule/no-extra-mut-helper-argument.md)
 * [no-html-comments](rule/no-html-comments.md)
 * [no-implicit-this](rule/no-implicit-this.md)
--- a/test/unit/rules/lint-no-element-event-actions-test.js
+++ b/test/unit/rules/lint-no-element-event-actions-test.js
@@ -1,0 +1,53 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const ERROR_MESSAGE = require('../../../lib/rules/lint-no-element-event-actions').ERROR_MESSAGE;
+
+generateRuleTests({
+  name: 'no-element-event-actions',
+
+  config: true,
+
+  good: [
+    '<button></button>',
+    '<button type="button" on={{action "myAction"}}></button>',
+    '<button type="button" onclick="myFunction()"></button>',
+    '<button type="button" {{action "myAction"}}></button>',
+    '<button type="button" value={{value}}></button>',
+    '{{my-component onclick=(action "myAction") someProperty=true}}',
+    '<SiteHeader @someFunction={{action "myAction"}} @user={{this.user}} />',
+  ],
+
+  bad: [
+    {
+      template:
+        '<button onclick={{action "myAction"}} onfocus={{action "myAction"}} otherProperty=true></button>',
+
+      results: [
+        {
+          message: ERROR_MESSAGE,
+          source: 'onclick={{action "myAction"}}',
+          line: 1,
+          column: 8,
+        },
+        {
+          message: ERROR_MESSAGE,
+          source: 'onfocus={{action "myAction"}}',
+          line: 1,
+          column: 38,
+        },
+      ],
+    },
+
+    {
+      template: '<SiteHeader onclick={{action "myAction"}} @user={{this.user}} />',
+
+      result: {
+        message: ERROR_MESSAGE,
+        source: 'onclick={{action "myAction"}}',
+        line: 1,
+        column: 12,
+      },
+    },
+  ],
+});
--- a/docs/rule/no-element-event-actions.md
+++ b/docs/rule/no-element-event-actions.md
@@ -1,0 +1,30 @@
+## no-element-event-actions
+
+Using HTML element event properties such as `onclick` for Ember actions is not recommended for the following reasons:
+
+* It doesn't work for SVGs (since there is no `onclick` property of `SVGElement`).
+* It can lead to confusing and unexpected behavior when mixed with normal action usage. For a comprehensive explanation of why, read [Deep Dive on Ember Events].
+
+This rule **forbids** the following:
+
+```hbs
+<button onclick={{action "submit"}}>Submit</button>
+```
+
+This rule **allows** the following:
+
+```hbs
+<button {{action "submit"}}>Submit</button>
+```
+
+```hbs
+<button {{action "submit" on="doubleClick"}}>Submit On Double Click</button>
+```
+
+### References
+
+* [Deep Dive on Ember Events]
+* Ember [Template Action](https://guides.emberjs.com/release/templates/actions/) documentation
+* [List of DOM Events](https://developer.mozilla.org/en-US/docs/Web/Events)
+
+[Deep Dive on Ember Events]: https://medium.com/square-corner-blog/deep-dive-on-ember-events-cf684fd3b808
--- a/lib/helpers/ast-node-info.js
+++ b/lib/helpers/ast-node-info.js
@@ -12,6 +12,10 @@ function isTextNode(node) {
   return node.type === 'TextNode';
 }
 
+function isAttrNode(node) {
+  return node.type === 'AttrNode';
+}
+
 function isCommentStatement(node) {
   return node.type === 'CommentStatement';
 }
@@ -121,4 +125,5 @@ module.exports = {
   isMustacheStatement: isMustacheStatement,
   isNonConfigurationHtmlComment: isNonConfigurationHtmlComment,
   isTextNode: isTextNode,
+  isAttrNode: isAttrNode,
 };
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -18,6 +18,7 @@ module.exports = {
   'no-bare-strings': require('./lint-no-bare-strings'),
   'no-debugger': require('./lint-no-debugger'),
   'no-duplicate-attributes': require('./lint-duplicate-attributes'),
+  'no-element-event-actions': require('./lint-no-element-event-actions'),
   'no-extra-mut-helper-argument': require('./lint-no-extra-mut-helper-argument'),
   'no-html-comments': require('./lint-no-html-comments'),
   'no-implicit-this': require('./lint-no-implicit-this'),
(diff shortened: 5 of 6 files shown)


docs/rule/no-dom-events.md
```diff
@@ -0,0 +1,27 @@
+## no-dom-events
+
+Mixing Ember actions and DOM events can lead to confusing and unexpected behavior. For a comprehensive explanation of why, read [Deep Dive on Ember Events]. As a result, it is recommended to stick entirely to Ember actions.
+
+This rule **forbids** the following:
+
+```hbs
+<button onclick={{action (action "submit")}}>Submit</button>
```
rwjblue: The nested `(action` is not needed here (AFAICT it doesn't do anything...).

I'd write this as:

```hbs
<button onclick={{action "submit"}}>Submit</button>
```

And the allowed version would be:

```hbs
<button {{action "submit"}}>Submit</button>
```

docs/rule/no-dom-events.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+This rule **allows** the following:
+
+```hbs
+<button {{action (action "submit")}}>Submit</button>
+```
+
+```hbs
+<button {{action (action "submit") on="doubleClick"}}>Submit On Double Click</button>
```
rwjblue: Here again, the `{{action (action` bit is not needed/useful when passing an actual string action (it would be useful if passing a function though).

docs/rules.md
```diff
@@ -13,6 +13,7 @@
 * [no-attrs-in-components](rule/no-attrs-in-components.md)
 * [no-bare-strings](rule/no-bare-strings.md)
 * [no-debugger](rule/no-debugger.md)
+* [no-dom-events](rule/no-dom-events.md)
```
rwjblue: Not sure where to do it, so I picked here 😄 

We should bikeshed the name of this rule a bit. We certainly don't want to infer that "dom events" are bad (indeed `{{action` actually _uses_ DOM events). Some names off the cuff that strike a better balance:

* `no-element-event-callbacks`
* `no-element-property-events`

docs/rule/no-dom-events.md
```diff
@@ -0,0 +1,27 @@
+## no-dom-events
```
rwjblue: I'd also like it if this rule pointed out some of the other pitfalls to using `onclick=` style. One common one is that it doesn't work for svg (since there is no `onclick` property of `SVGElement`)...

lib/rules/lint-no-dom-events.js
```diff
@@ -0,0 +1,36 @@
...
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+const ERROR_MESSAGE = 'Use Ember actions instead of DOM events.';
+
+module.exports = class NoDomEvents extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        const domEventAttribute = node.attributes.find(isDomEventAttributeWithAction);
```
rwjblue: I could be wrong, but doesn't `[].find` return the first item? I think we would actually want to `[].filter` and report _all_ of them...

Imagine:

```hbs
<button onclick={{action 'foo'}} ondblclick={{action 'bar'}}></button>
```

We would want both of those reported, not just the first one.

lib/rules/lint-no-dom-events.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+      },
+    };
+  }
+};
+
+function isDomEventAttributeWithAction(node) {
+  return (
+    AstNodeInfo.isAttrNode(node) &&
+    node.name.startsWith('on') &&
```
rwjblue: I'm not sure this is good enough 🤔, how certain are we that there are no legitimate attribute names starting with `on`?

docs/rule/no-dom-events.md
bmish: Fixed.

docs/rule/no-dom-events.md
bmish: Fixed.

docs/rules.md
bmish: I renamed it to `no-element-event-attribute-actions`. How does that sound?

docs/rule/no-dom-events.md
bmish: Added a list with that example.

lib/rules/lint-no-dom-events.js
bmish: Good point. Switched to `filter` and added a test for this.

lib/rules/lint-no-dom-events.js
bmish: My understanding from looking at the [full list](https://www.w3schools.com/tags/ref_attributes.asp) of HTML attributes is that all event attributes and only event attributes begin with `on`. 

I implemented the rule with this attribute prefix check for simplicity so that the rule doesn't have to know the full list of nearly 100 [event attributes](https://www.w3schools.com/jsref/dom_obj_event.asp).

Alternatively, if I wanted to check for full attribute name matches, I could store the full event attribute list inside the rule (and hope the list doesn't change often), or get the list from another package like [event-attributes](https://www.npmjs.com/package/event-attributes) (although that package seems to have very few users).

Assuming my initial assumption is correct, the alternatives seem like overkill.

docs/rules.md
rwjblue: I worry about muddying the waters between attributes and properties even further. The specific thing that we are trying to avoid is setting these as _properties_ (not attributes). This is because Ember (_currently_) uses a "properties first" mentality.

<details>
<summary>Specifically we do _something_ akin to this...</summary>

```js
let normalizedPropertiesByTag = new Map();

function selectAttributeOrPropertyFor(element, name) {
  let propertiesForTag = normalizedPropertiesByTag.get(element.tagName);
  
  if (propertiesForTag === undefined) {
    let propertiesForTag = new Map();
    for (let key in element) {
      propertiesForTag.set(key.toLowerCase(), key);
    }
    normalizedPropertiesByTag.set(element.tagName, propertiesForTag);
  }

  let propertyName = propertiesForTag.get(name.toLowerCase());
  if (propertyName) {
    return { type: 'property', name: propertyName };
  }

  return { type: 'attribute', name };
}

function populateAttributeOrProperty(element, attributeOrProperty, value) {
  let details = selectAttributeOrPropertyFor(element, attributeOrProperty);

  if (details.type === 'property') {
    element[details.name] = value;
  } else {
    element.setAttribute(details.name, value);
  }
}

// when rendering for <div data-foo="asdf"></div>
let el = document.createElement('div');
populateAttributeOrProperty(el, 'data-foo', 'asdf');
```
</details>
<br>

As you dig into the difference between attributes and properties, you see that the specific thing that we are trying to avoid is using `onclick` as a _property_ **not** an attribute. 

You can see that by testing this out:

```js
  let el = document.createElement("button");

  // setup the button text
  let text = document.createTextNode("Click Me -- Attribute!");
  el.appendChild(text);

  // trying to set a "real" function to an attribute
  el.setAttribute("onclick", () => alert("clicked -- attribute!"));
  document.body.appendChild(el);
```

Note that the alert does not fire. If you inspect the DOM you will see that what is done is that the function is `.toString()`'ed before appending to the attribute value. The only way to preserve the actual function is if we set as a property!

(I made a [codepen here](https://codepen.io/rwjblue/pen/wNKKrY) to make it easier to test this out).

rwjblue: How do you feel about `no-element-event-actions` or `no-element-event-property-actions`?

docs/rule/no-element-event-attribute-actions.md
```diff
@@ -0,0 +1,30 @@
+## no-element-event-attribute-actions
+
+Using HTML element event attributes such as `onclick` for Ember actions is not recommended for the following reasons:
+
+* It doesn't work for SVGs (since there is no `onclick` attribute of `SVGElement`).
```
rwjblue:
```suggestion
* It doesn't work for SVGs (since there is no `onclick` property of `SVGElement`).
```

docs/rules.md
bmish: Thanks for the explanation of the difference! I'm fine with either name you suggested. I renamed it to `no-element-event-actions`. I also updated all usages of `attribute` to `property` throughout the files for the new rule.

docs/rule/no-element-event-attribute-actions.md
bmish: Fixed.

bmish:
@rwjblue are these correct/good examples for angle bracket component invocation? I added these.

Good: `<SiteHeader click={{action "myAction"}} @user={{this.user}} />`
Bad: `<SiteHeader onclick={{action "myAction"}} @user={{this.user}} />`

test/unit/rules/lint-no-element-event-actions-test.js
```diff
@@ -0,0 +1,53 @@
...
+  config: true,
+
+  good: [
+    '<button></button>',
+    '<button type="button" on={{action "myAction"}}></button>',
+    '<button type="button" onclick="myFunction()"></button>',
+    '<button type="button" {{action "myAction"}}></button>',
+    '<button type="button" value={{value}}></button>',
+    '{{my-component onclick=(action "myAction") someProperty=true}}',
+    '<SiteHeader click={{action "myAction"}} @user={{this.user}} />',
```
rwjblue: Any of the things where the attribute name is not prefixed by `@` will be set as attribute / property (depending on the name and whatnot like I mentioned before) on whatever element has `...attributes` on it. 

If the idea here is to pass in a click function that the component as an argument that it can use in its JS file (ala `this.click()`) then you would do:

```suggestion
    '<SiteHeader @click={{action "myAction"}} @user={{this.user}} />',
```

bmish: Updated!

rwjblue:
Thank you @bmish!

bmish:
FYI @mariechatfield you'll be interested in this new rule!

PR #618, opened 2019-01-20
fix: allow `{{else}} {{#unless ...}}` in the 'simple-unless' rule and make corresponding update to 'no-negated-condition' rule.

bmish:
There are two relevant cases:
1. `{{#if condition}} ... {{else unless condition}} ... {{/if}}`
2. `{{#if condition}} ... {{else}} {{#unless condition}} ... {{/unless}} {{/if}}`

Currently, they are both banned by the 'simple-unless' rule, but I suspect that (2) was not originally intended to be banned (banning (2) may have been a mistake).

Here are the reasons why I don't think (2) was intended to be banned:
1. These two cases are difficult to distinguish with the AST.
2. The error message returned by both of these cases mentions (1) but not (2): 'Using an `{{else unless}}` block should be avoided.'
3. (1) is obviously undesirable whereas (2) isn't. In fact, I've heard multiple engineers voice surprise that (2) is currently banned.

As a result, I have updated 'simple-unless' to allow (2).

In addition, I have also update the related 'no-negated-condition' rule to allow suggesting (2), whereas previously it specifically avoided suggesting (2) to avoid conflicting with the 'simple-unless' rule.

### DIFF
--- a/test/unit/rules/lint-no-negated-condition.js
+++ b/test/unit/rules/lint-no-negated-condition.js
@@ -26,6 +26,7 @@ generateRuleTests({
 
     // if ... else if ...
     '{{#if condition}}<img>{{else if condition}}<img>{{/if}}',
+    '{{#if condition}}<img>{{else if (not condition2)}}<img>{{/if}}', // we ignore `if ... else if ...`
     '{{#if (not condition)}}<img>{{else if (not condition2)}}<img>{{/if}}', // we ignore `if ... else if ...`
 
     // if ... else if ... else ...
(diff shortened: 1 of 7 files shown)


test/unit/rules/lint-no-negated-condition.js
```diff
@@ -181,6 +176,23 @@ generateRuleTests({
       },
     },
 
+    // Nested inside the body of an `else`:
+    {
+      template: `{{#if condition}}
+                 {{else}}
+                   {{! some comment }}
```
rwjblue: Can you add another "bad" test without this comment (I'd like one with the comment and one without it) and without whitespace separating the curlies?

```hbs
{{#if condition}}{{else}}{{#if (not condition)}}<div></div>{{/if}}{{/if}}
```

In this circumstance the resulting AST is _nearly identical_ to:

```hbs
{{#if condition}}{{else if (not condition)}}<div></div>{{/if}}{{/if}}
```

And I want to make sure we are still allowing the latter (since it seems like that was your intent) while disallowing the former...

bmish: Updated tests as you suggested.

PR #624, opened 2019-01-23
Add require-valid-alt-text rule.

lifeart:
refs:
https://github.com/evcohen/eslint-plugin-jsx-a11y/tree/master/src/rules
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/alt-text.md

@rwjblue can you please specify proper a11y rules location?


rwjblue:
> @rwjblue can you please specify proper a11y rules location?

In general all rules go in `lib/rules` (the only exception is for deprecations). I'd say we stick with `lib/rules`...

lifeart:
should a11y rules names to be prefixed somehow?

rwjblue:
> should a11y rules names to be prefixed somehow?

Each rule should be named for what it does. If we need to somehow group them, we can do that in various ways in the README or the main rules documentation.

docs/rule/a11y/alt-text.md
```diff
@@ -0,0 +1,79 @@
...
+
+Enforce that all elements that require alternative text have meaningful information to relay back to the end user. This is a critical component of accessibility for screenreader users in order for them to understand the content's purpose on the page. By default, this rule checks for alternative text on the following elements: `<img>`, `<area>`, `<input type="image">`, and `<object>`.
+
+
+This rule **forbids** the following:
+
+
+### `<img>`
+
+An `<img>` must have the `alt` prop set with meaningful text or as an empty string to indicate that it is an image for decoration.
```
MelSumner: the rule is that if it's a meaningful image, it should have a non-empty alt value. If it's not a meaningful image, it should have an empty alt attribute value and have the role of presentation or none. 

Okay to have:
``
OR
``
OR
`a stylized graphic of a female hamster`

Not okay: 
``
OR 
``

bmish:
There's an existing rule [img-alt-attributes](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/img-alt-attributes.md). Shouldn't we enhance that one / make the existing rule more generic instead of creating a new one that overlaps so much?

rwjblue:
@bmish - I was thinking that `img-alt-attributes` would be replaced by the new rule, ya.

lifeart:
@bmish yeah, it was my reference for this rule)
@MelSumner logic updated, tests added

docs/rule/require-valid-alt-text.md
```diff
@@ -0,0 +1,81 @@
...
+This rule **allows** the following:
+
+```hbs
+<area shape="poly" coords="113,24,211,0" href="inform.html" alt="Inform">
+```
+
+### References
+
+* See [WCAG Suggestion H37](https://www.w3.org/TR/WCAG20-TECHS/H37.html)
+* Ref [eslint-plugin-jsx-a11y/alt-text](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/alt-text.md)
```
MelSumner: please remove references to other eslint libraries- we should only be referencing the spec itself in our rules. 
It can be replaced with this: 
* Ref [WCAG Criterion 1.1.1 - Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -0,0 +1,98 @@
...
+          return;
+        }
+        if (isImg) {
+          const hasAltAttribute = AstNodeInfo.hasAttribute(node, 'alt');
+          const altValue = AstNodeInfo.elementAttributeValue(node, 'alt');
+          const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');
+          if (!hasAltAttribute) {
+            this.logNode({
+              message:
+                'img tags must have an alt attribute, either with meaningful text, or an empty string for decorative images.',
```
MelSumner: img elements must have an alt attribute. See https://www.w3.org/WAI/WCAG21/Techniques/html/H37 for more details.

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -0,0 +1,98 @@
...
+              node,
+            });
+          } else {
+            if (altValue === '') {
+              if (['presentation', 'none'].includes(roleValue)) {
+                return;
+              }
+              this.logNode({
+                message:
+                  'img tags must have an alt attribute, either with meaningful text, or an empty string for decorative images.',
```
MelSumner: img elements must have an alt attribute. See https://www.w3.org/WAI/WCAG21/Techniques/html/H37 for more details.

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -0,0 +1,98 @@
...
+          const hasValidAttributes = AstNodeInfo.hasAnyAttribute(node, [
+            'aria-label',
+            'aria-labelledby',
+            'alt',
+          ]);
+
+          if (!hasValidAttributes) {
+            this.logNode({
+              message:
+                '<input> elements with type="image" must have a text alternative through the `alt`, `aria-label`, or `aria-labelledby` attribute.',
```
MelSumner: '`<input>` elements with type="image" are no longer standard use. An `` element should be used instead. However if you must use it, it must have a text alternative through the use of one of the following: the `alt`, `aria-label`, or `aria-labelledby` attribute.'

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -0,0 +1,98 @@
...
+          this.logNode({
+            message:
+              'Embedded <object> elements must have alternative text by providing inner text, aria-label or aria-labelledby attributes.',
+            node,
+          });
+        } else if (isArea) {
+          if (!AstNodeInfo.hasAnyAttribute(node, ['aria-label', 'aria-labelledby', 'alt'])) {
+            this.logNode({
+              message:
+                'Each area of an image map must have a text alternative through the `alt`, `aria-label`, or `aria-labelledby` attribute.',
```
MelSumner: please add "See https://www.w3.org/WAI/WCAG21/Techniques/html/H24 for more information."

test/unit/rules/lint-require-valid-alt-text-test.js
```diff
@@ -0,0 +1,128 @@
...
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'require-valid-alt-text',
+
+  config: true,
+
+  good: [
+    '<img alt="hullo">',
```
MelSumner: For this to be a useful example, it should mimic meaningful text that would also be useful- like a description of something that could be purchased online, perhaps?

`a shirt with the ember logo across the front`

test/unit/rules/lint-require-valid-alt-text-test.js
```diff
@@ -0,0 +1,128 @@
...
+    '<object aria-labelledby="some-alt"></object>',
+    '<object aria-label="some-alt"></object>',
+    '<object>some text</object>',
+
+    '<area alt="some-alt">',
+    '<area aria-labelledby="some-alt">',
+    '<area aria-label="some-alt">',
+  ],
+
+  bad: [
```
MelSumner: please update the messages in this section to reflect the review above.

docs/rule/require-valid-alt-text.md
```diff
@@ -0,0 +1,81 @@
...
+This rule **forbids** the following:
+
+```hbs
+<img src="rwjblue.png">
+```
+
+This rule **allows** the following:
+
+```hbs
+<img src="rwjblue.png" alt="picture of Robert Jackson">
```
MelSumner: `a man standing in front of a group of people, giving a presentation about Ember.`

docs/rule/require-valid-alt-text.md
```diff
@@ -0,0 +1,81 @@
...
+```hbs
+<object width="128" height="256"></object>
+```
+
+This rule **allows** the following:
+
+```hbs
+<object width="128" height="256" title="Middle-sized"></object>
+<object width="128" height="256" aria-label="Middle-sized"></object>
+<object width="128" height="256" aria-labelledby="Middle-sized"></object>
```
MelSumner: the value of aria-labelledby needs to be an ID and we should likely make that clear by having it be a different value that the other two.

MelSumner: @lifeart I tried to make a change myself- but I'm not sure if it worked, so I'm going to wait for your feedback. :)

PR #630, opened 2019-01-24
feat: [a11y] img-redundant-alt

lifeart:
ref: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/img-redundant-alt.md)


lifeart:
@rwjblue done

lifeart:
> Changes look good, thank you!
> 
> Do you happen to know if there is a specific reference for this that we should add to the documents references section?

only https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/img-redundant-alt.md (
@MelSumner ?

PR #631, opened 2019-01-24
Add rule: [a11y] iframe-has-title

lifeart:
https://dequeuniversity.com/rules/axe/1.1/frame-title

https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/iframe-has-title.md


lifeart:
@rwjblue done

docs/rule/iframe-has-title.md
```diff
@@ -1,4 +1,4 @@
-## iframe-has-title
+## require-iframe-title
```
rwjblue: The file name here is wrong now

lib/rules/lint-require-iframe-title.js
```diff
@@ -28,12 +37,23 @@ module.exports = class IframeHasTitle extends Rule {
...
                 message: errorMessage,
                 node,
               });
               return;
+            } else if (this.knownTitles.includes(value)) {
+              this.logNode({
+                message:
+                  errorMessage +
+                  ' ' +
+                  `Value title="${title.value.chars}" already used for different iframe.`,
```
rwjblue: Do you think we could keep track of both the preexisting titles **and** their locations? I think it might be nice to include the location that we found the alternate title in this error message. What do you think?

lifeart: yeah, it make sense, will do it

lifeart: done

PR #633, opened 2019-01-24
Add rule: [a11y] no-positive-tabindex

lifeart:
https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_focus_03
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/tabindex-no-positive.md


lifeart:
@rwjblue done

linearza:
Hi - we have an external component library which needs a dynamic tabindex value set, defaulting to -1. With this rule thats unfortunately not possible - surely it should allow dynamic values like: `tabindex={{or @tabindex "-1"}}` considering we can have multiple components (which wrap the actual inputs) on one page, so we need to set it dynamically from outside

bmish:
@linearza can you open an issue about that?

PR #634, opened 2019-01-24
Add new rule: [a11y] no-distracting-elements

lifeart:
1. [Deque University - marquee](https://dequeuniversity.com/rules/axe/1.1/marquee)
2. [Deque University - blink](https://dequeuniversity.com/rules/axe/1.1/blink)
3. [eslint-plugin-jsx-a11y/no-distracting-elements](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-distracting-elements.md)

### DIFF
--- a/docs/rule/no-distracting-elements.md
+++ b/docs/rule/no-distracting-elements.md
@@ -20,6 +20,7 @@ This rule **forbids** the following:
 
 ### References
 
+1. [w3.org/TR/WCAG20-TECHS/F47](https://www.w3.org/TR/WCAG20-TECHS/F47.html)
 1. [Deque University - marquee](https://dequeuniversity.com/rules/axe/1.1/marquee)
-2. [Deque University - blink](https://dequeuniversity.com/rules/axe/1.1/blink)
-3. [eslint-plugin-jsx-a11y/no-distracting-elements](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-distracting-elements.md)
+1. [Deque University - blink](https://dequeuniversity.com/rules/axe/1.1/blink)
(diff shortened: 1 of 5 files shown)


MelSumner:
I think these need to link directly to the specification, or else we run the risk of getting it wrong because someone else did. 

In this case, we should direct users to https://www.w3.org/TR/WCAG20-TECHS/F47.html

MelSumner:
I like this rule- it's also a deprecated HTML element, so I'm not sure which path forward would be preferred for this addon. @rwjblue what do you think?

lifeart:
@rwjblue, @MelSumner  looks like we need an library with this set of features / restrictions, I believe it exists somewhere. Something like https://www.npmjs.com/package/html-validator-cli or https://www.npmjs.com/package/html-validator

rwjblue:
Hmm, why do we need an external library? The list I linked to should be good enough, what am I missing?

lifeart:
at https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features alot of cases and messages ( > 54 ), or we need to create kinda local library for stuff like this.

rwjblue:
Ya, I guess I'd start with a local array (not too bad), and if we find out that isn't good enough we can try to find a better source.

MelSumner:
There are more extensive things that I think we should put into Ember Language Server but I if you separate out attributes and elements separately this rule isn't too bad.

rwjblue:
Right. I'm totally happy to have _another_ rule that is about `no-obsolete-attributes` or some such, but we don't have to start there...

MelSumner:
I think we can close this one now that no-obsolete-elements is merged. What do you think? @lifeart @rwjblue

rwjblue:
Closed by #733

PR #636, opened 2019-01-24
Add new rule `require-context-role`

### LINKED ISSUE #626: [A11y] required-context-role
## Rule Requested
The required context role defines the owning container where this role is allowed. If a role has a required context, authors MUST ensure that an element with the role is contained inside (or owned by) an element with the required context role. For example, an element with `role="listitem"` is only meaningful when contained inside (or owned by) an element with `role="list"`.

## Roles to check
Format: role | required context role
- columnheader | row
- gridcell | row
- listitem | group or list
- menuitem | group, menu, or menubar
- menuitemcheckbox | menu or menubar
- menuitemradio | group, menu, or menubar
- option | listbox
- row | grid, rowgroup, or treegrid
- rowgroup | grid
- rowheader | row
- tab | tablist
- treeitem | group or tree

## Example(s)
Conformant Code: 
```
<div role="list">
  <div role="listitem">Item One</div>
  <div role="listitem">Item Two</div>
</div>
```

Non-conformant code:
```
<div>
  <div role="listitem">Item One</div>
  <div role="listitem">Item Two</div>
</div>
```

## Suggested Error Message
(Replace items in brackets with the item that violated the rule)

"You have an element with the role of {{role}} but it is missing the required (immediate) parent element of {{parent-element-options}}. Reference:  https://www.w3.org/TR/wai-aria-1.0/roles#{{role}}."

## Reference(s)
- https://www.w3.org/TR/wai-aria-1.0/roles#scope 
- https://www.w3.org/TR/wai-aria-1.0/roles#columnheader 
- https://www.w3.org/TR/wai-aria-1.0/roles#gridcell
- https://www.w3.org/TR/wai-aria-1.0/roles#listitem
- https://www.w3.org/TR/wai-aria-1.0/roles#menuitem
- https://www.w3.org/TR/wai-aria-1.0/roles#menuitemcheckbox
- https://www.w3.org/TR/wai-aria-1.0/roles#menuitemradio
- https://www.w3.org/TR/wai-aria-1.0/roles#option
- https://www.w3.org/TR/wai-aria-1.0/roles#row
- https://www.w3.org/TR/wai-aria-1.0/roles#rowgroup
- https://www.w3.org/TR/wai-aria-1.0/roles#rowheader
- https://www.w3.org/TR/wai-aria-1.0/roles#tab
- https://www.w3.org/TR/wai-aria-1.0/roles#treeitem

lifeart:
Fixes #626.


lifeart:
```hbs
<div role="list">
  {{#each someList as |item|}}
    {{list-item item=item}}
  {{/each}}
</div>
```
I think end-users may rewrite it like

```hbs
<div role="list">
  {{#each someList as |item|}}
    <ListItem role="listitem" @item={{item}}/>
  {{/each}}
</div>
```

lifeart:
```hbs
<div role="list">
  {{#each someList as |item|}}
    {{list-item item=item}}
  {{/each}}
</div>
```
and 
```hbs
<div role="list">
  {{#each someList as |item|}}
    <ListItem role="listitem" @item={{item}}/>
  {{/each}}
</div>
```
added as valid testcases

MelSumner:
> Do we think its relatively uncommon to separate the two things like this?

It's incredibly uncommon. They will likely have many other issues if they hit this scenario.

MelSumner:
I need to do a little more research before we merge this one- I want to make 100% sure that they need to be sibling elements.

lib/rules/lint-required-context-role.js
```diff
@@ -0,0 +1,105 @@
...
+              value: childRoleValue,
+              node: child,
+            };
+          })
+          .filter(item => item !== false);
+
+        //VALID_PARENTS_FOR_CHILD
+        childs.forEach(child => {
+          if (VALID_PARENTS_FOR_CHILD[child.value].indexOf(roleValue) === -1) {
+            child.node[CHILD_PROP_KEY] = {
```
Turbo87: it seems bad to me to mutate the AST here. is there no other way of tracking state like using `enter/exit` visitor callbacks? 🤔

test/unit/rules/lint-required-context-role-test.js
```diff
@@ -0,0 +1,204 @@
...
+  return result;
+}, {});
+Object.keys(VALID_PARENTS_FOR_CHILD).forEach(key => {
+  VALID_PARENTS_FOR_CHILD[key].sort();
+});
+const errorMessage = role => {
+  return `You have an element with the role of "${role}" but it is missing the required (immediate) parent element of "[${VALID_PARENTS_FOR_CHILD[
+    role
+  ].join(', ')}]". Reference: https://www.w3.org/TR/wai-aria-1.0/roles#${role}.`;
+};
```
Turbo87: we should `require()` the original function instead of duplicating the logic here.

lib/rules/lint-required-context-role.js
```diff
@@ -0,0 +1,105 @@
...
+    if (!(childName in result)) {
+      result[childName] = [];
+    }
+    result[childName].push(keyName);
+  });
+  return result;
+}, {});
+Object.keys(VALID_PARENTS_FOR_CHILD).forEach(key => {
+  VALID_PARENTS_FOR_CHILD[key].sort();
+});
```
Turbo87: we might want to add a Jest snapshot test to ensure that `VALID_PARENTS_FOR_CHILD` matches our expectations 🤔

Turbo87:
I tend to agree with @rwjblue. It seems to me that it is a quite common pattern to extract the loop content into a dedicated component, whether that uses `role=listitem` or not.

I think the rule generally is useful though, so including it here would seem fine to me, but IMHO we should not add it to the recommended ruleset due to the concerns above.

MelSumner:
@rwjblue @Turbo87 what's the path forward for this PR?

MelSumner:
@lifeart would you be able to update this PR and get it ready for review? I'd like to have it for the 3.0 release if possible.

lib/rules/lint-required-context-role.js
lifeart: as I understand, we could note use weakMaps for node refs, because of proxies :(

lifeart: just checked, it works, code updated

lifeart:
@MelSumner done

lifeart:
have no idea why CI is failing
image

bmish:
Run `yarn update` to fix the README.

lib/rules/require-context-role.js
```diff
@@ -0,0 +1,110 @@
...
+    ', '
+  )}]". Reference: https://www.w3.org/TR/wai-aria-1.0/roles#${role}.`;
+}
+
+const metaForNode = new WeakMap();
+
+module.exports = class RequireContextRole extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
```
rwjblue, line 49: Can we refactor this use the path ancestors stuff instead?

rwjblue (commented):
Also needs a rebase

lib/rules/require-context-role.js
lifeart: done

lifeart:
@rwjblue should I add it to list of `a11y` rules?

lifeart:
@bmish, @rwjblue any thoughts here?

docs/rule/require-context-role.md
```diff
@@ -0,0 +1,58 @@
+# require-context-role
+
+## `<* role><* role /></*>`
```
bmish, line 3: What is this header?

bmish (approved):
This looks ready to merge to me.

PR #641, opened 2019-01-26
docs: add template for rule docs.

bmish:
Goal: help authors of new rules write better and more consistent documentation.


bmish:
@Turbo87 want to have a look?

PR #651, opened 2019-02-10
Added a related rules section

keanedawg:
`no-element-event-actions` and `no-action-modifiers` appear to be duplicate rules. I am adding a `related rules` section below.


keanedawg:
Moving to a separate pull request

PR #659, opened 2019-02-25
feat(bin/ember-template-lint.js): Add toggle flag for ignored files

velrest:
Add a flag to the command so the by default ignored files `['**/dist/**', '**/tmp/**', '**/node_modules/**']` can be toggled. This is usefull if for instance your editor lint plugin (vim/ale) copies the file to be linted into `/tmp`. This prevents the template from beeing linted as `'**/tmp/**'` applies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ Example usage:
 
 # define custom config path
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --config-path .my-template-lintrc.js
+
+# dont ignore `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
+./node_modules/.bin/ember-template-lint app/templates/application.hbs --no-ignore .my-template-lintrc.js
 ```
 
 ### ESLint
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -69,7 +69,10 @@ function getRelativeFilePaths() {
     .reduce((filePaths, fileArg) => {
       return filePaths.concat(
         globby.sync(fileArg, {
-          ignore: ['**/dist/**', '**/tmp/**', '**/node_modules/**'],
+          ignore:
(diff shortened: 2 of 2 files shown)


velrest:
If someone else is having this problem specifically with vim and ale this is a valid workaround for now:
```vim
let g:ale_handlebars_embertemplatelint_use_global=1
let g:ale_handlebars_embertemplatelint_executable="path_to_your_ember-template-lint@v1.0.0"
```

rwjblue:
When the wildcards are handled by the terminal they are explicitly expanded into specific files, in terminals that I tested `zsh` and `bash` this only happened when specifically tab completing inside the glob itself. In that case we will only ever see the list of actual paths to `.hbs` files and not the glob, but globby will also not traverse additional directories either.  In other words, if the shell expands the glob before invoking our script there is nothing we can do and we do not need the ignores at all.

velrest:
Im sorry if i misunderstood but AFAIK the glob is not only expanded when using tab completion. If the glob argument is not quoted, template-lint never receives the glob string as the shell (i only tested with zsh and bash) only passes the result of the glob (in this case the absolute path of the templates) to the command.

If we remove the ignore by default if there is, for example, no `**` in the file path (if i understood you correctly), this would mean that `ember-template-lint app/**/*.js` would also go through the `node_modules` etc which is currently prevented as `/home/user/app/node_modules/template.hbs` gets ignored because of `**/node_modules/**`.

rwjblue:
FWIW, I think the changes in https://github.com/ember-template-lint/ember-template-lint/pull/783 allow a better scenario for tools like `ale` (no need to write to a temp file at all now).

I still would be happy to add a flag to disable globbing completely, and I think that it should be much easier to implement now (post #783) since we now have a shared/unified arguments parser.

rwjblue:
Going to close for now, but I'm still very happy to add this if you have time to rebase / cleanup the commits.

PR #671, opened 2019-03-12
Add command for generating pending list to Readme

TristanToye:
I had to track this command down multiple times for multiple projects. It is super useful, but hard to find - only mentioned in issues, commits, not docs.

I know it is only in the `ember-cli-template-lint` package, but the docs redirect here to look for config and rules. Happy to update the readme for the other package instead.


TristanToye:
@rwjblue makes sense. I will take a look at the bin file 👍

PR #672, opened 2019-03-12
feat: add `allowDynamicStyles` option to `no-inline-styles` lint rule

bmish:
Dynamic values for inline styles often serve a useful purpose, such as in these examples (taken from the [style-concatenation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/style-concatenation.md) rule documentation examples):

```
<div style={{html-safe (concat "background-image: url(" url ")")}}>
```

```
<div style={{background-image url}}>
```

```
<div style={{background-color myBackgroundColor}}></div>
```

Note that the similar [no-inline-styles](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md) lint rule for React has the default behavior of only disallowing inline styles with literal values while allowing those with dynamic values.

### DIFF
--- a/docs/rule/no-inline-styles.md
+++ b/docs/rule/no-inline-styles.md
@@ -14,6 +14,19 @@ This rule **allows** the following:
 <div class="wide-element"></div>
 ```
 
+```hbs
+{{! only allowed when `allowDynamicStyles` is enabled }}
+<div style={{html-safe (concat "background-image: url(" url ")")}}></div>
+```
+
+### Configuration
+
+ The following values are valid configuration:
+
+  * boolean - `true` to enable / `false` to disable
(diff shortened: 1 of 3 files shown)


bmish:
@Turbo87

Turbo87:
few thoughts:

- I'm not sure why inline styles are even considered that bad at all given that e.g. tailwind is essentially inline styles but with class names instead 😅 

- as you know, dynamic styles need to be html-safe, which might be problematic if people forget. we should probably explain this a bit more in the docs about this rule/option

- even though I would probably not use this option I would be okay with merging it if you think it's useful

nice work 👍 

/cc @rwjblue

Issue #588, opened 2019-01-01
Not possible to disable specific rules in specific folders


bmish:
For example, I might want to disable the `no-bare-strings` rule inside the `my-addon/tests/dummy/app/templates/` directory as I obviously don't need to internationalize the strings in my test dummy app.

I would imagine being able to disable a specific rule in a specific directory with an `overrides` field in `.template-lintrc.js` (just like with [eslint](https://eslint.org/docs/user-guide/configuring)):
```
overrides: [
  {
    files: ['tests/dummy/app/templates/**/*.hbs'],
    rules: {
      'no-bare-strings': false
    }
  }
]
```
or perhaps in an additional `.template-lintrc.js` file in the specific directory itself.

To my knowledge, this is not possible yet, so I'm checking here to confirm this and ask whether increased configuration flexibility like this has/is being considered.

rwjblue:
> To my knowledge, this is not possible yet, 

Correct, this is not currently possible.

> I'm checking here to confirm this and ask whether increased configuration flexibility like this has/is being considered.

Yes, we would definitely like to add this functionality. Someone just needs to flesh out the details and actually implement... 😸

rwjblue:
Not sure if anyone has had time yet, but I'd definitely love someone to push this forward.

rwjblue:
@bmish - Is this something you are interested in pushing forward?

bmish:
I'm interested in this still but unfortunately I don't expect to have time to work on it any time soon.

PR #673, opened 2019-03-16
feat: add `ignoreDummyTestApp` option to `no-bare-strings` rule

bmish:
This change solves the common blocker of not being able to enable this rule inside an Ember addon when its dummy test app has lots of untranslated strings that don't need to be translated.

### DIFF
--- a/lib/rules/lint-no-bare-strings.js
+++ b/lib/rules/lint-no-bare-strings.js
@@ -20,6 +20,7 @@
      * `whitelist` -- An array of whitelisted strings
      * `globalAttributes` -- An array of attributes to check on every element.
      * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
+     * `ignoreDummyTestApp` -- A boolean indicating whether the rule should ignore templates in the dummy test app inside Ember addons as translations may not be needed there (defaults to `false`)',
  */
 
 const Rule = require('./base');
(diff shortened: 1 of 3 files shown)


bmish:
@Turbo87

Turbo87:
hmm, can we make that configurable? some addons use the dummy app as a documentation page and potentially might want to translate it using `ember-intl`. I guess a simple on/off might be sufficient, or maybe an `ignorePaths` array?

bmish:
@Turbo87 good idea. I added an `ignoreDummyTestApp` option that defaults to `false`. This option feels valuable to me because I'm assuming that many/most dummy test app strings don't need to be translated, but you bring up a good point that there are some situations where they do need to be translated.

Regarding your `ignorePaths` option suggestion, I actually think that there needs to be a new global configuration option allowing rules to be disabled for certain paths (like `overrrides` from eslint). I suggested this as a new feature in issue #588.

Turbo87:
> I actually think that there needs to be a new global configuration option allowing rules to be disabled for certain paths (like `overrides` from eslint)

yep, I totally agree on that. that would be great!

lib/rules/lint-no-bare-strings.js
```diff
@@ -141,6 +148,13 @@ module.exports = class LogStaticStrings extends Rule {
   }
 
   visitor() {
+    if (
+      this.config.ignoreDummyTestApp &&
+      this.templateEnvironmentData.moduleName.startsWith('tests/dummy/app/templates/')
```
rwjblue: This doesn't work when used via ember-cli-template-lint (the module name would be `dummy/templates/` not `tests/dummy/app/templates/`)

PR #674, opened 2019-03-16
no-bare-strings: Allow options to be specified without losing the default values of the other options

bmish:
For example, I might want to set the `whitelist` option, but let the other options retain their default values. Before this change, that was not possible, as the rule would either use all the defaults or none of them.

The old behavior was not scalable and was a blocker to adding new options to this rule, as it would become very tedious to specify any single option without affecting the other options.


bmish:
@Turbo87 @rwjblue

Turbo87:
I'm wondering if we should normalize the config first before using it. Something like `let config = Object.assign({}, DEFAULT_CONFIG, rule.config)` 🤔

PR #678, opened 2019-03-18
Additional ignore documentation

DingoEatingFuzz:
More details on module patterns. Explain the current caveats between
the eslint resolver and the ember-cli/ember resolver.

I'm pretty sure I got the details right, but in all honesty, I probably didn't.

### DIFF
--- a/docs/ignore.md
+++ b/docs/ignore.md
@@ -1,0 +1,32 @@
+## Ignore support
+
+You can tell the linter to ignore individual files or entire directories with the `ignore` option.
+
+The ignore option takes an array of strings that either match exact modules or glob-match multiple modules.
+
+* **module** -- `'app/templates/exceptional-page'`
+* **glob** -- `'app/templates/components/odd-ones/**'`
+
+### Sample configuration
+
+```javascript
+module.exports = {
+  extends: 'recommended',
+
+  ignore: [
+    'project-name/templates/login',
+    'project-name/templates/components/odd-ones/**',
+    'app/templates/login',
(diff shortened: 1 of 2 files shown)


docs/ignore.md
```diff
@@ -0,0 +1,32 @@
...
+    'app/templates/components/odd-ones/**',
+  ]
+};
+```
+
+### Why are patterns duplicated?
+
+You may have noticed in the sample configuration that the four patterns are actually two patterns repeated with different prefixes.
+
+This is because `ember-template-lint` needs to work with eslint's filesystem resolver as well
```
rwjblue:
```suggestion
This is because `ember-template-lint` works directly with the filesystem and is not aware
of the applications module prefix, however `ember-cli-template-lint` works within the build
pipeline of your application which is not aware of the original file paths since all modules
are within your applications module prefix by the time it is ran.
```

docs/ignore.md
```diff
@@ -0,0 +1,32 @@
...
+as Ember's module resolver. The `app/...` patterns match the filesystem (app directory). The `project-name/...` patterns match Ember's resolver (`ENV.modulePrefix` in `config/environment.js`).
```
rwjblue:
```suggestion
**_tldr;_** The `app/...` patterns match the filesystem (app directory). The `project-name/...` patterns match Ember's resolver (`ENV.modulePrefix` in `config/environment.js`).
```

DingoEatingFuzz:
Thanks! Suggestions accepted.

rwjblue:
Thank you!

PR #681, opened 2019-03-28
Allow yield statements as children of table

### LINKED ISSUE #680: Table group rule does not account for component composition
I have run into an issue with the table group rule. I am creating a contextual table component that will look something like this:
```hbs
<table>
  {{yield
    (hash
      Body=(component "table/body")
      Head=(component "table/head")
      Header=(component "table/header")
      Row=(component "table/row")
    )
  }}
</table>
```

Consuming contexts would do something like:
```hbs
<Table as |T|>
  <T.Head>
     ...
  </T.Head>
</Table>
```
As you can see, this is meant to be invoked in such away that the invoking context will employ a table group, but of course there is nothing forcing them to do so. 

I could silence the lint error using a comment of course, but I almost think that any template that uses a `yield` as a first child of the `table` tag should not result in a lint error. Thoughts?

sohara:
Fixes #680

### DIFF
--- a/lib/rules/lint-table-groups.js
+++ b/lib/rules/lint-table-groups.js
@@ -35,6 +35,8 @@ function isAllowedTableChild(node) {
       tagNamePair = node.hash.pairs.find(pair => pair.key === 'tagName');
       if (tagNamePair) {
         return ALLOWED_TABLE_CHILDREN.includes(tagNamePair.value.value);
+      } else if (node.path.original === 'yield') {
+        return true;
       }
       break;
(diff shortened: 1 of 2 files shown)


rwjblue:
Oh, no sorry about the delay in reviewing here! Looks like we have a merge conflict now, would you mind rebasing so we can land these changes?

sohara:
@rwjblue Just rebased, should be good to go.

rwjblue:
Thank you!

Issue #261, opened 2017-09-06
Can't disable a plugin


psbanka:
If you create a plugin for ember-template-lint and then try to disable it, this won't work, for the following reasons:

1. if you disable only the rule that is associated with a plugin, the code here: https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/base.js#L72 will error out because plugins are not recognized as rules (only the rules that ship with `ember-template-lint` are recognized as rules.
2. If you disable all rules, the plugin's configuration does not seem to change in time for it to process the directive. I tried to log the events that were happening and their sequence, and this is what I found:

```
(output from the plugin, in its visitor() method, logging its config:)
this.config { min: 20, max: 100 }
-> rules/base.js, line 401 (https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/base.js#L401)
-> rules/base.js line 413 (https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/base.js#L401)
-> returning { tree: false, value: false }
-> rules/base.js line 337 (https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/base.js#L337)

/Users/psbanka/dev/copy-tango/app/pods/support/docs-search/template.hbs
  -:-  error  Template length of 12 is smaller than 20  length-validation
```

For some reason the pugins seem to be processed *before* instructions are processed and therefore their configs do not get updated.

rwjblue:
I haven't had much time to dig into this yet, sorry for the delays.  Would you mind throwing together a demo app that I could pull down and test?

rwjblue:
Finally dug into this a bit, and we need to update the implementation for `_isRuleName` so that it is aware of all configured plugins (not just those that are hard coded in the repo).

Up for taking a crack at it?

rwjblue:
@psbanka - Up for taking a shot at this?

patocallaghan:
@psbanka @rwjblue I just came across this issue today.

I'm keen to have this functionality soon-ish so unless there is a PR close to completion I'd like to take a stab at it? Would that be okay?

rwjblue:
absolutely!

patocallaghan:
Awesome, I’ll put something together 👍

patocallaghan:
@rwjblue I finally got around to looking at this and am not entirely certain I'm going the right way about it.

[...]

PR #718, opened 2019-05-15
Fix being able to inline disabling of plugin rules

patocallaghan:
Working towards #261 

This is a spike to fix the issue where we currently can't inline disable plugin rules.

@rwjblue currently this works in an extremely janky way (hard-coded config) as a way to validate that it works but not 100% sure how to proceed. Any suggestions for how to get the correct config path when getting the plugin rules from `getConfig`?

### DIFF
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -68,13 +68,30 @@ module.exports = class BaseRule {
     this.source = env.rawSource.match(reLines);
   }
 
+  _isPluginRule(name) {
+    let getConfig = require('../get-config');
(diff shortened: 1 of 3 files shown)


rwjblue:
> Any suggestions for how to get the correct config path when getting the plugin rules from getConfig?

Hmm, the approach here doesn't use `getConfig`, is that part relevant still?

rwjblue:
I've submitted https://github.com/ember-template-lint/ember-template-lint/pull/750 which incorporates the tests added in this PR and an alternative fix.

PR #721, opened 2019-05-16
Allow `{{some-component tagName="tbody"}}` to be a child of a `table`

raycohen:


lib/rules/lint-table-groups.js
```diff
@@ -24,13 +24,18 @@ function hasDisallowedChildren(tableNode) {
...
     )
-    .some(
-      child =>
-        !(
-          (child.type === 'ElementNode' && ALLOWED_TABLE_CHILDREN.indexOf(child.tag) > -1) ||
-          (child.type === 'TextNode' && !/\S/.test(child.chars))
-        )
-    );
+    .some(child => {
+      if (child.type === 'MustacheStatement') {
```
mongoose700: At this point, I think it would be cleaner to do a `switch` on `child.type`.

test/unit/rules/lint-table-groups-test.js
```diff
@@ -44,6 +44,12 @@ generateRuleTests({
     {{/unless}}
     </table>
     `,
+    '<table>{{some-component tagName="tbody"}}</table>',
```
mongoose700: Could you also add some tests that fail?

raycohen: Do you have something specific in mind? I'm not expecting any new failures - this change makes some things that were failing no longer fail.

lib/rules/lint-table-groups.js
raycohen: updated

test/unit/rules/lint-table-groups-test.js
raycohen: ok this is allowing anything with tagName through. i'll add more cases

mongoose700: You could add something like `<table>{{some-component tagName="table"}}</table>`. There should be at least one failure to make sure the new code isn't simply `child.type === 'MustacheStatement'`, without the appropriate `tagName` check. You may also want to add something like `<table>{{some-component someOtherProperty="tbody"}}</table>`

raycohen: updated, let me know if you think I should add any more

lib/rules/lint-table-groups.js
```diff
@@ -24,13 +24,24 @@ function hasDisallowedChildren(tableNode) {
...
-      child =>
-        !(
-          (child.type === 'ElementNode' && ALLOWED_TABLE_CHILDREN.indexOf(child.tag) > -1) ||
-          (child.type === 'TextNode' && !/\S/.test(child.chars))
-        )
-    );
+    .some(child => {
+      let tagNamePair;
+
+      switch (child.type) {
```
bmish: This function is getting complicated so let's extract this switch statement into a new function `isAllowedTableChild(node)`. Then we can simply use it like: `.every(isAllowedTableChild)`. Note that I'm intentionally suggesting `isAllowedTableChild` instead of `isDisallowedTableChild` to get rid of the negative.

lib/rules/lint-table-groups.js
```diff
@@ -24,13 +24,24 @@ function hasDisallowedChildren(tableNode) {
...
+        case 'MustacheStatement':
+          tagNamePair = child.hash.pairs.filter(pair => pair.key === 'tagName')[0];
+          if (tagNamePair) {
+            return ALLOWED_TABLE_CHILDREN.indexOf(tagNamePair.value.value) === -1;
```
bmish: Use `includes()` instead of `indexOf()`. Same for the other check below.

lib/rules/lint-table-groups.js
raycohen: updated

lib/rules/lint-table-groups.js
```diff
@@ -19,18 +19,31 @@ function hasDisallowedChildren(tableNode) {
...
-    );
+    .every(isAllowedTableChild);
+}
+
+function isAllowedTableChild(node) {
+  let tagNamePair;
+
+  switch (node.type) {
+    case 'MustacheStatement':
+      tagNamePair = node.hash.pairs.filter(pair => pair.key === 'tagName')[0];
```
rwjblue:
```suggestion
      tagNamePair = node.hash.pairs.find(pair => pair.key === 'tagName');
```

lib/rules/lint-table-groups.js
```diff
@@ -19,18 +19,31 @@ function hasDisallowedChildren(tableNode) {
...
+      if (tagNamePair) {
+        return ALLOWED_TABLE_CHILDREN.includes(tagNamePair.value.value);
+      }
+      break;
+    case 'ElementNode':
+      return ALLOWED_TABLE_CHILDREN.includes(node.tag);
```
rwjblue: This case needs some of the same treatment as above (for angle bracket invocation), can you also add a test with:

```hbs
<table>
  <SomeThing @tagName="thead" />
</table>
```

raycohen: updated

lib/rules/lint-table-groups.js
raycohen: updated

rwjblue:
👏 thanks again!

PR #733, opened 2019-06-05
Add new rule: no-obsolete-elements

MelSumner:
If merged, this PR adds a new rule to check for obsolete elements and throw an error if they are used.

### DIFF
--- a/docs/rule/no-obsolete-elements.md
+++ b/docs/rule/no-obsolete-elements.md
@@ -9,34 +9,34 @@ This rule forbids the use of obsolete elements.
 This rule **forbids** the following:
 
 ```hbs
-  <applet></applet>
-  <acronym></acronym>
-  <bgsound></bgsound>
-  <dir></dir>
-  <frame></frame>
-  <frameset></frameset>
-  <noframes></noframes>
-  <isindex></isindex>
-  <keygen></keygen>
-  <listing></listing>
-  <menuitem></menuitem>
-  <nextid></nextid>
-  <noembed></noembed>
-  <plaintext></plaintext>
-  <rb></rb>
-  <rtc></rtc>
-  <strike></strike>
-  <xmp></xmp>
-  <basefont></basefont>
-  <big></big>
-  <blink></blink>
-  <center></center>
-  <font></font>
-  <marquee></marquee>
-  <multicol></multicol>
-  <nobr></nobr>
-  <spacer></spacer>
-  <tt></tt>
+<applet></applet>
+<acronym></acronym>
+<bgsound></bgsound>
+<dir></dir>
+<frame></frame>
+<frameset></frameset>
+<noframes></noframes>
+<isindex></isindex>
+<keygen></keygen>
+<listing></listing>
+<menuitem></menuitem>
+<nextid></nextid>
+<noembed></noembed>
+<plaintext></plaintext>
+<rb></rb>
+<rtc></rtc>
+<strike></strike>
+<xmp></xmp>
+<basefont></basefont>
+<big></big>
+<blink></blink>
+<center></center>
+<font></font>
+<marquee></marquee>
+<multicol></multicol>
+<nobr></nobr>
+<spacer></spacer>
+<tt></tt>
 ```
 
 This rule **allows** the following:
--- a/docs/rule/no-obsolete-elements.md
+++ b/docs/rule/no-obsolete-elements.md
@@ -1,0 +1,66 @@
+## no-obsolete-elements
+
+Some elements are entirely obsolete and must not be used by authors. 
+
+This rule forbids the use of obsolete elements.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+  <applet></applet>
+  <acronym></acronym>
+  <bgsound></bgsound>
+  <dir></dir>
+  <frame></frame>
+  <frameset></frameset>
+  <noframes></noframes>
+  <isindex></isindex>
+  <keygen></keygen>
+  <listing></listing>
+  <menuitem></menuitem>
+  <nextid></nextid>
+  <noembed></noembed>
+  <plaintext></plaintext>
+  <rb></rb>
+  <rtc></rtc>
+  <strike></strike>
+  <xmp></xmp>
+  <basefont></basefont>
+  <big></big>
+  <blink></blink>
+  <center></center>
+  <font></font>
+  <marquee></marquee>
+  <multicol></multicol>
+  <nobr></nobr>
+  <spacer></spacer>
+  <tt></tt>
+```
+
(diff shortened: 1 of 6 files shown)


docs/rule/no-obsolete-elements.md
```diff
@@ -0,0 +1,66 @@
...
+
+This rule forbids the use of obsolete elements.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+  <applet></applet>
+  <acronym></acronym>
```
rwjblue: There is some odd indentation in this list, can you make them all indented the same way?

docs/rule/no-obsolete-elements.md
```diff
@@ -0,0 +1,66 @@
...
+  <marquee></marquee>
+  <multicol></multicol>
+  <nobr></nobr>
+  <spacer></spacer>
+  <tt></tt>
+```
+
+This rule **allows** the following:
+
+```hbs
```
rwjblue: I'd probably remove this listing, its nearly impossible to be exhaustive and the main point is "it works with everything else"

lib/rules/lint-no-obsolete-elements.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const Rule = require('./base');
+
+const errorMessage = element => `Use of <${element}> detected. Do not use deprecated elements.`;
+
+const OBSOLETE_ELEMENTS = [
```
rwjblue: Can you put an inline comment just above here with the link to the whatwg listing?

lib/rules/lint-no-obsolete-elements.js
```diff
@@ -0,0 +1,53 @@
...
+  'nobr',
+  'spacer',
+  'tt',
+];
+
+module.exports = class NoObsoleteElements extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        if (OBSOLETE_ELEMENTS.includes(node.tag)) {
```
rwjblue: We also need to check if the `node.tag` is a block param.

Can you add a "good" test for:

```hbs
{{#let (component 'whatever-here') as |plaintext|}}
  <plaintext />
{{/let}}
```

^ breaks the `no-shadowed-elements` rule, but should **not** fail this rule

yarn.lock
```diff
@@ -96,42 +96,34 @@
   integrity sha512-tHq6qdbT9U1IRSGf14CL0pUlULksvY9OZ+5eEgl1N7t+OA3tGvNpxJCzuKQlsNgCVwbAs670L1vcVQi8j9HjnA==
 
 "@types/node@*":
-  version "10.12.18"
-  resolved "https://registry.yarnpkg.com/@types/node/-/node-10.12.18.tgz#1d3ca764718915584fcd9f6344621b7672665c67"
-  integrity sha512-fh+pAqt4xRzPfqA6eh3Z2y6fyZavRIumvjhaCL753+TVkGKGhpPeyrJG2JftD0T9q4GF00KjefsQ+PQNDdWQaQ==
+  version "12.0.4"
```
rwjblue: Can you reset the changes to this file? It seems unrelated to adding the rule.

If `yarn.lock` on master is out of sync we should update in a separate (focused) PR.

lib/rules/lint-no-obsolete-elements.js
MelSumner: @rwjblue just to be clear- we both want to prevent let-as-(obsolete element), right?

lib/rules/lint-no-obsolete-elements.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const Rule = require('./base');
+
+const errorMessage = element => `Use of <${element}> detected. Do not use deprecated elements.`;
```
bmish: I would recommend exporting this error message function and then importing it from the test file to avoid duplication. See `no-negated-condition` rule files for an example.

test/unit/rules/lint-no-obsolete-elements-test.js
```diff
@@ -0,0 +1,258 @@
...
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-obsolete-elements',
+
+  config: true,
+
+  good: ['<div></div>'],
+
+  bad: [
```
bmish: This long array of test cases is pretty tedious, and of course duplicates the list of deprecated elements from the rule file, making it more work to update the list later. Instead, you could export the `OBSOLETE_ELEMENTS` array from the rule file, and then import it here to generate the test cases via a `forEach` loop.

lib/rules/lint-no-obsolete-elements.js
rwjblue: Kinda, yes. _This_ rule should **not** flag any block params at all, but the no-shadowed-elements rule prevents using _any_ lowercased block param names in the element position. 

So under the no-shadowed-element rule (which is enabled by default), the example I gave above would be disallowed.

lib/rules/lint-no-obsolete-elements.js
rwjblue: Ya, good idea

test/unit/rules/lint-no-obsolete-elements-test.js
rwjblue: Ya, something like:

```js
bad: OBSOLETE_ELEMENTS.map(element => {
  return {
    /* generated POJO for this element */
  }
}),
```

lib/rules/lint-no-obsolete-elements.js
MelSumner: Hmmm. I think it should block use of these elements, block params or not. Let's schedule a call to discuss more, so I can understand the logic.

rwjblue: Seems good, but we won't be landing this rule until we have tests ensuring that this example does not fail the new rule being added:

```hbs
{{#let (component 'whatever-here') as |plaintext|}}
  <plaintext />
{{/let}}
```

MelSumner: I see. I still would like to understand it better. Thanks! :+1:

docs/rule/no-obsolete-elements.md
```diff
@@ -0,0 +1,53 @@
...
+<center></center>
+<font></font>
+<marquee></marquee>
+<multicol></multicol>
+<nobr></nobr>
+<spacer></spacer>
+<tt></tt>
+```
+
+This rule **allows** the following:
```
rwjblue: Can probably remove this line completely, and bold allows in the sentence just below.

rwjblue:
Thank you!

Issue #726, opened 2019-05-24
[Rule idea] No empty non void elements


ansmonjol:
Hello,

I would love to implement a new rule to avoid non-empty void elements in templates.

Something that could lead to raise an error on `<p></p>` or `<div></div>` for example. 
Of course nothing would happen on `<p>{{myVar}}</p>` or any void elements (`<br >`).

Do you think it's ok to call this rule `no-empty-non-void-elements` ?

In order to submit an acceptable final MR, do you have some auto generated files (like Changelog ?) or stuff I must perform during new rule implementation ? In plus than Documentation and tests.

Thanks

rwjblue:
I don't fully grok the request here, can you update the description to show more good and bad examples?

ansmonjol:
Done ! let me know if that's cleat enough :)

rwjblue:
Gotcha, I was misinterpreting the term "void" in your original post. I typically use the term "void elements" to refer to things like:

```html
<input>
<br>
<img>
```

Which apparently are actually called [empty elements](https://developer.mozilla.org/en-US/docs/Glossary/empty_element). 

I'd vaguely prefer to avoid confusing my future self by not using `void` in the name here, what do you think about something like `blacklist-childless-elements`?

ansmonjol:
Great ! Thanks for your comment.

I'm a bit afraid by the `blacklist` wording here WDYT ?
I was trying to keep consistency with the `no-` rule name. Also, the 'config' item list is directly inspired from the `lint-self-closing-void-elements` rule.

I propose to rename the config list to `BLOCK_TAGS` and the linter rule to `no-childless-elements` if that sounds good to you ?

rwjblue:
Seems great!

ansmonjol:
Hello @rwjblue, you can find the related PR https://github.com/ember-template-lint/ember-template-lint/pull/735

Please let me know if I forgot anything or something is not clear enough !

MelSumner:
'<div></div>` is pretty common to leave blank for reasons. I'd probably recommend removing that from the disallowed list.

PR #735, opened 2019-06-09
No childless elements

ansmonjol:
Closes https://github.com/ember-template-lint/ember-template-lint/issues/726


lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,68 @@
...
+ The following values are valid configuration:
+
+   * boolean -- `true` for enabled / `false` for disabled
+ */
+
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+
+/**
+ * [Specs of Void Elements]{@link https://www.w3.org/TR/html-markup/syntax.html#void-element}
```
bmish: This should go in the `References` section of the doc.

Here's the doc template: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/_TEMPLATE_.md

test/unit/rules/lint-no-childless-elements-test.js
```diff
@@ -0,0 +1,172 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-childless-elements',
+
+  config: true,
+
+  good: [
```
bmish: The good and bad test cases can be generated by exporting `BLOCK_TAGS` from the rule file and then doing: 
```
good: BLOCK_TAGS.map(blockTag => {
  return ...
}),

bad: BLOCK_TAGS.map(blockTag => {
  return { ... };
})
```

test/unit/rules/lint-no-childless-elements-test.js
```diff
@@ -0,0 +1,172 @@
...
+    '<ul>something</ul>',
+    '<ol>something</ol>',
+    '<li>something</li>',
+    '<div>something</div>',
+  ],
+
+  bad: [
+    {
+      template: '<h1></h1>',
+      message: 'Empty non-void elements are not allowed',
```
bmish: The error message should be imported from the rule file. See other rules like `no-negated-condition` for an example.

lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,68 @@
...
+ * @type {Object}
+ */
+const BLOCK_TAGS = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'span', 'a', 'ul', 'ol', 'li', 'div'];
+
+module.exports = class NoChildlessElements extends Rule {
+  parseConfig(config) {
```
bmish: This rule has no extra configuration so this entire `parseConfig` function should be omitted.

lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,68 @@
...
+      ['  * boolean - `true` to enable / `false` to disable'],
+      config
+    );
+
+    throw new Error(errorMessage);
+  }
+  visitor() {
+    return {
+      ElementNode(node) {
+        const isRuleEnabled = Boolean(this.config);
```
bmish: This should be removed. You can assume the rule is enabled if this function executes.

lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,68 @@
...
+        const isNonVoidTag = BLOCK_TAGS.includes(node.tag);
+        const isElementEmpty = !node.children.length;
+        const shouldDisplayMessage = isRuleEnabled && isNonVoidTag && isElementEmpty;
+
+        if (shouldDisplayMessage) {
+          let source = this.sourceForNode(node).trim();
```
bmish: This variable is unnecessary. And the `trim` is also unnecessary.

Just do `source: this.sourceForNode(node)` below.

ansmonjol:
Hello @bmish !

Thank you so much for your feedback.

They are all integrated, please tell me is something else can be improved !

docs/rule/no-childless-elements.md
```diff
@@ -0,0 +1,49 @@
...
+<p>something</p>
+<span>something</span>
+<a>something</a>
+<ul>something</ul>
+<ol>something</ol>
+<li>something</li>
+<div>something</div>
+```
+
+### Configuration
```
bmish: I would just remove this `Configuration` section since there is no special configuration for this rule.

lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+/*
+ Disallows childless elements
+
+ ```
+{{! good }}
```
bmish: I would remove this entire comment. The documentation file already provides the good and bad examples. It's not worth duplicating the list again here.

lib/rules/index.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
   'no-unbound': require('./lint-no-unbound'),
   'no-unnecessary-concat': require('./lint-no-unnecessary-concat'),
   'no-unused-block-params': require('./lint-no-unused-block-params'),
+  'no-childless-elements': require('./lint-no-childless-elements'),
```
bmish: This needs to be added alphabetically.

lib/rules/lint-no-childless-elements.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const Rule = require('./base');
+
+const BLOCK_TAGS = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'span', 'a', 'ul', 'ol', 'li', 'div'];
```
bmish: Where exactly does this list come from? Is it actually complete?

Would it be easier to flip this around and specify the void elements here instead? I see the list `area, base, br, col, embed, hr, img, input, link, meta, param, source, track, wbr` from https://www.w3.org/TR/html/syntax.html#void-elements

docs/rule/no-childless-elements.md
```diff
@@ -0,0 +1,45 @@
...
+
+```hbs
+<h1></h1>
+<h2></h2>
+<h3></h3>
+<h4></h4>
+<h5></h5>
+<h6></h6>
+<p></p>
+<span></span>
```
MelSumner: the span is an inline element and is often left empty while styled.

lib/rules/lint-no-childless-elements.js
MelSumner: These are the block level elements in HTML:
`<address><article><aside><blockquote><canvas><dd><div><dl><dt><fieldset><figcaption><figure><footer><form><h1>-<h6><header><hr><li><main><nav><noscript><ol><p><pre><section><table><tfoot><ul><video>`

ansmonjol:
Sorry for delay.

Great idea ! I'll address both points.

Will flip the tag of array to display void elements inside. 
If someone send a config with `block` or `allow` array, config array will be updated

```js
// array of tags to config the rule trigger
const ALLOWED_CHILDLESS_TAGS = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input', 'link', 'meta', 'param', 'source', 'track', 'wbr']

// config
{
    allow: ['div', 'h2], // Elements that will be added to ALLOWED_CHILDLESS_TAGS
    block: ['h2'] // Elements that will be removed from ALLOWED_CHILDLESS_TAGS
}
```
Then we'll trigger the rule it the `node.tag` of not included in `ALLOWED_CHILDLESS_TAGS`

bmish:
@ansmonjol how is this going?

ansmonjol:
Hello @bmish !

Sorry for delay. 
I do have finished the implementation with a `block` and `allow`.
I'm stuck with tests right now.

I do have a list of elements to allow as childless elements.
But in tests, I have to write the exact template to write.

For instance I'm looping threw my array of elements:
```
let ALLOWED_CHILDLESS_TAGS = Object.freeze([
  'area',
  'base',
  'br',
  'col',
  .....
]);
```

Then in tests I've a 
```
    ...ALLOWED_CHILDLESS_TAGS.map(blockTag => {
      return {
        template: `<${blockTag}>`,
      };
    }),
```
But if a tag is a block element (`div` for example), the test is failing, as `<div>` is not closed.
For instance, I want to add `span` in the list regarding [Melanie Sumner comment above](https://github.com/ember-template-lint/ember-template-lint/pull/735#pullrequestreview-251749343), but if I do so, tests are failing. (`"message": "Unclosed element 'span' (on line 1)."`)

I'm not sure what is the best way to adapt the test to make it understand when a tag is an inline or a block one.

As this is a test purpose feature, I do not want to update the way I build my array of `ALLOWED_CHILDLESS_TAGS`. 

Do you have any idea about how to solve that ? I had a look at other tests but it's always hardcoded

ansmonjol:
I'll push the code with failing test so you can have a look

lib/rules/lint-no-childless-elements.js
ansmonjol: Resolving discussion, as list and logic behind evolved

docs/rule/no-childless-elements.md
ansmonjol: Added it to the list ! But it lead to having failing tests. The reason is explained here https://github.com/ember-template-lint/ember-template-lint/pull/735#issuecomment-521995961

ansmonjol:
Oh I've an idea, create two list and merge them on top of the rule file. A `BLOCK_ELEMENTS` and `INLINE_ELEMENTS`.

Then if I import the `BLOCK_ELEMENTS` in tests I can do something like
```js
...ALLOWED_CHILDLESS_TAGS.map(blockTag => {
  if (BLOCK_ELEMENTS.includes(blockTag) {
    template: `<${blockTag}></${blockTag}>`,
  }

  return {
    template: `<${blockTag}>`,
  };
}),
```

What do you guys think ! 

Will push something similar this afternoon

docs/rule/no-childless-elements.md
ansmonjol: Fixed

ansmonjol:
Done, please review 🙏

docs/rule/no-childless-elements.md
```diff
@@ -0,0 +1,45 @@
+## no-childless-elements
+
+Childless elements
```
bmish: Remove this line that just says `Childless elements`?

ansmonjol: Yes @bmish and it also tag home made components, like if I had a `<Sidebar />` 😓 

I suggest to inverse the logic here and rename this rule.

The idea would be to to have a rule named `flag-childless-elements`.
It would only work with a custom config passed, no tag flagged by default.

Then you could have in your `.template-lintrc.js` something like:
```
'no-childless-elements': {
  tags: ['div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'],
},
```

Then you'll only have and error if those tags' node do not have children (`!node.children.length`)

What do you think of this implementation ?

ansmonjol:
Had no time to work on it. Closing

PR #737, opened 2019-06-10
Add no-invalid-interactive-img exception for "onerror" on img tags

### LINKED ISSUE #663: no-invalid-interactive capturing onError
The following triggers this rule, where I would expect it not to as `onError` isn't a user interaction such as clicking

```hbs
<img
  src={{src}}
  alt={{alt}}
  onError={{action "onError"}}
>
```

elidupuis:
Closes #663.


elidupuis:
Should probably add a similar exception for `` but I figured that should be a separate PR.

rwjblue:
Thank you!

PR #743, opened 2019-06-14
Add failing test for tabindex rule

amk221:
I need to support `tabindex="{{if this.show "-1" "0"}}"`

My use case for this is an expandable/collapsable component.

When it is collapsed it should be possible to _tab_ to it. (The focus causes expansion)

When it is expanded the first input inside is focused.

Therefore, I do not _also_ want it to be possible to focus the expanded container.


rwjblue:
👍 We should be able to make inline if work as long as the resulting values are static strings in the template (which is your use case).

rwjblue:
@lifeart - Any chance you'd have time to poke at this?

lifeart:
@rwjblue, done

rwjblue:
Fixed by #788

rwjblue:
Thank you @amk221!****

PR #744, opened 2019-06-15
Tweak error message for `no-element-event-actions` rule.

bmish:


lib/rules/lint-no-element-event-actions.js
```diff
@@ -3,7 +3,8 @@
 const Rule = require('./base');
 const AstNodeInfo = require('../helpers/ast-node-info');
 
-const ERROR_MESSAGE = 'Do not use HTML element event properties with Ember actions.';
+const ERROR_MESSAGE =
+  'Use Ember `{{action ...}}` helper instead of HTML element event properties like `onclick`.';
```
rwjblue: I'd actually prefer to suggest `{{on}}`, though it doesn't have the same interop benefits with the event dispatcher.

PR #750, opened 2019-06-19
Ensure custom rules can be disabled with inline comments.

### LINKED ISSUE #261: Can't disable a plugin

rwjblue:
Builds off of @patocallaghan's work in #718 (thank you!), but tweaks the implementation a bit to avoid re-parsing configuration.

Closes #718
Fixes #261

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -59,6 +59,7 @@ class Linter {
         console: this.console,
         log: addToResults,
         defaultSeverity: this._defaultSeverityForRule(pluginName, config.pending),
+        ruleNames: Object.keys(rules),
       });
 
       astPlugins.push(env => {
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -43,6 +43,7 @@ module.exports = class BaseRule {
     this.ruleName = options.name;
     this._console = options.console || console;
     this._log = options.log;
+    this._ruleNames = options.ruleNames;
     this.severity = options.defaultSeverity;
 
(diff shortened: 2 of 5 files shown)


patocallaghan:
Thanks @rwjblue 👏

PR #753, opened 2019-06-24
Add `no-unnecessary-component-helper` rule

bmish:
@Turbo87 @rwjblue


Turbo87:
is that officially supported? I don't think I've ever seen that documented anywhere 🤔

rwjblue:
> is that officially supported?

Dunno, but _I_ consider it supported (and know of a number of projects that rely on it). There are also tests in ember-cli/ember-resolver that ensure it continues to work.

FWIW, I'm **totally** happy to just decide that if you need those exotic types of things that you'd have to disable this rule. I just wanted to make sure we are explicit about it (and possibly mention it in the docs for the rule).

bmish:
I added an exemption to this rule for component names of the form `addon-name@component-name` since it's a trivial change and there's no point to making life harder for anyone using that (regardless of whether or not it's officially supported).

bmish:
@rwjblue this PR should be ready to merge.

PR #759, opened 2019-07-02
Deprecate `img-alt-attributes` rule in favor of `require-valid-alt-text`

### LINKED ISSUE #739: Deprecate 'img-alt-attributes' rule
Deprecate 'img-alt-attributes' rule since it's been replaced by a more complete/universal rule, 'require-valid-alt-text'.

bmish:
Fixes #739.

CC: @MelSumner

### DIFF
--- a/docs/rule/img-alt-attributes.md
+++ b/docs/rule/img-alt-attributes.md
@@ -1,5 +1,7 @@
 ## img-alt-attributes
 
+**NOTE**: This rule is deprecated in favor of the more comprehensive rule [require-valid-alt-text](./require-valid-alt-text.md).
+
 An `<img>` without an `alt` attribute is essentially invisible to assistive/accessibility technology (i.e. screen readers).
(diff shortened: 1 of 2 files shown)


bmish:
Here's the behavior that we can expect from adding this deprecation to `DEPRECATED_RULES`:

1. Currently, since both the old and new rules exist, there will be no impact.
2. When we actually delete the old `img-alt-attributes` rule in the next major release (assuming we want to delete it), the deprecation warning (see below) will be displayed for anyone still specifying the old rule name in their config, and the new rule will automatically be enabled for them.

```
Deprecated rule names found: [
    {
        "oldName": "img-alt-attributes",
        "newName": "require-valid-alt-text"
    }
]
```

Note that I did notice that `DEPRECATED_RULES` is typically used only for renaming rules and not for replacing an old rule with a new rule that has different behavior.

rwjblue:
Perfect! So this will only have an effect when we _actually_ remove the other rule (which we'll do when we queue up the major bump).

Issue #700, opened 2019-04-17
no-outlet-outside-routes error showing for route template


marshy2201:
The no-outlet-outside-routes error is showing for a file that is a route template.

Is this because of the file directory where 'components' is named?

```hbs
{{!-- app/templates/products/components/dc.hbs --}}
{{title "DC" replace=false prepend=false}}
{{outlet}}
```

marshy2201:
Here is the errors that are showing up:

![eslint-errors](https://user-images.githubusercontent.com/20504912/61528038-ba692980-aa15-11e9-83ba-c84285d049ae.png)

rwjblue:
I believe this was fixed by #762

PR #762, opened 2019-07-03
Fix false positive of no-outlet-outside-routes

### LINKED ISSUE #761: False positive when app's name is "components"
I just discovered there's a false positive when the app's name is _"components"_:

```
components/templates/application.hbs
  12:0  error  Unexpected {{outlet}} usage. Only use {{outlet}} within a route template.  no-outlet-outside-routes
```

This is actually the application route's template but the app happens to be named "components".

Reproduction available here: https://github.com/simplabs/ember-workshop/tree/master/examples/components

marcoow:
This checks the module name includes `templates/components` and thus makes sure the template is actually a component template vs. any templates where the app names happens to be _"components"_ (e.g. `components/templates/application`).

closes #761

### DIFF
--- a/lib/helpers/is-route-template.js
+++ b/lib/helpers/is-route-template.js
@@ -19,6 +19,10 @@ module.exports = function isRouteTemplate(moduleName) {
   var isPartial = baseName.indexOf('-') === 0;
 
   return (
-    !isPartial && moduleName.indexOf('/templates/components/') === -1 // classic component
+    !isPartial &&
+    // check that this is not a component template - [^/]+ is the app's module name
+    // which we don't know here
+    !/^[^/]+\/templates\/components\//.test(moduleName) && // classic component
+    !/^[^/]+\/components\//.test(moduleName) // co-located component template
   );
 };
--- a/lib/helpers/is-route-template.js
+++ b/lib/helpers/is-route-template.js
@@ -19,6 +19,6 @@ module.exports = function isRouteTemplate(moduleName) {
   var isPartial = baseName.indexOf('-') === 0;
 
   return (
-    !isPartial && moduleName.indexOf('components/') === -1 // classic component
+    !isPartial && moduleName.indexOf('/templates/components/') === -1 // classic component
   );
 };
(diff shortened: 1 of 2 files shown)


lib/helpers/is-route-template.js
```diff
@@ -19,6 +19,6 @@ module.exports = function isRouteTemplate(moduleName) {
   var isPartial = baseName.indexOf('-') === 0;
 
   return (
-    !isPartial && moduleName.indexOf('components/') === -1 // classic component
+    !isPartial && moduleName.indexOf('templates/components/') === -1 // classic component
```
bmish: To be even safer, could you change it to `/templates/components/` with the leading slash?

marcoow: sure, done

lib/helpers/is-route-template.js
```diff
@@ -19,6 +19,6 @@ module.exports = function isRouteTemplate(moduleName) {
   var isPartial = baseName.indexOf('-') === 0;
 
   return (
-    !isPartial && moduleName.indexOf('components/') === -1 // classic component
+    !isPartial && moduleName.indexOf('/templates/components/') === -1 // classic component
```
rwjblue: Hmm, with the upcoming changes for co-location (emberjs/rfcs#481) we won't be using `app/templates/components` any longer.

Specifically, these are both valid component template locations:

```js
app/components/foo.hbs
app/components/foo/index.hbs
```

marcoow: So could we do

```es6
!isPartial && moduleName.indexOf('<appname>/templates/components/') === -1 && moduleName.indexOf('<appname>/components/') === -1
```

instead then?

efx: @marcoow @rwjblue What about just checking for the `/components/` path?:
```js
!isPartial && !moduleName.includes('/components/')
```

efx: Well, reading the original issue we cannot do just the `/components` string given that example route.

marcoow: `<appname>` would be `app` in most tests here, yes - in reality it's the app's actual module name, e.g. `my-app`. I updated this to support co-located component templates as well.

efx:
Also closes #700

PR #601, opened 2019-01-08
Add no-implicit-this to recommended config

rwjblue:

PR #763, opened 2019-07-03
Add no-implicit-this to recommended rules

mehulkar:
I don't know all the caveats to doing this, i'm sure there's a reason it's not recommended yet, but I didn't see any open issues or PRs talking about this. So getting it started!

Adding this rule would save me a lot of time in code reviews! (I don't care about the codemod too much)


mehulkar:
OH wow, this is totally a dupe of #601 . CLosing

Issue #765, opened 2019-07-04
[New rule proposal] no-classic-invocation-component


patocallaghan:
Recently I created a rule at work to lint against using classic invocation syntax in templates and it seems to work fine. Any thoughts on whether you'd be open to have it contributed back to ember-template-lint? The alternative is to create a separate template rule plugin and open source that.

It currently handles the following cases:

### Bad
```hbs
{{bad-code}}
{{#bad-code}}{{/bad-code}}
{{nested/bad-code}}
{{#nested/bad-code}}{{/nested/bad-code}}
```

### Good
```hbs
<GoodCode />
<GoodCode></GoodCode>
<Nested::GoodCode />
<Nested::GoodCode></Nested::GoodCode>

{{! whitelisted helpers}}
{{some-valid-helper param}}

{{! in-built helpers}}
{{if someProperty "yay"}}
{{#each items as |item|}}
  {{item}}
{{/each}}
```

Just to point out, for new projects you more than likely won't have this rule on but for our use case as we are migrating our existing codebase to angle brackets, we are using it to help with the migration to remind folks what style they should be using for components.

bmish:
Sounds like a good rule to me! Definitely submit it here.

patocallaghan:
@bmish cool, I'll submit a PR and we can go from there 👍

patocallaghan:
@bmish added it here https://github.com/ember-template-lint/ember-template-lint/pull/768

PR #768, opened 2019-07-09
Add no-curly-component-invocation lint rule

patocallaghan:
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/765

So in our app we started the process of migrating 1000s of templates over to use angle bracket syntax. As part of that process we created our own template-lint rule to prevent instances of classic invocation component syntax sneaking back in. As we found it useful I thought I'd see if you'd like to have it contributed back to ember-template-lint? I'm not tied to any of the solution here so please let me know what you think.

* It detects if block or mustache statement name has a `/` or `-` in it. If so this is possibly a component.
* Checks against a whitelist of helpers (see config section in the rule README) to differentiate between helpers and components
* If the name isn't in the whitelist it will log a template lint error like:
> You are using the component `{{awesome-code}}` with classic invocation syntax. You should use `<AwesomeCode>` instead. If it is actually a helper you must manually add it to the 'no-classic-invocation-component' rule configuration, e.g. `'no-classic-invocation-component': { allow: ['awesome-code'] }`.
* It currently doesn't support the case where you have a component which doesn't have a `-` or a `/` in it (which _I think_ is legal) as we had none of those in our codebase. If we need to support it we could always add a `disallow` config option so these could be added there.

#### Bad

```hbs
{{bad-code}}
{{#bad-code}}{{/bad-code}}
{{nested/bad-code}}
{{#nested/bad-code}}{{/nested/bad-code}}
```

#### Good

```hbs
<GoodCode />
<GoodCode></GoodCode>
<Nested::GoodCode />
<Nested::GoodCode></Nested::GoodCode>
{{! whitelisted helpers}}
{{some-valid-helper param}}
{{some/some-valid-helper param}}
{{! in-built helpers}}
{{if someProperty "yay"}}
{{#each items as |item|}}
  {{item}}
{{/each}}
```


docs/rule/no-classic-invocation-component.md
```diff
@@ -0,0 +1,64 @@
...
+{{some/some-valid-helper param}}
+
+{{! in-built helpers}}
+{{if someProperty "yay"}}
+{{#each items as |item|}}
+  {{item}}
+{{/each}}
+```
+
+### Whitelisting helpers
```
rwjblue: Can you tweak this to make it clear that it only impacts helpers directly used in mustaches (as the `PathExpression`, but don't say that since it's way too "in the weeds")? 

I think _for the most part_ folks don't use many helpers like that...

lib/helpers/classic-invocation-component.js
```diff
@@ -0,0 +1,37 @@
...
+  let nestedComponentName = /\//g;
+  return nestedComponentName.test(tagName);
+}
+
+function transformNestedTagName(tagName) {
+  let paths = tagName.split('/');
+  return paths.map(name => capitalizedTagName(name)).join('::');
+}
+
+function capitalizedTagName(tagname) {
```
rwjblue: Can you add some tests for this function? It doesn't quite match [what Ember does](https://github.com/emberjs/ember.js/blob/b31998b6a0cccd22a8fb6fab21d24e5e7f2cb70d/packages/ember-template-compiler/lib/system/dasherize-component-name.ts), so I'd like to make sure that at least [Ember's test input/outputs match](https://github.com/emberjs/ember.js/blob/b31998b6a0cccd22a8fb6fab21d24e5e7f2cb70d/packages/ember-template-compiler/tests/system/dasherize-component-name-test.js#L8-L24).

lib/rules/index.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   'no-action-modifiers': require('./lint-no-action-modifiers'),
   'no-attrs-in-components': require('./lint-no-attrs-in-components'),
   'no-bare-strings': require('./lint-no-bare-strings'),
+  'no-classic-invocation-component': require('./lint-no-classic-invocation-component'),
```
rwjblue: I'm not 100% sure RE: using "classic" here. I absolutely agree that this is in "bike shed" territory, but I think I'd personally prefer `no-curly-component-invocation`.

lib/helpers/classic-invocation-component.js
```diff
@@ -0,0 +1,37 @@
+function transformTagName(tagName) {
+  if (isNestedComponentTagName(tagName)) {
+    return transformNestedTagName(tagName);
+  }
+
+  return capitalizedTagName(tagName);
+}
+
+function isComponentTagName(tagName) {
```
rwjblue: Components do **not** need dashes in their names (in either curly or angle forms) as of something around Ember 3.8ish. I think this needs to be tweaked...

rwjblue:
Thanks for working on this! I absolutely think its a great rule to add...

lib/rules/index.js
patocallaghan: I wasn't crazy about the "classic" part either but my initial thinking was that  using it was consistent with the [syntax conversion guide](https://guides.emberjs.com/release/reference/syntax-conversion-guide/#toc_angle-bracket-syntax) which also refers to it as "classic invocation".

"curly" is more descriptive though so if you want me to rename I can.

lib/helpers/classic-invocation-component.js
patocallaghan: So thinking about this I'm not sure how to make this work without having this rule throwing a warning for every property in a template, e.g. `{{myRandomProperty}}`. At least when we're just looking for `-` and `/` components we know they are at least components or helpers?

I was initially thinking of adding another config option similar to `allow`, called `disallow` which will allow users to list ambiguously named components that they'd like to throw lint errors on.

patocallaghan: Just to clarify I use this helper in the lint rule here https://github.com/ember-template-lint/ember-template-lint/pull/768/files#diff-9ccffddebfd018d7d7d2aab10ffb3428R22

lib/rules/index.js
rwjblue: Yeah, I'm a bit torn here. "classic components" very specifically means `Ember.Component`, to me it would be pretty confusing to have "classic invocation" to use a different definition of "classic". For example, you can absolutely invoke `@glimmer/component` based components with curlies but they are _for sure_ not "classic components".

Lets do the swap to `no-curly-component-invocation`, and I'll open up an issue over on the guides repo to chat through the issue.

lib/helpers/classic-invocation-component.js
rwjblue: > I was initially thinking of adding another config option similar to allow, called disallow which will allow users to list ambiguously named components that they'd like to throw lint errors on.

Yep, thats the sticky wicket for sure. I'd like to see this go a few different directions, lets do this:

* Ignore `-` for all `BlockStatement` invocations (other than the built in's which we can enumerage (`if`,`each`,`with`,`let`, ???)), we **know for sure** those are components 
* Add the `allow` / `disallow` to the config like you mentioned above to allow us to check (or ignore) specific tags
* Add another boolean config flag that allows you to choose the default value (do we only restrict things with a `-` or all "possible components")

lib/rules/index.js
rwjblue: https://github.com/ember-learn/guides-source/issues/929

lib/helpers/classic-invocation-component.js
patocallaghan: FYI sorry for the silence but I've working away on this when I get time. I'm almost finished all your suggested changes.

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
...
+        let path = name.split('/helpers/');
+        return path.pop();
+    }).filter(name=>!name.includes('/')).uniq();
+
+copy(JSON.stringify(componentLikeHelpers))
+```
+
+Hat tip to @lifeart for [this code](https://github.com/lifeart/ember-ast-hot-load#how-to-use-this-addon).
+
+### Blacklisting components without dashes
```
patocallaghan: I've introduced two new config options:

1. `disallow` - list component names which you want to throw lint errors on. Handy for capturing components with no dashes in their name e.g. `{{heading}}`
2. `noDashInName` - defaults to `undefined`. By default we assume that components have a dash in their name but setting this to `true` will allow you capture all `{{curlies}}`. It's loose so could result in a lot of false positives.

lib/rules/lint-no-curly-component-invocation.js
```diff
@@ -0,0 +1,98 @@
+const Rule = require('./base');
+const { transformTagName } = require('../helpers/curly-component-invocation');
+const IN_BUILT_HELPERS = [
+  'action',
```
patocallaghan: I pulled this list from https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers

patocallaghan:
@rwjblue I've updated the rule per your feedback 🎉

rwjblue:
Sorry for the delay, was on vacation last week (and therefore super swampped this week...).

.gitignore
```diff
@@ -15,3 +15,4 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
+.vscode
```
rwjblue: In _general_ I think editor specific ignores should either be in `.git/info/exclude` (which has the same format as `.gitignore` but isn't checked into the repo) or a global `.gitignore`.

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
+## Use angle bracket syntax for components
+
+### Rule name: `no-curly-component-invocation`
+
+There are two ways to invoke a component in a template: curly compoment syntax (`{{my-component}}`), and angle bracket syntax (`<MyComponent />`). The difference between them is syntactical. You should favour angle bracket syntax as it improves readability of templates, i.e. disambiguates components from helpers, and is also the future direction Ember is going with Glimmer components.
```
rwjblue:
```suggestion
There are two ways to invoke a component in a template: curly compoment syntax (`{{my-component}}`), and angle bracket syntax (`<MyComponent />`). The difference between them is syntactical. You should favour angle bracket syntax as it improves readability of templates, i.e. disambiguates components from helpers, and is also the future direction Ember is going with the Octane Edition.
```

Just removing "Glimmer components" here, since they are orthogonal to using curly/angle invocation (you can invoke GC's with curlies).

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
...
+    },
+  },
+};
+```
+
+To get a list of of all the helpers in your app run the following code in your browser's Developer Tools Console when your app has loaded:
+
+``` js
+var componentLikeHelpers = Object.keys(require.entries)
+    .filter(name=>(name.includes('/helpers/')|| name.includes('/helper')))
```
rwjblue: Mind removing the "pods" / MU support stuff from this snippet? (I'm assuming the `name.includes('/helper')` supports pods, and just below `/-` stuff supports MU)

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
...
+    .filter(name=>!name.includes('/-')).map(name=>{
+        let path = name.split('/helpers/');
+        return path.pop();
+    }).filter(name=>!name.includes('/')).uniq();
+
+copy(JSON.stringify(componentLikeHelpers))
```
rwjblue: 😻

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+Hat tip to @lifeart for [this code](https://github.com/lifeart/ember-ast-hot-load#how-to-use-this-addon).
+
+### Blacklisting components without dashes
+Since Ember 3.8 components have not required dashes in their name, e.g. `{{datepicker}}`. To help the linter throw an error on these curly component invocations, which it would otherwise have thought to be a helper or property, you can explicitly add it to the `disallow` section of the rule's config. Any curly statements matching an entry in `disallow` will throw a lint error.
```
rwjblue: I think I was wrong, and this was actually changed in 3.10 not 3.8 (sorry).

```suggestion
Since Ember 3.10 components have not required dashes in their name, e.g. `{{datepicker}}`. To help the linter throw an error on these curly component invocations, which it would otherwise have thought to be a helper or property, you can explicitly add it to the `disallow` section of the rule's config. Any curly statements matching an entry in `disallow` will throw a lint error.
```

docs/rule/no-curly-component-invocation.md
rwjblue: I'd love to add a snippet like the one above for folks to use here, seems _mostly_ similar (with small tweaks to the filtering) right?

docs/rule/no-curly-component-invocation.md
```diff
@@ -0,0 +1,94 @@
...
+        'heading',
+        'datepicker',
+      ],
+    },
+  },
+};
+```
+
+### Matching on components without dashes in their name
+Before Ember 3.8 components were required to have at least one dash, `-`, in their name. By default this rule assumes that all components have `-` in their name. If you'd like to change this behaviour to match all curly invocations, even those without dashes then set the `noDashInName` option to `true`.
```
rwjblue:
```suggestion
Before Ember 3.10 components were required to have at least one dash, `-`, in their name. By default this rule assumes that all components have `-` in their name. If you'd like to change this behaviour to match all curly invocations, even those without dashes then set the `noDashInName` option to `true`.
```

rwjblue: Lets change the name of the option to `requireDash` where the default value is `true` (for now), what do you think?

lib/rules/lint-no-curly-component-invocation.js
```diff
@@ -0,0 +1,98 @@
+const Rule = require('./base');
+const { transformTagName } = require('../helpers/curly-component-invocation');
+const IN_BUILT_HELPERS = [
```
rwjblue: Mind changing this to `BUILT_IN_HELPERS`? I know its kinda pedantic, but its a bit easier for my mind to grok...

patocallaghan:
No problem at all @rwjblue, thanks for the review 🙇 I think I've addressed all your feedback

lib/rules/lint-no-curly-component-invocation.js
```diff
@@ -31,8 +32,40 @@ const IN_BUILT_HELPERS = [
   'with',
   'yield',
 ];
+const DEFAULT_CONFIG = {
+  allow: [],
+  disallow: [],
+  requireDash: true,
```
rwjblue: Note for future: I'd like to change this to `false` in the next major bump (which will be focused on Octane support).

PR #779, opened 2019-07-22
Allow <img onload={{action 'foo'}}> in no-invalid-interactive rule

joankaradimov:
This PR does three things which I believe improve the consistency of the `no-invalid-interactive` rule:

1. Currently it is allowed to add `onerror={{action "foo"}}` on non-interactive elements. Which makes perfect sense, since this is not a user interaction. However, `{{action "foo" on="error"}}` is not allowed. This PR allows both forms.

2. It no longer treats the `load` events as interactions. They are now allowed - same as with `error` events.

3. It removes the limitation for `img` tag. The `load` and `error` events can be used with various other tags.

### DIFF
--- a/lib/rules/lint-no-invalid-interactive.js
+++ b/lib/rules/lint-no-invalid-interactive.js
@@ -111,8 +111,8 @@ module.exports = class InvalidInteractive extends Rule {
           return;
         }
 
-        // Allow onerror={{action "foo"}} on img tags
-        if (this._element.tag === 'img' && node.name === 'onerror') {
+        // Allow onerror={{action "foo"}} and onload={{action "foo"}} on any tag
+        if (this._element.tag === 'img' && ['onerror', 'onload'].includes(node.name)) {
           return;
         }
 


lib/rules/lint-no-invalid-interactive.js
```diff
@@ -81,6 +81,11 @@ module.exports = class InvalidInteractive extends Rule {
             return;
           }
 
+          // Allow {{action "foo" on="error"}} and {{action "foo" on="load"}} on any tag
```
rwjblue: Hmm, I don't think we actually support `{{action 'foo' on="load"}}` or `{{action 'foo' on="error"}}` in Ember's event dispatcher.

Here is the list of the events we listen to (this is the list of valid `{{action "foo" on="<event name here>"}}` and event handler methods on `Ember.Component`):

https://github.com/emberjs/ember.js/blob/9951eea116491f5878eef2211b5b95625261e444/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L70-L98

joankaradimov: Someone could have used `customEvents`. And the behaviour of the `no-invalid-interactive` rule can be configured based on tag, but not based on event name. So if load/error are in `customEvents` there would be no way to configure their ignoring.

rwjblue: > Someone could have used customEvents. 

Ya, definitely true 🤔. I think using custom events is pretty obscure and uncommon, and I'd prefer to avoid adding exceptions for these things all the time when in reality they are very very rarely allowed.

joankaradimov: Ok, I got rid of the `{{action 'foo' on="...."}}` handling.

joankaradimov:
I've removed the `{{action 'foo' on="...."}}` handling and I've forced pushed.

lib/rules/lint-no-invalid-interactive.js
```diff
@@ -111,8 +111,8 @@ module.exports = class InvalidInteractive extends Rule {
           return;
         }
 
-        // Allow onerror={{action "foo"}} on img tags
-        if (this._element.tag === 'img' && node.name === 'onerror') {
+        // Allow onerror={{action "foo"}} and onload={{action "foo"}} on any tag
+        if (node.name === 'onerror' || node.name === 'onload') {
```
rwjblue: What do you think about this:

```suggestion
        if (this._element.tag === 'img' && ['onerror', 'onload'].includes(node.name)) {
```

Specifically, we only want to allow `onload` and `onerror` properties on ``'s (AFAIK) and using `[].includes()` makes the logic a tiny bit easier to read.

joankaradimov: I've applied the change and squashed.

bmish:
@joankaradimov would you mind following up with another PR to add tests for this change? :)

joankaradimov:
Ok, I will.

Issue #354, opened 2018-02-15
feat: line specific rule toggling


FalkF:
Like other linters, ember-template-lint should allow users to enable and disable rules per line:

```{{! template-lint-enable-line bare-strings }}```

```{{! template-lint-disable-line bare-strings }}```

```{{! template-lint-enable-next-line bare-strings }}```

```{{! template-lint-disable-next-line bare-strings }}```

**Comment by rwjblue** [comment]
It is already possible to enable / disable via comments. Are you specifically saying that you want the addition of a single line comment?

**Comment by FalkF** [comment]
Yes, disabling per line would be nice.
This way one can make exceptions to a rule without disabling it for a whole file.
Maybe I manage to implement this :)

**Comment by rwjblue** [comment]
You can already make exceptions to a single rule without disabling it for a whole file. 🤔

**Comment by rwjblue** [comment]
Note: I’m not opposed to the capability, I’m just not sure how to make it work properly. “Next line” is pretty ambiguous when talking about ASTs...

**Comment by rwjblue** [comment]
I'm still curious about what you have in mind here.

@FalkF - Are you still interested in this one?

**Comment by Redsandro** [comment]
> You can already make exceptions to a single rule without disabling it for a whole file. :thinking:

How?

**Comment by rwjblue** [comment]
It is documented in [this section of the README](https://github.com/ember-template-lint/ember-template-lint#per-template) though maybe we need to make it even clearer. Would definitely like help.


---

An instruction will apply to all later siblings and their descendants:

```hbs
<!-- disable for <p> and <span> and their contents, but not for <div> or <hr> -->
<div>
  <hr>
  {{! template-lint-disable }}
  <p>
    <span>Hello!</span>
  </p>
</div>
```

An in-element instruction will apply to only that element:

```hbs
<!-- enable for <p>, but not for <div>, <hr> or <span> -->
<div>
  <hr>
  <p {{! template-lint-enable }}>
    <span>Hello!</span>
  </p>
</div>
```

An in-element instruction with the `-tree` suffix will apply to that element and all its descendants:

```hbs
<!-- configure for <p>, <span> and their contents, but not for <div> or <hr> -->
<div>
  <hr>
  <p {{! template-lint-configure-tree block-indentation "tab" }}>
    <span>Hello!</span>
  </p>
</div>
```

Redsandro:
@rwjblue Hmm, I must have missed that. Apologies. 

> maybe we need to make it even clearer. Would definitely like help.

[...]

PR #780, opened 2019-07-22
Spike template-lint-disable-next-line behaviour

22a:
Hey folks, following on from https://github.com/ember-template-lint/ember-template-lint/issues/354#issuecomment-499115812 I've spiked out what the implementation for `{{!— template-lint-disable-next-line no-foo —}}` could look like. We're interested in the functionality because it neatly brings ember-template-lint to parity with eslint and stylelint.

Here we introduce the notion of `config.disableUntil` to make the handlers aware of what "next line" means so that they're able to pop the temporary config off the stack at the right time.

There are a number of quirks this spike doesn't tackle:
- [ ] Support attaching this temporal pop handler to `before.enter` on BlockStatement and MustacheStatement, not just ElementNode
- [ ] Consider value of adding support for an accompanying `temlate-lint-enable-next-line` construct

If we were to iron out the remaining quirks with this implementation and back it up with some specs, is this something that you think would be valuable to add to this addon?

### DIFF
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -233,14 +233,25 @@ module.exports = class BaseRule {
 
       pluginContext._pushConfig(config.value);
 
-      // Pop the config once we exit the comment's siblings
-      let parentNode = scope.currentNode;
-      astHandlers.ElementNode.keys.children.exit.before.unshift(function(elementNode) {
-        if (elementNode === parentNode) {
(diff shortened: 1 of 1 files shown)


rwjblue:
@22a - Sorry for the long radio silence here, but overall this looks pretty good. I think I'd like to see a series of tests showing the usage to get a bit more comfortable here though.

PR #783, opened 2019-07-24
Add support for reading a file to lint from `process.stdin` (e.g `cat some-file.hbs | ember-template-lint`)

### LINKED ISSUE #782: Unable to use with VIM ALE plugin and other external tools
Pull Request #607 introduced a breaking change where tools that call ember-template-lint with a temporary file are unable to lint because the glob rules filter out the file.

Example usage: `ember-template-lint /tmp/random-dir/temporary-template.hbs`

A built in tool like [ALE](https://github.com/w0rp/ale) will execute this command because it needs to lint the unsaved buffer.

Possible solutions:

1. disable the glob rules if the argument is a file
2. disable the rule if the path is absolute
3. allow a flag to disable globbing
4. allow reading for stdin **best solution**

sukima:
You can now pipe content to ember-template-lint by passing a lone - as a file argument or the more descriptive /dev/stdin:

    $ ember-template-lint - < /some/path/to/file.hbs
    $ cat /some/path/to/file.hbs | ember-template-lint -
    $ ember-template-lint /dev/stdin <<<"<p>"
    $ cat <<EOF | ember-template-lint /dev/stdin
    <p>
    EOF

This will:

Fix #782 
Close #586

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -2,7 +2,7 @@
 
 'use strict';
 
-const rw = require('rw');
+const fs = require('fs');
 const path = require('path');
 const globby = require('globby');
 const Linter = require('../lib/index');
@@ -60,7 +60,11 @@ function printErrors(errors, invocationOptions) {
 }
 
 function lintFile(linter, filePath, moduleId) {
-  let source = rw.readFileSync(filePath, { encoding: 'utf8' });
+  let toRead = filePath === STDIN ? process.stdin.fd : filePath;
+
+  // TODO: swap to using get-stdin when we can leverage async/await
+  let source = fs.readFileSync(toRead, { encoding: 'utf8' });
+
   return linter.verify({ source, moduleId });
 }
 
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -116,7 +116,8 @@ function run() {
     linter = new Linter({ configPath });
   } catch (e) {
     console.error(e.message);
-    return process.exit(1); // eslint-disable-line no-process-exit
+    process.exitCode = 1;
+    return;
   }
 
   let errors = {};
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -112,6 +112,10 @@ function checkConfigPath() {
   return getArgumentValue('--config-path');
 }
 
+function filePathFromArgs() {
+  return getArgumentValue('--filename') || '';
+}
+
 function run() {
   let configPath = checkConfigPath();
   let linter;
@@ -125,7 +129,8 @@ function run() {
 
   let errors = getRelativeFilePaths().reduce((errors, relativeFilePath) => {
     let filePath = path.resolve(relativeFilePath);
-    let fileErrors = lintFile(linter, filePath, relativeFilePath.slice(0, -4));
+    let fileName = relativeFilePath === STDIN ? filePathFromArgs() : relativeFilePath;
+    let fileErrors = lintFile(linter, filePath, fileName.slice(0, -4));
 
     if (
       fileErrors.some(function(err) {
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -10,8 +10,8 @@ const chalk = require('chalk');
 
 const STDIN = '/dev/stdin';
 
-function printErrors(errors) {
-  const quiet = process.argv.indexOf('--quiet') !== -1;
+function printErrors(errors, invocationOptions) {
+  let { quiet, json, verbose } = invocationOptions.named;
 
   let errorCount = 0;
   let warningCount = 0;
@@ -30,14 +30,14 @@ function printErrors(errors) {
     errors[filePath] = errorsFiltered.concat(warnings);
   });
 
-  if (process.argv.indexOf('--json') + 1) {
+  if (json) {
     console.log(JSON.stringify(errors, null, 2));
   } else {
     Object.keys(errors).forEach(filePath => {
       let options = {};
       let fileErrors = errors[filePath] || [];
 
-      if (process.argv.indexOf('--verbose') + 1) {
+      if (verbose) {
         options.verbose = true;
       }
 
@@ -121,6 +121,27 @@ function parseArgv(_argv) {
       }
 
       options.named.filename = filename;
+    } else if (arg === '--quiet') {
+      // TODO: add error handling when named args are out of order
+      if (!shouldHandleNamed) {
+        continue;
+      }
+
+      options.named.quiet = true;
+    } else if (arg === '--json') {
+      // TODO: add error handling when named args are out of order
+      if (!shouldHandleNamed) {
+        continue;
+      }
+
+      options.named.json = true;
+    } else if (arg === '--verbose') {
+      // TODO: add error handling when named args are out of order
+      if (!shouldHandleNamed) {
+        continue;
+      }
+
+      options.named.verbose = true;
     } else if (arg === '--') {
       // named arguments are not allowed after `--`
       shouldHandleNamed = false;
@@ -132,10 +153,12 @@ function parseArgv(_argv) {
 }
 
 function run() {
+  let options = parseArgv(process.argv.slice(2));
+
   let {
     named: { configPath, filename: filePathFromArgs = '' },
     positional: fileArgs,
-  } = parseArgv(process.argv.slice(2));
+  } = options;
 
   let linter;
   try {
@@ -168,7 +191,7 @@ function run() {
   }
 
   if (Object.keys(errors).length) {
-    printErrors(errors);
+    printErrors(errors, options);
   }
 }
 
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -60,16 +60,7 @@ function printErrors(errors) {
 }
 
 function lintFile(linter, filePath, moduleId) {
-  let source;
-  try {
-    source = rw.readFileSync(filePath, { encoding: 'utf8' });
-  } catch (error) {
-    if (error.code === 'ENXIO') {
-      return [];
-    } else {
-      throw error;
-    }
-  }
+  let source = rw.readFileSync(filePath, { encoding: 'utf8' });
   return linter.verify({ source, moduleId });
 }
 
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -64,24 +64,20 @@ function lintFile(linter, filePath, moduleId) {
   return linter.verify({ source, moduleId });
 }
 
-function expandFileGlobs(fileArgs) {
-  return fileArgs.reduce((filePaths, fileArg) => {
-    let files;
-    if (['-', STDIN].includes(fileArg)) {
-      if (filePaths.includes(STDIN)) {
-        return filePaths;
-      }
-      files = [STDIN];
-    } else {
-      files = globby
-        .sync(fileArg, {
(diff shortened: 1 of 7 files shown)


bin/ember-template-lint.js
```diff
@@ -67,14 +67,18 @@ function getRelativeFilePaths() {
...
           ignore: ['**/dist/**', '**/tmp/**', '**/node_modules/**'],
           gitignore: true,
-        })
-      );
+        });
+      }
+      return filePaths.concat(files);
     }, [])
-    .filter(filePath => filePath.slice(-4) === '.hbs');
+    .filter(filePath => filePath === '/dev/stdin' || filePath.slice(-4) === '.hbs');
```
rwjblue: I think we can move this filter up into the `else` of the conditional just above, and then we won't need to have `filePath === '/dev/stdin' ||` (since the `else` branch is only ran when we **aren't** using `/dev/stdin` AFAICT).

sukima:
> Naively, I would assume ember-template-lint < foo.hbs should work

Will look into it.

> What do you think about this instead? [using array instead of `&&`]

:+1: 

> I think we can move this filter up into the else of the conditional just above

On it. :+1: 

> I think [this](https://github.com/ember-template-lint/ember-template-lint/issues/782#issuecomment-515130446) is correct, but we do probably want to add a new argument to the CLI to give us the file name

Oh.... :hankey: um. ok looking into it....

sukima:
@rwjblue Ready!

bin/ember-template-lint.js
```diff
@@ -58,35 +60,62 @@ function printErrors(errors) {
...
-    }, [])
-    .filter(filePath => filePath.slice(-4) === '.hbs');
+        .filter(filePath => filePath.slice(-4) === '.hbs');
+    }
+    return filePaths.concat(files);
+  }, []);
+}
+
+function getRelativeFilePaths() {
+  var fileArgs = process.argv.slice(2).filter(arg => arg.slice(0, 2) !== '--');
```
rwjblue: I'm a tad confused how we are avoiding reading the value of `--filename` as one of the glob paths here. AFAICT with the following invocation:

```
ember-template-lint --filename app/templates/application.hbs < /some/other/path/app/templates/application.hbs
```

This line of code would result in:

```js
var fileArgs = ['app/templates/application.hbs'];
```

Which I believe is incorrect.

I think in order to combat these sorts of issues, I'd like to see us move to a smallish `parseArgv` helper function (but would prefer to avoid pulling in `yargs` or `commander`). 

_Something_ like:

```js
function parseArgv(argv) {
  let options = {
    positional: [],
    named: {},
  };

  for (let i = 0; i < argv.length; i++) {
    let arg = argv[i];
    
    if (arg.startsWith('--')) {
      let value = argv[i + 1];
	  let optionName = arg.substring(2);

      options.named[optionName] = value;

      i++;
    }

    options.positional.push(arg);
  }
}

const options = parseArgv(process.argv.slice(2));
```

Then throughout the rest of the script we can refer to `options.positional` as the `fileArgs` and our passed flags are in `options.named`.  

I _think_ this will both fix the bug I mentioned above and simplify the code overall. What do you think?

test/unit/bin/ember-template-lint-test.js
```diff
@@ -89,6 +89,93 @@ describe('ember-template-lint executable', function() {
...
+          'node',
+          [
+            '../../../bin/ember-template-lint.js',
+            '--filename',
+            'app/templates/application.hbs',
+            '<',
+            'app/templates/application.hbs',
+          ],
+          {
+            cwd: './test/fixtures/with-errors',
```
rwjblue: I think we want these "implicit stdin" tests to be done in a directory that is **not** the same as the tests above. We want to confirm that the linter is properly linting the stdin and _cannot_ be reading the file from the `--filename`.

bin/ember-template-lint.js
sukima: Seems legit to me. I was attempting to copy previous precedence from `--config-path`. An argv parser is much nicer.

bin/ember-template-lint.js
```diff
@@ -58,35 +60,62 @@ function printErrors(errors) {
 }
 
 function lintFile(linter, filePath, moduleId) {
-  var source = fs.readFileSync(filePath, { encoding: 'utf8' });
+  var source;
+  try {
+    source = rw.readFileSync(filePath, { encoding: 'utf8' });
+  } catch (error) {
+    if (error.code === 'ENXIO') {
```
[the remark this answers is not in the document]
sukima: `ENXIO` is the system error code for when we attempt to read from `/dev/stdin` and it is not available.

> “No such device or address.” The system tried to use the device represented by a file you specified, and it couldn’t find the device. This can mean that the device file was installed incorrectly, or that the physical device is missing or not correctly attached to the computer.
> https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html#index-ENXIO

The general user land condition this is handling is when the user has asked to read from stdin but has not provided a stdin pipe. Pretty sure this is the case if they run `ember-template-lint` with no arguments or pass in `/dev/stdin` but stdin is attached to their TTY.

I tried looking further and discovered that this error showed up when [running this test](https://github.com/ember-template-lint/ember-template-lint/blob/4ffbc81077a3582aca4e40b4ae05b9986d8ffcb8/test/unit/bin/ember-template-lint-test.js#L10). because it offers no args the assumption is stdin but nothing has been piped to it so the OS throws an ENXIO error. I added that conditional to get the test to pass. Maybe my understanding is inaccurate?

bin/ember-template-lint.js
rwjblue: Ya, I definitely saw that you were following the existing style. It just felt like “time” to move to a better parsing situation.

rwjblue:
@sukima - Any updates on this? No giant rush, I'd just like to get this fix out for folks...

sukima:
Eeekk!!! so sorry! I went on vacation and forgot all my commitments. On it....

rwjblue:
No worries! Hope you had a great time off!

rwjblue:
@sukima - 👋

sukima:
Ugh yeah im sorry. I am a bad dev. Parsing args manually (no dependency) is scary. I promise to look into this soon.

rwjblue:
> I am a bad dev.

Noooo, definitely not. We are all busy. :smile:

sukima:
@rwjblue I think I addressed your concerns. Should be ready now. Sorry for the delay; once I started playing with it the work wasn't as bad as I had thought it would be.

rwjblue:
Looks like some linting failures are the only issue in CI:

```
/home/travis/build/ember-template-lint/ember-template-lint/bin/ember-template-lint.js
   99:27  error  Replace `arg·=·argv.shift(` with `(arg·=·argv.shift()`           prettier/prettier
   99:27  error  Expected a conditional expression and instead saw an assignment  no-cond-assign
  120:25  error  Insert `,`                                                       prettier/prettier
```

sukima:
@rwjblue resolved.

rwjblue:
@sukima - I pushed some commits here, mind reviewing?

bin/ember-template-lint.js
```diff
@@ -170,9 +166,15 @@ function run() {
   }
 
   let errors = {};
-  let relativeFilePaths = fileArgs.length === 0 ? [STDIN] : expandFileGlobs(fileArgs);
+  let filesToLint;
+
+  if (positional.length === 0 || positional.includes('-') || positional.includes(STDIN)) {
+    filesToLint = new Set([STDIN]);
```
sukima: This will strait up block any globbing. Which means the use of STDIN is an exclusive operation.

```
ember-template-lint app/**/*.js - --filename foobar.js
```

Will not be supported. If that is intentional then it might be worth redoing the arguments to be explicit about this. For example not relying on `-` on its own. Maybe this to turn on STDIN **mode**

```
ember-template-lint --read-from-stdin --filename foobar.js
```

Thoughts?

bin/ember-template-lint.js
```diff
@@ -60,7 +60,11 @@ function printErrors(errors, invocationOptions) {
 }
 
 function lintFile(linter, filePath, moduleId) {
-  let source = rw.readFileSync(filePath, { encoding: 'utf8' });
+  let toRead = filePath === STDIN ? process.stdin.fd : filePath;
+
+  // TODO: swap to using get-stdin when we can leverage async/await
+  let source = fs.readFileSync(toRead, { encoding: 'utf8' });
+
```
sukima: There is a *lot* more going on [in the `rw` package](https://github.com/mbostock/rw/blob/master/lib/rw/read-file-sync.js) are we **absolutely** sure this is correct? My instinct is that DYI pipes could attract dragons that a package like `rw` has already tamed.

bin/ember-template-lint.js
```diff
@@ -121,6 +121,27 @@ function parseArgv(_argv) {
       }
 
       options.named.filename = filename;
+    } else if (arg === '--quiet') {
+      // TODO: add error handling when named args are out of order
```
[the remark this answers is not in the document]
rwjblue: I should update the comment to be clearer, the specific thing that we should throw on here is if you use `--` **and _then_** use `--quiet` (or any other named argument). This is an error condition because everything after `--` should be filenames to lint.

bin/ember-template-lint.js
rwjblue: > This will strait up block any globbing. 

Yes, that is intentional from my perspective. Using both stdin and filename globs together seems very bizarre. Either we lint stdin or we lint globs, not both.

bin/ember-template-lint.js
rwjblue: Ya, I reviewed that code a bit. There are a few specific things that we should avoid that are done in `rw` but the largest one is that if a file that is specified does not exist, we should **not** just assume it means that we want stdin.

bin/ember-template-lint.js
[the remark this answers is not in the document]
sukima: What if the file name **is** `--quiet`? I would expect this

```
ember-template-lint --quiet -- --quiet
```

to be quiet and read a file named `--quiet` Then again we are filtering out by `.hbs` extension so I don't know. Seems like more work for an edge case that is a little nonsensical.

sukima:
This is pedantic I know. I prefer the while/switch loop syntax. It just reads better for me. I know, switches are evil and in most all cases that is true but there are a few cases when it is appropriate and arg parsing is one of them.

It is your project so go however you like. It won't block a :+1: from me. I only wanted to voice my personal opinion on the matter.

rwjblue:
>  I know, switches are evil and in most all cases that is true but there are a few cases when it is appropriate and arg parsing is one of them.

No no, I really preferred the switch setup (with the exception of the named block, which is just a personal issue because my mental JS parser can't handle the labels).

I'll bring back that bit...

bin/ember-template-lint.js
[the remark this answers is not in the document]
rwjblue: > Seems like more work for an edge case that is a little nonsensical.

Yeah, probably. 🤔

bin/ember-template-lint.js
rwjblue: Yeah, possibly. I want to use [get-stdin](https://github.com/sindresorhus/get-stdin) instead (its a bit simpler to reason about, seems more "popular", etc), and will make an issue to migrate to that (only reason I didn't do it here was because it is promise based, and I'd rather rewrite this to use `async`/`await` than `.then`).

rwjblue:
Thank you again for all of your hard work here @sukima!

PR #788, opened 2019-07-28
Add support for handling `<button tabindex={{if this.foo 0 -1}}><button>` to no-positive-tabindex rule

lifeart:
https://github.com/ember-template-lint/ember-template-lint/pull/743


test/unit/rules/lint-no-positive-tabindex-test.js
```diff
@@ -14,6 +14,9 @@ generateRuleTests({
     '<span tabindex={{"-1"}}>baz</span>',
```
MelSumner: Ideally these would be buttons and not spans. Spans are not interactive elements.

rwjblue: Great point! Updated during pairing session with @MelSumner...

bmish:
@lifeart can you add `unless` support to this to?

lifeart:
@bmish oh, sorry, just found your comment. Thank you for PR!

Issue #746, opened 2019-06-19
Update recommended rule set


MelSumner:
## Primary objectives:

* Remove "style" related rules from the recommended set
* Add Octane related rules to recommended set
* Update list of A11Y rules in recommended set

### Add

* no-abstract-roles
* no-action
* no-curly-component-invocation
* no-extra-mut-helper-argument
* no-implicit-this
* no-obsolete-elements
* no-positive-tabindex
* no-unnecessary-component-helper
* require-iframe-title
* require-valid-alt-text

### Remove

* block-indentation
* img-alt-attributes
* linebreak-style
* no-unnecessary-concat
* quotes
* self-closing-void-elements

### Change Default Config

* link-rel-noopener -> `strict` config

rwjblue:
After reviewing with @MelSumner, our current recommendations for a good "recommended" set that satisfies the objectives above are:

### Added

* no-abstract-roles
* no-action-modifiers
* no-curly-component-invocation
* no-element-event-actions
* no-extra-mut-helper-argument
* no-implicit-this
* no-obsolete-elements
* no-positive-tabindex
* no-unnecessary-component-helper
* require-iframe-title
* require-valid-alt-text

### Removed

* block-indentation
* img-alt-attributes
* linebreak-style
* no-unnecessary-concat
* quotes
* self-closing-void-elements

### Change Default Config

* link-rel-noopener -> `strict` config

bmish:
Overall looks good but I have a few questions.

Regarding the additions:

1. My understanding is that the [no-action-modifiers](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-action-modifiers.md) and [no-element-event-actions](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-element-event-actions.md) rules are exact opposites of each other. So why are they both enabled?

Regarding the removals:

1. Are the style rules (`block-indentation`, `linebreak-style`, `quotes`) removed because prettier for handlebars will take the place of these? Because otherwise, I see a lot of value in these rules for enforcing consistent, clean, and nit-free code.

To recap the other removals:

1. `img-alt-attributes` is deprecated and replaced by `require-valid-alt-text`.
2. I assume `no-unnecessary-concat` is removed because of the discussion in #604.
3. I assume `self-closing-void-elements` is removed because of the discussion in #691.

rwjblue:
[...]

PR #790, opened 2019-07-31
Add "octane" configuration preset.

rwjblue:
This config represents what we believe will be the "recommended" config for the next major version of ember-template-lint which will target Ember's Octane Edition.

At the moment we are adding the configuration to a specific `octane` preset which will allow users to begin using it immediately, but once Octane "ships" we will release a major version bump of ember-template-lint changing `recommended` to use this new config.

For easier tracking here are the primary changes from the current `recommended` config (as of 1.5.0):

### Add

* no-abstract-roles
* no-action-modifiers
* no-curly-component-invocation
* no-element-event-actions
* no-extra-mut-helper-argument
* no-implicit-this
* no-obsolete-elements
* no-positive-tabindex
* no-unnecessary-component-helper
* require-iframe-title
* require-valid-alt-text

### Remove

* block-indentation
* img-alt-attributes
* linebreak-style
* no-unnecessary-concat
* quotes
* self-closing-void-elements

### Change Default Config

* link-rel-noopener -> `strict` config

---

Implements the changes (currently) suggested in https://github.com/ember-template-lint/ember-template-lint/issues/746.


mydea:
Hmm, not sure, but I get `Cannot find configuration for extends: octane`, with this config:

```js
'use strict';

module.exports = {
  extends: 'octane',
  rules: {
    'block-indentation': true,
    'no-bare-strings': true
  }
};
```

(on 1.5.1) - am I doing something wrong?

rwjblue:
Hmm, no that looks good to me. Would you mind opening an issue so we can track down whats wrong?

PR #791, opened 2019-08-01
Ensure table-group rule allows `each`, `each-in`, `let`, and comments

mongoose700:
With this change, we allow for `#each` and nesting `#each`, `#if`, and `#unless` blocks in a table (as long as the children are valid). It also allows for comments (the previous rule for allowing comments only worked if there were no other children).

### DIFF
--- a/lib/rules/lint-table-groups.js
+++ b/lib/rules/lint-table-groups.js
@@ -1,30 +1,23 @@
 'use strict';
 
+const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
 const message = 'Tables must have a table group (thead, tbody or tfoot).';
 
 const ALLOWED_TABLE_CHILDREN = ['thead', 'tbody', 'tfoot', 'caption', 'colgroup'];
 
-function hasDisallowedChildren(tableNode) {
-  const validBlocks = tableNode.children.filter(node => {
-    const isBlock = node.type === 'BlockStatement';
-    if (!isBlock) {
-      return false;
+// For effective children, we skip over any control flow helpers,
+// since we know that they don't render anything on their own
+function getEffectiveChildren(node) {
+  const unflattenedEffectiveChildren = AstNodeInfo.childrenFor(node).map(node => {
+    if (AstNodeInfo.isControlFlowHelper(node)) {
+      return getEffectiveChildren(node);
+    } else {
+      return [node];
     }
-    const isIf = node.path.original === 'if';
-    const isUnless = node.path.original === 'unless';
-    return isIf || isUnless;
   });
-  const childrenFromValidBlocks = [].concat.apply(
-    [],
-    validBlocks.map(block => [].concat(block.program.body, block.inverse ? block.inverse.body : []))
-  );
-  return ![]
-    .concat(
-      childrenFromValidBlocks,
-      tableNode.children.filter(item => validBlocks.indexOf(item) === -1)
-    )
-    .every(isAllowedTableChild);
+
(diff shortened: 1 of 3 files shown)


lib/rules/lint-table-groups.js
```diff
@@ -49,31 +45,22 @@ function isAllowedTableChild(node) {
...
+    case 'MustacheCommentStatement':
+      return true;
     case 'TextNode':
       return !/\S/.test(node.chars);
   }
 
   return false;
 }
 
-function hasChildTags(tableNode) {
```
mongoose700: This special check is what caused it to fail for comments next to valid children. It seems cleaner to remove this preemptive check entirely, so we don't have to keep two different checks consistent.

lib/rules/lint-table-groups.js
```diff
@@ -1,30 +1,26 @@
...
-    .concat(
-      childrenFromValidBlocks,
-      tableNode.children.filter(item => validBlocks.indexOf(item) === -1)
-    )
-    .every(isAllowedTableChild);
+
+  return [].concat.apply([], unflattenedEffectiveChildren);
+}
+
+function hasAllowedChildren(tableNode) {
```
bmish: This seems more like: `hasOnlyAllowedChildren`

lib/rules/lint-table-groups.js
```diff
@@ -49,31 +45,22 @@ function isAllowedTableChild(node) {
         return ALLOWED_TABLE_CHILDREN.includes(tagNameAttribute.value.chars);
       }
       break;
+    case 'CommentStatement':
+    case 'MustacheCommentStatement':
+      return true;
```
bmish: Can you add tests for comments?

mongoose700: There are tests for comments, but the extra check we had before was causing them to go through a special case. I removed that check entirely, so I don't think we need another test of a comment adjacent to a valid child.

bmish: You fixed a real bug with ignoring comments so I do think it's worthwhile to add a test to show that bug is fixed :)

mongoose700: I generally don't like adding tests that make sure code that has been deleted isn't there, but it was easy enough to add it to an existing test.

test/unit/rules/lint-table-groups-test.js
```diff
@@ -316,5 +356,15 @@ generateRuleTests({
         column: 0,
       },
     },
+    {
+      template: '<table>some text</table>',
```
mongoose700: Previously this would have been accepted, since the check in `hasChildTags` didn't have the whitespace check. There just weren't any tests for it.

lib/rules/lint-table-groups.js
```diff
@@ -1,30 +1,22 @@
...
 
-function hasDisallowedChildren(tableNode) {
-  const validBlocks = tableNode.children.filter(node => {
-    const isBlock = node.type === 'BlockStatement';
-    if (!isBlock) {
-      return false;
+// For effective children, we skip over any if, unless, and else blocks
+function getEffectiveChildren(node) {
+  const unflattenedEffectiveChildren = AstNodeInfo.childrenFor(node).map(node => {
+    if (AstNodeInfo.isIf(node) || AstNodeInfo.isUnless(node) || AstNodeInfo.isEach(node)) {
```
rwjblue: Mind adding `each-in`, `let`, and `with` support (along with tests) also?

mongoose700: Added. Could you take another look?

PR #794, opened 2019-08-06
Add missing rule to `rules.md`

bmish:
Forgotten in #768.

CC: @patocallaghan


rwjblue:
Yikes. I keep forgetting this in reviews 😭. We should add a test to ensure this kind of thing doesn't slip. Basically to assert the following:

* All files under `lib/rules/` (except `base.js` and `index.js`) exist as values in the `lib/rules/index.js` reexport.
* All files under `lib/rules/` (same exception) have a corresponding `docs/rule/` file.
* All files under `docs/rule/` have a link from `docs/rules.md`.

rwjblue:
I created https://github.com/ember-template-lint/ember-template-lint/issues/797 to track that suggestion...

Issue #797, opened 2019-08-06
Add test ensuring new rule setup is correct.


rwjblue:
I keep forgetting some of the basic setup work for adding a new rule (thanks to @bmish for catching my mistakes). In order to save us from my forgetfulness we should add a test to ensure this kind of thing doesn't slip. 

Basically to assert the following:

* All files under `lib/rules/` (except `base.js` and `index.js`) exist as values in the `lib/rules/index.js` reexport.
* All files under `lib/rules/` (same exception) have a corresponding `docs/rule/` file.
* All files under `docs/rule/` have a link from `docs/rules.md`.

lifeart:
On it

rwjblue:
I think this was resolved by #798

PR #798, opened 2019-08-07
test: ensuring new rule setup is correct #797

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/797


bmish:
Can you also add a test that each rule has its own test file?

lifeart:
@bmish, checks for tests existence added

PR #803, opened 2019-08-14
Add 'octane' config to export list so it can be found

### LINKED ISSUE #799: "octane" preset does not work
I get `Cannot find configuration for extends: octane`, with this config:

```js
'use strict';

module.exports = {
  extends: 'octane',
  rules: {
    'block-indentation': true,
    'no-bare-strings': true
  }
};
```

on 1.5.1. (see #790)

pgengler:
This should fix #799.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -2,5 +2,6 @@
 
 module.exports = {
   '0-8-recommended': require('./0-8-recommended'),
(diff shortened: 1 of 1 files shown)


rwjblue:
🤦‍♂

rwjblue:
Thank you!

PR #805, opened 2019-08-19
Update `no-action-modifiers` and `no-element-event-actions` rules to suggest using the `on` modifier

### LINKED ISSUE #804: Default rules for 'octane' config contain conflicting recommendations
Using the `octane` config, both the "no-action-modifiers" and "no-element-event-actions" rules are enabled; I'm guessing the idea is to encourage use of `{{on ...}}` but the recommendations for these rules don't get along:

The no-action-modifiers error says to use `onclick`:
> `action` modifiers should not be used (use `onclick` attributes instead).  no-action-modifiers

While no-element-event-action says to use `{{action ...}}`:
> Use Ember `{{action ...}}` helper instead of HTML element event properties like `onclick`.  no-element-event-actions

bmish:
Fixes #804. This was also discussed in #746.

`on` was introduced in [Ember 3.11](https://blog.emberjs.com/2019/07/15/ember-3-11-released.html).

### DIFF
--- a/docs/rule/no-action-modifiers.md
+++ b/docs/rule/no-action-modifiers.md
@@ -2,16 +2,18 @@
 
 This rule forbids the use of `{{action}}` modifiers on elements.
 
+The recommended alternative is the `on` modifier. `on` is available in Ember 3.11+ and by [polyfill](https://github.com/buschtoens/ember-on-modifier) for earlier versions.
+
 This rule **forbids** the following:
 
 ```hbs
-<button {{action 'handleClick'}}>
+<button {{action 'submit'}}>Submit</button>
 ```
 
 This rule **allows** the following:
 
 ```hbs
-<button onclick={{action 'handleClick'}}>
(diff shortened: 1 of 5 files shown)


docs/rule/no-action-modifiers.md
```diff
@@ -2,6 +2,8 @@
 
 This rule forbids the use of `{{action}}` modifiers on elements.
 
+In Ember 3.11+, the recommended alternative is the `on` modifier. Prior to Ember 3.11, the alternative is to use HTML element event properties such as `onclick`.
```
rwjblue: Maybe we should just suggest `{{on` all the time? [ember-on-modifier](https://github.com/buschtoens/ember-on-modifier) is tested all the way back to 2.13 (uses the `ember-modifier-manager-polyfill` for older version support), what do you think?

bmish: Agreed, updated to only suggest `on` and link to the polyfill.

rwjblue:
Thank you!

PR #814, opened 2019-08-28
[rule] add new no-class-concat rule

gabrielcsapo:
*Why?*

We have had a lot of issues with people using concat for classes because finding and replacing class is extremely difficult. 

Since there is a lint rule against styles being concated, this seems like a natural fit.

### DIFF
--- a/docs/rule/no-class-concat.md
+++ b/docs/rule/no-class-concat.md
@@ -1,0 +1,15 @@
+## no-class-concat
+
+Concating class names causes issues with maintainability. Disallowing concating class names will help ensure that class names are defined statically. 
+
+This rule **forbids** the following:
+
+```hbs
+<div class={{concat "foo" bar}}>
+```
+
+This rule **allows** the following:
+
+```hbs
+<div class="foo">
+```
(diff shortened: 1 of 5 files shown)


rwjblue:
Yes, exactly. My point more broadly is just that `concat` _itself_ is not needed unless you are already within curlies. I’m not specifically trying to debate the use of “compound classes” by that statement, just that _either way_ using `concat` to do is is much harder to read and wasteful.

rwjblue:
> no-unnecessary-concat

I think we already have this rule, and it’s for a slightly different thing. (Though there has been a bit of debate on that particular rules merits). I suppose we could put this logic in there there. It does seem to fit.

gabrielcsapo:
Would you be interested in this changing to a no-dynamic-classes rule? That would help the issue we are having right now.

rwjblue:
@gabrielcsapo - Sure, I guess that's fine (can't be enabled by default IMHO) but this implementation isn't doing that either. It only prevents using `{{concat` and doesn't do anything to prevent `class="foo-{{bar}} baz"`.

rwjblue:
@gabrielcsapo - How would you like to proceed here?

gabrielcsapo:
@rwjblue going to try and update to create the no-dynamic-classes-rule. Will close this out and open a new PR.

PR #819, opened 2019-09-10
Added `no-whitespace-for-layout` rule

MelSumner:
If merged, this rule will prevent the use of whitespace for formatting in templates.


test/unit/rules/lint-no-whitespace-for-layout-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-whitespace-for-layout',
+  config: true,
+
+  good: ['START FINISH', 'START&nbsp;FINISH'],
```
bmish: Perhaps add some test cases with
* elements like `<div>`
* linebreaks?

lib/rules/lint-no-whitespace-for-layout.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = class noWhitespaceForLayout extends Rule {
+  visitor() {
+    return {
+      TextNode(node) {
+        let source = this.sourceForNode(node);
+        let matches = source.match(/(( )|(\&nbsp\;))(( )|(\&nbsp\;))/g);
+        if (matches !== null) {
+          this.log({
+            message: 'Excess space detected',
```
bmish: Can you store this error message in a constant `ERROR_MESSAGE` and export it for use in the tests cases to avoid duplication? Many other rules follow that pattern.

docs/rule/no-whitespace-for-layout.md
```diff
@@ -0,0 +1,42 @@
+# no-whitespace-for-layout
```
bmish: Would you be able to add some valid / invalid code samples in the documentation? Specifically, it would help to suggest some alternatives (good examples) for linebreaks such as `<br>`, CSS padding/margin, etc.

There's a `_TEMPLATE_.md` with a suggested documentation outline including the good/bad example section.

MelSumner: updated, sorry 'bout that.

lib/rules/lint-no-whitespace-for-layout.js
```diff
@@ -0,0 +1,26 @@
+// lint-no-whitespace-for-layout.js
```
rwjblue: lets kill this comment

lib/rules/lint-no-whitespace-for-layout.js
```diff
@@ -0,0 +1,26 @@
...
+'use strict';
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'Excess whitespace detected.';
+
+module.exports = class noWhitespaceForLayout extends Rule {
```
rwjblue:
```suggestion
module.exports = class NoWhitespaceForLayout extends Rule {
```

lib/rules/lint-no-whitespace-for-layout.js
```diff
@@ -0,0 +1,26 @@
...
+  visitor() {
+    return {
+      TextNode(node) {
+        let source = this.sourceForNode(node);
+        let matches = source.match(/(( )|(\&nbsp\;))(( )|(\&nbsp\;))/g);
```
rwjblue: Can you put a comment here to say what this is doing? I'm thinking something like:

```
// check for two ` ` or `&nbsp;` in a row
```

lib/rules/lint-no-whitespace-for-layout.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'Excess whitespace detected.';
+
+module.exports = class NoWhitespaceForLayout extends Rule {
+  visitor() {
+    return {
+      TextNode(node) {
```
lifeart: should this rule be ignored inside `<pre>`, `<script>`, `<style>` tags?

lifeart: Should it warn for common cases like
`{{this.name}} years`

`age is {{this.years}`

PR #826, opened 2019-09-21
Start testing with Node 12

bmish:
We might as well start testing with Node 12.

https://nodejs.org/en/about/releases/


rwjblue:
Thank you!

PR #827, opened 2019-09-21
Add CI check to ensure yarn.lock is up-to-date

bmish:
This will ensure `yarn install` is run after any changes to package.json.

Example output of the CI failure caused by forgetting to run `yarn install`:

```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

Similar example: https://github.com/ember-cli/eslint-plugin-ember/pull/519


rwjblue:
Thanks!

PR #828, opened 2019-09-22
Add eslint-plugin-filenames to enforce kebab-case filenames internally

bmish:
Prevents adding incorrectly-named files.

https://github.com/selaux/eslint-plugin-filenames


rwjblue:
Thanks!

PR #830, opened 2019-09-23
Enable optional eslint rules internally

bmish:
Almost all of the included changes are autofixes, plus a few manual fixes.

These are the same rules used internally by [eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember/blob/master/.eslintrc.js), minus a few rules that had too many violations.


rwjblue:
> The most common autofix is converting let to const because of the prefer-const rule

I significantly prefer the "liberal let" style ([from this blog post](https://madhatted.com/2016/1/25/let-it-be)). Specifically, using `const` for "true constants" (e.g. require'd things).

bmish:
Interesting, I'll go ahead and remove `prefer-const` and the fixes for that shortly.

rwjblue:
Thank you, and sorry for the extra work.

bmish:
Updated!

PR #829, opened 2019-09-22
Add eslint-plugin-eslint-comments internally

bmish:
This neat plugin checks for poor practices around `// eslint-disable` comments.

In particular, it prevents adding unnecessary disable comments, among other things.

https://github.com/mysticatea/eslint-plugin-eslint-comments

PR #831, opened 2019-09-23
Enforce eslint-plugin-eslint-comments recommended rules internally

bmish:
This was supposed to be part of #829.

https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/


rwjblue:
Whoops! Thanks for following up with the fix!

PR #836, opened 2019-09-27
Expose internal errors from erroneous config when consuming it

### LINKED ISSUE #833: nonexistent plugin rules breaks Linter
I'm facing an issue where the rules are not working when a plugin is involved.

```
module.exports = {
  plugins: [
    {
      name: 'someRUles',
      rules: {
        'someRules': require('./template-lint-rules/someRules'),
        ...
      },
    },
  ],

  extends: 'recommended',

  rules: {
    'no-partial': true,
    'no-bare-strings': true,
    },
```

If I remove the plugin, the error are triggered as intended. Any ideas on why this is happening would be appreciated :)

ygongdev:
- Modified `get-config.js` to now throw properly if the consumed config has errors.
- Added a test.

Closes #833

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -27,8 +27,9 @@ function resolveConfigPath(configPath, pluginPath) {
   }
 
   let filePath = pluginPath || usedPath;
+  // Making sure that the filePath exists
   try {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

PR #846, opened 2019-10-06
Update `table-groups` to enforce required ordering of table children

initram:
This makes sure that a table groups are in the correct order according to the spec. It only verifies the order, not the cardinality of the table-groups.

Fixed the ordering part of #285. Not sure if we should also lint the carinality of the groups (some groups are only allowed once, some must appear, some zero to many times), e.g., should the linter warn when a table has two tbody groups?

### DIFF
--- a/lib/rules/lint-table-groups.js
+++ b/lib/rules/lint-table-groups.js
@@ -4,7 +4,8 @@ const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
 
 const message = 'Tables must have a table group (thead, tbody or tfoot).';
-const orderingMessage = 'Tables must have table groups in the correct order (caption, colgroup, thead, tbody then tfoot).';
+const orderingMessage =
+  'Tables must have table groups in the correct order (caption, colgroup, thead, tbody then tfoot).';
 
 const ALLOWED_TABLE_CHILDREN = ['caption', 'colgroup', 'thead', 'tbody', 'tfoot'];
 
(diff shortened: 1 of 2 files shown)


rwjblue:
> should the linter warn when a table has two tbody groups?

I believe that the spec allows you to have multiple `tbody` groups (please double check me on this), but we should warn/error in this case (duplicate `tfoot`):

```hbs
<table>
  <tfoot></tfoot>
  <tfoot></tfoot>
</table>
```

We should also warn if we are missing a required group and there are no dynamic invocations in the "correct" position (if there are dynamic invocations we can't tell if those will introduce the missing group):

```hbs
{{! should warn, no `tbody` }}
<table>
</table>

{{! can't warn RE: missing `tbody` because it could be added by `anything` }}
<table>
  {{anything}}
</table>
```

initram:
I think it would be difficult to make a good implementation of limiting the number of of different groups.

The following examples would be hard to verify (i feel)
`{{! should not warn, only one `thead` }}
<table>
  {{#if foo}}
    <thead>true</thead>
  {{else}}
    <thead>false</thead>
  {{/if}}
</table>

{{! should warn, potentially multiple `thead` }}
<table>
  {{#each foo as |bar|}}
    <thead>{{bar}}</thead>
  {{/each}}
</table>
`

I guess it could turn of the checking if it sees any conditional block or loop. but I feel that would be hard to understand. We may have to look at some real world examples of people working with tables to better understand the situations we want to catch.

rwjblue:
Ya, good point. No need to couple that to these fixes...

PR #848, opened 2019-10-07
Add `no-whitespace-within-word` rule

MelSumner:
If merged, this adds a new rule to prevent the use of whitespace within a word for styling purposes, and should catch the most common offenders.


lib/rules/lint-no-whitespace-within-word.js
```diff
@@ -0,0 +1,66 @@
+// lint-no-whitespace-within-word.js
+'use strict';
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+const ERROR_MESSAGE = 'Excess whitespace in layout detected.';
+
+// Entity format: nbsp -> &nbsp; -> '\\&nbsp\\;'
+// Number format: 160 -> &#160; -> '\\&\\#160\\;'
+const blackList = [
```
rwjblue: Can we make this a single column (instead of 2 and sometimes 3 on a line)?

lib/rules/lint-no-whitespace-within-word.js
```diff
@@ -0,0 +1,66 @@
+// lint-no-whitespace-within-word.js
+'use strict';
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+const ERROR_MESSAGE = 'Excess whitespace in layout detected.';
+
+// Entity format: nbsp -> &nbsp; -> '\\&nbsp\\;'
```
rwjblue: Might be nice to link to the reference for these entities

lib/rules/lint-no-whitespace-within-word.js
```diff
@@ -0,0 +1,66 @@
...
+// (anyBlackListItem)(anyWhiteListItem)(anyBlackListItem)(anyWhiteListItem)(anyBlackListItem)
+const blackListRegExp = new RegExp(blackListString+whiteListString+blackListString+whiteListString+blackListString, 'g');
+
+
+module.exports = class noWhitespaceWithinWord extends Rule {
+  visitor() {
+    return {
+      TextNode(node) {
+        let source = this.sourceForNode(node);
+        let matches = source.match(blackListRegExp);
```
rwjblue: I find this mega regexp hard to reason about. It might be easier to reason about this as a loop, where you keep track of the last seen item (whitespace vs non-whitespace) and then increment a counter if the next char is different. 

Something like

```js
let alternationCount = 0;
let source = this.sourceForNode(node);

for (let i = 0; i < source.length; i++) {
  let currentChar = source[i];
  let previousChar = i > 0 ? source[i - 1] : undefined;
  
  if (
    (isWhitespace(currentChar) && !isWhitespace(previousChar)) || 
    (!isWhitespace(currentChar) && isWhitespace(previousChar))
  ) {
    alternationCount++;
  }
}

if (alternationCount >= 3) {
 this._log({ ... });
}
```

MelSumner: This seems like it will be slower but I can try it.

rwjblue: > This seems like it will be slower but I can try it.

Hmm, based on what?

docs/rule/no-whitespace-within-word.md
```diff
@@ -0,0 +1,53 @@
+## no-whitespace-within-word
+
+Using in-line whitespace word formatting results in pages that are in violation of the Web Content Accessibility Guidelines (WCAG 2.0). This approach reflects a choice to ignore [WCAG's officially-documented techniques](https://www.w3.org/TR/WCAG20-TECHS/intro.html) that ensure web accessibility criteria are met, and, unsurprisingly, produces results that are explicitly mentioned in [WCAG's list of common sources of web accessibility failures](https://www.w3.org/TR/WCAG20-TECHS/failures.html).
```
rwjblue: > Using in-line whitespace word formatting results in pages that are in violation of the Web Content Accessibility Guidelines (WCAG 2.0).

Can we make the case better than "XYZ spec says so"? In theory, the spec is well reasoned, and we can inline some of that basic reasoning here while still referencing the spec. 

> This approach reflects a choice to ignore [WCAG's officially-documented techniques](https://www.w3.org/TR/WCAG20-TECHS/intro.html) that ensure web accessibility criteria are met, and, unsurprisingly, produces results that are explicitly mentioned in [WCAG's list of common sources of web accessibility failures](https://www.w3.org/TR/WCAG20-TECHS/failures.html)

Lets avoid:

* assumption of malice or willful ignorance
* condescending tone

docs/rule/no-whitespace-within-word.md
```diff
@@ -0,0 +1,53 @@
...
+
+Specifically, this common whitespace-within-word-induced web accessibility failure is -- from a WCAG perspective -- a failure to successfully achieve [WCAG Success Criterion 1.3.2: Meaningful Sequence](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-sequence.html). In practice, the predominant issue raised by inline whitespace styling is that the resultant text 'formatting' is entirely visual in nature; the ability to discern the correct manner in which to read the text, and therefore, to correctly comprehend its meaning, is restricted to sighted users.
```
rwjblue: > In practice, the predominant issue raised by inline whitespace styling is that the resultant text 'formatting' is entirely visual in nature; the ability to discern the correct manner in which to read the text, and therefore, to correctly comprehend its meaning, is restricted to sighted users

This seems like an **excellent** introduction, and should be _**first**_. Then we can back it up with the spec that agrees with us.

lib/rules/lint-no-whitespace-within-word.js
MelSumner: I think this approach doesn't work, I tried it a few different ways - I had to timebox it because the submitted approach does function as intended. Let's add this to our review session and dig in a bit more.

docs/rule/no-whitespace-within-word.md
```diff
@@ -0,0 +1,53 @@
...
+
+```hbs
+<div>Welcome</div>
+
+<div>Yes&nbsp;I am.</div>
+```
+
+### Migration
+
+The rule's RegExp-based detection mechanism can be used within a codebase to identify potential instances where in-line whitespace-within-word formatting will result in an F32-related loss of meaningful sequence. Flags / violations largely need to be diagnosed / addressed on a case-by-case basis, but the WCAG-approved vehicle for controlling letter-spacing within a word is CSS styling -- specifically, the implementation of [WCAG Technique C8: Using CSS letter-spacing to control spacing within a word](https://www.w3.org/WAI/WCAG21/Techniques/css/C8).
```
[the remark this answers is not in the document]
MelSumner: yes, will update

lib/rules/lint-no-whitespace-within-word.js
rwjblue: We resolved this together, updated the code below.

PR #849, opened 2019-10-08
Improve `require-valid-alt-text` rule to catch additional common failure scenarios

MelSumner:
If merged, this PR updates the `require-valid-alt-text` rule to catch common failures as listed in https://www.w3.org/WAI/WCAG21/Techniques/failures/F30.html 

- catches number-only alt text
- adds a few new words to the redundant words list
- catches instances where the src value is equal to the alt value

### DIFF
--- a/lib/rules/lint-require-valid-alt-text.js
+++ b/lib/rules/lint-require-valid-alt-text.js
@@ -8,8 +8,6 @@ function hasAccessibleChild(node) {
 }
 
 const REDUNDANT_WORDS = ['image', 'photo', 'picture', 'logo', 'spacer'];
-//TODO check for numbers in alt text
-
 
 const ERROR_MESSAGE =
   'Invalid alt attribute. Screen-readers already announce `img` tags as an image. Words such as `image`, `photo,` or `picture` (or any specified custom words) are considered placeholder text and as such are not valid.';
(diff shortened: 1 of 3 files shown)


lib/rules/lint-require-valid-alt-text.js
```diff
@@ -80,16 +87,25 @@ module.exports = class A11yLintAltText extends Rule {
...
                     line: node.loc && node.loc.start.line,
                     column: node.loc && node.loc.start.column,
                     source: this.sourceForNode(node),
                   });
+                } else {
+                  const existingWords = REDUNDANT_WORDS.filter(
+                    word => normalizedAltValue.indexOf(word) !== -1
+                  );
+                  if (existingWords.length) {
+                    this.log({
```
rwjblue: Lets swap this to `this.logNode` instead

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -40,13 +40,20 @@ module.exports = class A11yLintAltText extends Rule {
...
+          const srcValue = AstNodeInfo.elementAttributeValue(node, 'src');
 
           if (!hasAltAttribute) {
             this.logNode({
               message: 'img tags must have an alt attribute',
               node,
             });
+          } else if (altValue === srcValue) {
+            this.logNode({
+              message: 'the alt text must not be the same as the image source',
```
rwjblue: This message doesn't really fit with the other error messages we issue here (they are sentences, but this one isn't).

lib/rules/lint-require-valid-alt-text.js
```diff
@@ -80,16 +87,25 @@ module.exports = class A11yLintAltText extends Rule {
...
 
               if (normalizedAltValue !== null) {
-                const existingWords = REDUNDANT_WORDS.filter(
-                  word => normalizedAltValue.indexOf(word) !== -1
-                );
-                if (existingWords.length) {
+                if (normalizedAltValue.match(/^[0-9]+$/g)){
                   this.log({
-                    message: errorMessage,
+                    message: 'a number is not valid alt text',
```
rwjblue: This message doesn't really fit with the other error messages we issue here (they are sentences, but this one isn't).

rwjblue:
Also, looks like CI is failing due to a linting error:

```
/home/travis/build/ember-template-lint/ember-template-lint/lib/rules/lint-require-valid-alt-text.js
  90:59  error  Insert `·`  prettier/prettier
```

Should be able to fix via `yarn lint --fix` I think.

PR #850, opened 2019-10-08
Changed parser mode to 'codemod'

### LINKED ISSUE #120: Incorrect indentation on line starting with &nbsp;

initram:
This allows the linter to lint indentation correctly even if the line starts with `&nbsp;`

Fixes #120

Though some changes were needed to rules that work on TextNodes as they now behave slightly different. The changes I made to the block indentation rule also made it a bit more consistent in my opinion.

Additional test may be needed to cover what ever corners cases there may be, but all current tests should pass.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -145,7 +145,9 @@ class Linter {
     let templateAST;
 
     try {
-      templateAST = preprocess(source);
+      templateAST = preprocess(source, {
+        mode: 'codemod',
+      });
     } catch (error) {
       let message = buildErrorMessage(options.moduleId, error);
       messages.push(message);
--- a/lib/rules/lint-no-bare-strings.js
+++ b/lib/rules/lint-no-bare-strings.js
@@ -38,7 +38,7 @@ const TAG_ATTRIBUTES = {
   img: ['alt'],
 };
 
-// Charrecter entity refeence chart: https://dev.w3.org/html5/html-author/charref
+// Character entity reference chart: https://dev.w3.org/html5/html-author/charref
 const DEFAULT_CONFIG = {
   whitelist: [
     '&lpar;', // (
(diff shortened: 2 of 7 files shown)


lib/index.js
```diff
@@ -123,6 +123,7 @@ class Linter {
 
     try {
       precompile(stripBom(options.source), {
+        mode: 'codemod',
```
rwjblue: 👍I absolutely had this package in mind when adding this feature upstream.

rwjblue:
> We could consider adding more items the the default white list and test that it works for string that are currently being transformed into whitespace.

Right, I think the main thing we need to ensure is that the items in the current default whitelist that _could_ have been converted from an HTML entity have a corresponding whitelist entry for the entity as well.

rwjblue:
@initram - Would you mind rebasing? There have been changes to `lib/index.js` which are causing conflicts, but also the definition of `codemod` mode changed in `@glimmer/syntax@0.43.0` (see https://github.com/glimmerjs/glimmer-vm/pull/983 for details on that breakage WRT `codemod` mode). We may need more tweaks to accommodate those changes.

lib/rules/lint-no-bare-strings.js
```diff
@@ -38,8 +38,47 @@ const TAG_ATTRIBUTES = {
   img: ['alt'],
 };
 
+// Charrecter entity refeence chart: https://dev.w3.org/html5/html-author/charref
```
rwjblue:
```suggestion
// Character entity reference chart: https://dev.w3.org/html5/html-author/charref
```

rwjblue:
I went ahead and rebased for you, looks like no additional failures crop up so this is good to land!

rwjblue:
Thank you again @initram!

Issue #835, opened 2019-09-27
Rule: Disallow {{action}} (both element and helper forms)


rwjblue:
As part of the Octane efforts, we should add a new rule that disallows the usage of `{{action` in any location. This would supersede the usage of `no-element-action` in the `octane` preset.

Bad examples:

```hbs
<button {{action 'submit'}}>Submit</button>
```

```hbs
<FooBar @baz={{action 'submit'}} />
```

```hbs
{{yield (action 'foo')}}
```

```hbs
{{yield (action this.foo)}}
```

The rule should suggest the usage of `on` for the element modifier form, and `fn` for the helper form.

lifeart:
I can take it, how we will call this rule?
`deprecations/lint-deprecated-action-helper`?
`no-action-helper`?

bmish:
Note that we have two existing rules:
* [no-action-modifiers](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-action-modifiers.md)
* [no-element-event-actions](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-element-event-actions.md)

I would suggest calling the new rule simply `no-action` to indicate that it's more broad than the existing rules.

rwjblue:
Agree with @bmish

rwjblue:
Closed by #853

PR #853, opened 2019-10-09
Add `no-action` rule

lifeart:
#835 add no `{{action}}` rule

### DIFF
--- a/docs/rule/no-action-helper.md
+++ b/docs/rule/no-action-helper.md
@@ -1,0 +1,111 @@
+## no-action-helper
+
+What's wrong with `{{action}}`?
+
+"Action" is an overloaded term in Ember parlance. Actions are:
+
+1.) Methods on the `actions` hash
+```js
+actions: {
+  helloWorld() {}
+}
+```
+
+2.) But also modifiers that setup event handlers?
+```hbs
+<div {{action 'helloWorld'}}></div>
+```
+
+3.) Oh, and they are partial applied functions too:
+```hbs
+{{some-component onClick=(action 'addNumbers' 1 2 3)}}
+```
+
+4.) Also, they usually receive strings, but can also receive functions?
+```hbs
+<div {{action this.someMethod}}></div>
+```
+
+They're something we pass downward, but also something we send back up (as in Data Down, Actions Up). They are the way you interact with the DOM, except when you need to actually access the event itself, or when you use the `click` method on a classic component class. The point is, if you try to ask a group of experienced Ember devs to define what an "action" is, you may get a few contradicting/overlapping opinions.
+
+Actions have served many different purposes over the years, and this makes them difficult to learn, to teach, and to repurpose.
+
+```hbs
+<button {{action "increment" 5}}>Click</button>
+<button {{action this.increment 5}}>Click</button>
+<button onclick={{action "increment" 5}}>Click</button>
+<button onclick={{action this.increment 5}}>Click</button>
+<button {{action (action "increment" 5)}}>Click</button>
+<button {{action (action this.increment 5)}}>Click</button>
+```
+
+We wanted to separate out the different responsibilities here, and the `@action` decorator was the first part of the puzzle, and the other two pieces are the `{{fn}}` helper and `{{on}}` modifier.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+<button {{action 'submit'}}>Submit</button>
+```
+
+```hbs
+<FooBar @baz={{action 'submit'}} />
+```
+
+```hbs
+{{yield (action 'foo')}}
+```
+
+```hbs
+{{yield (action this.foo)}}
+```
+
+This rule **allows** the following:
+
+```hbs
+{{#let (fn this.foo bar) as |action|}}<Component @baz={{action}} />{{/let}}
+```
+
+```hbs
+<MyScope as |action|><Component @baz={{action}} /></MyScope>
+```
+
+```hbs
+<button {{on "submit" @action}}>Click Me</button>
+```
+
+```hbs
+<button {{on "submit" this.action}}>Click Me</button>
+```
+
+### Migration
+
+```hbs
+<select onchange={{action this.updateSelection this.options}}>
+  {{#each this.options as |opt|}}
+    <option>{{opt.value}}<option>
+  {{/each}}
+</select>
+```
+to
+
+```hbs
+<select {{on 'change' (fn this.updateSelection this.options)}}>
+  {{#each this.options as |opt|}}
+    <option>{{opt.value}}<option>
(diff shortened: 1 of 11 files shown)


lib/rules/lint-no-action-helper.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+      if (node.path.data === true || node.path.this === true) {
+        return;
+      }
+      if (params.includes(maybeAction)) {
+        return;
+      }
+
+      this.log({
+        message: params.length ? JSON.stringify(params) : ERROR_MESSAGE,
```
bmish: What is `JSON.stringify(params)` for? It doesn't look tested either.

lib/rules/lint-no-action-helper.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE =
+  'Do not use the `action` helper. Instead, use the `on` modifier and `fn` helper.';
+
+module.exports = class NoActionModifiers extends Rule {
```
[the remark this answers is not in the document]
lifeart: `no-action` may confuse developers, due to overlapping name with `@action` decorator.

Should it be `no-template-action` or `no-action-helper-modifier`?

docs/rule/no-action-helper.md
```diff
@@ -0,0 +1,111 @@
...
+"Action" is an overloaded term in Ember parlance. Actions are:
+
+1.) Methods on the `actions` hash
+```js
+actions: {
+  helloWorld() {}
+}
+```
+
+2.) But also modifiers that setup event handlers?
```
rwjblue:
```suggestion
2.) But also element modifiers that setup event handlers?
```

docs/rule/no-action-helper.md
```diff
@@ -0,0 +1,111 @@
...
+<button {{action this.increment 5}}>Click</button>
+<button onclick={{action "increment" 5}}>Click</button>
+<button onclick={{action this.increment 5}}>Click</button>
+<button {{action (action "increment" 5)}}>Click</button>
+<button {{action (action this.increment 5)}}>Click</button>
+```
+
+We wanted to separate out the different responsibilities here, and the `@action` decorator was the first part of the puzzle, and the other two pieces are the `{{fn}}` helper and `{{on}}` modifier.
+
+### Examples
```
rwjblue: I think we should have an example using `<button onclick={{action 'foo'}}></button>` also

docs/rule/no-action-helper.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+This rule **allows** the following:
+
+```hbs
+{{#let (fn this.foo bar) as |action|}}<Component @baz={{action}} />{{/let}}
+```
+
+```hbs
+<MyScope as |action|><Component @baz={{action}} /></MyScope>
```
rwjblue: In general, we should not suggest using `action` as a param name like this. Ember is still working on fixing some precedence issues here, and some versions will actually invoke the action helper in this example. It's best to avoid reusing a helper/modifier/component name as a block param name for now...

lib/rules/lint-no-action-helper.js
rwjblue:
```suggestion
module.exports = class NoAction extends Rule {
```

lib/rules/lint-no-action-helper.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE =
+  'Do not use the `action` helper. Instead, use the `on` modifier and `fn` helper.';
```
rwjblue: I'd like for the error message to show how they were using `action` including the curlies or paren. We should also avoid calling it a helper (since sometimes it's a helper and sometimes it's a modifier).

lib/rules/lint-no-action-helper.js
```diff
@@ -0,0 +1,66 @@
...
+        enter(node) {
+          addParams(node.program ? node.program.blockParams : []);
+        },
+        exit(node) {
+          addParams(node.program ? node.program.blockParams : []);
+        },
+      },
+      ElementNode: {
+        enter(node) {
+          addParams(node.blockParams);
```
rwjblue: We already have a built in mechanism for tracking block params (`this.isLocal(node)`), we shouldn't need to add another.

lib/rules/lint-no-action-helper.js
```diff
@@ -0,0 +1,66 @@
...
+
+    function detectActionHelper(node) {
+      let maybeAction = node.path.original;
+      if (maybeAction !== 'action') {
+        return;
+      }
+      if (node.path.data === true || node.path.this === true) {
+        return;
+      }
+      if (params.includes(maybeAction)) {
```
rwjblue: Should swap to using the built-in param tracking:

```suggestion
      if (this.isLocal(node)) {
```

test/unit/rules/lint-no-action-helper-test.js
```diff
@@ -0,0 +1,60 @@
...
+  config: true,
+
+  good: [
+    '{{#let (fn this.foo bar) as |action|}}<Component @baz={{action}} />{{/let}}',
+    '<MyScope as |action|><Component @baz={{action}} /></MyScope>',
+    '<button {{on "submit" @action}}>Click Me</button>',
+    '<button {{on "submit" this.action}}>Click Me</button>',
+  ],
+
+  bad: [
```
rwjblue: Can you add `bad` examples for:

* `<button onclick={{action 'foo'}}></button>`

docs/rule/no-action.md
```diff
@@ -0,0 +1,108 @@
...
+
+```hbs
+<select {{on 'change' (fn this.updateSelection this.options)}}>
+  {{#each this.options as |opt|}}
+    <option>{{opt.value}}<option>
+  {{/each}}
+</select>
+```
+
+### Related Rules
```
bmish: For each of these related rules, can you go into their documentation and add `no-action` as a related rule for them too?

lib/rules/lint-no-action.js
```diff
@@ -0,0 +1,50 @@
...
+
+    function detectAction(node, usageContext) {
+      if (isLocal(node.path)) {
+        return;
+      }
+      let maybeAction = node.path.original;
+      if (maybeAction !== 'action') {
+        return;
+      }
+      if (node.path.data === true || node.path.this === true) {
```
rwjblue: Doesn't the `node.path.original` guard just above handle both of these cases? For `{{@action}}` (the `node.path.data` check) and `{{this.action}}` (the `node.path.this` check) `node.path.original` should not be `"action"`...

Issue #857, opened 2019-10-15
[BUG] `no-curly-component-invocation` rule throwing unexpectedly


alexdiliberto:
After enabling this rule, in my template files I'm seeing yielded arguments all throw the corresponding lint error

Example, from a template only glimmer-component:

```
{{#each @options as |option|}}
  {{option}}
{{/each}}
```

```
  2:2  error  You are using the component {{option}} with curly component syntax. You should use <Option> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['option'] }`.  no-curly-component-invocation
```

`{{option}}` isn't a component or helper, it's simply a POJO or some raw data passed into the component.

Is it really necessary to whitelist all yielded arguments like this?

---

Versions:

```
ember-template-lint@^1.1.0:
  version "1.5.3"
ember-cli-template-lint@^1.0.0-beta.3:
  version "1.0.0-beta.3"
```

alexdiliberto:
some quick debugging, don't really have the time to dig in fully

```
node= { type: 'MustacheStatement',
  path:
   { type: 'PathExpression',
     original: 'option',
     this: false,
     parts: [ 'option' ],
     data: false,
     loc:
      SourceLocation { source: undefined, start: [Object], end: [Object] } },
  params: [],
  hash:
   { type: 'Hash',
     pairs: [],
     loc: { source: '(synthetic)', start: [Object], end: [Object] } },
  escaped: true,
  loc:
   { source: null,
     start: { line: 2, column: 2 },
     end: { line: 2, column: 12 } },
  strip: { open: false, close: false } }
entered block: "if (this.isDisallowed(name) || !this.config.requireDash) {"
this.config.requireDash= false
this.isDisallowed(name) false
message= You are using the component {{option}} with curly component syntax. You should use <Option> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['option'] }`.
```

rwjblue:
This is a tricky one. For a block param, we can't _know_ for sure if it is a component invocation, some text, or heck even a DOM element. The rule does not make any different assumptions about yielded block params at the moment.

We could tweak things a bit so that the rule does not apply to block params when used without any arguments (e.g. `{{foo}}` vs either `{{foo whatever}}` or `{{foo some=thing}}`). 🤔 

[...]

PR #861, opened 2019-10-16
Update `no-curly-component-invocation` to allow non-component yielded values to be used with mustaches

patocallaghan:
Potential fix for #857 

Opt out block param variables from curly invocation checks.

@rwjblue I noticed in that issue you mention

> We could tweak things a bit so that the rule does not apply to block params when used without any arguments (e.g. {{foo}} vs either {{foo whatever}} or {{foo some=thing}}). 

Is just checking if it's a block param enough or should we do additional checks?

### DIFF
--- a/lib/rules/lint-no-curly-component-invocation.js
+++ b/lib/rules/lint-no-curly-component-invocation.js
@@ -80,7 +80,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
 
   getCurlyInvocationSyntaxError(node) {
     let name = node.path.original;
-    if (node.path.type !== 'PathExpression') {
+    if (node.path.type !== 'PathExpression' || this.isLocal(node)) {
       return '';
     }
     if (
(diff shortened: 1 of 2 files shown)


lib/rules/lint-no-curly-component-invocation.js
```diff
@@ -80,7 +80,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
 
   getCurlyInvocationSyntaxError(node) {
     let name = node.path.original;
-    if (node.path.type !== 'PathExpression') {
+    if (node.path.type !== 'PathExpression' || this.isLocal(node)) {
```
rwjblue:
```suggestion
    let hasArguments = node.params.length > 0 || node.hash.pairs.length > 0;
    let isAmbiguousLocalInvocation = this.isLocal(node) && !hasArguments;
    if (node.path.type !== 'PathExpression' || isAmbiguousLocalInvocation) {
```

patocallaghan: I've updated the logic based on your feedback but needed to tweak it slightly. I had to move calculating `isAmbiguousLocalInvocation` check to happen lazily because the `this.isLocal` throws an exception when the type in the mustache is not a `PathExpression`, e.g. `NumberLiteral.

So can `{{whatever some=arg}}` or `{{whatever arg}}` ever be a helper or just a component?

PR #862, opened 2019-10-18
✨Add option --print-pending from ember-cli-template-lint

TristanToye:
This task was previously mentioned when the docs were updated earlier this year to port over the ability to print a pending list by @rwjblue https://github.com/ember-template-lint/ember-template-lint/pull/671#pullrequestreview-213931116

> Seems good to me, but I'd really like to migrate that to our built-in bin script (perhaps with a flag like `--print-pending`). Any chance you'd want to work on it?

This adds the option to easily populate your `.template-lintrc.js` file without needing [ember-cli-template-lint](https://github.com/ember-template-lint/ember-cli-template-lint). Should make adopting and getting started a lot easier.

To easily run against your whole project you can now do: 
```
./node_modules/.bin/ember-template-lint app/templates/** --print-pending
```

This is mostly copied over from the existing logic here https://github.com/ember-template-lint/ember-cli-template-lint/blob/master/lib/commands/print-failing.js

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -135,7 +135,7 @@ function run() {
   let options = parseArgv(process.argv.slice(2));
 
   let {
-    named: { configPath, filename: filePathFromArgs = '', printPending },
+    named: { configPath, filename: filePathFromArgs = '', printPending, json },
     positional,
   } = options;
 
@@ -165,13 +165,9 @@ function run() {
     let fileErrors = lintFile(linter, filePath, moduleId);
 
     if (printPending) {
-      let failingRules = fileErrors.reduce((memo, error) => {
-        if (memo.indexOf(error.rule) === -1) {
-          memo.push(error.rule);
-        }
-
-        return memo;
-      }, []);
+      let failingRules = Array.from(
+        fileErrors.reduce((memo, error) => memo.add(error.rule), new Set())
+      );
 
       if (failingRules.length > 0) {
         filesWithErrors.push({ moduleId, only: failingRules });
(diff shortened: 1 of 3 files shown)


bin/ember-template-lint.js
```diff
@@ -157,7 +161,22 @@ function run() {
...
+    let fileErrors = lintFile(linter, filePath, moduleId);
+
+    if (printPending) {
+      let failingRules = fileErrors.reduce((memo, error) => {
+        if (memo.indexOf(error.rule) === -1) {
+          memo.push(error.rule);
+        }
+
+        return memo;
+      }, []);
```
rwjblue: If you make this `new Set()` then you can avoid checking and manually deduping (and since `set.add` returns the `set` itself we can use a simpler one liner):

```js
      let failingRules = fileErrors.reduce((memo, error) => memo.add(error.rule), new Set());
```

bin/ember-template-lint.js
```diff
@@ -157,7 +161,22 @@ function run() {
...
+
+      if (failingRules.length > 0) {
+        filesWithErrors.push({ moduleId, only: failingRules });
```
rwjblue: If we make `failingRules` a `Set`, this would end up being something like:

```suggestion
        filesWithErrors.push({ moduleId, only: Array.from(failingRules) });
```

bin/ember-template-lint.js
```diff
@@ -172,6 +191,13 @@ function run() {
...
   }
 
+  if (printPending && filesWithErrors.length > 0) {
+    console.log(
+      'Add the following to your `.template-lintrc.js` file to mark these files as pending.\n\n'
+    );
+    console.log(`pending: ${JSON.stringify(filesWithErrors, null, 2)}`);
+  }
+
   if (Object.keys(errors).length) {
```
rwjblue: I don't think we should emit the actual errors in addition to printing the pending list when `--print-pending` is used. Can you tweak things to avoid that (I'm thinking of a `return;` added to the `if` just above)...

bin/ember-template-lint.js
```diff
@@ -172,6 +191,13 @@ function run() {
     }
   }
 
+  if (printPending && filesWithErrors.length > 0) {
+    console.log(
```
rwjblue: We should honor `--json` flag as well here, so that its possible to get things in a parsable JSON format:

```js
let pendingList = JSON.stringify(filesWithErrors, null, 2);

if (options.named.json) {
  console.log(pendingList);
} else {
  console.log(
    'Add the following to your `.template-lintrc.js` file to mark these files as pending.\n\n'
  );

  console.log(`pending: ${pendingList}`);
}
```

TristanToye:
Thanks for the feedback @rwjblue - updated and added some basic tests. 

Didn't have a better idea for comparing this string though https://github.com/ember-template-lint/ember-template-lint/pull/862/files#diff-4e539a2c8765ac2d1af384fc39e287c8R460

rwjblue:
Thank you!

Issue #731, opened 2019-06-04
`no-invalid-interactive` does not trigger when using the `on` modifier


Windvis:
The `no-invalid-interactive` rule doesn't throw an error when using the `on` modifier instead of the action helper / modifier.

```
    <div {{action this.actionName}}>...</div>{{!-- This triggers a lint error --}}
    <div {{on "click" this.actionName}}>...</div>{{!-- This does not --}}
```

rwjblue:
Yep, we need to fix this! Thanks for reporting!

lifeart:
on it

lifeart:
PR created

PR #864, opened 2019-10-19
bugfix: #731 no-invalid-interactive for on modifier

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/731

### DIFF
--- a/lib/rules/lint-no-invalid-interactive.js
+++ b/lib/rules/lint-no-invalid-interactive.js
@@ -70,14 +70,14 @@ module.exports = class InvalidInteractive extends Rule {
 
         let modifierName = node.path.original;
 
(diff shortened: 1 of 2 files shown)


lib/rules/lint-no-invalid-interactive.js
```diff
@@ -155,6 +166,14 @@ function isActionName(node, name) {
   return false;
 }
 
+function isActionName(node, name) {
```
rwjblue: Now that I'm reading this diff, I realize that the name of these (preexisting) functions aren't quite correct. We aren't looking for an "action name" we are looking for an event name.

Would you mind cleaning that up a bit? I think this function should be called `isEventHandlerFor`...

Issue #841, opened 2019-10-02
New Rule: prefer-named-arguments


buschtoens:
Disallow the following:

- `{{this.args.foo}}`
- `{{args.foo}}`: in case `no-implicit-this` is not enabled

Prompt the user to prefer `{{@foo}}`.

But still allow:

- `{{this.args}}`: in case you want to pass through all args to a helper, e.g.
  `{{validate-args this.args}}`

rwjblue:
👍 I'm definitely in favor, with some tweaks:

* lets call it `prefer-named-arguments`
* `{{this.args}}` should be disallowed (it would just render `[object Object]` AFAICT)
* passing `this.args` to a helper is allowed

lifeart:
allowed cases?

```hbs
{{foo (name this.args)}}
{{foo name=this.args}}
{{foo name=(extract this.args)}}
<Foo @params={{this.args}} />
<Foo {{mod this.args}} />
<Foo {{mod items=this.args}} />
<Foo {{mod items=(extract this.args)}} />

```

rwjblue:
Yes, using `this.args` in a `PathExpression` should be allowed, but only if that is the full path (e.g. `this.args.foo` should be disallowed)).

Turbo87:
looks like @lifeart resolved this with #865

PR #865, opened 2019-10-19
Add new `no-args-paths` rule.

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/841


lib/rules/index.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   'no-abstract-roles': require('./lint-no-abstract-roles'),
   'no-action-modifiers': require('./lint-no-action-modifiers'),
   'no-attrs-in-components': require('./lint-no-attrs-in-components'),
+  'no-args': require('./lint-no-args'),
```
rwjblue: Should this be `no-args-paths` instead? We actually _do_ allow `this.args` so saying `no-args` seems a tad bit off.

lib/rules/lint-no-args.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const Rule = require('./base');
+
+module.exports = class LintNoAttrs extends Rule {
```
rwjblue: This class name is wrong

lib/rules/lint-no-args.js
```diff
@@ -0,0 +1,36 @@
...
+  visitor() {
+    function checkPathForAttrs(node) {
+      if (node.parts && node.parts[0] === 'attrs') {
```
rwjblue: Seems like this should be `args` not `attrs`

rwjblue:
@lifeart - Can you submit a follow up PR to add this to the octane config?

lifeart:
Done https://github.com/ember-template-lint/ember-template-lint/pull/893

PR #868, opened 2019-10-23
Add `no-meta-redirect-with-time-limit` rule

MelSumner:
When merged, this PR will provide a check to prevent the use of a timed delay on `<meta http-equiv="refresh"/>' where the `content` delay value is greater than 0.

### DIFF
--- a/lib/rules/lint-no-meta-redirect-with-time-limit.js
+++ b/lib/rules/lint-no-meta-redirect-with-time-limit.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+module.exports = class noMetaRedirectWithTimeLimit extends Rule {
+  logNode({ node, message }) {
+    return this.log({
+      message,
+      line: node.loc && node.loc.start.line,
+      column: node.loc && node.loc.start.column,
+      source: this.sourceForNode(node),
+    });
+  }
+  visitor() {
+    return {
+      ElementNode(node) {
+        if (AstNodeInfo.hasAttribute(node, 'hidden')) {
+          return;
+        }
+
+        if (AstNodeInfo.hasAttributeValue(node, 'aria-hidden', 'true')) {
+          return;
+        }
+
+        const isMeta = AstNodeInfo.isElementNode(node, 'meta');
+        const hasMetaRedirect = AstNodeInfo.hasAttribute(node, 'http-equiv');
+
+        if (isMeta && hasMetaRedirect) {
+          const contentAttr = AstNodeInfo.hasAttribute(node, 'content');
+          const contentAttrValue = AstNodeInfo.elementAttributeValue(node, 'content');
+          if (contentAttr) {
+            if (contentAttrValue.charAt(0) > 0) {
+              this.log({
+                message: 'a meta redirect should not have a delay value greater than zero',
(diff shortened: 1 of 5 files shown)


lib/rules/lint-no-meta-redirect-with-time-limit.js
```diff
@@ -0,0 +1,46 @@
...
+      ElementNode(node) {
+        if (AstNodeInfo.hasAttribute(node, 'hidden')) {
+          return;
+        }
+
+        if (AstNodeInfo.hasAttributeValue(node, 'aria-hidden', 'true')) {
+          return;
+        }
+
+        const isMeta = AstNodeInfo.isElementNode(node, 'meta');
```
lifeart: `isElementNode` accept only one argument

```suggestion
        const isMeta = node.tag === 'meta';
```

Is it possible to move this assert on the top because of perf reasons? (we don't need to check element visibility if this is not `meta` node.

lib/rules/lint-no-meta-redirect-with-time-limit.js
```diff
@@ -0,0 +1,46 @@
...
+        const hasMetaRedirect = AstNodeInfo.hasAttribute(node, 'http-equiv');
+
+        if (isMeta && hasMetaRedirect) {
+          const contentAttr = AstNodeInfo.hasAttribute(node, 'content');
+          const contentAttrValue = AstNodeInfo.elementAttributeValue(node, 'content');
+          if (contentAttr) {
+            if (contentAttrValue.charAt(0) > 0) {
```
lifeart: I think it could be simplyfied to 
```suggestion
            if (!AstNodeInfo.hasAttributeValue(node, 'content', '0')) {
```

lib/rules/lint-no-meta-redirect-with-time-limit.js
```diff
@@ -0,0 +1,47 @@
...
+      message,
+      line: node.loc && node.loc.start.line,
+      column: node.loc && node.loc.start.column,
+      source: this.sourceForNode(node),
+    });
+  }
+  visitor() {
+    return {
+      ElementNode(node) {
+        if (AstNodeInfo.hasAttribute(node, 'hidden')) {
```
rwjblue: Does this attribute actually have effect on `meta`? Aren't they always hidden (AFAICT they have no display)?

lib/rules/lint-no-meta-redirect-with-time-limit.js
```diff
@@ -0,0 +1,47 @@
...
+          return;
+        }
+
+        if (AstNodeInfo.hasAttributeValue(node, 'aria-hidden', 'true')) {
```
rwjblue: Same here, what does this have to do with `meta`? I don't see `hidden` or `aria-hidden` mentioned in the references...

lib/rules/lint-no-meta-redirect-with-time-limit.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+module.exports = class noMetaRedirectWithTimeLimit extends Rule {
```
rwjblue:
```suggestion
module.exports = class NoMetaRedirectWithTimeLimit extends Rule {
```

Issue #851, opened 2019-10-08
Running limited set of rules does not work properly.


rwjblue:
If you would like to run a limited set of rules against a file, you _should_ be able to do:

```
echo "module.exports = { rules: { 'no-shadowed-elements': true } };" > temp-templatelint-rc.js
./node_modules/.bin/ember-template-lint --config-path ./temp-templatelint-rc.js some/file/here.hbs
```

However, due to the fact that the linter sets up _all_ known plugins here we are still running rules that are not supposed to be enabled:

https://github.com/ember-template-lint/ember-template-lint/blob/e10fed889cb6a17fd4d3cffbaa17fbbbbd46e813/lib/index.js#L51-L54

That section above, should be tweaked to something like:

```js
    let rules = this.config.loadedRules;
    let astPlugins = [TransformDotComponentInvocation];

    let configuredRules = this.config.rules;
    for (let pluginName in configuredRules) {
```

This would ensure that only plugins with configured rules would run.

lifeart:
on it

rwjblue:
Fixed by #870

PR #870, opened 2019-10-24
Fix issue preventing running a subset of rules

lifeart:
#851

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -55,11 +55,6 @@ class Linter {
       if (configuredRules[pluginName] === false) {
         continue;
       }
-      if (!(pluginName in rules)) {
-        throw new Error(
-          `ember-template-lint: Config rules error - unable to find linter for rule, named: "${pluginName}".`
-        );
-      }
       let Plugin = rules[pluginName];
       let plugin = new Plugin({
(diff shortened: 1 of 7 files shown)


lifeart:
@rwjblue I'm not confident about functionality of https://github.com/ember-template-lint/ember-template-lint/blob/eda935c9909a3d4b40173e6822ae1e4a034ee018/lib/index.js#L65

should It be replaced to `ruleNames: Object.keys(configuredRules)`?

If I replace it, I see 7 errors in tests

lifeart:
I'm not confident how to test `loaded ast plugins cont`, should it be an `console.log` in case of special flag passed?

rwjblue:
> I'm not confident how to test loaded ast plugins count, should it be an console.log in case of special flag passed?

Sorry, I'm not following, can you help me understand what you mean?

lifeart:
@rwjblue I can't find way to compare loaded ast (rules) plugins count (loaded into linter) and specified in config.

I wish to have assert kinda:

```js
buildASTPlugins().length === Object.keys(this.config.rules).length
```

rwjblue:
Thank you @lifeart!

PR #871, opened 2019-10-24
Add new rule `element-attributes-order`

### LINKED ISSUE #684: [Rule Request] Element Attribute Order
Not exactly sure how to call this rule.
Pretty much what I want is a rule that would throw on incorrect order inside of Element/Component invocation.

For example:
Take this component:
```handlebars
<MyComp @name="test" data-test="123" @doSomething={{action "gogogo"}} class="flex" />
```
There is no order to the attributes/modifiers etc

I think there should be an order such that `class` is first, then `data-test` then `props` then `actions`
Any thoughts?

lifeart:
element-attributes-order rule

@Alonski @gossi 

I suppose I need some help with documentation for this rule.

Fixes #684


gossi:
This is amazing! Checking the doc markdown seems solid to me. What are areas you would expect help?

lifeart:
@gossi thanks! I was thinging about some motivation text, references and related rules.

lifeart:
@rwjblue, could you please take a look?

rwjblue:
Ya, will try to get it reviewed this weekend

gossi:
So, the motiviation for this is:

Always expect the same type of attributes grouped and at the same location, for consistency amongst multiple developers and DX improvement for faster scanning of particular attributes when reading code.

I don't have references, related rules might be: [Attribute indentation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/attribute-indentation.md)

Alonski:
I forgot about this 😁

docs/rule/element-attributes-order.md
```diff
@@ -0,0 +1,38 @@
+## element-attributes-order
+
+Element attributes must be ordered as:
+
+* arguments
+* attributes
+* element modifiers
+* ...attributes
+* ?attributes
```
Turbo87: what is this?

docs/rule/element-attributes-order.md
```diff
@@ -0,0 +1,38 @@
+## element-attributes-order
+
+Element attributes must be ordered as:
+
+* arguments
+* attributes
+* element modifiers
+* ...attributes
```
Turbo87: it can be valuable to put some attributes **after** `...attributes`, but it seems that this rule doesn't allow that? 🤔

docs/rule/element-attributes-order.md
```diff
@@ -0,0 +1,38 @@
+## element-attributes-order
+
+Element attributes must be ordered as:
+
+* arguments
+* attributes
+* element modifiers
```
Turbo87: why are element modifiers before `...attributes`?

docs/rule/element-attributes-order.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+```hbs
+<MyComponent ...attributes {{did-render this.someAction}} />
+```
+
+This rule **allows** the following:
+
+```hbs
+<MyComponent @name="1" bar="baz" {{did-render this.someAction}} ...attributes aria-role="button" />
```
Turbo87: this does not seem to match the "must be ordered as" order above 🤔

docs/rule/element-attributes-order.md
gossi: `attributes` = html attributes, such as `class="foo"`
`...attributes` = ember/glimmer `...attributes` splat
`?attributes` = same as html attributes, yet optional. It must be allowed to have you own html attributes after the `...attributes` splat to allow overrides.

Alonski:
@Turbo87 Regarding a few of your questions we discussed this with @rwjblue in this issue:
https://github.com/ember-template-lint/ember-template-lint/issues/684

docs/rule/element-attributes-order.md
lifeart: this behavour supported https://github.com/ember-template-lint/ember-template-lint/pull/871/files#diff-780c04b6663870e4456bae29ebc13842R16

docs/rule/element-attributes-order.md
lifeart: order - https://github.com/ember-template-lint/ember-template-lint/issues/684#issuecomment-480830414

docs/rule/element-attributes-order.md
lifeart: https://github.com/ember-template-lint/ember-template-lint/issues/684#issuecomment-480830414

Alonski:
@rwjblue @Turbo87 Can we rereview this?

docs/rules.md
```diff
@@ -3,6 +3,7 @@
 * [attribute-indentation](rule/attribute-indentation.md)
 * [block-indentation](rule/block-indentation.md)
 * [eol-last](rule/eol-last.md)
+* [element-attributes-order](rule/element-attributes-order.md)
```
bmish, line 6: This file has been deleted so you will want to rebase on the latest master and run `yarn update` to update the README instead.

lib/rules/element-attributes-order.js
```diff
@@ -0,0 +1,128 @@
...
+        attributes_last_index,
+        attributes_first_index,
+        modifiers_last_index,
+        modifiers_first_index,
+        splattributes_last_index,
+        splattributes_first_index,
+      } = nodeTokens;
+      if (has_arguments) {
+        if (
+          (arguments_last_index > attributes_first_index && has_attributes) ||
```
bmish, line 74: Could you define the default ordering as a constant array? It would make it much easier to follow the code and also enable us to make the ordering configurable in the future.

```js
const ATTRIBUTE_ORDER = ['arguments', 'attributes', 'modifiers', ...];
```

lib/rules/element-attributes-order.js
```diff
@@ -0,0 +1,128 @@
...
+
+    function extractElementTokens(node, nodeText) {
+      const keys = [];
+      node.attributes.forEach(attr => {
+        const source = sourceForNode(attr);
+        const type = attr.name.startsWith('@')
+          ? 'arguments'
+          : attr.name.startsWith('...')
+          ? 'splattributes'
+          : 'attributes';
```
bmish, line 18: A nice-to-have improvement would be to extract the attribute types into an enum:

```js
const AttributeType = {
    ARGUMENTS: 'arguments',
    ATTRIBUTES: 'attributes',
    MODIFIERS: 'modifiers',
    ...
}

// example usage
if (type === AttributeType.ARGUMENTS) { ... }
```

bertdeblock:
Hi! 👋 I'm excited to see this merged. Would it be possible to have another look at this? Can I help out in any way?

Alonski:
@lifeart Alex do you might fixing conflicts and trying to get this ready to go?
Would be cool to have this rule available :)

lib/rules/element-attributes-order.js
lifeart: good catch! added

lib/rules/element-attributes-order.js
lifeart: should we follow this implementation? https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/attributes-order.js

bmish:
Shall we still pursue this rule? How close is it to being ready?

bertdeblock:
I still think this is a great idea and would like to see this rule land.

lifeart:
let's close it for now, I think final implementation could be better

bmish:
Agreed, it would be cool to get this in even if it's not perfect.

PR #872, opened 2019-10-25
lib/index: Split buildASTPlugins() to prepare ember-template-recast usage

dcyriller:
`ember-template-recast` won't take `ASTPlugin`s. Instead it will take `ASTPluginBuilder`s.

```
export interface ASTPluginBuilder {
  (env: ASTPluginEnvironment): NodeVisitor;
}
```

vs.

```
export interface ASTPlugin {
  name: string;
  visitor: NodeVisitor;
}
```

To prepare the switch, this PR proposes a refactoring: 
- split `this.buildASTPlugins` method into `this.buildRules`, `buildPlugin` and `buildASTPlugins`
- remove `preprocess`'s options.metadata usage

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -55,7 +55,7 @@ class Linter {
     return 2;
   }
 
-  buildASTPlugins(config) {
+  buildASTPlugins(config, pluginConfig) {
     function addToResults(result) {
       config.results.push(result);
     }
@@ -74,7 +74,7 @@ class Linter {
         ruleNames: Object.keys(rules),
       });
 
-      astPlugins.push(buildPlugin(rule, config));
+      astPlugins.push(buildPlugin(rule, pluginConfig));
     }
 
     return astPlugins;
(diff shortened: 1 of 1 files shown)


dcyriller:
For clarity purpose, I've removed some commits. :)

This is an extraction of https://github.com/ember-template-lint/ember-template-lint/pull/873 meant to be reviewed / merged before. The goal is to simplify the whole review.

rwjblue:
@dcyriller - I'm going to test an alternative approach (should still serve the same purposes for #873), will update here once I have more info.

dcyriller:
> I'm going to test an alternative approach 

Let me know when it's ready :)

I've pushed a few other commits 🙈. It is a refactoring I was going to propose after #873. It might interest you @rwjblue.

rwjblue:
@dcyriller - I rebased this and added my changes in https://github.com/ember-template-lint/ember-template-lint/pull/892. I'll probably land that (we can always iterate in follow up PRs).

dcyriller:
Awesome @rwjblue. I didn’t think about using traverse (from @glimmer/syntax) before migrating and using transform (from template-recast). The migration is almost done now 😄. I will update my PR early next week.

Issue #760, opened 2019-07-03
[Feature Request] Add --fix option


Alonski:
Adding the feature to autofix template lint rule errors would be nice.

We should add the underlying infrastructure so that rules can be autofixed similar to how ESLint does autofixing with --fix.

Related to:
https://github.com/ember-template-lint/ember-template-lint/issues/305

rwjblue:
Ya, I'm definitely in favor. I'd like to at least _try_ to leverage the work done in ember-template-lint/ember-template-recast to support "safer" fixers. 

The initial rough scope of this would be to:

* Add a hook to the plugin class that folks can implement to handle fixing
* Add a new `--fix` flag to the executable

rwjblue:
Would definitely love for help experimenting / pushing this forward.

rwjblue:
With the (relatively) recent revamp of ember-template-recast, I'd like to consider using it as the fixing mechanism for rules.

Basically, a rule would indicate that a given error is fixable (something akin to what ESLint rules do), and when the `--fix` option is passed ember-template-lint would:

1) gather the errors (without fixing)
2) run the fixer for each of the rules that had errored (in series)
3) gather remaining errors and print to the console

dcyriller:
:wave: I've started to work on this.

rwjblue:
FWIW, I chatted a bit with @dcyriller on this in discord this week. I think we have a rough path forward. I'll try to summarize here:

* Migrate this library to use ember-template-recast as its underlying parser (from `@glimmer/syntax`)
* Add a new optional `fixer` property to the object that is passed to `this.log`. 
* Add a new `--fix` flag to the bin script
* When `ember-template-lint --fix **/*.hbs` is ran, do _roughly_ the following algorithm:
  * assume `--json` in the bin script
  * gather the JSON results
  * get a list of the failures that have a `fixer` property
  * for each file that has an available `fixer`, for each rule
    * instantiate the Rule again
    * use `ember-template-recast`'s `traverse` passing it the result of `ruleInstance.getVisitor()`
    * immediately invoke `fixer()`
    * loop :D
  * run the normal linting again, printing the results (or honoring the original --json flag state)
* Create a meta issue to evaluate each rule for "fixability"

I _believe_ that @dcyriller is still planning on working on this, but figured it would be good to get the general shape of the proposed solution "out there" in case there are fundamental issues that others spot...

rwjblue:
[...]

PR #873, opened 2019-10-25
Leverage ember-template-recast

dcyriller:
This PR proposes to use ember-template-recast to switch to a non-destructive parser. The goal is to eventually implement the `--fix` option. It would be the first step of #760.

#### four small comments:
- [merged ✅] I've extracted 3 commits into a separate PR that may be merged first: https://github.com/ember-template-lint/ember-template-lint/pull/872
- [resolved ✅] `template-recast`'s `transform` method seems to take a function that returns a visitor (not a plugin) as a second argument. Whereas the README states the opposite (it should return a plugin not a visitor). Should I make an adjustement in `template-recast`? Or is it fine to go on with a visitor as it is in this PR? (see https://github.com/ember-template-lint/ember-template-recast/pull/158)
- I need to adjust three rules to make them compatible with `template-recast` (one test failing each): `block-indentation` (✅), `no-inline-styles` (✅ with a 🙊-patch) and `linebreak-style` (✅)
- this PR relies on https://github.com/ember-template-lint/ember-template-recast/pull/154 that will need to be resolved first ✅


dcyriller:
Hopefully it should be good for review now :)

rwjblue:
A few notes here:

* Ready for a rebase after #892
* I think we will need #850 to land first (`ember-template-recast` uses `codemod` mode, and we have rules that need adjustment before we are ready). Some of the commits here seem to replicate some of those changes, but I'd rather land #850 (@initram did a bunch of work on it over there).

dcyriller:
I didn’t see that #850 had to solve similar issues in the tests 😅. I’ll wait for it to be merged and rebase after!

rwjblue:
Ok, just landed #850

dcyriller:
@rwjblue it should be ready for review now 🙂

test/unit/rules/lint-no-inline-styles-test.js
```diff
@@ -36,7 +36,20 @@ generateRuleTests({
     'style={{foo}}',
     'style="{{foo}} {{bar}}"',
     'style',
-    'style=""',
     'style="color:blue;margin-left:30px;"',
-  ].map(mapBadStyleExampleToTestRule),
+  ]
+    .map(mapBadStyleExampleToTestRule)
```
dcyriller: You might want to see https://github.com/ember-template-lint/ember-template-lint/pull/873/commits/c4e29f6360d60e5f6fd9d63fcaad8d01dc1acb88 commit message

lib/index.js
```diff
@@ -163,7 +155,7 @@ class Linter {
 
     for (let rule of rules) {
       try {
-        traverse(templateAST, buildPlugin(rule));
+        transform(templateAST, () => rule.getVisitor());
```
rwjblue: Ya, wasn’t sure if you’d still want / need `buildPlugin` (now that `rule.templateEnvironmentData` isn’t used). Glad to see this simplified. :)

rwjblue:
Even though CI passes for Node 6 (not 100% sure why, ember-template-recast must not use Node 8 specific syntax?), I still think we should block landing this on bumping majors to drop Node 6.

dcyriller:
> Even though CI passes for Node 6 (not 100% sure why, ember-template-recast must not use Node 8 specific syntax?), I still think we should block landing this on bumping majors to drop Node 6.

Indeed, it should have failed on the step `yarn install` since `template-recast`‘s engines field states: `8 || 10 || >=12` (see https://github.com/ember-template-lint/ember-template-recast/blob/v3.2.8/package.json#L56).

The reason why it doesn’t fail is because this yarn flag in the `travis.yml` file: `—ignore-engines` (see https://github.com/ember-template-lint/ember-template-lint/blob/v1.7.0/.travis.yml#L19). I guess it is there to be able to cache the install step between various CI jobs (testing against different node versions).

So, this PR is definitely blocked by a Node requirement upgrade. :)

rwjblue:
FWIW, I added `—ignore-engines` so that I can use `release-it` (it requires Node 8).

dcyriller:
I now understand your previous comment 🙈

rwjblue:
Rebased and targeted `next` branch (follow along in #897).

PR #874, opened 2019-10-25
test-harness: Make it available for (external) plugins

dcyriller:
With this PR, external plugins may be able to use the test-harness.

An example of usage can be found here: https://github.com/dcyriller/ember-template-lint-plugin-prettier/pull/41

side-note: I would not be opposed to move `ember-template-lint-plugin-prettier` to ember-template-lint organization.


rwjblue:
> side-note: I would not be opposed to move ember-template-lint-plugin-prettier to ember-template-lint organization.

Sounds good, lets do it 😄

Issue #854, opened 2019-10-10
Refactor fixture based tests to use `broccoli-test-helper`'s `createTempDir`


rwjblue:
This will allow the tests to be reasoned about much easier. 

For a concrete example, I'd like for this test:

https://github.com/ember-template-lint/ember-template-lint/blob/996456138ed78926b40cf7bba109f281ebbf1e45/test/unit/get-config-test.js#L25-L34

To be rewritten like:

```js
  it('if config is provided, it is returned', function() {
    let expected = {
      rules: {
        foo: 'bar',
        baz: 'derp',
      },
    };

    project.write({
      '.template-lintrc.js': `module.exports = ${JSON.stringify(expected)};`,
    });

    process.chdir(project.path());
   
    let actual = getConfig({});
    expect(actual.rules).to.deep.equal(expected.rules);
  });
```

locks:
This has been implemented by linked https://github.com/ember-template-lint/ember-template-lint/pull/876 PR.

PR #876, opened 2019-10-27
Begin migrating tests to inline fixtures (via `broccoli-test-helper`)

lifeart:
@rwjblue 

* should we remove nodejs 6 support?
* how deal with long fixtures?

related issue: https://github.com/ember-template-lint/ember-template-lint/issues/854


bmish:
We should drop Node 6 support in the next major V2 release (not sure when @rwjblue plans to do that). We did that for the recent major release of eslint-plugin-ember too (https://github.com/ember-cli/eslint-plugin-ember/pull/485).

package.json
```diff
@@ -23,6 +23,7 @@
     "test": "mocha"
   },
   "dependencies": {
+    "broccoli-test-helper": "^2.0.0",
```
rwjblue: Should be in `devDependencies`

test/acceptance-test.js
```diff
@@ -135,9 +178,27 @@ describe('public api', function() {
     });
   });
 
-  describe('Linter.prototype.verify', function() {
-    let basePath = path.join(fixturePath, 'with-errors');
+  describe('Linter.prototype.verify', async function() {
+    project = await createTempDir();
```
rwjblue: Can you move this setup into a `before` (instead of in the root of the `describe`), and make the `describe` block back into a non-async function?

rwjblue:
Thanks @lifeart, looks great!

PR #882, opened 2019-10-28
Fix issues when converting dasherized curly component invocation into angle bracket invocation

suchitadoshi1987:
This fixes the capitalization issues caused due to nonAlphanumeric characters.
- `/` would be converted to `::`
- any alpha-numeric character encountered after a non-alpha-numeric character would be capitalized.
- Removes all occurrences of `-`

### DIFF
--- a/test/unit/helpers/curly-invocation-component-test.js
+++ b/test/unit/helpers/curly-invocation-component-test.js
@@ -10,6 +10,7 @@ describe('#transformTagName', function() {
   it(`it works as expected`, function() {
     expect(transformTagName('foo')).to.equal('Foo');
     expect(transformTagName('foo-bar')).to.equal('FooBar');
+    expect(transformTagName('Foo-Bar')).to.equal('FooBar');
     expect(transformTagName('f3-bar')).to.equal('F3Bar');
     expect(transformTagName('foo3-bar')).to.equal('Foo3Bar');
     expect(transformTagName('foo3bar-baz')).to.equal('Foo3barBaz');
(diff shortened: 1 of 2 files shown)


test/unit/helpers/curly-invocation-component-test.js
```diff
@@ -18,6 +18,10 @@ describe('#transformTagName', function() {
     expect(transformTagName('foo@bar-baz')).to.equal('Foo@BarBaz');
     expect(transformTagName('foo/bar-baz')).to.equal('Foo::BarBaz');
     expect(transformTagName('foo/bar-baz/bang')).to.equal('Foo::BarBaz::Bang');
+    expect(transformTagName('foo$bar-bang')).to.equal('Foo$BarBang');
+    expect(transformTagName('foo$bar-bang::baz::foo3bar')).to.equal('Foo$BarBang::Baz::Foo3bar');
+    expect(transformTagName('foo$bar-bang::baz??foo3bar')).to.equal('Foo$BarBang::Baz??Foo3bar');
+    expect(transformTagName('foo$bar-bang::baz/foo3bar')).to.equal('Foo$BarBang::Baz::Foo3bar');
```
Turbo87: was `Foo-Bar` a valid component name before? I'm wondering if we should add test cases for that too 🤔

rwjblue: @Turbo87 - you mean in curlies?

suchitadoshi1987: +1 on @rwjblue's question. If you mean in curlies, yes I think it is supported, I can add a test case for that.

Turbo87: yeah, I would assume that `{{Foo-Bar}}` would be a valid invocation for `components/foo-bar.js`, right?

I'm asking because I noticed that `SIMPLE_DASHERIZE_REGEXP` only considered lowercase characters and was curious if that's correct or not :)

suchitadoshi1987: @Turbo87 added the test case :)

Turbo87: sweet, thanks!

Issue #883, opened 2019-10-28
no-curly-component-invocation mistaking helper usage within angle-bracket component


Panman8201:
```hbs
<MyComponent @arg={{my-helper this.foobar}} />
```

The `no-curly-component-invocation` rule is mistaking the `{{my-helper}}` usage above as a component.

> You are using the component `{{my-helper}}` with curly component syntax. You should use `<MyHelper>` instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. ...

While I could add an exemption to the rule, this is valid usage and IMO should not error anyway.

Panman8201:
Looks like the fix #884 was merged and released. Thanks! Closing.

PR #884, opened 2019-10-29
Fix `no-curly-component-invocation` to not warn about invocations within an existing angle bracket invocation

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/883

### DIFF
--- a/test/unit/rules/lint-no-curly-component-invocation-test.js
+++ b/test/unit/rules/lint-no-curly-component-invocation-test.js
@@ -29,6 +29,9 @@ generateRuleTests({
      {{/each}}`,
     `{{#-in-element}}Hello{{/-in-element}}`,
     `{{#in-element}}Hello{{/in-element}}`,
+    '<MyComponent @arg={{my-helper this.foobar}} />',
+    '<MyComponent @arg="{{my-helper this.foobar}}" />',
+    '<MyComponent {{my-modifier this.foobar}} />',
   ],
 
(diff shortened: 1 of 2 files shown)


test/unit/rules/lint-no-curly-component-invocation-test.js
```diff
@@ -29,9 +29,18 @@ generateRuleTests({
      {{/each}}`,
     `{{#-in-element}}Hello{{/-in-element}}`,
     `{{#in-element}}Hello{{/in-element}}`,
+    '<MyComponent @arg={{my-helper this.foobar}} />',
+    '<MyComponent @arg="{{my-helper this.foobar}}" />',
+    '<MyComponent {{my-modifier this.foobar}} />',
```
Panman8201: Should the single quotes be ticks instead? Or perhaps the ticks should be single quotes (looks like there are both). 🤷‍♂

rwjblue: yeah the consistency here is a bit wonky (not introduced by @lifeart in this PR though), we should do a cleanup pass and see about enforcing consistent quoting via eslint...

bmish: Fixed: https://github.com/ember-template-lint/ember-template-lint/pull/891

rwjblue:
Awesome thank you!

Issue #881, opened 2019-10-28
Error after updating from 1.5.3 to 1.6.0


mydea:
After trying to update, I now get:

```
 -:-  error  Cannot read property '0' of undefined  undefined
```

a few times in my Octane app.

My config is like this:

```js
'use strict';

module.exports = {
  extends: 'octane',
    // enable some custom rules...
  }
};
```

The error goes away if I switch `extends: 'octane'` with `extends: 'recommended'`.

rwjblue:
Seems bad 🤔, do you (or does anyone) have a template that causes the issue handy?

alexdiliberto:
Seeing the same thing here after bumping to 1.6.1, looks like may be related to octane's new default `no-action` rule

alexdiliberto:
@rwjblue here are a couple templates from an app I work on which causes

```
<div class="form-group">
  <PInput::Text
    data-test-external-url-card-title
    @value={{@hudUrl.title}}
    @label="Title"
    @placeholder="HUD URL Title"
  />
  <ErrorMessages @errorMessages={{@hudUrl.errors.title}} class="mt-2" />
</div>
<div class="form-group">
  <PInput::Text
    data-test-external-url-card-description
    @value={{@hudUrl.description}}
    @label="Description"
    @placeholder="HUD URL Description"
  />
  <ErrorMessages @errorMessages={{@hudUrl.errors.description}} class="mt-2" />
</div>
<div class="form-group">
  <PInput::Text
    data-test-external-url-card-url
    @value={{@hudUrl.url}}
    @label="URL (must include http:// or https://)"
    @placeholder="HUD URL Value"
  />
  <ErrorMessages @errorMessages={{@hudUrl.errors.url}} class="mt-2" />
</div>
<div
  class="row no-gutters justify-content-end align-items-center form-group mb-0"
>
  {{#liquid-if (not @hudUrl.isNew)}}
    <PButton
      data-test-external-url-delete-btn
      @text="Delete"
      @cancel={{true}}
      @onClick={{action @deleteHudUrl @hudUrl}}
    >
      <FaIcon @icon="trash" @fixedWidth={{true}} />
    </PButton>
  {{/liquid-if}}

  <PInput::Checkbox
    data-test-external-url-enabled-checkbox
    @inline={{true}}
    @label="Enabled"
    @checked={{@hudUrl.enabled}}
    local-class="enabled-checkbox"
  />
</div>
```

```
<div class="container-fluid p-0">
[...]

PR #887, opened 2019-10-30
Ensure `no-action` does not error when encountering literals (StringLiteral, BooleanLiteral, etc).

lifeart:
error from 
```hbs
 <PButton
      @naked={{true}}
 />
```
and in `getLocalName` from `rules/internal/scope`

`no-action` uses `isLocal` method, and it's fails for Bool, String, Undef, Null paths

// https://github.com/ember-template-lint/ember-template-lint/issues/881

* pathExpression has empty parts if it's `{{this}}`


rwjblue:
Also, thank you for jumping on this so quickly!

Issue #886, opened 2019-10-30
Add support to inline templates(inside ts/js files)


flashios09:
Hi,
The unstable ember language server(https://github.com/flashios09/ember-language-server/commit/7d3cbc33a4b60b7bf602aad8cfe0b8d316e95178) now supports the linting for inline templates(inside ts/js).
I think that we can move it to the `ember-template-lint`, just extract the inline templates before:
```js
verify(options) {
    //...
    const source = (getExtension(options.moduleId) === '.hbs') ? 
      options.source : 
      searchAndExtractHbs(options.source);
    try {
      precompile(stripBom(source), {
        moduleName: options.moduleId,
        rawSource: stripBom(source),
        plugins: {
          ast: this.buildASTPlugins(pluginConfig),
        },
      });
    } catch (error) {
      let message = {
        fatal: true,
        moduleId: options.moduleId,
        message: error.message,
        source: error.stack,
        severity: 2,
      };
    }
    // ...
  }
```
like this:
- we don't need to use any extra plugin or another linter(eslint ...).
- we don't need to duplicate the rules, **one unified config file** will be used for hbs, js and ts.
- the **ember-cli-template-lint** will also support the inline templates.

rwjblue:
I think it would take a few more changes, but seems _generally_ good to me. 

Some things that this will have to address when implemented:

* The bin script globing would have to change (it defaults to `**/*.hbs`)
* Ensure that the reported line and column numbers in all of the reported failures offsets are correct to what is actually in the file
* We’d need to figure out how to parse the JS/TS files (possibly requiring “parser” config like eslint does) so that we don’t have issues with “custom syntax” (e.g. decorators)

flashios09:
> The bin script globing would have to change (it defaults to **/*.hbs).

I know about this and i think it can be updated to support inline template too. Just remove/update the *extension* check.

> Ensure that the reported line and column numbers in all of the reported failures offsets are correct to what is actually in the file

The `searchAndExtractHbs` function works fine with indentation(column, line), you can check the tests here https://github.com/flashios09/extract-tagged-template-literals/blob/master/src/__tests__/searchAndExtract.test.ts

> We’d need to figure out how to parse the JS/TS files (possibly requiring “parser” config like eslint does) so that we don’t have issues with “custom syntax” (e.g. decorators)

Sorry i didn't get this point, can you explain more please ?

gabrielcsapo:
Any update on this? This would be super great to have added.

[...]

Issue #305, opened 2017-10-18
Autofix option


RedTn:
Great addon! I'm having a tough time finding if there is a --fix option like in eslint (eslint . --fix) where it will automatically fix files that fail the lint rules. Is there an option for this or is it missing?

rwjblue:
Sadly no, but it is definitely something that we want to make happen at some point.

PR #888, opened 2019-10-30
feat: ability to lint inline templates in .js/.ts files

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/886

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -55,7 +55,7 @@ async function buildLinterOptions(workingDir, filePath, filename = '', isReading
 }
 
 function executeGlobby(workingDir, pattern, ignore) {
-  let supportedExtensions = new Set(['.hbs', '.handlebars']);
+  let supportedExtensions = new Set(['.hbs', '.handlebars', '.js', '.ts']);
 
   // `--no-ignore-pattern` results in `ignorePattern === [false]`
   let options =
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -1,6 +1,8 @@
 const path = require('path');
 
 const { parse, transform } = require('ember-template-recast');
+const { searchAndExtractHbs } = require('@lifeart/ember-extract-inline-templates');
+const { parse: rawParseJS } = require('@babel/parser');
 
 const ModuleStatusCache = require('./-private/module-status-cache');
 const TodoHandler = require('./-private/todo-handler');
@@ -314,6 +316,21 @@ class Linter {
 
     let hasBOM = options.source.charCodeAt(0) === 0xfeff;
     let source = hasBOM ? options.source.slice(1) : options.source;
+
+    if ((options.filePath || '').endsWith('.js') || (options.filePath || '').endsWith('.ts')) {
+      try {
+        source = searchAndExtractHbs(source, {
+          parse(code) {
+            return rawParseJS(code, {
+              sourceType: 'module',
+              plugins: [],
+            });
+          },
+        });
+      } catch {
+        // EOL
+      }
+    }
 
     let templateAST;
 
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "@ember-template-lint/todo-utils": "^10.0.0",
+    "@lifeart/ember-extract-inline-templates": "^2.0.0",
     "chalk": "^4.1.2",
     "ci-info": "^3.2.0",
     "date-fns": "^2.24.0",
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   "dependencies": {
     "@ember-template-lint/todo-utils": "^10.0.0",
     "@lifeart/ember-extract-inline-templates": "^2.0.0",
+    "@babel/parser": "^7.15.7",
     "chalk": "^4.1.2",
     "ci-info": "^3.2.0",
     "date-fns": "^2.24.0",
@@ -65,7 +66,6 @@
     "yargs": "^16.2.0"
   },
   "devDependencies": {
-    "@babel/parser": "^7.15.7",
     "@babel/traverse": "^7.15.4",
     "@microsoft/jest-sarif": "^1.0.0-beta.0",
     "common-tags": "^1.8.0",
--- a/package.json
+++ b/package.json
@@ -47,9 +47,9 @@
     ]
   },
   "dependencies": {
+    "@babel/parser": "^7.15.7",
     "@ember-template-lint/todo-utils": "^10.0.0",
     "@lifeart/ember-extract-inline-templates": "^2.0.0",
-    "@babel/parser": "^7.15.7",
     "chalk": "^4.1.2",
     "ci-info": "^3.2.0",
     "date-fns": "^2.24.0",
--- a/test/acceptance/public-api-test.js
+++ b/test/acceptance/public-api-test.js
@@ -1450,4 +1450,125 @@ describe('public api', function () {
       expect(result.isFixed).toEqual(true);
     });
   });
+
+  describe('Linter able to lint and fix .[js, ts] files', function () {
+    let linter;
+
+    beforeEach(function () {
+      project.setConfig({
+        rules: {
+          quotes: ['error', 'double'],
+          'require-button-type': 'error',
+        },
+      });
+
+      project.write({
+        app: {
+          templates: {
+            'application.js': `
+              import { hbs } from 'ember-cli-htmlbars';
+              export default hbs\`<input class='mb4'>\`;
+            `,
+            'other.ts': `
+              import { hbs } from 'ember-cli-htmlbars';
+              interface Foo {};
+              export default hbs\`<button>LOL, Click me!</button>\`;
+            `,
+          },
+        },
+      });
+
+      linter = new Linter({
+        console: mockConsole,
+        configPath: project.path('.template-lintrc.js'),
+      });
+    });
+
+    afterEach(async function () {
+      await project.dispose();
+    });
+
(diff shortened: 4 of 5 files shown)


lifeart:
@flashios09 I added you as collaborator for `lifeart/ember-template-lint`, could you check node 6 ci issues?

lifeart:
some updates
image

lib/index.js
```diff
@@ -5,6 +5,7 @@ const Minimatch = require('minimatch').Minimatch;
 const getConfig = require('./get-config');
 const stripBom = require('strip-bom');
 const chalk = require('chalk');
+const { searchAndExtractHbs } = require('extract-tagged-template-literals');
```
rwjblue: Hmm, I'm vaguely concerned with a few things in this lib:

* Why does it use regexp's instead of parsing the JS/TS? (using regexp's _seems_ error prone)
* How does the lib support things like `hbs('some template here')`? I made https://github.com/flashios09/extract-tagged-template-literals/issues/2 to discuss.
* I'd like to see an increased bus factor before using it here (I think only @flashios09 has commit + release permissions atm)...
* How do we plan to support updates? We are actively working to move ember-template-lint over to using ember-template-recast (see #305 and #873) so that we can support a `--fix` concept and I'd like to make sure we have a path to support the same functionality in inline templates.
* I would have assumed that we want to process each inline template independently, I'm not super tied to that but its just what I thought we'd want
* I wonder if we should be correcting the location info here in ember-template-lint or in the underlying lib. There are a number of rules that I think could fail given the current setup (things like indentation thinking that the first block should be completely outdented, or that your template doesn't have a bunch of trailing newlines, etc)

lib/index.js
```diff
@@ -122,7 +123,12 @@ class Linter {
     };
 
     try {
-      preprocess(stripBom(options.source), {
+      const source = stripBom(options.source);
+      const preprocessStr =
+        options.moduleId.endsWith('.js') || options.moduleId.endsWith('.ts')
+          ? searchAndExtractHbs(source)
+          : source;
```
rwjblue: Perhaps we should target this at `ember-template-recast` itself since that lib wants to be able to update inline templates and we are actively working on updating here to use `ember-template-recast`, see #873

bin/ember-template-lint.js
```diff
@@ -77,7 +77,12 @@ function expandFileGlobs(positional) {
         ignore: ['**/dist/**', '**/tmp/**', '**/node_modules/**'],
         gitignore: true,
       })
-      .filter(filePath => filePath.slice(-4) === '.hbs')
+      .filter(
+        filePath =>
+          filePath.slice(-4) === '.hbs' ||
```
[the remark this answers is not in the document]
flashios09: > no need to slice, since we aren't really using the result


I think there is no difference between this `filePath.slice(-4) ===  '.hbs'` and this `filePath.endsWith('.hbs')`, **slice** will **not modify** the original string **filePath**.

But here, i prefer using `endsWith` for readability reason :)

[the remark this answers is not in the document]
rwjblue: > I think there is no difference between this filePath.slice(-4) === '.hbs' and this filePath.endsWith('.hbs'), slice will not modify the original string filePath.

Yep, I totally understand this is how `.slice` works. I just don't think we need `.slice` because we aren't using the result and we want **exactly* what `.endsWith` is made for 😸

lib/index.js
flashios09: > Why does it use regexp's instead of parsing the JS/TS? (using regexp's seems error prone)

Why we need a **parser** to **just** extract **tagged literal templates** ? 
and what is the parser to use to go from **js/ts** to **hbs** ?
I have tried:
- **javascript/babylon** -> transform **js/ts** to **js/ts**
- **handlebars/ember-template-recast/glimmer** -> transform **hbs** to **hbs**

> How does the lib support things like hbs('some template here')? I made flashios09/extract-tagged-template-literals#2 to discuss.

The goal was to **add ember template linting to tagged literal templates** inside **ts/js** **component** or **test/component**, like this we can write SFC(personally i prefer SFC especially if we have **short template**)
-> so we extract **inline templates**, replace the script lines with `' '` or `' '`(to preserve the indentation or `\n`) then pass **the converted to hbs source** to `Glimmer::preprocess(...)` to get linting errors/warning.

I don't **think/know** that we **can/have to** do the same for **non tagged literal templates** formats, i mean adding support for **ember template lint** !!
How we can do this with **CallExpression** ? 
What about concatenation ? using vars for templates ?
```js
const input = '<input />'; // ember-template-error self-closing-void-elements
const template = hbs('<div>' + input + '</div>' + '<br>'); 
```
what we will extract here ? the error **self-closing-void-elements** will be attached  to line 1 or line 2 ?

> I'd like to see an increased bus factor before using it here (I think only @flashios09 has commit + release permissions atm)...

I will send you an invitation if you want :trollface: 

> How do we plan to support updates? We are actively working to move ember-template-lint over to using ember-template-recast (see #305 and #873) so that we can support a --fix concept and I'd like to make sure we have a path to support the same functionality in inline templates.

The extracted inline templates(just hbs source) will have the same locations(line, column), the indentations will be preserved(same as written in the original file), so i think it can be parsed with **glimmer** then if we have some errors that we can fix, e.g. from `<input />` to `<input >`, it can be done :)
Otherwise, what is the input format that we have to pass ?

> I would have assumed that we want to process each inline template independently, I'm not super tied to that but its just what I thought we'd want

Why we want to process each inline template **independently** ?

> I wonder if we should be correcting the location info here in ember-template-lint or in the underlying lib.

What location should be corrected ?

> There are a number of rules that I think could fail given the current setup (things like indentation thinking that the first block should be completely outdented, or that your template doesn't have a bunch of trailing newlines, etc)

We can skip/ignore some rules for inline templates like `eol-last` but for the indentation, i don't think that we can touch/remove/outdent it, how to deal with something like this:
```js
const inputTemplate = hbs`<input value={{@value}} />`; // error self-closing-void-elements
const buttonTemplate = hbs`
    <button type={{this.type}}>
         {{yield}}
    </button>
`;
```
will be converted to:
```hbs
                                                <input value={{@value}} />

    <button type={{this.type}}>
         {{yield}}
    </button>
```
If we will force a **de-indentation** or remove the **left whitespaces**, the lint error will not have the correct location(line and column number) ! 

For **trailing spaces**, i have already fix this in `extract-tagged-literal-templates`.
The empty lines at the end(after the last template), will not be added(appended) to the extracted hbs source.

lib/index.js
flashios09: I have tried `ember-template-recast`, correct me if i'm wrong but this will parse and transform **hbs** source to **hbs** source, here the input isn't **hbs**, we need to convert the js/ts content to hbs first !

lifeart: you need to extract template text using (for example - ember-meta-explorer), and after - use it as souce for template-lint logic, including ember-template-recast

flashios09: @lifeart I have tried `ember-meta-explorer` and it seems just a **wrapper** of `parser`(babylon..., to AST) + `traverser`(babel-traverser with some pre-defined params, plugins ...) + `generator`(babel-generator, from AST to the original code format).

I have scanned all the test examples, i can't find any thing that go from `js/ts` to `hbs` :(

First, we need to figure out how to deal with **Call Expression** formats (e.g. `hbs('template content')`), please take a look at https://github.com/ember-template-lint/ember-template-lint/pull/888#discussion_r342510569

-> i'm waiting for Robert response, if we will not support **Call Expression** formats, i don't think that we have to use a parser to **just extract tagged literal templates** !

lifeart: flashios09

```js
const traverse = require("@babel/traverse");
const { parseScriptFile } = require("ember-meta-explorer");

const fileContent = 'hbs`hello`';
const templates = [];
const ast = parseScriptFile(fileContent);

traverse(ast, {
  TaggedTemplateExpression(path) {
    const node = path.node;
    if (node.tag && node.tag.type === 'Identifier' && node.tag.name === 'hbs') {
      const location = node.quasi.quasis[0].value.loc;
      const templateText = node.quasi.quasis[0].value.raw;
      templates.push([templateText , location]);
    }
  }
});
 
 
console.log(templates);
```

lib/index.js
rwjblue: > BTW, is it possible to use more than one template for a component/route file ?

No, not really. It is however very common to have many many inline templates in test files.

gabrielcsapo:
Any momentum on this @lifeart? We are seeing our tests and production code not align when it comes to the rules we have enabled and would love to see that fixed.

lifeart:
@gabrielcsapo one of the things to figure out - how to deal with multiple templates (for example - in tests), some A11Y rules may require uniq ID's or such. We need to split file to templates, and get linting results independently and it's not quite a good aligned with existing architecture and we need to figute out best approach, do you have any conserns, related to conversation above?

gabrielcsapo:
@lifeart I was think that adding a "context" option to pass to the lint rules that says "you are in a test block" or you are in a "production code block" could allow rules to decide what seems appropriate for them to check against. Also another option would be have the ability to turn rules on and off for tests and production code.

gabrielcsapo:
@lifeart any update on this PR? it would be great to get this in, most of the violations that we are seeing are coming from our integration tests and would be great to fix those 😞

lifeart:
@rwjblue do you have any suggestions on this?
Looks like we have ast template extractor, and custom extensions now supported, should we move forward with this?

lifeart:
Related one - https://github.com/flashios09/ember-extract-inline-templates/pull/5

NullVoxPopuli:
What's the tldr left on this? can we get TODO checkboxes added to the description?

lifeart:
closing as implemented in another PR's

PR #892, opened 2019-11-02
Remove reliance on `preprocess` (from `@glimmer/syntax`) ability to accept an array of AST plugins

rwjblue:
This builds on top of the work that @dcyriller started over in #872 (which is an awesome refactor) and takes it even further to remove the need to pass any arguments to `preprocess`.

This allows us to:

* have much more control on how our rules are instantiated
* report errors _per rule_ (and annotate them properly as coming from the specified rule)
* report errors upon parsing the file separately from those of running rules.
* report errors that occur when parsing a rules config separately from those of running rules.
* continue checking rules that _do_ have valid config if **any** errors occur during config parsing
* continue checking remaining rules when a single error occurs
* prepare for migration to ember-template-recast

Closes https://github.com/ember-template-lint/ember-template-lint/pull/872


test/unit/rules/lint-no-trailing-dot-in-path-expression-test.js
```diff
@@ -1,86 +0,0 @@
...
-    '{{#contact-details contact=contact}}\n' +
-      '   {{contact.displayName}}\n' +
-      '{{/contact-details}}',
-  ],
-
-  bad: [
-    {
-      template: '{{contact.}}',
-
-      fatal: {
```
rwjblue: These tests are removed because we now actually have the ability to receive **multiple** fatal errors (we were previously just swallowing them). I removed the tests since this is all just testing implementation details of `@glimmer/syntax` itself.

PR #893, opened 2019-11-02
Add `no-args-paths` to the `octane` preset

lifeart:
@rwjblue


Turbo87:
hmm, shouldn't a preset change be considered a breaking change? or do we consider the Octane preset experimental/unstable for now?

rwjblue:
> or do we consider the Octane preset experimental/unstable for now?

Confirm, like octane itself it is a "preview". I'll open an issue to track making our SemVer commitments clearer here.

Issue #894, opened 2019-11-02
Rule request: no-index-component-invocation


lifeart:
refs: https://github.com/NullVoxPopuli/coc-ember/issues/6#issuecomment-549075135

![image](https://user-images.githubusercontent.com/1360552/68076512-7c2a2700-fdc6-11e9-9a7f-7748a064f607.png)

rwjblue:
Seems reasonable to me...

PR #895, opened 2019-11-03
Add `no-index-component-invocation` rule

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/894

should it be in octane config?

// cc @rwjblue, @NullVoxPopuli


lib/rules/lint-no-index-component-invocation.js
```diff
@@ -0,0 +1,72 @@
...
+
+module.exports = class LintNoIndexComponentInvocation extends Rule {
+  visitor() {
+    const sourceForNode = this.sourceForNode.bind(this);
+    const log = this.log.bind(this);
+    function lintIndexUsage(node) {
+      if (node.type === 'ElementNode') {
+        if (node.tag.endsWith('::Index')) {
+          return log({
+            message: `Replace "<${node.tag} ..." to <${node.tag.replace('::Index', '')} ..."`,
```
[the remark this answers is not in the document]
lifeart: Good catch!

test/unit/rules/lint-no-index-component-invocation-test.js
```diff
@@ -0,0 +1,76 @@
...
+        message: 'Replace "{{foo/index ..." to "{{foo ..."',
+        source: 'foo/index',
+        line: 1,
+        column: 2,
+      },
+    },
+    {
+      template: '{{foo/bar (component "foo/index")}}',
+      result: {
+        message: 'Replace "(component "foo/index" ..." to "(component "foo" ..."',
```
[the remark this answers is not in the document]
lifeart: Done

lifeart:
//cc @rwjblue

docs/rule/no-index-component-invocation.md
```diff
@@ -0,0 +1,58 @@
+## no-index-component-invocation
+
+We propose to allow `index`
+files inside a directory to have the equivalent semantics. Components and Component Templates can be structured as `app/components/foo-bar/index.js` and
+`app/components/foo-bar/index.hbs`. This allows additional files related to the
+component (such as a `README.md` file) to be co-located on the filesystem.
```
Turbo87: While this is possible, it is not the default for Ember projects so IMHO we shouldn't be suggesting it here. We can mention that this works, but "propose" might be the wrong word here.

lib/rules/lint-no-index-component-invocation.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const Rule = require('./base');
+
+module.exports = class LintNoIndexComponentInvocation extends Rule {
+  visitor() {
+    const sourceForNode = this.sourceForNode.bind(this);
+    const log = this.log.bind(this);
+    function lintIndexUsage(node) {
```
Turbo87: if you make this a method on the `LintNoIndexComponentInvocation` class you don't have to explicitly bind the methods above 😉

lifeart: using function make easier to test logic in astexplorer

docs/rule/no-index-component-invocation.md
lifeart: Tobias, could you provide more suitable sentence?

lifeart:
image

MrChocolatine:
Does this rule only apply to Octane or what ? I am running a project with Ember 3.12 and if I rename a component `<Foo::Index />` to `<Foo />` it does not seem to be rendered anymore.

Actually in Ember Inspector I can see `<Foo />` its render is blank. If I keep the `::Index` the component renders correctly. What could be the issue ?

lifeart:
@MrChocolatine is component template located into `templates/components/foo/index.hbs` ?

MrChocolatine:
> @MrChocolatine is component template located into `templates/components/foo/index.hbs` ?

It is yes.

The actual files of my component are:
```
app/components/foo/bar/index.js
app/templates/components/foo/bar/index.hbs
```

When I invoke this component:
```hbs
-- works --
<Foo::Bar::Index @model={{this.model.whatever}} />

-- does not work / render nothing --
<Foo::Bar @model={{this.model.whatever}} />
```

lifeart:
@MrChocolatine could you try update ember-cli-babel version?

MrChocolatine:
I just updated it to its last version `ember-cli-babel@7.20.5` and I still have the same issue.

I even tried to simplify the process but it didn't help:
- A simple text in the component's HBS file
- Empty content of the component's JS class.
- Change invocation of the component `<Foo::Bar />` (without argument this time)

If I delete the component's JS file I get an error `Compile Error: Cannot find component foo/bar`.

lifeart:
@MrChocolatine yeah, it's related to template compilation logic updates (that's why it's working for octane)

cc  @rwjblue

MrChocolatine:
That is what I expected.
When I started to make components with `::Index` I tried to get rid of it without success. It in the end, if I remember correctly, I had to upgrade my Ember version to make it works without `::Index`. (and I will, but not immediately)

MrChocolatine:
We recently did a massive upgrade of dependencies to their latest version, except for `ember-cli`, `ember-source` and `ember-data` but nothing changed, `<Foo::Bar />` still renders nothing.

Still convinced this rule should only apply to Octane projects.

rwjblue:
You definitely do not need to be using Octane to invoke nested components like this (or to use colocation in general).

MrChocolatine:
> You definitely do not need to be using Octane to invoke nested components like this (or to use colocation in general).

We are speaking of `<Foo::Bar::Index />` and `<Foo::Bar />` right ? (I just edited my previous message)
Then I don't see what could the issue since I am using Ember 3.12 and every other dependencies are up-to-date 🤔 .

lifeart:
As I understand issue happens on such layout:
```
app/components/foo/index.js
app/templates/components/foo/index.hbs
```

PR #885, opened 2019-10-29
Drop support for NodeJs v6 and v11

dcyriller:
Close https://github.com/ember-template-lint/ember-template-lint/issues/724

Will be required for https://github.com/ember-template-lint/ember-template-lint/pull/873 (see https://github.com/ember-template-lint/ember-template-recast/blob/master/package.json#L56)

It will unblock https://github.com/ember-template-lint/ember-template-lint/pull/766 and https://github.com/ember-template-lint/ember-template-lint/pull/707

PR #844, opened 2019-10-06
Remove HTML comment config

initram:
Removed the code handling a legacy way of using HTML comments to configure the linter.

Fixes #749

PR #897, opened 2019-11-03
Breaking changes for v2.0.0

rwjblue:
This PR represents the future v2.0.0 release, and is useful as a way to land various pending PR's together for v2.0.0.

Currently includes (will be updated as we merge more PRs into this branch):

* Add `no-args-path`, `no-invalid-link-text`, `no-invalid-meta`, `require-button-type` into `recommended` configuration #1047
* Add `no-index-component-invocation` to `recommended` configuration #1059
* Add `no-invalid-role` and `no-negated-condition` rules to `recommended` config #1053
* Add `stylistic` configuration #1048
* Add support for detecting `eol-last` rule config from `.editorconfig` #1062
* Add utility function to allow reading from `.editorconfig` #906
* Begin migrating tests to inline fixtures (via `broccoli-test-helper`) #876
* Change `allowDynamicStyles` option to default to true in `no-inline-styles` rule #1052
* Drop support for Node 8 #1023
* Drop support for Node 6 and 11 #885
* Migrate from `@glimmer/syntax` to `ember-template-recast` (as a parser) #873 
* Move main linter engine to stand alone file #932
* Remove HTML comment config #844
* Remove deprecated `templateEnvironmentData` rule property #1067
* Remove deprecated array config for `no-nested-interactive` rule #1055
* Remove deprecated rule `img-alt-attributes` (replaced by `require-valid-alt-text`) #900
* Remove deprecated rule `no-meta-redirect-with-time-limit` which was replaced by `no-invalid-meta` #1054
* Remove deprecated rule `no-trailing-dot-in-path-expression` #1061
* Remove support for most deprecated rule names #1069
* Update dependencies to latest (requiring Node 8+) #898
* Update `no-invalid-meta` rule to catch incorrect meta attribute combinations #1088
* Update `recommended` config #899
* Use `.editorconfig` to choose default value for `block-indentation` #918
* Use `.editorconfig` to choose default value for `linebreak-style` #919


rwjblue:
I just published `2.0.0-beta.1` with these changes.

lifeart:
Could we merge https://github.com/ember-template-lint/ember-template-lint/pull/876/ ?

rwjblue:
> Could we merge #876 ?

Ya, a few comments over there to resolve first but should be good to land once those are addressed.

rwjblue:
Released v2.0.0-beta.2

MelSumner:
It would be great to get the semver policy and the a11y-strict preset in this version if possible.  :+1:

MelSumner:
@rwjblue maybe need to update https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/ignore.md to remove reference to MU.

rwjblue:
@Turbo87 / @dcyriller / @bmish - What blockers do we have to release here?

dcyriller:
I don’t see any blocker left

lib/config/recommended.js
```diff
@@ -2,31 +2,34 @@
...
-    'link-rel-noopener': true,
+    'link-rel-noopener': 'strict',
+    'link-href-attributes': true,
+    'no-abstract-roles': true,
     'no-attrs-in-components': true,
     'no-debugger': true,
     'no-duplicate-attributes': true,
+    'no-extra-mut-helper-argument': true,
     'no-html-comments': true,
+    'no-implicit-this': true,
```
Turbo87: are we sure we want to enable this by default?

I was under the impression that we only wanted to enable this for the `octane` preset for now. If we do enable this, then why not also the `no-curly...` rule? 🤔

Turbo87:
seeing the current discussion on Discord I'm wondering if we should change the recommended `no-inline-styles` config to use the `allowDynamicStyles` option 🤔 

see https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-inline-styles.md

/cc @bmish

rwjblue:
> seeing the current discussion on Discord I'm wondering if we should change the recommended no-inline-styles config to use the allowDynamicStyles option 

Not sure which convo you mean here, mind opening a specific issue for this (so we can discuss it)?

lib/config/recommended.js
rwjblue: @Turbo87 - Ya, agree. I'll update.

Turbo87:
@rwjblue https://github.com/ember-template-lint/ember-template-lint/issues/1044

rwjblue:
Thank you for making the issue! I'm going to decouple from the release of 2.0.0 though, since (based on our SemVer policy) we can make a rule set _less_ restrictive within a major version.

lib/config/recommended.js
rwjblue: Taken care of in #1045

rwjblue:
OK, will be releasing 2.0.0 "soon" (tomorrow morning most likely), please do a (hopefully) final review pass and let me know if there are any other issues y'all spot.

rwjblue:
https://github.com/ember-template-lint/ember-template-lint/pull/1144 should fix CI here, any objections to just shipping this thing!?

Turbo87:
definitely no objections. glad to finally get rid of the next branch :D

PR #910, opened 2019-11-06
Add `no-invalid-meta` rule

MelSumner:
If merged, this PR updates the no-meta-redirect-with-time-limit:
- deprecates the rule
- names the new rule `no-invalid-meta`
Added checks to:
- to differentiate between REFRESH and REDIRECT and accommodate F41 as well as F40 - if it's a refresh, the delay should be greater than 72000 seconds
- disallow maximum-scale
- disallow `user-scalable=no`


rwjblue:
Adding validation for this seems fine, but we either need to change the rule name (deprecating the original and adding a new one) or make a new rule.

I'd _rather_ change the name (e.g. `no-invalid-meta`), what do you think @MelSumner?

MelSumner:
`no-invalid-meta-refresh` since there are other meta tags and we're not targeting those? Or do we want to leave room for future meta rules?

rwjblue:
My thought was that we could go with `no-invalid-meta`, and add configuration for each of the various scenarios to check.

MelSumner:
Okay, sounds good, I'll work on it.

Turbo87:
I think I personally would prefer `no-invalid-meta-refresh`. That way the implementation of the rule is easier to maintain and the rule can be enabled/disabled individually from other meta-related rules.

MelSumner:
@rwjblue updated to be `no-invalid-meta` - this works out because there are other rules about meta tags to write as well.

lib/rules/no-invalid-meta.js
```diff
@@ -0,0 +1,66 @@
...
+                  message: 'a meta refresh should have a delay greater than 72000 seconds',
+                  line: node.loc && node.loc.start.line,
+                  column: node.loc && node.loc.start.column,
+                  source: this.sourceForNode(node),
+                });
+              }
+            }
+          }
+        }
+        if (contentAttrValue.includes('user-scalable=no')) {
```
rwjblue: Small note RE: parsing here, we have to detect `user-scalable =    no` (arbitrary whitespace is allowed surrounding the `=`).

docs/rule/no-meta-redirect-with-time-limit.md
```diff
@@ -1,30 +0,0 @@
-## (no-meta-redirect-with-time-limit)
```
rwjblue: Since we have released this one, we can't remove it here. Can you add an entry to the deprecated rules list for it?

See https://github.com/ember-template-lint/ember-template-lint/commit/2016af3b8698fe28d931a667c2a3bb34a7920a6a for an example.

PR #911, opened 2019-11-07
Clarify that logo and spacer are forbidden words

gojefferson:
I got foot gunned by this rule and had to look at the source code to see why. Would be nice to document which words are blacklisted in alt text.


rwjblue:
Thank you!

PR #913, opened 2019-11-07
Update `no-unnecessary-component-helper` rule to allow the `component` helper as an angle bracket component argument

buschtoens:
This adds a failing test case for `no-unnecessary-component-helper` for the usage of the curly `{{component}}` helper as an argument to an angle bracket component invocation.

```hbs
<Foo @bar={{component "my-component"}} />
```

It fixes the test by recording all valid usages of the `{{component}}` helper as attributes to an `ElementNode` in a Set and then exempting these nodes in `checkNode`.

I'm not sure whether this is the best / most elegant solution, but I did not find a different way to get the parent of the node passed to `checkNode`. I'm happy to implement this a different way, if you can point me towards a better solution.

### DIFF
--- a/lib/rules/lint-no-unnecessary-component-helper.js
+++ b/lib/rules/lint-no-unnecessary-component-helper.js
@@ -7,6 +7,16 @@ const ERROR_MESSAGE = 'Invoke component directly instead of using `component` he
 
 module.exports = class NoUnnecessaryComponentHelper extends Rule {
   visitor() {
+    let inSafeNamespace = false;
+    const markAsSafeNamespace = {
+      enter() {
+        inSafeNamespace = true;
+      },
+      exit() {
+        inSafeNamespace = false;
+      },
+    };
+
     function isComponentHelper(node) {
       return (
         node &&
@@ -17,19 +27,12 @@ module.exports = class NoUnnecessaryComponentHelper extends Rule {
       );
     }
 
-    const allowedUsages = new Set();
-    function recordAngleBracketArguments(node) {
-      node.attributes
-        .filter(attribute => attribute.name.startsWith('@') && isComponentHelper(attribute.value))
-        .forEach(attribute => allowedUsages.add(attribute.value));
-    }
-
     function checkNode(node) {
       if (
         isComponentHelper(node) &&
         AstNodeInfo.isStringLiteral(node.params[0]) &&
         !node.params[0].value.includes('@') &&
-        !allowedUsages.has(node)
+        !inSafeNamespace
       ) {
         this.log({
           message: ERROR_MESSAGE,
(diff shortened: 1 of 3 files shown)


lib/rules/lint-no-unnecessary-component-helper.js
```diff
@@ -7,13 +7,29 @@ const ERROR_MESSAGE = 'Invoke component directly instead of using `component` he
 
 module.exports = class NoUnnecessaryComponentHelper extends Rule {
   visitor() {
-    function checkNode(node) {
-      if (
+    function isComponentHelper(node) {
+      return (
+        node &&
```
bmish: These null checks (`node && node.path`) should be removed. If someone passes null to this function, it's a problem and we should fail fast. Instead, you can use type checks from AstNodeInfo on node if needed.

test/unit/rules/lint-no-unnecessary-component-helper-test.js
```diff
@@ -31,6 +31,15 @@ generateRuleTests({
...
     '(component "my-component")',
 
+    // Curly inline usage in an angle bracket component:
+    '<Foo @bar={{component SOME_COMPONENT_NAME}} />',
+    '<Foo @bar={{component "my-component"}} />',
+    '<Foo @bar={{component SOME_COMPONENT_NAME}}></Foo>',
+    '<Foo @bar={{component "my-component"}}></Foo>',
+
+    // Static arguments in angle bracket components don't crash the rule:
+    '<Foo @arg="foo" />',
```
bmish: Can you add another test case with an if statement like:
```
<Foo @arg={{if this.user.isAdmin "admin"}}/>
```

bmish: Can you add a non-`@` parameter test case?
```
<Foo some-property={{component "bar"}}/>
```

lib/rules/lint-no-unnecessary-component-helper.js
buschtoens: I added them in 1b5892ea738a726ccf08542dff4f2f68aac1c7dd, because static attributes don't have a `path`, causing a crash in L14, since `AstNodeInfo.isPathExpression(node.path)` does not handle nullish values.

https://github.com/ember-template-lint/ember-template-lint/blob/bc849e2376990ca11ec71396fa90f8e449b028f3/lib/helpers/ast-node-info.js#L31-L33

buschtoens: I can either:

- keep everything as it is
- add `node &&` to the `isPathExpression` (and other helpers?)
- move the `node && node.path` out of `isComponentHelper` and into the `filter` arrow function in [`recordAngleBracketArguments`](https://github.com/buschtoens/ember-template-lint/blob/no-unnecessary-component-helper-with-angle-brackets/lib/rules/lint-no-unnecessary-component-helper.js#L23)

buschtoens:
@rwjblue @bmish Thanks for guiding me through it! 😊

Issue #284, opened 2017-10-06
linebreak-style should auto-detect from `.editorconfig`


rwjblue:
The `true` configuration value for the `linebreak-style` rule should be modified to check for `.editorconfig` file and then fall back to `unix` setting if no `.editorconfig` is present (or the `.editorconfig` does not have line ending configuration.

rwjblue:
@initram - What do you think?

initram:
I think that makes sense. This could also be used for the indentation rule.

rwjblue:
I wonder if there is a package for parsing `.editorconfig`?

ro0gr:
> I wonder if there is a package for parsing .editorconfig?

Is that what you are looking for ? - https://www.npmjs.com/package/editorconfig

not sure if that's helpful, but I worked on a POC for editorconfig aware blueprints a while ago https://github.com/ember-cli/blprnt/pull/29/files

rwjblue:
Yep, thats perfect!

rwjblue:
I believe that this was resolved in #918 / #906, and is released in 2.0.0-beta.2.

PR #919, opened 2019-11-11
Use `.editorconfig` to choose default value for `linebreak-style`

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/pull/906

fix: https://github.com/ember-template-lint/ember-template-lint/issues/284


rwjblue:
Thank you again!

Issue #914, opened 2019-11-08
`no-bare-strings` should consider literals in mustaches (e.g. `{{"foo"}}`)


rwjblue:
When the `no-bare-strings` rule is enabled, the following is not allowed:

```hbs
<p>Hello!</p>
```

The rule should **also** prevent:

```hbs
<p>{{"Hello!"}}</p>
```

lifeart:
how about this?

```hbs
<CustomInput @type={{"range" | "string" | "password"}} />
```

rwjblue:
Component arguments are not validated by `no-bare-strings`.

rwjblue:
Closed by #920

PR #920, opened 2019-11-11
Ensure `no-bare-strings` consider literals in mustaches (e.g. `{{"foo"}}`)

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/issues/914

### DIFF
--- a/lib/rules/lint-no-bare-strings.js
+++ b/lib/rules/lint-no-bare-strings.js
@@ -211,7 +211,11 @@ module.exports = class LogStaticStrings extends Rule {
           this._elementStack.pop();
         },
       },
-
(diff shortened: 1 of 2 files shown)


lib/rules/lint-no-bare-strings.js
```diff
@@ -211,7 +211,11 @@ module.exports = class LogStaticStrings extends Rule {
           this._elementStack.pop();
         },
       },
-
+      MustacheStatement(node) {
+        if (!this._currentAttrNode) {
```
rwjblue: This also needs to avoid checking element modifiers right?

rwjblue:
Thanks @lifeart!

PR #923, opened 2019-11-14
fix: allow `getLocalName` to work with numbers

### LINKED ISSUE #922: Template number literal `Cannot read property 'length' of undefined` (glimmer-engine)
Seeing this error when using a raw number literal as an assigned argument property in my template.

```hbs
      <AnimatedText
        @text={{this.errorMessage}}
        @animationDelay={{300}} <-- this is the issue
      />
```

```md
my-app/templates/components/foo.hbs
  -:-  error  Cannot read property 'length' of undefined  undefined
```

`@animationDelay={{300}}` is causing the issue here, removing that line makes the error go away

Here's my `.template-lintrc.js` file

```js
'use strict';

module.exports = {
  extends: 'octane',

  rules: {
    'attribute-indentation': true,
    'inline-link-to': false,
    quotes: 'double',
    'block-indentation': true,
    'img-alt-attributes': true,
    'linebreak-style': true,
    'no-unnecessary-concat': true,
    'self-closing-void-elements': true,
      requireDash: false,
      allow: [
		// ... a bunch of components
      ]
    }
  }
};

```

```
ember-template-lint@^1.1.0:
  version "1.8.1"
ember-cli-template-lint@^1.0.0-beta.3:
  version "1.0.0-beta.3"
```

Let me know if more information is necessary to reproduce.

alexlafroscia:
While many types of literals were handled within `getLocalName`, numbers were forgotten!

Closes #922

### DIFF
--- a/lib/rules/internal/scope.js
+++ b/lib/rules/internal/scope.js
@@ -16,6 +16,7 @@ function getLocalName(node) {
     case 'NullLiteral':
     case 'BooleanLiteral':
     case 'StringLiteral':
+    case 'NumberLiteral':
       return undefined;
     case 'PathExpression':
     default:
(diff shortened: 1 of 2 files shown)


alexdiliberto:
Thank you!

Issue #926, opened 2019-11-14
Rule Request: No arguments for html elements


lifeart:
For template:

```hbs
<option @value="1" />
```

we have such error:

```
VM6678 vendor.js:44766 Uncaught (in promise) TypeError: func is not a function
    at Compilers.compile (VM6678 vendor.js:44766)
    at compile (VM6678 vendor.js:45462)
    at LazyCompiler.add (VM6678 vendor.js:45656)
    at CompilableBlock.compile (VM6678 vendor.js:45451)
    at Object.evaluate (VM6678 vendor.js:49455)
    at AppendOpcodes.evaluate (VM6678 vendor.js:48807)
    at LowLevelVM.evaluateSyscall (VM6678 vendor.js:52029)
    at LowLevelVM.evaluateInner (VM6678 vendor.js:51975)
    at LowLevelVM.evaluateOuter (VM6678 vendor.js:51967)
    at VM.next (VM6678 vendor.js:54033) 
```

This is creepy and untracable (for end-users), I think we can guess (is it component or not) and emit error if argument assigned to lowercased element without same block name and without dots in notation.

https://codesandbox.io/s/my-app-m9s63?fontsize=14&hidenavigation=1&theme=dark

Disallow:
```hbs
<foo @name />
<foo-bar @name="1" />`
```

Allow:
```hbs
{{#let (component "bar" as |foo|}}
  <foo @name="1" />
{{/let}}

<foo.some.name @name="1"/>
<@foo @name="2" />
<@foo.bar @name="2" />
```

rwjblue:
Seems good, lets do it!

jenweber:
I also opened an issue in Glimmer https://github.com/glimmerjs/glimmer.js/issues/226 to improve the error message.

lifeart:
landed in https://github.com/ember-template-lint/ember-template-lint/pull/927

PR #927, opened 2019-11-14
Add `no-arguments-for-html-elements` rule

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/issues/926
//cc @jenweber, @rwjblue


lifeart:
@rwjblue should we split "checks" to different rules or figure out new rule name to cover all cases? 
Or just remove some "checks"..

Alonski:
@lifeart Can the name of this PR be changed to something like:
`Feature: No arguments for html elements`?

test/unit/rules/lint-no-arguments-for-html-elements-test.js
```diff
@@ -0,0 +1,80 @@
...
+    <:slot @name="Header"></:slot>
+  </MyComponent>`,
+    '<@foo.bar @name="2" />',
+    '<@foo @name="2" />',
+    '<foo.some.name @name="1"/>',
+  ],
+
+  bad: [
+    {
+      template: '<div as |a|></div>',
```
Turbo87: this looks unrelated to a rule called `no-arguments-for-html-elements`. IMHO we should extract that into a `no-block-params-for-html-elements` rule.

test/unit/rules/lint-no-arguments-for-html-elements-test.js
```diff
@@ -0,0 +1,80 @@
...
+      result: {
+        message: ERROR_MESSAGE_3.replace('%', 'a'),
+        moduleId: 'layout.hbs',
+        source: '<div as |a|></div>',
+        line: 1,
+        column: 0,
+      },
+    },
+    {
+      template: '<div name=@value></div>',
```
Turbo87: same as above, this should probably be extracted to a dedicated rule, because

1. it would be useful for component invocations too
2. there might be cases where you actually want to pass a `@`-prefixed string as the attribute and in that case it should be possible to disable the rule

lifeart: any suggestions how this roule could be named? (require-mustache-wrap-variables) ?

 for prefixed string it should be `name="@" or name = '@' ` and we can check it using `sourceForNode` method + indexOf

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,97 @@
+## no-arguments-for-html-elements
```
bmish: Can you mention how "angle bracket components" use the `@argument` style and that's why users are more likely to start making this mistake?

There's also a rule documentation template FYI: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/_TEMPLATE_.md

lib/rules/lint-no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'Argument "%" won\'t work for HTML elements ("<%>").';
```
bmish: Instead of creating this custom string interpolation scheme with the percent signs, you should just use and export functions that look like:

```
function makeErrorMessage(argumentName) {
  return `...${argumetName}`;
}
```

test/unit/rules/lint-no-arguments-for-html-elements-test.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const ERROR_MESSAGE = require('../../../lib/rules/lint-no-arguments-for-html-elements')
+  .ERROR_MESSAGE;
+const ERROR_MESSAGE_2 = require('../../../lib/rules/lint-no-arguments-for-html-elements')
+  .ERROR_MESSAGE_2;
+const ERROR_MESSAGE_3 = require('../../../lib/rules/lint-no-arguments-for-html-elements')
```
bmish: No need to require the same file several times. Just do:
```
const rule = require(...);
const { ERROR_MESSAGE_1, ERROR_MESSAGE_2, ... } = rule; 
```

test/unit/rules/lint-no-arguments-for-html-elements-test.js
Turbo87: maybe `no-at-prefixed-string-values` or something like that

I'm not sure I understand the second question 🤔

lifeart: 2.) solved using this approach -> https://github.com/ember-template-lint/ember-template-lint/pull/927/files#diff-407e0d43d58a271ba8e91f1876ea034fR37

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+    at CompilableBlock.compile
+    at Object.evaluate
+    at AppendOpcodes.evaluate
+    at LowLevelVM.evaluateSyscall
+    at LowLevelVM.evaluateInner
+    at LowLevelVM.evaluateOuter
+    at VM.next
+```
+
+This is creepy and untracable (for end-users).
```
MelSumner:
```suggestion
This is untraceable for end-users.
```

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+Using this linting rule we can guess (is it component or not) and emit error if argument assigned to lowercased element without same block name and without dots in notation.
```
MelSumner:
```suggestion
This linting rule determines if it is a component or not and emits an error if argument is assigned to a lower-case element without the same block name and without dots in notation.
```

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+
+This rule also emit error if we trying to specify "path-like" string as attribute argument.
```
MelSumner:
```suggestion
This rule will also emit an error for these other conditions: 

- trying to specify "path-like" string as attribute argument
- trying to use an HTML element with block params

```

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+Also, error will be thrown if we trying to use HTML element with block-params.
```
MelSumner:
```suggestion
```

Added this line to the bullet points above

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+
+```hbs
+<foo.some.name @name="1"/>
+<@foo @name="2" />
+<@foo.bar @name="2" />
+```
+
+```hbs
+<MyComponent>
+  <:slot @name="Header"></:slot>
```
MelSumner: is this legit valid code?

lib/rules/lint-no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGES = {
+  1: 'Argument "%" won\'t work for HTML elements ("<%>").',
+  2: 'Attribute value "%" won\'t work as expected, you need to wrap it as "%".',
+  3: 'Block param "%" won\'t work for HTML element.',
```
MelSumner:
```suggestion
  3: 'Block params ("%") should not be used on HTML elements.',
```

lib/rules/lint-no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGES = {
+  1: 'Argument "%" won\'t work for HTML elements ("<%>").',
```
MelSumner:
```suggestion
  1: 'Arguments ("%") should not be used on HTML elements ("<%>").',
```

lib/rules/lint-no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,103 @@
...
+  2: 'Attribute value "%" won\'t work as expected, you need to wrap it as "%".',
```
MelSumner: I'm not sure what you're trying to say here?

docs/rule/no-arguments-for-html-elements.md
```diff
@@ -0,0 +1,100 @@
...
+</MyComponent>
+```
```
MelSumner: You're missing some of the template.md sections here- having a references section is important at the very least.

lifeart: @MelSumner could you mention some references? It's removed because I can't figure out matched refs.

docs/rule/no-arguments-for-html-elements.md
lifeart: yes, this is "slots" (already available in glimmer-vm)

lib/rules/lint-no-arguments-for-html-elements.js
lifeart: @MelSumner in handlebars AST this code is equal:

`<div name=@foo />` and `<div name="@foo" />`, but first one have no sence (likely an error), and second one is "string param @foo"

lifeart: https://github.com/ember-template-lint/ember-template-lint/pull/927#discussion_r349895000

lib/rules/no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,103 @@
...
+      }, []);
+      if (!safeValues.includes(originalNode)) {
+        log({
+          message: makeError(2, `${value.chars}`, `${name}={{${value.chars}}}`),
+          line: value.loc && value.loc.start.line,
+          column: value.loc && value.loc.start.column,
+          source: sourceForNode(value),
+        });
+      }
+    }
```
rwjblue, line 49: I don't fully understand the point of this set of checks, can you explain?

Note, we already have [no-quoteless-attributes](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-quoteless-attributes.md) in the recommended rule set.

lib/rules/no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,103 @@
...
+    const sourceForNode = this.sourceForNode.bind(this);
+
+    function looksLikeHTMLElement(node) {
+      if (isLocal(node)) {
+        return false;
+      }
+      const tagName = node.tag;
+      if (tagName.startsWith('@') || tagName.startsWith(':') || tagName.includes('.')) {
+        return false;
+      }
```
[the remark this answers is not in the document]
lifeart: is angle-baracket component return false for `<foo.some.name />`

lib/rules/no-arguments-for-html-elements.js
lifeart: good catch! removed

docs/rule/no-arguments-for-html-elements.md
fusion2004: :O Is there documentation on how to use or examples of this? Or is this not recommended to use in production apps yet?

test/unit/rules/lint-no-arguments-for-html-elements-test.js
Turbo87: gotta say I'm surprised that this got merged without splitting it into several rules, especially now that it's planned to be included in the recommended ruleset...

rwjblue: hmmm, ya I must have missed this. Good catch, we should absolutely split it out.

rwjblue: Made https://github.com/ember-template-lint/ember-template-lint/issues/1451 to track, and added to v3.0.0 milestone.

PR #930, opened 2019-11-18
updated `require-valid-alt-text` to check for common a11y failure #39

MelSumner:
If merged, this PR would update the require-valid-alt-text lint rule to check for common failure 39 - the `alt` tag should exist and be empty if `role="none"` or `role="presentation"` exists.


MelSumner:
- added documentation updates
- added good tests

docs/rule/require-valid-alt-text.md
```diff
@@ -28,16 +27,18 @@ This rule **forbids** the following:
...
 <img src="b52.jpg" alt="52" />
+<img src="foo" alt="foo as a banana" role="presentation">
 ```
 
 ```hbs
 <img src="rwjblue.png" alt="A man standing in front of a room of people, giving a presentation about Ember.">
 <img src="foo" alt="YourCompany Home Page" />
-<img src="bar" aria-hidden alt="Picture of me taking a photo of an image" /> // Will pass because it is hidden.
```
rwjblue: Hehe, we really should have a rule to enforce this one too 😝 (no-invalid-aria-attributes :wink: ).

rwjblue:
OK, I think the primary remaining question is:

> This implementation only adds the requirement to img, shouldn't it apply to other element types?

MelSumner:
Agree- I will need to test them and see, they aren't commonly used so I don't know off the top of my head.

docs/rule/require-valid-alt-text.md
MelSumner: Yes, probably should :) I'll write a new issue.

MelSumner:
@rwjblue since the procedure for the test specifically indicates `` elements, I'd like to stick to that for now, and add more later if it becomes clear that there are additional HTML elements that this common failure would also apply to.

PR #931, opened 2019-11-18
Add new rule: `no-invalid-role`

MelSumner:
Started a new rule to check for known invalid element/role combinations. 

The first one added is to prevent common A11y failure 92- use of role presentation/none on elements that have semantic meaning.


docs/rule/no-invalid-role.md
```diff
@@ -0,0 +1,42 @@
+## no-invalid-role
+
+Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content.
```
rwjblue: Since this rule is a bit more generic than checking `role="presentation"`, can you tweak the prose here to discuss what `role` is for instead? Then we can add sub-headers for the common types of errors that the rule checks for (which ATM would mostly be only one thing, but in the future will grow).

docs/rule/no-invalid-role.md
```diff
@@ -0,0 +1,42 @@
...
+<span role="none"></span>
+```
+
+### Migration
+
+* If violations are found, remediation should be planned to replace the semantic HTML with the `div` element. Additional CSS will likely be required.
+
+### References
+
+* [HTML Semantic Elements](https://html.spec.whatwg.org/#semantics-2)
```
rwjblue: This link does not describe "semantic elements" it discusses "an elements semantics", I know that seems like a semantic argument (😝) but IMHO it's quite an important distinction.

docs/rules.md
```diff
@@ -59,3 +60,7 @@
 * [deprecated-each-syntax](rule/deprecations/deprecated-each-syntax.md)
 * [deprecated-inline-view-helper](rule/deprecations/deprecated-inline-view-helper.md)
 * [deprecated-render-helper](rule/deprecations/deprecated-render-helper.md)
+ 
+## New Rules
```
rwjblue: Why is this added?

lib/rules/lint-no-invalid-role.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+function ERROR_MESSAGE_INVALID_ROLE(element) {
+  return `Use of presentation role on <${element}> detected. Semantic elements should not be used for presentation.`;
+}
+
+const ALLOWED_ELEMENTS = ['div', 'img', 'span', 'svg'];
```
rwjblue: I'd like a spec link to explain why these specific elements are "allowed". I didn't see that in any of the links in the rule documentation (though maybe I just missed it).

lib/rules/lint-no-invalid-role.js
```diff
@@ -0,0 +1,44 @@
...
+          if (
+            roleValue
+              .trim()
+              .toLowerCase()
+              .includes('presentation') ||
+            roleValue
+              .trim()
+              .toLowerCase()
+              .includes('none')
+          ) {
```
rwjblue: I think it is probably easier / simpler to do:

```js
const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');

if (['presentation', 'none'].includes(roleValue)) {
```

lib/rules/lint-no-invalid-role.js
MelSumner: Sure, I'll look for some explicit links- my plan was to start a rule that covers common failure [92](https://www.w3.org/WAI/WCAG21/Techniques/failures/F92) with ones I already know for sure, and then expand from there.

rwjblue: Hmm, but the implementation is quite different. The current implementation prevents using `presentation` for _any_ element other than this explicit list. That's a **ton** of elements (including ones that are custom defined by users e.g. `<my-awesome-thing></my-awesome-thing>` which should probably be treated like a `div`), and it makes me worried 😟

MelSumner: @rwjblue so I do think that I should reverse this and do the other way that we talked about this morning, even though the list is longer, it _is_ a complete list of semantic elements. That way custom components wouldn't get caught up in the mix. What do you think?

rwjblue:
Looks like this needs a rebase

lib/rules/lint-no-invalid-role.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
rwjblue: Need to rename the file now that we forbid `lint-` in the name

lib/rules/lint-no-invalid-role.js
rwjblue: Yes, I think that would be much less risky.

MelSumner: Going to close this issue and write it the other way.

PR #932, opened 2019-11-18
Move main linter engine to stand alone file.

dcyriller:
Please note that the target branch is `next`.


lib/index.js
```diff
@@ -1,239 +1,8 @@
...
-}
-
-module.exports = Linter;
+module.exports = require('./linter');
 module.exports.Rule = require('./rules/base');
 module.exports.ASTHelpers = require('./helpers/ast-node-info');
 
-module.exports.WARNING_SEVERITY = WARNING_SEVERITY;
-module.exports.ERROR_SEVERITY = ERROR_SEVERITY;
+module.exports.WARNING_SEVERITY = require('./linter').WARNING_SEVERITY;
```
rwjblue: FWIW, this mutates the `Linter` class, so we might as well make these things static fields in the Linter file...

lib/linter.js
```diff
@@ -0,0 +1,234 @@
...
+    if (options.verbose) {
+      message += `\n${error.source}`;
+    }
+
+    return message;
+  }
+}
+
+module.exports = Linter;
+module.exports.WARNING_SEVERITY = WARNING_SEVERITY;
```
rwjblue: If you make these as static class fields you don't have to export them here or in lib/index.js

lib/index.js
dcyriller: Interesting 👍.

I've updated the PR with your suggestion.

I wondered what to do with these constants. Having them as static class fields looks good.

dcyriller: Unfortunately, static class fields are not supported yet in eslint itself (because it is in stage 3?). So I have to use `babel-eslint` as an eslint parser. Does it look okay?

dcyriller: Using static class fields actually requires to add `babel` to the project 😫.

What I've done is refactoring the exports not to mutate the `Linter`. This is explained in the commits. (I'll clean up the history.)

rwjblue: eeeck, I was thinking:

```
class Foo {
  static get WARNING_SEVERITY() {
    return WARNING_SEVERITY;
  }
}
```

which should work without transpilation

lib/index.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-module.exports = require('./linter');
+module.exports = require('./linter').Linter;
```
rwjblue: The lines just below still mutate the class.

It isn't a big deal though, I would just like to fix it in the long run. Fine for a follow up...

lib/index.js
dcyriller: okayy, ah sorry, I will follow up

PR #933, opened 2019-11-18
Linter class: Use named severity

dcyriller:
In the main Linter class, switch : `2` for `ERROR_SEVERITY` variable.

I'm targetting `next` branch. This is a https://github.com/ember-template-lint/ember-template-lint/pull/932 follow-up.


rwjblue:
Changes generally seem good, we can merge after #932

Issue #942, opened 2019-11-21
add `a11y-strict` config


MelSumner:
Add a config that turns on all of the A11y-related linting rules.

PR #946, opened 2019-11-22
Add new comprehensive presets (`all` and `a11y-strict`)

MelSumner:
If merged, this PR adds two presets: 

- all 
- a11y-strict 

Addresses #942


rwjblue:
Needs a few tweaks:

* reexport from `lib/config/index.js`
* add _something_ that explains our semver commitments for these presets, specifically that their public API is to include all related rules and as new rules are added they _will_ be added to these presets (IOW we don't consider adding rules here to be a breaking change)

bmish:
Would be a good time to add a nice list of configs to the README.

Turbo87:
I think `a11y-strict` is a good idea, but I'm not convinced about the `all` preset, because it suggests that all of those rules might be suitable for everyone. 🤔

Turbo87:
I just realized, another issue with `all` is that if we consider it part of our public "API" then adding basically any new rule would be a breaking change from now on :-/

bmish:
There are other eslint plugins that have an `all` configuration (such as [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin)) and they explicitly say in their README that it's not recommended for production use because new rules can be added any time. So that isn't a problem as long as we document it.

Turbo87:
@bmish sure, but then I'm curious what reasons there are for adding it at all? 🤔

dcyriller:
A solution to that could be to extract a11y-related rules into an external plugin. This way the plugin could have its own semver versioning.

Turbo87:
@dcyriller just to be clear, I'm okay with the a11y preset, just worried about the "all" preset

dcyriller:
🙈 off topic, I read `all` as `all a11y rules` which is `a11y-strict`.

MelSumner:
> A solution to that could be to extract a11y-related rules into an external plugin. This way the plugin could have its own semver versioning.

I would prefer to not do this, since the goal has been to write rules in a useful ways so that they are not simply about accessibility but well-crafted code as well. There are only a few that are specific to accessibility success criteria. Does that make sense?

MelSumner:
@Turbo87 @bmish @rwjblue so should I remove the "all" preset?

Turbo87:
> so should I remove the "all" preset?

IMHO yes :)

Turbo87:
or at least split it from this PR and then we can discuss it in a dedicated PR.

I think the a11y preset is ready to be merged once the `all` preset is removed from this PR :)

bmish:
I don’t have a strong opinion on the `all` preset. I’m fine with it if there’s a good reason to add it (but we should be careful since additional presets means additional maintenance burden for us).

A separate PR would be preferred.

MelSumner:
Ok, I'm going to close this PR and do something a little different.

PR #947, opened 2019-11-22
Add tests confirming extending multiple configs works properly.

### LINKED ISSUE #924: New Feature: allow multiple presets
Allow multiple presets to be included, so we can make multiple recommended sets. 

Example: I'd like to make a a11y-strict preset, which will check for a lot more things and is a lot more strict about how things are done. 

This will also require tags/meta data to be added to each rule, to define which preset it belongs to.

rwjblue:
Co-Authored-By: @melsumner

Closes #924


dcyriller:
Great! I thought extending multiple sets was not supported yet.

Issue #867, opened 2019-10-21
New Rule: require-button-type


Turbo87:
It is very easy to forget `type="button"` on `<button>` elements and accidentally create a submit button. I think a rule to enforce the usage of an explicit `type` attribute on buttons might make sense.

@rwjblue thoughts?

rwjblue:
Sure, seems good to me. Some questions:

* Should it be a default rule (when we bump majors)?
* What do we link to as supporting references?

Turbo87:
I probably wouldn't make it a default rule. It's fine to skip `type` if you know what you're doing. But if we decide to make it a default I'd be fine with that too.

I don't know about any references, but https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button explains the `type` attribute quite well IMHO.

lifeart:
ref: https://github.com/emberjs/ember.js/issues/18232

should this lint rule pass if `...attributes` passed?

Turbo87:
not sure, I think it probably should still set an explicit default value.

johanrd:
@rwjblue @Turbo87: The `require-button-type` rule makes sense as a safeguard for buttons that are inside/associated with a `form`, but for buttons that are not inside/associated with a form, this rule leads to a lot of clutter. 

What do you think about changing the default to something like `require-button-type-inside-forms`, instead of requiring a button type for all buttons?

Turbo87:
how do you know if your button in the component is used in a form or not?

johanrd:
Not sure, but I guss inspiration could be drawn from the implementation of `no-nested-interactive`?

Turbo87:
the problem is that this only works within single template files, but not across component boundaries. I'd rather be safe and just add it to all buttons instead of having to think about it all the time.

johanrd:
OK, I understand.

A sensible approach in my case could then be to disable the global `require-button-type`, and instead add an instruction manually before every relevant `form` tag:

```hbs
{{!-- template-lint-enable require-button-type --}}
<form>
```

…I guess it is a matter of taste whether `require-button-type` should be `true` or `false` in the default config. Since the rule goes beyond its intended scope, I'd lean towards defaulting it to `false`, but I guess it may also be a good trade-off for most people to just write `type="button"` on all buttons.

PR #950, opened 2019-11-23
Implement `require-button-type` rule

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/867

### DIFF
--- a/lib/rules/lint-require-button-type.js
+++ b/lib/rules/lint-require-button-type.js
@@ -1,0 +1,45 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'All `<button>` elements should have a valid `type` attribute';
+
+module.exports = class RequireButtonType extends Rule {
+  logNode({ node, message }) {
+    return this.log({
+      message,
+      line: node.loc && node.loc.start.line,
+      column: node.loc && node.loc.start.column,
+      source: this.sourceForNode(node),
+    });
+  }
+
+  visitor() {
+    return {
+      ElementNode(node) {
+        let { tag, attributes } = node;
+
+        if (tag !== 'button') {
+          return;
+        }
+
+        let typeAttribute = attributes.find(it => it.name === 'type');
+        if (!typeAttribute) {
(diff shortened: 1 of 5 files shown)


test/unit/rules/lint-require-button-type-test.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'require-button-type',
+
+  config: true,
+
+  good: [
```
bmish: Should add valid test cases to ensure this doesn't affect other elements:

```
<div></div>
<div type="randomType"></div>
```

lib/rules/lint-require-button-type.js
```diff
@@ -0,0 +1,45 @@
...
+        let { chars } = value;
+        if (!['button', 'submit', 'reset'].includes(chars)) {
+          return this.logNode({ node, message: ERROR_MESSAGE });
+        }
+      },
+    };
+  }
+};
+
+module.exports.ERROR_MESSAGE = ERROR_MESSAGE;
```
bmish: You're exporting the error message but you aren't actually importing it from the test file. Should update the test cases to use this.

test/unit/rules/lint-require-button-type-test.js
Turbo87: good idea, thanks :)

lib/rules/lint-require-button-type.js
Turbo87: done ✅

lib/rules/lint-require-button-type.js
```diff
@@ -0,0 +1,45 @@
...
```
dcyriller: What do you think of using a static class field to export ERROR_MESSAGE?

Turbo87: I think that's essentially what the code is currently doing. I don't think static class fields are supported by all of our currently supported Node versions though.

PR #952, opened 2019-11-23
Add GitHub Actions CI workflow

Turbo87:
because it's usually quite a bit faster and more powerful :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,102 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - 'v*'
+  pull_request: {}
+  schedule:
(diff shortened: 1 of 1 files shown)


bmish:
How does this work / how was this file generated?

Turbo87:
I manually created it based on the contents of the TravisCI config

bmish:
Will this eventually completely replace Travis for running our tests?

Turbo87:
yes, that would be the plan

bmish:
Curious to try it! Too bad it looks quite a bit lengthier than our .travis.yml.

Turbo87:
yeah, most of it is due to the verbose caching implementation, but it's the recommended solution for yarn that I mostly just copied.

PR #954, opened 2019-11-24
Ignoring existing pending modules that have no linting errors

gmurphey:
Was changing my rule configuration and running `ember-template-lint . --print-pending` to generate an updated pending list of files and noticed that there were some entries that looked like this:

```
{
 "moduleId": "my/path/to/a/template",
 "only": [
    null
  ]
},
```

Debugging led me to realize these were being added to the output for `--print-pending` due to an error being thrown during linting saying `Pending module ('my/path/to/a/template') passes all rules. Please remove 'my/path/to/a/template' from pending list.`

This PR adds a check when building the output for `--print-pending` to make sure that the error thrown by the linter has a rule.

Thanks for this great tool!

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -116,7 +116,13 @@ function run() {
 
     if (printPending) {
       let failingRules = Array.from(
-        fileErrors.reduce((memo, error) => memo.add(error.rule), new Set())
+        fileErrors.reduce((memo, error) => {
+          if (error.rule) {
+            memo.add(error.rule);
+          }
+
+          return memo;
+        }, new Set())
       );
 
       if (failingRules.length > 0) {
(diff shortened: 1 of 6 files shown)


bin/ember-template-lint.js
```diff
@@ -166,7 +166,13 @@ function run() {
 
     if (printPending) {
       let failingRules = Array.from(
-        fileErrors.reduce((memo, error) => memo.add(error.rule), new Set())
+        fileErrors.reduce((memo, error) => {
+          if (error.rule) {
```
rwjblue: Instead of making this kindof implicit, can we add a special `.code` / `.rule` value that we can check for?

I'm thinking we can tweak this:

https://github.com/ember-template-lint/ember-template-lint/blob/549775f2bc09eb9b7822ed0272ac67460515bc7f/lib/index.js#L174-L178

And add a `rule: 'invalid-pending-module'` or something.

What do you think?

rwjblue:
Friendly ping @gmurphey 😸

bin/ember-template-lint.js
gmurphey: Great suggestion! Should be more clear now.

Issue #949, opened 2019-11-22
Cleanup: remove `lint-` prefix from rules. 


MelSumner:
We know it's linting, we don't need the prefix.

rwjblue:
:thumbsup: this has been annoying for a long time :smile:

Turbo87:
yes, please! 😅

Turbo87:
resolved by #957

PR #957, opened 2019-11-25
removed the `lint-` prefix and updated related files so tests pass

MelSumner:
If merged, this PR:
- removes the `lint-` prefix from all of the rules (including deprecated ones)
- updates references to the rules

Addresses #949


test/unit/rules/no-element-event-actions-test.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const generateRuleTests = require('../../helpers/rule-test-harness');
-const ERROR_MESSAGE = require('../../../lib/rules/lint-no-element-event-actions').ERROR_MESSAGE;
+const ERROR_MESSAGE = require('../../../lib/rules/no-element-event-actions').ERROR_MESSAGE;
```
Turbo87: this seems wrong 🤔

MelSumner: how so?

Turbo87: because `../../../lib/rules/no-element-event-actions../../../lib/rules/no-element-event-actions` looks like something in your find-and-replace went wrong 😉

Turbo87: but now I'm confused because the notification email showed:

```
-const ERROR_MESSAGE = require('../../../lib/rules/lint-no-element-event-actions').ERROR_MESSAGE;
+const ERROR_MESSAGE = require('../../../lib/rules/no-element-event-actions').ERROR_MESSAGE;
```

is GitHub displaying broken diffs to me right now?! 🤔

MelSumner: I didn't use f&r because I was trying to avoid screwing up- but anyway here's what I see, maybe we can figure it out this way: 
image

Which seems accurate (and the tests suggest the same, since they are passing). 

Is there something else I am missing?

Turbo87: I think I figured it out. I'm using https://addons.mozilla.org/de/firefox/addon/octolinker/ and it seems that their latest version has a bug which sometimes duplicates text... 😱

bmish:
@rwjblue we do have a test for that already in ` test/unit/base-plugin-test.js` and she updated it.

PR #959, opened 2019-11-26
Add validation that all files in `lib/config/` have been reexported in `lib/config/index.js`

### LINKED ISSUE #948: Add validation that all files in `lib/config/` have been reexported in `lib/config/index.js`
The current implementation was added in https://github.com/ember-template-lint/ember-template-lint/pull/798, we need to add `lib/config/index.js` validation to that setup. This is basically the same as the validation that we do for adding a new rule, but for configs.

lifeart:
Resolves #948

/cc @rwjblue


rwjblue:
Thank you!

PR #964, opened 2019-11-29
Implement `no-passed-in-event-handlers` rule

### LINKED ISSUE #962: New Rule: `no-passed-in-event-handlers`
It is possible to pass e.g. `@click` to an Ember component to override the default `click` event handler. For tagless components this will trigger an assertion though and can't be used as legitimate API, and for Glimmer components it will not work out of the box like in Ember components either.

I would like to propose a new rule, `no-passed-in-event-handler`, that scans potential component invocations for these patterns and flags them as issues:

- `<Foo @click={{this.handleClick}} />`
- `{{foo click=this.handleClick}}`

Turbo87:
Resolves #962 🎉

### DIFF
--- a/docs/rule/no-passed-in-event-handlers.md
+++ b/docs/rule/no-passed-in-event-handlers.md
@@ -1,0 +1,47 @@
+## no-passed-in-event-handlers
+
+It is possible to pass e.g. `@click` to an Ember component to override the
+default `click` event handler. For tagless components this will trigger an
+assertion though and can't be used as legitimate API, and for Glimmer
+components it will not work out of the box, like in Ember components, either.
+
+This rule scans potential component invocations for these patterns and flags
+them as issues.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+<Foo @click={{this.handleClick}} />
+```
+
+```hbs
+<Foo @keyPress={{this.handleClick}} />
+```
+
+```hbs
+{{foo click=this.handleClick}}
+```
+
+This rule **allows** the following:
+
+```hbs
+<Foo @onClick={{this.handleClick}} />
+```
+
+```hbs
+<Foo @myCustomClickHandler={{this.handleClick}} />
+```
+
+```hbs
+<Foo @onKeyPress={{this.handleClick}} />
+```
+
+```hbs
+{{foo onClick=this.handleClick}}
+```
+
+### Migration
+
+* create explicit component APIs for these events (e.g. `@click` -> `@onClick`)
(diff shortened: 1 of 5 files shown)


docs/rule/no-passed-in-event-handlers.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+```hbs
+{{foo click=this.handleClick}}
+```
+
+This rule **allows** the following:
+
+```hbs
+<Foo @onClick={{this.handleClick}} />
```
bmish: I'm worried that this example could make `onClick` seem like a real Ember API. It may help to either add a comment explaining that it's a custom name or use a name like `myCustomClickHandler={{this.handleClick}}`.

lib/rules/no-passed-in-event-handlers.js
```diff
@@ -0,0 +1,86 @@
...
+            });
+          }
+        }
+      },
+    };
+  }
+};
+
+function makeErrorMessage(eventHandler) {
+  return `\`${eventHandler}\` should not be used as a component argument`;
```
bmish: I would update this to be more clear that the reason this component argument is disallowed is because it's an event handler: 
```
DOM events including `${eventHandler}` should not be used as a component argument
```

test/unit/rules/no-passed-in-event-handlers-test.js
```diff
@@ -0,0 +1,85 @@
...
+generateRuleTests({
+  name: 'no-passed-in-event-handlers',
+
+  config: 'true',
+  good: [
+    '<Foo />',
+    '<Foo @onClick={{this.handleClick}} />',
+    '<Foo @onclick={{this.handleClick}} />',
+    '<Foo @Click={{this.handleClick}} />',
+    '<Foo @touch={{this.handleClick}} />',
```
bmish: Perhaps add an example with a literal argument value like `@random=true`.

Turbo87:
> Wouldn't `no-dom-event-component-arguments` be more accurate as the rule name?

I've taken the naming "event handlers" from https://api.emberjs.com/ember/release/classes/Component#event-handler-methods. I'm open to renaming it to `no-passed-in-event-handler-methods` if you think that that's better 🤔

docs/rule/no-passed-in-event-handlers.md
Turbo87: good point, I've added your `myCustomClickHandler` example

lib/rules/no-passed-in-event-handlers.js
Turbo87: yep, I've adjusted the warning message

test/unit/rules/no-passed-in-event-handlers-test.js
Turbo87: done

bmish:
The rule name makes more sense to me now that you shared that article. Could you link to the article in a `### References` section in the documentation?

PR #966, opened 2019-11-30
Report errors as GitHub Actions Annotations

Turbo87:
Bildschirmfoto 2019-11-30 um 10 52 10

more info available at https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-error-message-error

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -206,6 +206,8 @@ function run() {
   if (Object.keys(errors).length) {
     printErrors(errors, options);
   }
+
+  linter.reportGitHubActionAnnotations(errors);
 }
 
 // exports are for easier unit testing
--- a/lib/index.js
+++ b/lib/index.js
@@ -228,6 +228,52 @@ class Linter {
 
     return message;
   }
+
+  reportGitHubActionAnnotations(errors) {
+    if (!process.env.GITHUB_ACTIONS || process.env.DISABLE_GITHUB_ACTIONS_ANNOTATIONS) {
+      return;
+    }
+
+    let files = Object.keys(errors);
+    for (let file of files) {
+      let fileErrors = errors[file];
+      for (let error of fileErrors) {
+        let line = error.line;
+        let col = error.column;
+
+        let annotation = Linter._formatGitHubActionAnnotation(error.message, { file, line, col });
+        this.console.log(annotation);
+      }
+    }
+  }
+
+  static _formatGitHubActionAnnotation(message, options = {}) {
+    message = message || '';
+
+    let output = '::error';
+
+    let outputOptions = Object.keys(options)
+      .map(key => `${key}=${escape(String(options[key]))}`)
+      .join(',');
+
+    if (outputOptions) {
+      output += ` ${outputOptions}`;
+    }
+
+    return `${output}::${escapeData(message)}`;
+  }
+}
+
+function escapeData(s) {
+  return s.replace(/\r/g, '%0D').replace(/\n/g, '%0A');
+}
+
+function escape(s) {
+  return s
+    .replace(/\r/g, '%0D')
+    .replace(/\n/g, '%0A')
+    .replace(/]/g, '%5D')
+    .replace(/;/g, '%3B');
 }
 
 module.exports = Linter;
--- a/test/unit/bin/ember-template-lint-test.js
+++ b/test/unit/bin/ember-template-lint-test.js
@@ -127,6 +127,17 @@ describe('ember-template-lint executable', function() {
   });
 
   describe('errors and warnings formatting', function() {
+    let oldValue;
+
+    beforeEach(function() {
+      oldValue = process.env.GITHUB_ACTIONS;
+      delete process.env.GITHUB_ACTIONS;
+    });
+
+    afterEach(function() {
+      process.env.GITHUB_ACTIONS = oldValue;
+    });
+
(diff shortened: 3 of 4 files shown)


dcyriller:
Great addition for GH-hosted projects!

bin/ember-template-lint.js
```diff
@@ -57,6 +57,20 @@ function printErrors(errors, invocationOptions) {
       );
     }
   }
+
+  if (process.env.GITHUB_ACTIONS) {
+    let { issueCommand } = require('@actions/core/lib/command');
```
rwjblue: I'm a bit worried about requiring deeply like this. 🤔 

https://github.com/actions/toolkit/blob/1c12ced7ba4b5f2d21da0ddfea83b4885505503f/packages/core/src/command.ts#L3

Does not help assuage my concern...

Turbo87: indeed, but unfortunately there is no other way to include file metadata in the message 😞 

if we don't use this lib then we'll have to construct the `::error file={name},line={line},col={col}::{message}` message ourselves 🤔

rwjblue: It seems like they support emitting the console output directly as public API [see here](https://github.com/actions/toolkit/blob/master/docs/commands.md), though they don't document the specific output that `issueCommand` emits here.

I guess I'm happy to do it this way for now, but I would like to see if they can either export `issueCommand` as public API or document the `::error` syntax ([on this page](https://github.com/actions/toolkit/blob/master/docs/commands.md)) so we can just use `console.log`.

rwjblue: 🤦‍♂ I totally missed your link in the PR description which shows public documentation of the format.

@Turbo87 - Can you update this to drop the additional dependency, and directly `console.log` the appropriate format when `process.env.GITHUB_ACTIONS` is set?

Turbo87: yep, can do. but I think I remember that their tool did some escaping, which we would probably have to do too then.

rwjblue: Ya, I think since we know the line, col, and filename will not have any of the special chars they escape ([see here](https://github.com/actions/toolkit/blob/1c12ced7ba4b5f2d21da0ddfea83b4885505503f/packages/core/src/command.ts#L81-L87)) we don't have to worry about those ones, but we will need to escape any newlines in the message part.

I think the result would be something like:

```js
let message = error.message.replace(/\r/g, '%0D').replace(/\n/g, '%0A');

console.log(`::error file={${file}},line={${line}},col={${col}}::${message}`);
```

Turbo87: done

lib/index.js
```diff
@@ -228,6 +228,52 @@ class Linter {
 
     return message;
   }
+
+  reportGitHubActionAnnotations(errors) {
+    if (!process.env.GITHUB_ACTIONS || process.env.DISABLE_GITHUB_ACTIONS_ANNOTATIONS) {
```
Turbo87: @runspired I included `DISABLE_GITHUB_ACTIONS_ANNOTATIONS` just for you 😉

runspired: Thanks!

test/unit/bin/ember-template-lint-test.js
```diff
@@ -127,6 +127,17 @@ describe('ember-template-lint executable', function() {
...
   describe('errors and warnings formatting', function() {
+    let oldValue;
+
+    beforeEach(function() {
+      oldValue = process.env.GITHUB_ACTIONS;
+      delete process.env.GITHUB_ACTIONS;
+    });
+
+    afterEach(function() {
+      process.env.GITHUB_ACTIONS = oldValue;
```
rwjblue: If `process.env.GITHUB_ACTIONS` was originally not defined, this will define it and make it truthy. 😱

Screen Shot 2019-12-05 at 4 29 08 PM


---

I think we need to do something like:

```
// in module scope, eagerly
const isGithubActions = process.env.GITHUB_ACTIONS;

// ...snip...
beforeEach(function() {
  delete process.env.GITHUB_ACTIONS;
});

afterEach(function() {
  if (isGithubActions) {
    process.env.GITHUB_ACTIONS = 'true';
  } else {
    delete process.env.GITHUB_ACTIONS;
  }
});
```

test/unit/bin/ember-template-lint-test.js
```diff
@@ -380,6 +391,41 @@ describe('ember-template-lint executable', function() {
...
+      let oldValue;
+
+      beforeEach(function() {
+        oldValue = process.env.GITHUB_ACTIONS;
+        process.env.GITHUB_ACTIONS = 'true';
+      });
+
+      afterEach(function() {
+        process.env.GITHUB_ACTIONS = oldValue;
+      });
```
rwjblue: Same comment/issue as above

test/unit/bin/ember-template-lint-test.js
Turbo87: done, I've extracted a reusable `setupEnvVar()` test helper function :)

rwjblue:
Thank you @Turbo87!

runspired:
Indeed thanks! Especially thanks for making it opt-out. I'm sure this will be very useful to many, and hopefully the UI/UX for these annotations gets improved by github such that I get happy using them, at which time I'll be excited to have this and turn it on :D

PR #968, opened 2019-11-30
Replace Mocha with Jest

Turbo87:
Looks like I've tried this before in #364 but didn't succeed due to slow CI times. Let's try this again... 😂 

Advantages:
- built-in code coverage tracking (`--coverage`)
- built-in interactive watch mode (`--watchAll`)

**Update:** seems that this version of Jest is no longer having the same issues with slow CI runs as the last time I've tried. 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
       name: "Node 10"
       script:
         - yarn lint:js
-        - yarn test
+        - yarn test --coverage
 
     - stage: additional tests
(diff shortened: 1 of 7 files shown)


bmish:
I see `mocha: true` in `.eslintrc.js` that you could remove.

Turbo87:
@bmish it's still using the mocha APIs for now, so I guess that definition is still valid 😅

PR #971, opened 2019-11-30
no-curly-component-invocation: Do not warn for MustacheStatements with positional arguments

### LINKED ISSUE #925: Disable `no-curly-component-invocation` rule in `octane` configuration
The `no-curly-component-invocation` rule in the `octane` configuration is creating a lot of false positives as it has no knowledge of what is a helper or component and in some cases it also flags regular `{{foo}}` values as components.

To avoid this unnecessary noise for Octane users we should disable the rule until we figure these issues out.

The second issue is clearly a bug, the first one can't really be solved, but we could still use a better heuristic and e.g. consider curlies with positional args as helpers.

/cc @rwjblue @lifeart

Turbo87:
Resolves #925

/cc @patocallaghan @rwjblue

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -84,6 +84,10 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
   }
 
   getCurlyInvocationSyntaxError(node) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you @turbo87!

PR #973, opened 2019-12-01
Remove TravisCI configuration file

Turbo87:
because we use GitHub Actions now and it is significantly faster :)


Turbo87:
I don't have exact numbers but for all of the PRs that I opened since then, the GitHub Actions CI was finished significantly earlier than TravisCI.

PR #977, opened 2019-12-04
updated README to include a semantic versioning policy

MelSumner:
Updated the README to add a semantic versioning policy as previously discussed. Tweaked the one we looked at as a sample; please let me know if we need to change some of the specific details to fit this project more precisely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ It is possible to share a config (`extends`) or plugin (custom rules) across pro
 
 ## Semantic Versioning Policy
 
-The semver policy for this addon can be read here: [semver policy](https://github.com/ember-template-lint/dev/versioning.md).
+The semver policy for this addon can be read here: [semver policy](dev/versioning.md).
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ var results = linter.verify({ source: template, moduleId: 'template.hbs' });
 ```
 
 `results` will be an array of objects which have the following properties:
+
 * `rule` - The name of the rule that triggered this warning/error.
 * `message` - The message that should be output.
 * `line` - The line on which the error occurred.
@@ -227,6 +228,39 @@ You can define and use your own custom rules using the plugin system. See [plugi
 ### Sharing configs
 
 It is possible to share a config (`extends`) or plugin (custom rules) across projects. See [ember-template-lint-plugin-peopleconnect](https://github.com/peopleconnectus/ember-template-lint-plugin-peopleconnect) for an example.
+
+## Semantic Versioning Policy
+
+`ember-template-lint` follows [semantic versioning](http://semver.org/).
+
+## Patch Release
+
+A patch release should not break your lint build.
+
+* A bug fix in a rule that results in it reporting fewer errors.
+* Improvements to documentation.
+* Non-user-facing changes such as refactoring code, adding, deleting, or modifying tests, and increasing test coverage.
+* Re-releasing after a failed release (i.e., publishing a release that doesn't work for anyone).
+
+### Minor Release
+
+A minor release may break your lint build in some cases.
+
+* A bug fix in a rule that results in it reporting more errors.
+* A new rule is created.
+* A new option to an existing rule is created.
+* An existing rule is deprecated.
+
+### Major Release
+
+A major release is likely to break your lint build.
+
+* A support for old Node version is dropped.
+* A support for old ember-template-lint version is dropped.
+* An existing rule is changed in it reporting more errors.
+* An existing rule is removed.
+* An existing option of a rule is removed.
+* An existing config is updated.
 
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -228,6 +229,39 @@ You can define and use your own custom rules using the plugin system. See [plugi
...
+### Major Release
+
+A major release is likely to break your lint build.
+
+* A support for old Node version is dropped.
+* A support for old ember-template-lint version is dropped.
+* An existing rule is changed in it reporting more errors.
+* An existing rule is removed.
+* An existing option of a rule is removed.
+* An existing config is updated.
```
Turbo87: seems quite verbose to me to be included in the `README.md` file. maybe we could move the majority of it into a `VERSIONING.md` or `docs/versioning.md` and only keep a reference to that file in the README?

MelSumner: Sure! 

Just to make sure, is there a less verbose way to say this?

Turbo87: just to clarify, I think the content itself is fine, just too long to be included in the README itself. :)

rwjblue: Ya, that sounds good to me. 

@MelSumner - Any objections?

MelSumner: No objections! Will do!

bmish:
Is this based on eslint's policy? I noticed they considered a policy here: https://github.com/eslint/eslint/issues/6244

README.md
```diff
@@ -228,6 +229,10 @@ You can define and use your own custom rules using the plugin system. See [plugi
 
 It is possible to share a config (`extends`) or plugin (custom rules) across projects. See [ember-template-lint-plugin-peopleconnect](https://github.com/peopleconnectus/ember-template-lint-plugin-peopleconnect) for an example.
 
+## Semantic Versioning Policy
+
+The semver policy for this addon can be read here: [semver policy](https://github.com/ember-template-lint/dev/versioning.md).
```
Turbo87:
```suggestion
The semver policy for this addon can be read here: [semver policy](dev/versioning.md).
```

dev/versioning.md
```diff
@@ -0,0 +1,30 @@
...
+* A bug fix in a rule that results in it reporting more errors.
+* A new rule is created.
+* A new option to an existing rule is created.
+* An existing rule is deprecated.
+
+### Major Release
+
+A major release is likely to break your lint build.
+
+* A support for old Node version is dropped.
```
Turbo87:
```suggestion
* Support for an old Node version is dropped.
```

dev/versioning.md
```diff
@@ -0,0 +1,30 @@
...
+* A support for old ember-template-lint version is dropped.
```
Turbo87:
```suggestion
* Support for on old ember-template-lint version is dropped.
```

but what does that actually mean? 🤔

dev/versioning.md
```diff
@@ -0,0 +1,30 @@
...
+* A new rule is created.
+* A new option to an existing rule is created.
+* An existing rule is deprecated.
+
+### Major Release
+
+A major release is likely to break your lint build.
+
+* Support for an old Node version is dropped.
+* Support for on old ember-template-lint version is dropped.
```
rwjblue: Lets remove this line

dev/versioning.md
```diff
@@ -0,0 +1,52 @@
...
+A major release is likely to break your lint build.
+
+* Support for an old Node version is dropped.
+* An existing rule is changed in it reporting more errors.
+* An existing rule is removed.
+* An existing option of a rule is removed.
+
+## Config Files
+
+For dynamically generated config files such as `a11y-strict` and `all`, the config will change as needed for any release and therefore cannot be covered by a semver guarantee.
```
rwjblue: I think this part _could_ be added when we land the first of these (over
#946) if we want to punt on this part for now.

dev/versioning.md
```diff
@@ -0,0 +1,52 @@
...
+A major release is likely to break your lint build.
+
+* Support for an old Node version is dropped.
+* An existing rule is changed in it reporting more errors.
+* An existing rule is removed.
+* An existing option of a rule is removed.
+
+## Config Files
+
+For dynamically-generated config files such as `a11y-strict` and `all`, the config will change as needed for any release and therefore cannot be covered by a semver guarantee.
```
Turbo87: I think I would prefer to include `a11y-strict` in our semver guarantee 🤔 

also, from a user perspective I wouldn't know what "dynamically-generated" means in this context 😅

Turbo87: I just noticed that we have a bit of a deadlock here.

- This PR mentions `a11y-strict` already, but #978 isn't merged yet
- #978 is blocked on not being able to document its semver guarantee because this PR isn't merged yet

I would propose to remove this sentence for now until we have a preset that matches this description.

```suggestion
```

... or we should mention `octane` here, which is apparently still considered experimental and thus not has semver guarantees yet

PR #978, opened 2019-12-04
Add `a11y-strict` preset

### LINKED ISSUE #942: add `a11y-strict` config

MelSumner:
If merged, this PR adds an `a11y-strict` preset. 

Resolves #942

### DIFF
--- a/lib/config/a11y-strict.js
+++ b/lib/config/a11y-strict.js
@@ -1,0 +1,22 @@
+'use strict';
+
+module.exports = {
+  rules: {
+    'link-href-attributes': true,
+    'link-rel-noopener': 'strict',
+    'no-abstract-roles': true,
+    'no-bare-strings': true,
+    'no-input-block': true,
+    'no-input-tagname': true,
+    'no-invalid-interactive': true,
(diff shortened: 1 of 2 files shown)


MelSumner:
I have some work to do on updating the README as well as adding some additional info on RULES, so definitely expect some additional PRs related to this.

bmish:
Can we enable all of these as recommended rules in the next major release? Is it necessary for a separate config for them?

rwjblue:
> Can we enable all of these as recommended rules in the next major release? Is it necessary to have a separate config for them?

I think that the idea here is that we have a config that _can_ be used, but that we honestly don't think all apps **should** use all of these rules. We will absolutely continue to evaluate any/all rules that we land for addition to `recommended` (at a major bump, as appropriate) but doing so requires a decent amount of care to make sure we aren't flagging significant false positives or causing undue burden on devs (which would lead them to removing template linting).

Turbo87:
@MelSumner now that #977 is merged, can you rebase and include information about the semver policy in the new doc file?

MelSumner:
Closing this since we took a different approach.

PR #980, opened 2019-12-04
Add `no-invalid-role` rule

MelSumner:
If merged, this adds a new rule called `no-invalid-role`. It checks for `role="presentation"` or `role="none"` on a list of HTML elements that should not have this role.

It should not effect custom components.

### DIFF
--- a/docs/rule/no-invalid-role.md
+++ b/docs/rule/no-invalid-role.md
@@ -1,0 +1,41 @@
+## no-invalid-role
+
+This rule checks for invalid element/role combinations.
+
+Current list of checks:
+
+1. Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content. This rule checks semantic HTML elements for the presence of `role="none"` or `role="presentation"`. Elements that are permitted to explicitly have these role values are `<div>`, `<img>`, `<span>` and `<svg>`.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+<table role="presentation">
+</table>
+```
+
+```hbs
+<ul role="none">
+</ul>
+```
+
+This rule **allows** the following:
+
+```hbs
+<img role="presentation">
+```
+
+```hbs
+<span role="none"></span>
+```
+
+### Migration
+
+* If violations are found, remediation should be planned to replace the semantic HTML with the `div` element. Additional CSS will likely be required.
+
+### References
+
+* [HTML elements reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)
+* [WAI-ARIA presentation(role)](https://www.w3.org/TR/wai-aria/#presentation)
+* [Failure of Success Criterion 1.3.1 due to the use of role presentation on content which conveys semantic information](https://www.w3.org/WAI/WCAG21/Techniques/failures/F92)
--- a/docs/rule/no-invalid-role.md
+++ b/docs/rule/no-invalid-role.md
@@ -4,7 +4,7 @@ This rule checks for invalid element/role combinations.
 
 Current list of checks:
 
-1. Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content. This rule checks semantic HTML elements for the presence of `role="none"` or `role="presentation"`. Elements that are permitted to explicitly have these role values are `<div>`, `<img>`, `<span>` and `<svg>`.
+1. Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content. This rule checks semantic HTML elements for the presence of `role="none"` or `role="presentation"` and compares it to the list of disallowed elements. It should not effect custom elements.
 
 ### Examples
 
--- a/lib/rules/no-invalid-role.js
+++ b/lib/rules/no-invalid-role.js
@@ -1,0 +1,126 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+function ERROR_MESSAGE_INVALID_ROLE(element) {
+  return `Use of presentation role on <${element}> detected. Semantic elements should not be used for presentation.`;
+}
+
+// None of these elements can be marked with `role="presentation"` or `role="none"`.
+// List from https://developer.mozilla.org/en-US/docs/Web/HTML/Element.
+const DISALLOWED_ELEMENTS = [
+  'a',
+  'abbr',
+  'applet',
+  'area',
+  'audio',
+  'b',
+  'bdi',
+  'bdo',
+  'blockquote',
+  'br',
+  'button',
+  'caption',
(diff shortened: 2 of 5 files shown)


lib/rules/no-invalid-role.js
```diff
@@ -0,0 +1,126 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+function ERROR_MESSAGE_INVALID_ROLE(element) {
```
Turbo87:
```suggestion
function createErrorMessage(element) {
```

docs/rule/no-invalid-role.md
```diff
@@ -0,0 +1,41 @@
+## no-invalid-role
+
+This rule checks for invalid element/role combinations.
+
+Current list of checks:
+
+1. Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content. This rule checks semantic HTML elements for the presence of `role="none"` or `role="presentation"`. Elements that are permitted to explicitly have these role values are `<div>`, `<img>`, `<span>` and `<svg>`.
```
Turbo87: the last sentence is a little misleading. we use a blacklist of elements in the implementation, but this description makes it sound like a whitelist. that had be confused because I was wondering if it worked for `IndustryStandardCase` component invocations or not 😉

lib/rules/no-invalid-role.js
```diff
@@ -0,0 +1,126 @@
...
+
+module.exports = class NoInvalidRole extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        const hasRoleAttribute = AstNodeInfo.hasAttribute(node, 'role');
+        const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');
+        if (hasRoleAttribute) {
+          if (['presentation', 'none'].includes(roleValue)) {
+            if (DISALLOWED_ELEMENTS.includes(node.tag)) {
```
Turbo87: looks like a lot of nesting here. do you think it might be worth it to collapse that into a single condition with `&&`? or do something like `if (!hasRoleAttribute) return;`?

test/unit/rules/no-invalid-role-test.js
```diff
@@ -0,0 +1,59 @@
...
+    '<div role="presentation"></div>',
+    '<img alt="" role="none">',
+    '<img role="none">',
+    '<img alt="" role="presentation">',
+    '<img role="presentation">',
+    '<span role="none"></span>',
+    '<span role="presentation"></span>',
+    '<svg role="none"></svg>',
+    '<svg role="presentation"></svg>',
+  ],
```
Turbo87: can you add a component invocation here? 🙏

docs/rule/no-invalid-role.md
MelSumner: that's my fault, I flipped the implementation but missed this. I'll update, thank you for catching it!

lib/rules/no-invalid-role.js
MelSumner: Previous reviews have requested that I don't use `&&` but rather nest it for easier maintainability - I'll see if I can simplify though.

Turbo87: 🤔 

seems like an interesting conversation. not sure I'd support that opinion... 😅 

in this case I'd personally go with the early-return pattern, because I assume that more `role` related assertions will follow

lib/rules/no-invalid-role.js
```diff
@@ -0,0 +1,132 @@
...
+
+        const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');
+
+        if (!['presentation', 'none'].includes(roleValue)) {
+          return;
+        }
+
+        if (DISALLOWED_ELEMENTS.includes(node.tag)) {
+          this.log({
+            message: ERROR_MESSAGE_INVALID_ROLE(node.tag),
```
rwjblue:
```suggestion
            message: createErrorMessage(node.tag),
```

lib/rules/no-invalid-role.js
```diff
@@ -0,0 +1,132 @@
...
+            column: node.loc && node.loc.start.column,
+            source: this.sourceForNode(node),
+          });
+        }
+      },
+    };
+  }
+};
+
+module.exports.ERROR_MESSAGE_INVALID_ROLE = ERROR_MESSAGE_INVALID_ROLE;
```
rwjblue:
```suggestion
module.exports.createErrorMessage = createErrorMessage;
```

test/unit/rules/no-invalid-role-test.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const rule = require('../../../lib/rules/no-invalid-role');
+
+const ERROR_MESSAGE_INVALID_ROLE = rule.ERROR_MESSAGE_INVALID_ROLE;
```
rwjblue:
```suggestion
const { createErrorMessage } = rule;
```

test/unit/rules/no-invalid-role-test.js
```diff
@@ -0,0 +1,60 @@
...
+    '<svg role="none"></svg>',
+    '<svg role="presentation"></svg>',
+    '<custom-component role="none"></custom-component>',
+  ],
+
+  bad: [
+    {
+      template: '<ul role="presentation"></ul>',
+      result: {
+        message: ERROR_MESSAGE_INVALID_ROLE('ul'),
```
rwjblue:
```suggestion
        message: createErrorMessage('ul'),
```

test/unit/rules/no-invalid-role-test.js
```diff
@@ -0,0 +1,60 @@
...
+        moduleId: 'layout.hbs',
+        source: '<ul role="presentation"></ul>',
+        line: 1,
+        column: 0,
+      },
+    },
+    {
+      template: '<table role="presentation"></table>',
+      result: {
+        message: ERROR_MESSAGE_INVALID_ROLE('table'),
```
rwjblue:
```suggestion
        message: createErrorMessage('table'),
```

test/unit/rules/no-invalid-role-test.js
```diff
@@ -0,0 +1,60 @@
...
+        moduleId: 'layout.hbs',
+        source: '<table role="presentation"></table>',
+        line: 1,
+        column: 0,
+      },
+    },
+    {
+      template: '<table role="none"></table>',
+      result: {
+        message: ERROR_MESSAGE_INVALID_ROLE('table'),
```
rwjblue:
```suggestion
        message: createErrorMessage('table'),
```

PR #985, opened 2019-12-05
Update: Add meta info to rules

### LINKED ISSUE #943: add meta information to rules
For each rule, some meta information is needed: 

- category (1)
- tags (0-many)

This will help with automating presets and some other future features.

MelSumner:
This first commit starts to resolve #943 

If this is what we want, I'll do the rest.

### DIFF
--- a/lib/rules/attribute-indentation.js
+++ b/lib/rules/attribute-indentation.js
@@ -42,7 +42,7 @@ module.exports = class AttributeSpacing extends Rule {
     return {
       description: 'requires params & attributes to be indented',
       category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
-      recommended: false,
+      presets: {},
       url:
         'https://github.com/ember-template-lint/ember-template-lint/blog/master/docs/rules/attribute-indentation.md',
       fixable: false,
--- a/lib/rules/attribute-indentation.js
+++ b/lib/rules/attribute-indentation.js
@@ -38,6 +38,17 @@ function canApplyRule(node, type, config) {
 }
 
 module.exports = class AttributeSpacing extends Rule {
+  static get meta() {
+    return {
+      description: 'requires params & attributes to be indented',
+      category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
+      recommended: false,
+      url:
+        'https://github.com/ember-template-lint/ember-template-lint/blog/master/docs/rules/attribute-indentation.md',
+      fixable: false,
+    };
+  }
+
   getLineIndentation(node) {
     let currentLine = this.source[node.loc.start.line - 1];
     let leadingWhitespace = getWhiteSpaceLength(currentLine);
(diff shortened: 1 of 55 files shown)


lib/rules/attribute-indentation.js
```diff
@@ -38,6 +38,17 @@ function canApplyRule(node, type, config) {
 }
 
 module.exports = class AttributeSpacing extends Rule {
+  static get meta() {
+    return {
+      description: 'requires params & attributes to be indented',
+      category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
+      recommended: false,
```
Turbo87:
```suggestion
      presets: {},
```

in contrast to e.g.:

```js
      presets: { recommended: true, 'a11y-strict': 'strict' },
```

lib/rules/attribute-indentation.js
```diff
@@ -38,6 +38,17 @@ function canApplyRule(node, type, config) {
...
+      url:
+        'https://github.com/ember-template-lint/ember-template-lint/blog/master/docs/rules/attribute-indentation.md',
+      fixable: false,
```
Turbo87: maybe we can derive this one automatically based on the existance of a fixer method on the rule or something like that? I haven't looked into the fixer implementation yet.

rwjblue: The fixer implementation will not be "across the board" for each rule either, each logged failure would be independently fixable (meaning if we tried to detect if fixing is possible, we'd first have to actually lint a file!).

Since, this is a static getter on the class (with the intention to be able to easily use it for scripting things like building the README) I think we should keep a flag here to indicate that it _may_ be fixable.

Turbo87:
@MelSumner not sure what happened here, but this seems to include a lot of unrelated commits now 😅

MelSumner:
> @MelSumner not sure what happened here, but this seems to include a lot of unrelated commits now

hmmm. I don't know either, I will have to investigate. :female_detective:

MelSumner:
@Turbo87 looks like it was this: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch/26986320#26986320 (thank you to @jenweber for the tip!) 

I toggled the base branch to something else and then back to master, and then the unrelated commits went away. :woman_shrugging:

lib/rules/inline-link-to.js
```diff
@@ -5,6 +5,16 @@ const Rule = require('./base');
 const message = 'The inline form of link-to is not allowed. Use the block form instead.';
 
 module.exports = class InlineLinkTo extends Rule {
+  static get meta() {
+    return {
+      description: 'disallows the inline form of `link-to`',
+      category: 'Best Practices', // 'Stylistic Issues', 'Deprecated Rules', 'Possible Error', 'Best Practices',
+      presets: {},
+      url:
+        'https://github.com/ember-template-lint/ember-template-lint/blog/master/docs/rules/inline-link-to.md',
```
Turbo87: that URL leads to 404 for me

Turbo87: I'm wondering if we should infer the URL from the rule name and `package.json` version instead 🤔

MelSumner:
@rwjblue are we ready to work on this one and merge it now that 2.0 is out?

MelSumner:
So many conflicts. It'll be cleaner to re-do this PR.

PR #986, opened 2019-12-05
no-multiple-empty-lines: Fix `parseConfig()` implementation

### LINKED ISSUE #983: [1.9.0] error  Cannot use 'in' operator to search for 'max' in true  undefined
The new `no-multiple-empty-lines` rule outputs an error when enabled:

```
frontend/templates/foo.hbs
  -:-  error  Cannot use 'in' operator to search for 'max' in true  undefined
```

Turbo87:
Resolves #983

### DIFF
--- a/lib/rules/no-multiple-empty-lines.js
+++ b/lib/rules/no-multiple-empty-lines.js
@@ -2,6 +2,8 @@
 
 const Rule = require('./base');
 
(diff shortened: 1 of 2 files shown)


bmish:
We should really adopt JSON schema for rule schema validation, so the plugin can do this validation instead of each rule, just like eslint has. CC: @rwjblue 

https://eslint.org/docs/developer-guide/working-with-rules#options-schemas

PR #988, opened 2019-12-09
Add rule `no-invalid-link-text`

MelSumner:
If merged, this rule will check link text to prevent common accessibility failure 84- Failure of Success Criterion 2.4.9 due to using a non-specific link such as "click here" or "more".

### DIFF
--- a/lib/rules/no-invalid-link-text.js
+++ b/lib/rules/no-invalid-link-text.js
@@ -15,6 +15,11 @@ function hasInvalidLinkText(node) {
   // Extract the text content(s) from the TextNode child(ren)
   const nodeChildren = AstNodeInfo.childrenFor(node);
   const textChildren = nodeChildren.filter(child => AstNodeInfo.isTextNode(child));
+
+  if (nodeChildren.length !== textChildren.length) {
+    // do not flag an error when the link contains additional dynamic (non-text) children
+    return;
+  }
   const linkTexts = textChildren.map(linkText => linkText['chars'].toLowerCase().trim());
 
   // Check to see if the text content is too `generic` by checking it against
--- a/lib/rules/no-invalid-link-text.js
+++ b/lib/rules/no-invalid-link-text.js
@@ -3,7 +3,7 @@
 const Rule = require('./base');
 const AstNodeInfo = require('../helpers/ast-node-info');
 
-const disallowedGenericLinkTextList = [
+const DISALLOWED_LINK_TEXT = [
   // WCAG F84-Provided Examples
   'click here',
   'more info',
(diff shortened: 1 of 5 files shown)


test/unit/rules/no-invalid-link-text-test.js
```diff
@@ -0,0 +1,24 @@
...
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-invalid-link-text',
+
+  config: true,
+
+  good: ['<a href="https://myurl.com">Click here to read more about this amazing adventure</a>'],
+
+  bad: [
```
rwjblue: We need more examples here for `<LinkTo>` and `{{#link-to}}more{{/link-to}}` and `{{link-to 'more' 'route-name'}}` (confirm the order here, I hate this syntax :sob:).

lib/rules/no-invalid-link-text.js
```diff
@@ -0,0 +1,44 @@
...
+  'click here',
+  'more info',
+  'read more',
+  'more',
+];
+module.exports = class NoInvalidLinkText extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        if (node.tag !== 'a') {
```
rwjblue: Also need to add `LinkTo` support

lib/rules/no-invalid-link-text.js
```diff
@@ -0,0 +1,44 @@
...
+          return;
+        }
+        if (AstNodeInfo.isLinkElement(node)) {
+          // Extract the text content(s) from the TextNode child(ren)
+          const nodeChildren = AstNodeInfo.childrenFor(node);
+          const textChildren = nodeChildren.filter(child => AstNodeInfo.isTextNode(child));
+          const linkTexts = textChildren.map(linkText => linkText['chars'].toLowerCase().trim());
```
rwjblue: If `textChildren.length !== nodeChildren.length` we have to assume that the other dynamic bits have added the additional descriptive content.

Can you add additional "good" tests for:

```hbs
<a href="/asdf">{{someGreatDescription}} click here</a>
```

lib/rules/no-invalid-link-text.js
```diff
@@ -0,0 +1,44 @@
...
+
+          // Check to see if the text content is too `generic` by checking it against
+          // the reference list (array, above) of `disallowed` link text Strings/phrases
+          const genericLinkTexts = linkTexts.filter(linkText =>
+            DISALLOWED_LINK_TEXT.includes(linkText)
+          );
```
rwjblue: Might be able to use `linkTexts.some` instead of `.filter` + `.length` check

test/unit/rules/no-invalid-link-text-test.js
MelSumner: since the linter already prevents inline syntax, I didn't add it.

PR #993, opened 2019-12-10
updated rule to cover orientation lock failure

MelSumner:
If merged, this PR updates the `no-invalid-meta` to cover common failure 97- failure due to locking orientation.

### DIFF
--- a/docs/rule/no-invalid-meta.md
+++ b/docs/rule/no-invalid-meta.md
@@ -6,8 +6,10 @@ This rule checks for these meta tag issues:
 - a meta with a timed refresh- the timed delay should be greater than 72,000 seconds.
 - a meta that locks orientation to landscape or portrait view
 
+### Redirects & Refresh
 Sometimes a page automatically redirects to a different page. When this happens after a timed delay, it is an unexpected change of context that may interrupt the user. Redirects without timed delays are okay, but please consider a server-side method for redirecting instead (method will vary based on your server type).
 
+### Orientation Lock
 When content is presented with a restriction to a specific orientation, users must orient their devices to view the content in the orientation that the author imposed. Some users have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair), and if the content cannot be viewed in that orientation it creates problems for the user.
 
 ### Examples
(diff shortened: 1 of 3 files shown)


lib/rules/no-invalid-meta.js
```diff
@@ -46,6 +44,22 @@ module.exports = class NoInvalidMeta extends Rule {
...
         }
+        if (contentAttrValue.includes('user-scalable=no')) {
+          this.log({
+            message: 'a meta viewport should not restrict user-scalable',
+            line: node.loc && node.loc.start.line,
+            column: node.loc && node.loc.start.column,
+            source: this.sourceForNode(node),
+          });
+        }
+        if (contentAttrValue.includes('maximum-scale')) {
```
rwjblue: This one is fine without custom parsing, since we don't care about the specific value

docs/rule/no-invalid-meta.md
```diff
@@ -1,11 +1,16 @@
...
 
 - a meta with a redirect- if it exists, it checks for a timed delay greater than 0.
 - a meta with a timed refresh- the timed delay should be greater than 72,000 seconds.
+- a meta that locks orientation to landscape or portrait view
+
+### Redirects & Refresh
+Sometimes a page automatically redirects to a different page. When this happens after a timed delay, it is an unexpected change of context that may interrupt the user. Redirects without timed delays are okay, but please consider a server-side method for redirecting instead (method will vary based on your server type).
+
+### Orientation Lock
+When content is presented with a restriction to a specific orientation, users must orient their devices to view the content in the orientation that the author imposed. Some users have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair), and if the content cannot be viewed in that orientation it creates problems for the user.
```
Turbo87: these two things seem entirely unrelated to me. I would still prefer it if they were separate rules... :-/

rwjblue:
Hmm, how are they unrelated though? It is specifically about validating meta tags. Unfortunately meta has a bunch of seemingly unrelated behaviors, but we can’t really change that :thinking:

Turbo87:
I would prefer e.g. `no-timed-redirect` and `no-orientation-lock` in this case. Adding new cases to this rule will IMHO make it weird from a semver perspective too.

Unrelated: these kinds of static meta tags are usually in the `index.html` file. are we supposed to run `ember-template-lint` in that file too?

I'm asking because that file contains a doctype and that apparently isn't supported here yet. 🤔

PR #997, opened 2019-12-12
Add `require-button-type` to octane configuration

rwjblue:


Turbo87:
isn't that a breaking change? 🤔

rwjblue:
Octane preset is preview (as octane itself is)

Turbo87:
I see, okay. guess that should be included in https://github.com/ember-template-lint/ember-template-lint/pull/977 😄

PR #1006, opened 2019-12-13
no-curly-component-invocation: Ignore *all* curly invocations with positional arguments

### LINKED ISSUE #1004: no-curly-component-invocation: False positive for block invocation with positional parameter
Example: 

```hbs
{{#animated-if condition}}foo{{/animated-if}}
```

Turbo87:
Resolves #1004 

as originally intended in https://github.com/ember-template-lint/ember-template-lint/issues/925#issuecomment-554481335 🤦‍♂️

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -84,7 +84,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
   }
 
   getCurlyInvocationSyntaxError(node) {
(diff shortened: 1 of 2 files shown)


ef4:
This change is good. 

I'm concerned that I didn't get it automatically when using ember-cli-update to 3.15, which means people are going to be told to rewrite all their `{{#animated-if}}` to angle brackets, which is going to generate significant unnecessary confusion.

Turbo87:
@ef4 FYI #1027

Issue #616, opened 2019-01-17
junit xml output


brondsem:
Is junit-style xml output possible?  That would be nice so that Jenkins (and probably other tools too) could directly parse the output of ember-template-lint.

rwjblue:
Sounds great, definitely open to adding something like this.

For now, I think the best we have is `--json` flag to emit machine readable output...

lifeart:
@rwjblue looks like it can be solved adding one dependency https://www.npmjs.com/package/mocha-junit-reporter

PR #1008, opened 2019-12-13
Extract `Printer` classes

Turbo87:
This PR extracts several `Printer` classes from the code to simplify the console output logic.

This prepares to implement a `JUnitPrinter` class (see https://github.com/ember-template-lint/ember-template-lint/issues/616) in a later PR.

### DIFF
--- a/lib/printers/json.js
+++ b/lib/printers/json.js
@@ -1,10 +1,25 @@
+const Linter = require('../index');
+
 class JsonPrinter {
   constructor(options = {}) {
+    this.options = options;
     this.console = options.console || console;
   }
 
   print(errors) {
-    this.console.log(JSON.stringify(errors, null, 2));
+    let filteredErrors = {};
+    for (let filePath of Object.keys(errors)) {
+      let fileErrors = errors[filePath] || [];
+
+      let errorsFiltered = fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY);
+      let warnings = this.options.quiet
+        ? []
+        : fileErrors.filter(error => error.severity === Linter.WARNING_SEVERITY);
+
+      filteredErrors[filePath] = errorsFiltered.concat(warnings);
+    }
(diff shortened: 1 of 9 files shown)


lib/printers/junit.js
```diff
@@ -0,0 +1,26 @@
+const Linter = require('../index');
```
rwjblue: Let's remove this one for now? It doesn't seem correct (looks like a duplicate of the JsonPrinter), and maybe still WIP.

lib/printers/json.js
```diff
@@ -0,0 +1,26 @@
...
+      let errorsFiltered = fileErrors.filter(error => error.severity === Linter.ERROR_SEVERITY);
+      let warnings = this.options.quiet
+        ? []
+        : fileErrors.filter(error => error.severity === Linter.WARNING_SEVERITY);
+
+      filteredErrors[filePath] = errorsFiltered.concat(warnings);
+    }
+
+    this.console.log(JSON.stringify(filteredErrors, null, 2));
+  }
```
rwjblue: Should some of this get extracted to a helper method? Seems like both `pretty` and `json` want most of this same logic.

rwjblue: I guess alternatively, `pretty` could extend from `json` (since it can use the well known JSON structure to emit the pretty messages).

lib/printers/junit.js
Turbo87: whooops, that shouldn't have ended up here 😱

lib/printers/json.js
Turbo87: yeah, idk, the pretty implementation also counts the number of issues, so I figured the deduplication wasn't worth it for now. especially since the json/pretty printers are currently mutually exclusive.

rwjblue: Gotcha, seems totally fine to me.

Issue #29, opened 2016-05-03
Rule: Disallow `{{yield}}` only templates.


rwjblue:
Prevent `{{yield}}` only templates. They are not required, and only increase the total template payload size...

oriSomething:
are you sure they're totally unneeded?
i think in case you make components for styling purposes it's actually understood. if there is no component. the `{{yield}}` made automatically? if not, i'm not sure this rule should be on by default

rwjblue:
I don't understand what you mean. If you have a component JS file, but do not need a template then the {{yield}} is not needed (it is default behavior). If you have no JS **and** no template other than {{yield}} you shouldn't be using a component invocation at all (use normal DOM, it is faster and much clearer).

oriSomething:
i had a case that i created a button that had only `{{yield}}` in the template, that all the functionality was in JS, the `{{yield}}` was for the text on the button only.
also i know from a friend who works on some project that because of a technical need he had some components that their template is only `{{yield}}`.
maybe those are extreme cases. but i barely see this rule helpful, and more of annoying thing every time i'll create a new component.

rwjblue:
The cases you listed would have worked properly with no template.

oriSomething:
maybe, but it would damage the ease to extend the template if needed.
anyway, as i told, i don't think this rule can create gains as much as it would be annoying while creating a new component in the CLI while the server works

mmun:
I think this rule is interesting. I think it's worth exploring without enabling by default.

rwjblue:
@lifeart - I'm not sure I follow? Having a template with _only_ `{{yield}}` seems bad regardless of contextual components...

lifeart:
ah, I mean `{{yield (hash someProp=someValue)}}`.

Regarding `{{yield}}` only components it can be an usecase for component with tag wrapper for some styling or native DOM bindings for internal content.

ex:
```js
{
   classNames: ['some', 'nice', 'ui', 'panel'],
   tagName: 'div'
}
```
```hbs
   {{my-panel}} my content inside nice panel {{my-panel}}
```

-------------------------

```js
{
   classNames: ['some', 'nice', 'ui', 'panel'],
   tagName: 'div',
   didRender() {
      this.element.querySelector('button').addEventListener();
   }
}
```
```hbs
   {{my-panel}} <button>some native action-binded item</button> {{my-panel}}
```

rwjblue:
@lifeart - In your example:

```js
[...]

PR #1010, opened 2019-12-13
Add `no-yield-only` rule

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/29 (aka. the oldest still-open issue on this project 😅)

/cc @rwjblue

### DIFF
--- a/lib/rules/no-yield-only.js
+++ b/lib/rules/no-yield-only.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE = '{{yield}}-only templates are not allowed';
+
+module.exports = class NoMultipleEmptyLines extends Rule {
+  visitor() {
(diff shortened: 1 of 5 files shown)


lib/rules/no-yield-only.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE = '{{yield}}-only templates are not allowed';
+
+module.exports = class NoMultipleEmptyLines extends Rule {
+  visitor() {
+    if (this._rawSource.trim() !== '{{yield}}') {
+      return;
```
rwjblue: Ha! I didn't realize we allowed returning `undefined` here.

Turbo87: hehe, yeah. was wondering if it worked and the unit tests didn't fail so I considered it okay :D

Issue #751, opened 2019-06-19
Consider using `{{!-- template-lint-disable --}}` vs `{{! template-lint-disable }}` 🤔


rwjblue:

Turbo87:
isn't that the same thing from the AST perspective? 🤔

rwjblue:
Yes, but not _quite_. Also prettier prefers `{{!--` IIRC (@gavinjoyce made an issue somewhere about this?).

PR #1011, opened 2019-12-13
Add test cases for `{{!-- template-lint-disable --}}` comments

Turbo87:
I guess this resolves https://github.com/ember-template-lint/ember-template-lint/issues/751 ? 😅 

/cc @rwjblue

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -71,6 +71,7 @@ module.exports = function generateRuleTests({
   groupingMethod(name, function() {
     let DISABLE_ALL = '{{! template-lint-disable }}';
     let DISABLE_ONE = `{{! template-lint-disable ${name} }}`;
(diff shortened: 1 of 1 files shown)


rwjblue:
Yeah, seems good. Can you also update the docs to show `{{!--` style instead of `{{!` style (or show `{{!--` as the primary and then `{{!` as an alternate)?

Turbo87:
> Can you also update the docs

hmm... but... why? :D

any reason to prefer one over the other?

rwjblue:
> any reason to prefer one over the other?

Ya, because `prettier --fix` uses `{{!--` and `{{!--` covers more cases (it doesn't close when commenting out a thing that includes `}}`, since it requires `--}}` to end the comment).

rwjblue:
https://github.com/emberjs/rfcs/issues/504 was the issue I was thinking of

Turbo87:
hmm, I see. okay if I do that in a separate follow-up PR? I'd say this PR is still useful on its own like this :)

PR #1012, opened 2019-12-13
Implement JUnit XML printer

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/616

This implements a basic JUnit XML reporter that writes its output to the `junit-template-lint.xml` file if the `--junit` option is used. Unlike `--json`, this is not mutually exclusive with the regular pretty text reporter.

Ideally, we would have more CLI options to control the output but I figured this would probably be good enough for now.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -59,6 +59,9 @@ function parseArgv(_argv) {
         case '--json':
           options.named.json = true;
           break;
+        case '--junit':
+          options.named.junit = true;
+          break;
         case '--verbose':
           options.named.verbose = true;
           break;
--- a/lib/printers/default.js
+++ b/lib/printers/default.js
@@ -12,6 +12,11 @@ class DefaultPrinter {
       this.delegates.push(new PrettyPrinter(options));
     }
 
+    if (options.junit) {
+      let JUnitPrinter = require('./junit');
+      this.delegates.push(new JUnitPrinter(options));
+    }
+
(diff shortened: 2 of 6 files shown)


rwjblue:
> Unlike --json, this is not mutually exclusive with the regular pretty text reporter.

I would prefer not to do this. Can we instead allow `--pretty --json --junit` (basically adding each of the possible formatters)?

Turbo87:
yeah, that would probably be better 👍 

`--json=some-file.json` does that work currently? I don't remember seeing any code that would support this 🤔 

about that topic: any thoughts on introducing a proper args parsing lib?

rwjblue:
> --json=some-file.json does that work currently? I don't remember seeing any code that would support this 🤔

No, we only currently support `--json`. I'm proposing a more comprehensive conceptual setup.

> about that topic: any thoughts on introducing a proper args parsing lib?

if we need it we can, generally speaking our needs are pretty small here and the impl inrepo is simpler/lighterweight/etc than most parsers out there.

rwjblue:
@Turbo87 friendly ping 😸

Turbo87:
yeah, I haven't forgotten it 😉

rwjblue:
hehe, I know the feeling!

rwjblue:
👋 just ping me when you have a chance to pick this back up 😸

rwjblue:
I'd still love for this to move forward, but going to close for now since it seems like time is too tight to make progress.

Please reopen when / if you have the time @Turbo87!

Turbo87:
yep, will do. sorry... 😞

PR #1013, opened 2019-12-13
Fix missing `Linter.errorsToMessages()` API for ember-cli-template-lint

Turbo87:
turns out that `ember-cli-template-lint` was relying on that method to exist... 😩 

/cc @rwjblue


Turbo87:
@trabus confirmed that this fixes the issue. will merge and release this as a patch release.

PR #1019, opened 2019-12-18
Adding error when individual pending rules are passing

gmurphey:
This hopefully fixes the following scenario:

1. We have the following `pending` config:
        ```
{
  moduleId: 'some/path/here',
  only: ['no-bare-strings', 'quotes']
}
        ```
1. Someone "accidentally" fixes all the `quote` issues in `some/path/here` but doesn't update the `pending` config
1. `ember-template-lint` continues happily chugging along
1. Someone reintroduces `quote` issues  in `some/path/here` and `ember-template-lint` doesn't catch them
1. Repeat

The changes in this PR would catch the passing pending rule in step 2, prompting the dev to remove the rule from pending rules.

This new "meta" rule should be incorporated to the work done in #954 after it's merged.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -115,9 +115,10 @@ function run() {
     let fileErrors = lintFile(linter, filePath, moduleId);
 
     if (printPending) {
+      const ignoredPendingRules = ['invalid-pending-module', 'invalid-pending-module-rule'];
       let failingRules = Array.from(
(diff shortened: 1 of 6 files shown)


gmurphey:
@rwjblue this has been updated to ignore `invalid-pending-module-rule` failures when generating pending modules and added a test showing as much. Thanks!

PR #1024, opened 2019-12-19
Upgrade ember-template-recast

dcyriller:
As dependabot doesn't work for `next` branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "chalk": "^2.4.2",
-    "ember-template-recast": "^3.2.8",
+    "ember-template-recast": "^3.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "chalk": "^2.4.2",
-    "ember-template-recast": "^3.2.8",
+    "ember-template-recast": "^3.2.1",
```
rwjblue: This seems slightly odd? the before has ^3.2.8 and this PR downgrades it to 3.2.1, is that right?

dcyriller: Oups it’s a typo, I meant ^3.3.1 (as in yarn.lock)

PR #935, opened 2019-11-18
Implement more plumbing for `--fix` support (add `Linter.prototype.verifyAndFix`)

dcyriller:
This PR implements the `verifyAndFix` function. Next step will be to implement the `applyFixes` function.

### Tests
I've added an acceptance test for `verifyAndFix`.

### Manual tests
I've manually tested the PR in our codebases (in which we use template-lint@2-beta).

### Related
From this PR, I've extracted these three small PRs that may be reviewed / merged first: https://github.com/ember-template-lint/ember-template-lint/pull/932, https://github.com/ember-template-lint/ember-template-lint/pull/933, https://github.com/ember-template-lint/ember-template-lint/pull/934 (EDIT: ✅)

PR #1025, opened 2019-12-19
--fix: Implement fix algorithm (part 2/2)

dcyriller:
Should be reviewed after https://github.com/ember-template-lint/ember-template-lint/pull/935

In follow-ups (that are ready but no PR opened yet), I have:
- implemented the fs write operation
- implemented the --fix option for `require-button-type`
- setup Jest snapshots to test the above mentioned rule

## Caveheats
- it removes a potential BOM. But is should be easy to preserve it in a follow-up.
- ~~more important: the notion of range is not available yet in the AST, I'll have to add it in template-recast~~

## TODO
- ~~implement the range in template-recast (probably?)~~
- ~~document the fixer API, taken from eslint~~
- implement a few examples
- add more tests
- create a meta issue to implement fix option on fixable rules
- we will need documentation for: how to write a fixer?, how to use builders?, API doc for builders, ...?

### DIFF
--- a/lib/apply-fixes.js
+++ b/lib/apply-fixes.js
@@ -1,3 +1,26 @@
+const { print, transform } = require('ember-template-recast');
+
+/**
+ * Compares the locations of two objects in a source file
+ * @param {line: number, column: number} itemA The first object
+ * @param {line: number, column: number} itemB The second object
+ * @returns {number} A value less than 1 if itemA appears before itemB in the source file, greater than 1 if
+ * itemA appears after itemB in the source file, or 0 if itemA and itemB have the same location.
+ */
+function compareLocations(itemA, itemB) {
+  return itemA.line - itemB.line || itemA.column - itemB.column;
+}
+
+/**
+ * @param {string} source - The source code to fix.
+ * @param {Object} fixer - A visitor that will fix the template
+ * @returns {Object} result - The fixed source
+ */
+function applyFix(source, fixer) {
+  let { ast } = transform(source, fixer);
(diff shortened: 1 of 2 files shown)


dcyriller:
After a chat with @Turbo87, I reworked the fix mechanism. It was based on string manipulation (because I wanted to keep a somewhat consistent API with what ESLint does). It is now based on ember-template-recast's `transform` and `print` methods. This will probably require more work in template-recast but it looks more powerful indeed. Here is what the implementation looks like:

```js
function applyFix(source, fixer) {
  let { ast } = transform(source, fixer);
  return print(ast);
}
```

Where `fixer` returns a visitor akin:
```js
function fixer(env) {
  let { builders: b } = env.syntax;

  return {
    ElementNode(node) {
      let attribute = b.attr('type', b.text('button'));
      node.attributes.push(attribute);
      return node;
    },
  };
}
```

So, I don't need a range concept anymore 🎉.

dcyriller:
More work is required to actually support `--fix` option. Here is an example of a rule implementing the option: [require-type-button](https://github.com/ember-template-lint/ember-template-lint/pull/1132). But before engaging too much work, I would be more confortable if we could agree / settle / discuss this implementation.

rwjblue:
I've rebased and updated the base branch to `master` (now that the first bit already landed).

rwjblue:
@dcyriller I made an alternative proposal over in https://github.com/ember-template-lint/ember-template-lint/pull/1158, mind taking a look?

Issue #1026, opened 2019-12-21
no-curly-component-invocation creates false positives with default `octane` preset


Turbo87:
The `octane` preset is using `'no-curly-component-invocation': { requireDash: false }`, which appears to cause a lot of false positives:

```
/Users/tbieniek/Code/crates.io/app/templates/me/following.hbs
  12:24  error  You are using the component {{currentPageStart}} with curly component syntax. You should use <CurrentPageStart> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['currentPageStart'] }`.  no-curly-component-invocation
  12:47  error  You are using the component {{currentPageEnd}} with curly component syntax. You should use <CurrentPageEnd> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['currentPageEnd'] }`.  no-curly-component-invocation
  13:29  error  You are using the component {{totalItems}} with curly component syntax. You should use <TotalItems> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['totalItems'] }`.  no-curly-component-invocation
  22:8  error  You are using the component {{currentSortBy}} with curly component syntax. You should use <CurrentSortBy> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['currentSortBy'] }`.  no-curly-component-invocation

/Users/tbieniek/Code/crates.io/app/templates/me/index.hbs
  16:10  error  You are using the component {{model.user.name}} with curly component syntax. You should use <Model.User.Name> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['model.user.name'] }`.  no-curly-component-invocation
  18:10  error  You are using the component {{model.user.login}} with curly component syntax. You should use <Model.User.Login> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. `'no-curly-component-invocation': { allow: ['model.user.login'] }`.  no-curly-component-invocation
```

rwjblue:
[...]

PR #1027, opened 2019-12-21
no-curly-component-invocation: Fix false positives

Turbo87:
This PR resolves https://github.com/ember-template-lint/ember-template-lint/issues/1026 by basically rewriting the rule 🙈 

I've documented the assumptions that the rule now makes in the rule documentation.

I would personally consider this a bugfix since it clarifies some edge cases and should result in less false positive warnings.

### DIFF
--- a/docs/rule/no-curly-component-invocation.md
+++ b/docs/rule/no-curly-component-invocation.md
@@ -12,37 +12,49 @@ look like they could be component invocations.
 
 ### Examples
 
-This rule **forbids** the following:
+- `{{foo}}` ✅
+  (simple mustache without `-` in the path is likely to be a property; unless in `disallow` list and not a scoped variable)
 
-```hbs
-{{bad-code}}
-{{#bad-code}}{{/bad-code}}
-{{nested/bad-code}}
-{{#nested/bad-code}}{{/nested/bad-code}}
-```
+- `{{foo.bar}}` ✅
+  (simple mustache with nested path is likely to be a nested property)
 
-This rule **allows** the following:
+- `{{foo-bar}}` ❌ (angle brackets: `<FooBar />`)
+  (simple mustache with `-` in the path is more likely a component than a property or helper; unless in `allow` list)
 
-```hbs
-<GoodCode />
-<GoodCode></GoodCode>
-<Nested::GoodCode />
-<Nested::GoodCode></Nested::GoodCode>
-```
+- `{{nested/component}}` ❌ (angle brackets: `<Nested::Component />`)
+  (simple mustache with `/` in the path is more likely a component than a property or helper; unless in `allow` list)
 
-```hbs
-{{! whitelisted helpers}}
-{{some-valid-helper param}}
-{{some/some-valid-helper param}}
-```
+- `{{42}}` ✅
+  (literal value mustaches)
 
-```hbs
-{{! in-built helpers}}
-{{if someProperty "yay"}}
-{{#each items as |item|}}
-  {{item}}
-{{/each}}
-```
+- `{{foo bar}}` ✅
+  (mustache with positional parameters can't be converted to angle brackets syntax)
+
+- `{{foo bar=baz}}` ❌ (angle brackets: `<Foo @bar={{baz}} />`)
+  (mustache with only named parameters is likely a component; unless in `allow` list)
+
+  Setting `requireDash: true` lets `{{foo bar=baz}}` pass, but `{{foo-bar bar=baz}}` fails
+
+- `<div {{foo}} />` ✅
+  (mustache is a modifier)
+
+- `<Foo @bar={{baz}} />` ✅
+  (mustache is an argument or attribute)
+
+- `{{#foo}}{{/foo}}` ❌ (angle brackets: `<Foo></Foo>`)
+  (block mustache is considered a component unless it has positional parameters, an inverse block, or is in `allow` list)
+
+- `{{#foo bar}}{{/foo}}` ✅
+  (block mustache with positional parameters can't be converted to angle brackets syntax)
+
+- `{{#foo}}bar{{else}}baz{{/foo}}` ✅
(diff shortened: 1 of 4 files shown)


test/unit/rules/no-curly-component-invocation-test.js
```diff
@@ -6,71 +6,177 @@ function generateError(name) {
   return `You are using the component {{${name}}} with curly component syntax. You should use <${angleBracketName}> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. \`'no-curly-component-invocation': { allow: ['${name}'] }\`.`;
 }
 
+const SHARED_GOOD = [
+  '{{foo}}',
```
rwjblue: This should not be a "good" example. The octane configuration _already_ requires the template to avoid ambiguous paths making `{{foo}}` reserved for an invocation (either a helper or a component). When `foo` is not a local, it should print an error message that is a little different than the more obvious cases (e.g. it should mention that the path is ambiguous and they should either refactor to `{{this.foo}}` or `<Foo />`).

lib/rules/no-curly-component-invocation.js
```diff
@@ -78,89 +46,132 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
...
+
+        let hasNamedArguments = node.hash.pairs.length !== 0;
+        if (!hasNamedArguments) {
+          if (path.parts.length > 1) {
+            // {{foo.bar}}
+            return;
+          }
+
+          let name = path.original;
+          if (this.config.allow && this.config.allow.includes(name)) {
```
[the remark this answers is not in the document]
Turbo87: yeah, totally agree. I just used what was there without touching the `parseConfig()` method for now, but I can adjust that too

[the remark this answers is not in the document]
rwjblue: Hehe, you rewrote everything else so....

test/unit/rules/no-curly-component-invocation-test.js
Turbo87: This means we assume `{{foo}}` is a component invocation, not a helper invocation?

IMHO that should be configurable then. I've seen enough cases where apps would have been able to use the `no-curly-component-invocation` rule, but haven't migrated to `no-implicit-this` yet 🤔

Turbo87: also, see https://github.com/ember-template-lint/ember-template-lint/issues/1026 😉

rwjblue: Yes, I have seen #1026 and replied in detail there.

rwjblue: > IMHO that should be configurable then. 

Yep, sounds totally reasonable.

rwjblue: We can then make it default to making no assumptions about `{{foo}}` _if_ `no-implicit-this` is disabled.

lib/rules/no-curly-component-invocation.js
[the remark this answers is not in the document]
Turbo87: yeeeeaaaahh..... 🙈 

😆

test/unit/rules/no-curly-component-invocation-test.js
Turbo87: done!

rwjblue:
Thank you @Turbo87!

docs/rule/no-curly-component-invocation.md
```diff
@@ -1,115 +1,92 @@
...
+Edition.
+
+This rule checks all the curly braces in your app and warns about those that
+look like they could be component invocations.
+
+### Examples
+
+- `{{foo}}` ✅
+  (simple mustache without `-` in the path is likely to be a property; unless
+  `noImplicitThis` is set or in `disallow` list and not a scoped variable)
```
chancancode: Do we check for local variables when `noImplicitThis` is set? (`{{#let ... as |foo|}}{{foo}}{{/let}}` should not error.)

docs/rule/no-curly-component-invocation.md
```diff
@@ -1,115 +1,92 @@
...
+
+- `{{foo.bar}}` ✅
+  (simple mustache with nested path is likely to be a nested property; unless
+  `noImplicitThis` is set)
```
chancancode: same question re:above

docs/rule/no-curly-component-invocation.md
```diff
@@ -1,115 +1,92 @@
...
+
+- `{{foo-bar}}` ❌ (angle brackets: `<FooBar />`)
+  (simple mustache with `-` in the path is more likely a component than a
+  property or helper; unless in `allow` list)
+
+- `{{nested/component}}` ❌ (angle brackets: `<Nested::Component />`)
+  (simple mustache with `/` in the path is more likely a component than a
+  property or helper; unless in `allow` list)
+
+- `{{42}}` ✅
```
chancancode: I assume this covers `{{true}}`, `{{false}}`, `{{null}}` and `{{undefined}}` right?

chancancode:
Regression: this PR is currently rejecting `{{outlet}}` due to removal of the built-in list, @Turbo87 is looking into it! See also https://discordapp.com/channels/480462759797063690/514504648590098457/658796846881177601 in discord.

docs/rule/no-curly-component-invocation.md
Turbo87: yes, we do

docs/rule/no-curly-component-invocation.md
Turbo87: yes

docs/rule/no-curly-component-invocation.md
Turbo87: apparently not yet, but I'll fix that too

PR #1032, opened 2019-12-23
no-curly-component-invocation: Ignore built-ins and fix nested key scoped variable scenario

Turbo87:
This PR addresses https://github.com/ember-template-lint/ember-template-lint/pull/1027#discussion_r361016133 and https://github.com/ember-template-lint/ember-template-lint/pull/1027#issuecomment-568596607.

Sorry about that... 😔 

/cc @chancancode

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -97,10 +97,11 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
           this.logNode({ message: this.generateError(name), node });
         } else {
           let isExplicitThis = path.this || path.data;
+          let isLocal = this.isLocal(path);
 
           if (path.parts.length > 1) {
             // {{foo.bar}}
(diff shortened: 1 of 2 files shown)


lib/rules/no-curly-component-invocation.js
```diff
@@ -97,10 +132,11 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
           this.logNode({ message: this.generateError(name), node });
         } else {
           let isExplicitThis = path.this || path.data;
+          let isLocal = this.isLocal(path);
```
chancancode: Should this be `this.isLocal(path.parts[0]);`?

Turbo87: the `isLocal()` method figures that out by itself 😉

chancancode: Never mind, I think I read the source of the wrong `isLocal`!

PR #1046, opened 2020-01-06
Update `octane` config to extend from `recommended`.

rwjblue:
This makes it much much clearer what the difference is between `octane` and `recommended`, and also makes it easier to maintain as we progress through the 2.x series.

Note: this does not result in **any** actual changes in the resulting configuration


Turbo87:
@rwjblue thank you! very good idea 👍

lib/config/octane.js
```diff
@@ -1,46 +1,16 @@
...
-    'no-attrs-in-components': true,
     'no-curly-component-invocation': {
       noImplicitThis: true,
       requireDash: false,
     },
-    'no-debugger': true,
-    'no-duplicate-attributes': true,
-    'no-extra-mut-helper-argument': true,
-    'no-html-comments': true,
-    'no-implicit-this': true,
```
Turbo87: @rwjblue shouldn't this remain here? 🤔

rwjblue: Crappola, yeah. I'll fix. I removed it because I was comparing `next` at the time, _then_ I merged the other PR :sob:

PR #1047, opened 2020-01-06
Add `no-args-path`, `no-invalid-link-text`, `no-invalid-meta`, `require-button-type` into `recommended` configuration.

rwjblue:
These rules were originally added to `octane` configuration during the 1.x series (in order to preserve SemVer compatibility with `recommended` in 1.x), but as we approach 2.0.0 we should move all rules that are not conceptually part of "octane" back into the `recommended` set.


lib/config/recommended.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     'link-rel-noopener': 'strict',
     'link-href-attributes': true,
     'no-abstract-roles': true,
+    'no-args-paths': true,
```
rwjblue: This may seem octane related, but ultimately if anyone is dabbling with `@glimmer/component` but haven't yet updated to use `octane` config they should still be guided away from `{{this.args.foo}}`.

Turbo87: seems reasonable 👍

Issue #953, opened 2019-11-24
Create an easier migration path for stylistic linting in v2 release


bmish:
### Context
In the upcoming v2.0 release, the stylistic lint rules will be removed from the recommended configuration (as mentioned in #746), so that prettier for templates can take over responsibility for template styling.

### Problem
This is a pretty huge change, probably the most disruptive change slated for the v2 release, and could cause upgrade difficulties because:
1. It requires consumers to install a new plugin (ember-template-lint-plugin-prettier) to continue to get valuable stylistic linting. Many consumers will not be aware of this new plugin, not bother to install/adopt it, or perhaps not be interested to use it. 
2. Enabling prettier for templates causes significant churn just by touching/rewriting most template lines.
3. Enabling prettier for templates can be risky (by breaking things like significant whitespace) and thus take a lot of time and work to enable in larger codebases.

### Solution
My preferred solution is for ember-template-lint to keep stylistic lint rules enabled in the recommended configuration, and let ember-template-lint-plugin-prettier disable the conflicting stylistic lint rules itself. 

This is the same way that [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) works ("Turns off all rules that are unnecessary or might conflict with Prettier."). This allows eslint to provide basic style linting until you enable prettier.

This solution would allow ember-template-lint to continue to function on its own without being dependent on another plugin for stylistic linting, while also being compatible with prettier for templates for anyone who installs it.

If that solution isn't tractable, then we could at least provide a "stylistic" configuration for consumers who still want to get stylistic linting from ember-template-lint, and encourage consumers to use that until they're ready to adopt prettier for templates.

Interested to hear feedback on this.

dcyriller:
Related: https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/pull/52

bmish:
@dcyriller your change (having ember-template-lint-plugin-prettier disable the conflicting stylistic rules) looks excellent to me.

@rwjblue @MelSumner @Turbo87 thoughts about re-enabling our own stylistic rules in the recommended configuration as described in my proposal?

Turbo87:
@bmish your reasoning makes sense to me. we should try to not let those two code styles drift apart too much though. I'm generally okay with either solution.

bmish:
Ping on this? Just wondering if anyone else shares my concerns. If not, we can close the issue.

rwjblue:
I think there is a bit of an easier migration path here:

[...]

PR #1048, opened 2020-01-06
Add `stylistic` configuration

bmish:
These rules were removed from the `recommended` configuration in v2 (https://github.com/ember-template-lint/ember-template-lint/pull/899) but we are providing a separate configuration for those who want to continue to use them (as an easier migration path vs. adopting prettier https://github.com/ember-template-lint/ember-template-lint/issues/953).


rwjblue:
@Turbo87 - Yes, this config would fall under the "normal" SemVer rules for config files (only _less restrictive_ changes can be made without a major).

PR #1053, opened 2020-01-07
Add `no-invalid-role` and `no-negated-condition` rules to `recommended` preset

bmish:
BREAKING CHANGE

Thoughts about adding these rules to the `recommended` configuration for the v2 release?

1. [no-invalid-role](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-role.md)
2. [no-negated-condition](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-negated-condition.md)


lib/config/recommended.js
```diff
@@ -30,6 +32,7 @@ module.exports = {
     'no-unbound': true,
     'no-unnecessary-component-helper': true,
     'no-unused-block-params': true,
+    'no-yield-only': true,
```
Turbo87: this one is difficult because the default blueprint for components generates such a template

rwjblue: Agree, we shouldn't fail on `ember generate component`'s output...

rwjblue: updated to remove this one, the others are good though...

PR #918, opened 2019-11-11
Use `.editorconfig` to choose default value for `block-indentation`

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/pull/906

fix: https://github.com/ember-template-lint/ember-template-lint/issues/901

PR #1060, opened 2020-01-07
Add more rules to `stylistic` config

bmish:
Since we're adding this new preset in v2 (originally added in https://github.com/ember-template-lint/ember-template-lint/pull/1048), we might as well include all of the stylistic rules in it, especially since these additional rules are trivial to adopt, and can be disabled as needed.

* [eol-last](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/eol-last.md)
* [no-multiple-empty-lines](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-multiple-empty-lines.md)
* [no-trailing-spaces](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-trailing-spaces.md)

### DIFF
--- a/lib/config/stylistic.js
+++ b/lib/config/stylistic.js
@@ -4,6 +4,8 @@ module.exports = {
   rules: {
     'block-indentation': true,
     'linebreak-style': true,
+    'no-multiple-empty-lines': true,
+    'no-trailing-spaces': true,
     'no-unnecessary-concat': true,
     quotes: 'double',
     'self-closing-void-elements': true,
--- a/lib/config/stylistic.js
+++ b/lib/config/stylistic.js
@@ -3,6 +3,7 @@
 module.exports = {
   rules: {
     'block-indentation': true,
+    'eol-last': 'editorconfig',
     'linebreak-style': true,
     'no-multiple-empty-lines': true,
     'no-trailing-spaces': true,
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Name   | Description |
 |:-----|:------------|
 | [recommended](https://github.com/ember-template-lint/ember-template-lint/blob/next/lib/config/recommended.js) | enables the recommended rules |
 | [octane](https://github.com/ember-template-lint/ember-template-lint/blob/next/lib/config/octane.js) | extends the `recommended` preset by enabling Ember Octane rules |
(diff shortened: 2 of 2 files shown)


bmish:
@rwjblue I believe this is the last change I'd like to get in for the v2 release.

lib/config/stylistic.js
```diff
@@ -3,7 +3,11 @@
 module.exports = {
   rules: {
     'block-indentation': true,
+    'eol-last': 'always',
```
rwjblue: I think we should change this to honor the `.editorconfig`. As it stands, enabling this (via `always`) means we error for `ember g component foo` because the default ember-cli `.editorconfig` states that we should **not** have an EOL in `.hbs` files ([see here](https://github.com/ember-cli/ember-new-output/blob/v3.15.1/.editorconfig#L16-L17)).

I'd say we update this to use the editor config setup similar to #918 and #919. That way, whatever your `.editorconfig` value has will be used and essentially make `'eol-last': true` automatically use either `always` or `never` appropriately.

lib/config/stylistic.js
```diff
@@ -3,7 +3,11 @@
...
     'linebreak-style': true,
+    'no-multiple-empty-lines': true,
+    'no-trailing-dot-in-path-expression': true,
```
rwjblue: This rule should be removed from ember-template-lint, it was added to support a thing that is actually not possible (`{{some.path.here. whoops}}`) any more. IIRC, the guts of the rule are empty and it essentially does nothing.

bmish: Got it, I'll remove that from the config, and I'll open up a separate PR to remove that rule, if you're interested.

lib/config/stylistic.js
bmish: Agreed with that. I removed this rule from the config. We can consider adding it once it has been updated to read from `.editorconfig`.

rwjblue: Submitted https://github.com/ember-template-lint/ember-template-lint/pull/1062 to add that support.

rwjblue: @bmish - If you rebase, you can bring this back now (with the `'editorconfig'` setting added in #1062).

bmish: Added it back. Thanks!

PR #1062, opened 2020-01-07
Add support for detecting `eol-last` from `.editorconfig`

rwjblue:


lib/rules/eol-last.js
```diff
@@ -25,8 +40,10 @@ module.exports = class EolLast extends Rule {
     let errorMessage = createErrorMessage(
       this.ruleName,
       [
-        '  * "always" - enforces that files end with a newline',
-        '  * "never" - enforces that files do not end with a newline',
+        '  * boolean -- `true` for inferring to require or disallow final newlines based your projects `.editorconfig` settings (via `insert_final_newline`)',
```
dcyriller: nit: in that case, would it be not better to be able to specify `{ eol-last: "editorconfig"` instead of `{ eol-last: true }`?

rwjblue: Ya, will update.

rwjblue: @dcyriller - Updated!

PR #1063, opened 2020-01-07
CI: Remove next branch

dcyriller:
This is a very small cleanup to prepare v2 release. As `next` branch is going to disappear.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


Turbo87:
does it hurt to keep it? could be useful for the v3.x release at some point 😉

dcyriller:
Okay, I didn’t see that far. I’m closing then ahah

dcyriller:
I’m giving it a second thought: maybe we want to cleanup unused code?

PR #1067, opened 2020-01-08
Remove deprecated `templateEnvironmentData` rule property

bmish:
Follow-up to https://github.com/ember-template-lint/ember-template-lint/pull/896.

### DIFF
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const chalk = require('chalk');
 const DEPRECATED_RULES = require('../helpers/deprecated-rules');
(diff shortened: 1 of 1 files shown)


bmish:
@rwjblue so the recommended replacement for this is just using `this. _moduleName` or `this._rawSource` directly right? Seems a bit strange for those to have underscore prefixes if they are considered public now.

dcyriller:
The recomended replacement would be to read from the Rule constructor's arguments. You can find an example here: https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/blob/v1.1.0/lib/rules/prettier.js

PR #1071, opened 2020-01-10
Update `no-invalid-interactive` to allow `change` event with `<form>` elements

lifeart:
ref:

image

usecase example:
https://codepen.io/lifeart/pen/XWJqMbV

### DIFF
--- a/lib/rules/no-invalid-interactive.js
+++ b/lib/rules/no-invalid-interactive.js
@@ -80,7 +80,11 @@ module.exports = class InvalidInteractive extends Rule {
           // Allow {{action "foo" on="reset"}} on form tags
           // Allow {{action "foo" on="submit"}} on form tags
           if (this._element.tag === 'form') {
-            if (isSubmitEventBinding(node) || isResetEventBinding(node)) {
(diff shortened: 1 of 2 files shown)


lib/rules/no-invalid-interactive.js
```diff
@@ -80,7 +80,11 @@ module.exports = class InvalidInteractive extends Rule {
           // Allow {{action "foo" on="reset"}} on form tags
           // Allow {{action "foo" on="submit"}} on form tags
           if (this._element.tag === 'form') {
-            if (isSubmitEventBinding(node) || isResetEventBinding(node)) {
+            if (
+              isSubmitEventBinding(node) ||
+              isResetEventBinding(node) ||
+              isChangeEventBinding(node)
```
rwjblue: I'd like to unroll these small one off one-liner functions here, something like:

```js
if (
  isEventHandlerFor(node, 'submit') &&
  isEventHandlerFor(node, 'reset') &&
  isEventHandlerFor(node, 'change')
  ) {
}
```

PR #1072, opened 2020-01-12
Stop removing the extension from module ID

bobisjan:
This PR stops removing the `.hbs` extension from `moduleId`, which aligns implementation with the README now.

Test for EditorConfig shows that `[*.hbs]` section is not taken into account and `insert_final_newline` is `true` within resolved configuration.


rwjblue:
Hmm. I do think we need to stop removing the extension (future work will include supporting linting within .js / .ts files, for example).

bobisjan:
I've updated PR to stop removing extension, should update PR's title?

bobisjan:
@rwjblue, PR now targets master since next was merged recently  🎉, would you mind taking a look?

bobisjan:
Rebased, should be ready now ☺️.

rwjblue:
I think we can’t quite change this. I’d like to add `filePath` and keep setting `moduleId` as we are now. All of the internal references to `moduleId` would be updated to first check `filePath`, then we will deprecate using `moduleId` (when we deprecate ember-cli-template-lint).

What do you think @bobisjan?

rwjblue:
I’d like to see this land as a precursor before https://github.com/ember-template-lint/ember-template-lint/pull/1117 (which currently introduces `filePath` along with a few other things).

bobisjan:
@rwjblue, sounds good 👍, I’ll switch to `filePath` here.

test/acceptance/cli-test.js
```diff
@@ -430,6 +430,18 @@ describe('ember-template-lint executable', function() {
...
+    it('should exit without error and any console output', function() {
+      let result = run(['app/templates/application.hbs'], {
+        cwd: './test/fixtures/with-editorconfig',
+      });
+
+      expect(result.exitCode).toEqual(0, 'exits without error');
+      expect(result.stdout).toBeFalsy();
+      expect(result.stderr).toBeFalsy();
+    });
+  });
```
rwjblue: Would you mind leveraging the (newish) fixturify setup instead of adding new fixture files?

bobisjan: Yes, should I extend `without-errors` fixture with `.editorconfig`?

suchitadoshi1987:
I believe there are cases in `test/acceptance/public-api-test.js` test as well which might need to be updated as well for `moduleId` For eg: (https://github.com/ember-template-lint/ember-template-lint/blob/master/test/acceptance/public-api-test.js#L368)

test/acceptance/cli-test.js
rwjblue: Sorry for the confusion, I’m saying that we should just not use fixtures at all. I’ll drum up an example for you from the recent work in 2.0.

rwjblue: https://github.com/ember-template-lint/ember-template-lint/blob/4279eb30f9c7c20954951d26b74e1b34dcdb5b89/test/unit/get-config-test.js#L22

bobisjan: Thanks, I got it and will update later.

bin/ember-template-lint.js
```diff
@@ -119,10 +117,11 @@ function run() {
   }
 
   for (let relativeFilePath of filesToLint) {
-    let filePath = path.resolve(relativeFilePath);
-    let fileName = relativeFilePath === STDIN ? filePathFromArgs : relativeFilePath;
-    let moduleId = fileName.slice(0, -4);
-    let fileErrors = lintFile(linter, filePath, moduleId, fix);
+    let resolvedFilePath = path.resolve(relativeFilePath);
+    let toRead = resolvedFilePath === STDIN ? process.stdin.fd : resolvedFilePath;
```
rwjblue: 👍 this is a really nice bit of cleanup, thank you!

PR #1078, opened 2020-01-15
Add `modifier-name-case` rule

mansona:
Adding a rule that makes sure you always use the dasherized-case form for modifier invocations.

### DIFF
--- a/docs/rule/modifier-name-case.md
+++ b/docs/rule/modifier-name-case.md
@@ -1,0 +1,23 @@
+## modifier-name-case
+
+It is currently possible to invoke a modifier with multiple words in its name
+using camelCase: `{{didInsert}}` or using dasherized-case: `{{did-insert}}`.
+This means that you can potentially have a lot of inconsistency throughout your
+codebase.
+
+This rule enforces that you will always use the dasherized-case form of the
+modifier invocation.
+
+### Examples
+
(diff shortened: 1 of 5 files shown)


Turbo87:
@rwjblue any objections?

dcyriller:
Should it be added to octane set? (It would have to target next branch then)

bmish:
@dcyriller I'm fine with adding it to the `octane` preset, but it would still be best for us to add the rule to master so we can release it ahead of v2 if we want.

Turbo87:
I think it should even be in recommended, not just octane

rwjblue:
TBH, I'm not 100% sure this is the correct direction for recommended/octane configuration. Specifically, we will soon (hahaha "soon") have template imports and they **cannot** have hyphenated names (because it will use normal JS import syntax, and you can't have hyphenated identifiers in JS). 

I think we should actually address this in ember-resolver instead (similar to what we've done for helpers, templates, and components [here](https://github.com/ember-cli/ember-resolver/blob/59d6e19cfd19c8f16f6f1cada771112da6231b9a/addon/resolvers/classic/index.js#L179)).

rwjblue:
I created https://github.com/ember-cli/ember-resolver/issues/506 to track fixing ember-resolver.

PR #1090, opened 2020-01-21
Update `no-curly-component-invocation` rule to ignore all usages of the `yield` helper

### LINKED ISSUE #1089: no-curly-component-invocation gives false positive on yield helper with arguments
In the following template, the first `{{yield}}` is allowed, the second is not.

```hbs
{{#if this.isEnabled}}
  {{yield}}
{{else}}
  {{yield to="inverse"}}
{{/if}}
```

I will try to create a failing test soon.

wagenet:
Fixes #1089.

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -39,10 +39,11 @@ const BUILT_INS = [
   'unbound',
   'unless',
   'with',
-  'yield',
   '-in-element',
   'in-element',
 ];
+
+const ALWAYS_CURLY = ['yield'];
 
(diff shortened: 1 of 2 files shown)


lib/rules/no-curly-component-invocation.js
```diff
@@ -44,6 +44,8 @@ const BUILT_INS = [
   'in-element',
 ];
 
+const ALWAYS_CURLY = ['yield'];
```
Turbo87: I guess we can remove it from line 42 now, right?

wagenet: Good call. I can do that.

PR #1102, opened 2020-02-05
Add new rule `no-heading-inside-button`

MelSumner:
This adds a new rule that disallows heading elements inside of button elements.


bmish:
Out of curiosity, are there other types of elements that are not allowed inside a button? Asking in case we wanted to make this more generic like: `no-disallowed-button-content`.

MelSumner:
@bmish great question! You're right in that most elements shouldn't be there- they will be disregarded by assistive technology. The button's `aria-label` value or the button's text value is the bit that screen readers are looking for (they'll ignore the rest). 

However, since it's atypical to see other elements nested inside of a button element, I kept it scoped to the specific issue we're trying to address- the top level headings are required for users with assistive tech, and nesting those inside of a button element breaks that for them. 

What do you think?

test/unit/rules/no-heading-inside-button-test.js
```diff
@@ -0,0 +1,85 @@
...
+generateRuleTests({
+  name: 'no-heading-inside-button',
+
+  config: true,
+
+  good: ['<button></button>', '<button><span></span></button>', '<button><div></div></button>'],
+
+  bad: [
+    {
+      template: '<button><h1></h1></button>',
```
Turbo87: can you add a case where the h1 in the button is nested in another element?

MelSumner: It won't catch it.

docs/rule/no-heading-inside-button.md
```diff
@@ -0,0 +1,30 @@
+## no-header-inside-button
+
+Assistive technology allows users to browse a page by heading elements (h1-h6). However, if those heading elements are nested inside of button elements, they will automatically be marked as presentational by browsers. Any HTML element where ("children presentational" is true)[https://w3c.github.io/aria/#button] should be coerced by the browser to be presentational, and therefore not included in the accessibility tree.
```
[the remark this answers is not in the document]
MelSumner: Can you clarify? it seems okay?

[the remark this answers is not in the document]
bmish: Check the rendered version. I think you flipped the parenthesis and brackets.

test/unit/rules/no-heading-inside-button-test.js
Turbo87: but I guess it probably should catch it, right?

I think we could implement this similar to what we did in https://github.com/ember-template-lint/ember-template-lint/pull/1105, and then it should catch that case too

MelSumner:
@bmish "no-invalid-interactive" catches nested interactive elements, so we should be fine there already.

test/unit/rules/no-heading-inside-button-test.js
```diff
@@ -0,0 +1,90 @@
...
+generateRuleTests({
+  name: 'no-heading-inside-button',
+
+  config: true,
+
+  good: [
+    '<button></button>',
+    '<button><span></span></button>',
+    '<button><div></div></button>',
+    '<div></div>',
```
bmish: Can you add `<div><h1></h1></div>` as a valid test case to make sure headings are allowed inside other elements?

docs/rule/no-heading-inside-button.md
```diff
@@ -0,0 +1,30 @@
+## no-heading-inside-button
```
rwjblue: Ha! I always called these "header" elements, TIL!

docs/rule/no-heading-inside-button.md
```diff
@@ -0,0 +1,30 @@
...
+
+Assistive technology allows users to browse a page by heading elements (h1-h6). However, if those heading elements are nested inside of button elements, they will automatically be marked as presentational by browsers. Any HTML element where ["children presentational" is true](https://w3c.github.io/aria/#button) should be coerced by the browser to be presentational, and therefore not included in the accessibility tree.
```
rwjblue: > if those heading elements are nested inside of button elements, they will automatically be marked as presentational by browsers

1) can you add a link that states this?
2) does this apply to "heading inside of buttons" or does it apply to "heading inside of interactive elements"?

docs/rule/no-heading-inside-button.md
```diff
@@ -0,0 +1,30 @@
...
+
+This rule **allows** the following:
+
+```hbs
+<button><span>Button Text</span></button>
+```
+
+### Migration
+
+* Replace h1-h6 elements inside of buttons with classes that reflect the same styling.
```
rwjblue:
```suggestion
* Replace <h1> - <h6> elements inside of buttons with classes that reflect the same styling.
```

lib/rules/no-action-modifiers.js
```diff
@@ -33,39 +33,22 @@ module.exports = class NoActionModifiers extends Rule {
   }
 
   visitor() {
-    this._element = null;
```
rwjblue: These changes seem unrelated to the rule being added, please revert them (possibly separating to another PR if applicable)

lib/rules/no-bare-strings.js
```diff
@@ -75,6 +75,7 @@ const DEFAULT_CONFIG = {
     '&bull;', // •
     '&bullet;', // •
     '&mdash;', // —
+    '&ndash;', // –
```
rwjblue: The changes in this file are unrelated to this PR, please revert

lib/rules/no-curly-component-invocation.js
```diff
@@ -65,20 +65,12 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
...
-          inAttrNode = true;
-        },
-        exit() {
-          inAttrNode = false;
-        },
-      },
-
       // eslint-disable-next-line complexity
-      MustacheStatement(node) {
+      MustacheStatement(node, visitorPath) {
```
rwjblue: The changes in this file are unrelated to this PR, please revert

test/unit/rules/no-bare-strings-test.js
```diff
@@ -46,6 +46,7 @@ generateRuleTests({
     '{{t "foo"}}, {{t "bar"}} ({{length}})',
     '(),.&+-=*/#%!?:[]{}',
     '&lpar;&rpar;&comma;&period;&amp;&nbsp;',
+    '&mdash;&ndash;',
```
rwjblue: The changes in this file are unrelated to this PR, please revert

yarn.lock
```diff
@@ -18,28 +18,34 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
-"@glimmer/interfaces@^0.47.4":
-  version "0.47.4"
-  resolved "https://registry.yarnpkg.com/@glimmer/interfaces/-/interfaces-0.47.4.tgz#17772fae66571b606f9fa26d733d7211960a2560"
-  integrity sha512-Wbv6954fIOy1aWHh2IdneBmIGrGysyu+ZWrmAiuMQC92uf57a9AXFJW0DFY+X6gd8Biaw/Nx8EHE0bNeMYCp9Q==
+"@glimmer/env@0.1.7":
```
rwjblue: The changes in this file are unrelated to this PR, please revert

test/unit/rules/no-heading-inside-button-test.js
rwjblue: Confirm, we should ensure this is flagged:

```hbs
<button><div><div><h3></h3></div></div></button>
```

lib/rules/no-heading-inside-button.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports = class NoHeadingInsideButton extends Rule {
+  visitor() {
+    return {
+      ElementNode(node, path) {
+        // Only heading elements: check rule conditions
+        if (!HEADING_ELEMENTS.includes(node.tag)) {
+          return;
+        }
+        // if it's a heading, check to see if one of the parent elements is a button
+        if (HEADING_ELEMENTS.includes(node.tag)) {
```
rwjblue: I think we can remove this conditional (its a duplicate of the early return just above).

rwjblue: Just pushed a commit for that 😸

PR #1103, opened 2020-02-05
Add `new` CLI command for creating a new rule

MelSumner:
- Command is `yarn new rule-name` 
- it creates a new rule and the necessary files with basic template structure
- it prints messages to the console when the necessary files are created


dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,58 @@
...
+let newRuleName = process.argv[2];
+
+// Utility Filesystem: template files co-localized with this script
+let pathUtil = __dirname;
+let pathDocTemplate = path.join(pathUtil, 'new-rule-doc-template.md');
+let pathRuleTemplate = path.join(pathUtil, 'new-rule-template.js');
+let pathTestTemplate = path.join(pathUtil, 'new-rule-test-template.js');
+
+// Parent Filesystem: path/to/ember-template-lint
+let pathRoot = process.cwd();
```
Turbo87: this should probably use `__dirname` too to be more robust :)

dev/new-rule/new-rule-test-template.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
```
Turbo87: you might have to add these template files to the eslint ignore file to make CI happy 😉

dev/new-rule/new-rule-template.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+// *** Change template to 'Real Error Message to Report' ***
+const ERROR_MESSAGE = 'Error Message to Report';
+
+// *** Change template to `RealRuleName` ***
+module.exports = class MyNewRule extends Rule {
```
Turbo87: would it make sense to put a placeholder here instead of `MyNewRule` and replace it before writing the generated file?

dev/new-rule/new-rule.js
MelSumner: sure! learning moment for me- how does it make it more robust?

Turbo87: if you ran `yarn new foo` from a subfolder of the project it would create the files relative to that subfolder. but we do want it relative to the root folder of the project instead, which we can do via `path.join(__dirname, '..')` or something roughly like that.

bmish:
Can we move the templates into their own folder? That way, we can use simpler filenames for them:
* dev/new-rule/templates/doc.md
* dev/new-rule/templates/rule.js
* dev/new-rule/templates/test.js

And can you create a new file `docs/rule/_TEMPLATE_.md` that contains `Please see dev/new-rule/templates/doc.md to find the documentation template for new rules.` Why? Because I want to increase discoverability of this documentation template to increase the chance that developers follow it (especially if they don't know about this new CLI command).

dev/new-rule/new-rule-template.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+// *** Change template to 'Real Error Message to Report' ***
+const ERROR_MESSAGE = 'Error Message to Report';
```
bmish: Could you add `TODO:` prefixes in comments or messages that the developer will need to update? Using a common prefix like that makes it easier to search and recognize places that need to be updated (some IDEs even highlight TODOs).

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,58 @@
...
+let pathDocList = path.join(pathRoot, 'docs/', 'rules.md');
+let pathDocNew = path.join(pathRoot, 'docs/rule/', newRuleName + '.md');
```
bmish: When you use `path.join`, you should omit the slashes, just use: `path.join(pathRoot, 'docs', 'rules', newRuleName + '.md')` for example.

bmish:
Can you mention this command under the `Contributing` in the README?

dev/new-rule/new-rule-template.js
MelSumner: ok, did it. 👍

.eslintrc.js
```diff
@@ -113,6 +113,14 @@ module.exports = {
...
         'import/no-dynamic-require': 'off'
       }
+    },
+    {
+      files: ['dev/new-rule/templates/*.js'],
+      rules: {
+        'node/no-missing-require': 'off',
+        'no-unused-vars': 'off',
+        'no-unpublished-require': 'off',
+      }
```
rwjblue: Why do we need to override these rules?

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
```
rwjblue: I'd _vaguely_ prefer for this to be `dev/new-rule/index.js`

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,77 @@
...
+/* eslint-disable node/no-unsupported-features/node-builtins */
```
rwjblue: Why do we need these eslint overrides?

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,77 @@
...
+  // un-dasherize the input from the CLI(rule name)
+  let newRuleClassArray = newRuleName.split('-');
+  let newRuleClassArrayCaps = newRuleClassArray.map(element =>
+    element.replace(element.charAt(0), element.charAt(0).toUpperCase())
+  );
+  let newRuleClassName = newRuleClassArrayCaps.join('');
+  // search the rule file for the placeholder class name & replace it with the new class name
+  let outputContent = originalContent.replace('PlaceholderForRuleClass', newRuleClassName);
+  // write the modified template text to the rule file
+  fs.writeFileSync(pathRuleNew, outputContent);
```
rwjblue:
```suggestion
  fs.writeFileSync(pathRuleNew, outputContent, { encoding: 'utf8' });
```

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,77 @@
...
+}
+createNewRuleFile(newRuleName, pathRuleTemplate, pathRuleNew);
+
+// create the new rule test file
+fs.copyFileSync(pathTestTemplate, pathTestNew);
+// create the new rule doc file
+fs.copyFileSync(pathDocTemplate, pathDocNew);
+// adds the new rule doc file to the rule list in the docs
+fs.appendFileSync(pathDocList, docListAppend);
```
rwjblue: I don't think we can/should append directly to the bottom, I'd prefer this list of rules stay sorted alphabetically.

dev/new-rule/templates/doc.md
```diff
@@ -0,0 +1,53 @@
+## (rule name goes here)
```
rwjblue: Can you update the test suite that confirms new rules are created with all the right "things", to also check for these placeholders (and if found fail the test).

docs/rule/_TEMPLATE_.md
```diff
@@ -1,53 +1 @@
...
-
-* [related-rule-name1](related-rule-name1.md)
-* [related-rule-name2](related-rule-name2.md)
-
-### References
-
-* (link to relevant documentation goes here)
-* (link to relevant function spec goes here)
-* (link to relevant guide goes here)
+Please see dev/new-rule/templates/doc.md to find the documentation template for new rules.
```
rwjblue: Do we still need this file?

dev/new-rule/new-rule.js
```diff
@@ -0,0 +1,77 @@
...
+# /path/to/ember-template-lint
+$ yarn new my-new-rule
+$ yarn test my-new-rule
+*/
+
+const fs = require('fs');
+const path = require('path');
+
+// CLI Argument: `my-new-rule` (name of new rule to add)
+let newRuleName = process.argv[2];
```
rwjblue: Can you print out usage if `process.argv[2]` is missing or is `-h` or `--help`?

docs/rule/_TEMPLATE_.md
MelSumner: yes, since it's where people expect to find it (via @bmish earlier review)

dev/new-rule/index.js
```diff
@@ -0,0 +1,91 @@
...
+// Create rule file
+let pathRuleTemplate = path.join(pathTemplates, 'rule.js');
+let pathRuleNew = path.join(pathRoot, 'lib', 'rules', newRuleName + '.js');
+function createNewRuleFile(newRuleName, pathRuleTemplate, pathRuleNew) {
+  let originalContent = fs.readFileSync(pathRuleTemplate, { encoding: 'utf8' });
+  let newRuleClassArray = newRuleName.split('-');
+  let newRuleClassArrayCaps = newRuleClassArray.map(element =>
+    element.replace(element.charAt(0), element.charAt(0).toUpperCase())
+  );
+  let newRuleClassName = newRuleClassArrayCaps.join('');
```
bmish: Can you extract all this string manipulation into a separate function `toTitleCase(str)`?

dev/new-rule/new-rule.js
MelSumner: it's already done in `originalContent`

dev/new-rule/templates/doc.md
MelSumner: I'd like to split this into a different issue (more for my own sprint purposes) if that's okay.

MelSumner: https://github.com/ember-template-lint/ember-template-lint/issues/1119

dev/new-rule/index.js
bmish: I see you added a helper `toTitleCase` but haven't used it yet?

.eslintignore
```diff
@@ -0,0 +1 @@
+/dev/new-rule/templates/*
```
rwjblue: Should probably have a newline at the end of this file

.eslintrc.js
```diff
@@ -100,9 +100,10 @@ module.exports = {
 
   overrides: [
     {
-      files: ['bin/**/*.js'],
+      files: ['bin/**/*.js', 'dev/new-rule/index.js'],
       rules: {
         'no-console': 'off',
+        'no-unsupported-features': 'off',
```
rwjblue: I do not think we want to turn off `no-unsupported-features` in all of our bin scripts (and I'm not 100% sure why we'd want to turn it off for the new `dev/new-rule/index.js` either). 

Please revert this, and if we _do_ need an override for the new `dev/new-rule/index.js` please explain what feature is used that we are relying on, and add it in its own override section (so it does not allow us to accidentally use too new of features in the main bin script).

README.md
```diff
@@ -251,8 +251,14 @@ A few ideas for where to take this in the future:
 ### Installation
 
 * `git clone` this repository
-* `npm install`
+* `yarn install`
+
+### Creating A New Rule
```
rwjblue: Should be in a `CONTRIBUTING.md`, mind making an issue to track moving all of the "contributing related" items in the `README.md` into CONTRIBUTING.md?

dev/new-rule/helpers/to-title-case.js
```diff
@@ -0,0 +1,10 @@
+function toTitleCase(newRuleName) {
```
rwjblue: Should probably remove this helper file, it is not used (or update the actual script to use it)

dev/new-rule/index.js
```diff
@@ -0,0 +1,132 @@
...
+  2. a new test file will be created
+  3. a new doc file will be created
+  4. the rule will be appended to the list of rules in the index
+  5. the doc will be appended to the list of docs in the doc index
+
+  Example usage: 
+  --------------
+  Templates use the rule name \`my-new-rule\`, so this example
+  creates a functioning/immediately testable rule
+  # /path/to/ember-template-lint
```
rwjblue: Not sure what this line is telling us, maybe remove it?

docs/rule/_TEMPLATE_.md
```diff
@@ -1,53 +1,3 @@
...
-
-* [related-rule-name1](related-rule-name1.md)
-* [related-rule-name2](related-rule-name2.md)
-
-## References
-
-* (link to relevant documentation goes here)
-* (link to relevant function spec goes here)
-* (link to relevant guide goes here)
+Please see dev/new-rule/templates/doc.md to find the documentation template for new rules.
```
rwjblue: Can you make this a link?

package.json
```diff
@@ -74,7 +75,8 @@
   "jest": {
     "coveragePathIgnorePatterns": [
       "<rootDir>/node_modules/",
-      "<rootDir>/test/"
+      "<rootDir>/test/",
+      "<rootDir>/dev/new-rules/**"
```
rwjblue: Hmm, why?

README.md
MelSumner: https://github.com/ember-template-lint/ember-template-lint/issues/1164

.eslintrc.js
MelSumner: reverted it. sorry about that.

MelSumner:
I'm withdrawing this PR for now, the idea needs more work and it feels a little scattered. I'll resubmit the PR when I'm happier with the result. Thank you! (and sorry!)

PR #1105, opened 2020-02-06
Add new rule: `require-input-label`

MelSumner:
This adds a new rule that provides a minimum check to see if an input element has met the following conditions: 

1. it's wrapped in a label element
2. it has one of these three attributes: id, aria-label, or aria-labelledby


lib/rules/require-input-label.js
```diff
@@ -0,0 +1,53 @@
...
+              message: ERROR_MESSAGE,
+              line: node.loc && node.loc.start.line,
+              column: node.loc && node.loc.start.column,
+              source: this.sourceForNode(node),
+            });
+          }
+        },
+        exit(node) {
+          if (node.tag === 'label') {
+            this._parentStack.pop();
```
Turbo87: why are we only popping from the parent stack if it's a label?

Turbo87:
I couldn't help myself:

Bildschirmfoto 2020-02-07 um 10 27 29

I think this is what you initially wanted to do?

docs/rule/require-input-label.md
```diff
@@ -0,0 +1,48 @@
...
+This rule **forbids** the following:
+
+```hbs
+<div><input /></div>
+```
+
+This rule **allows** the following:
+
+```hbs
+<label><input /></label>
```
bmish: Shouldn't this example actually have some label text?

test/unit/rules/require-input-label-test.js
```diff
@@ -0,0 +1,34 @@
...
+    '<label>LabelText<Input /></label>',
+    '<label><Input />LabelText</label>',
+    '<label><div><input /></div></label>', // technically okay, hopefully no one does this though
+    '<input id="probablyHasLabel" />', // it's likely to have an associated label if it has an id attribute
+    '<input aria-label={{labelText}} />',
+    '<input aria-labelledby="someIdValue" />',
+  ],
+
+  bad: [
+    {
```
bmish: Can you add a lone `<input />` element as a test case?

test/unit/rules/require-input-label-test.js
```diff
@@ -0,0 +1,34 @@
...
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const ERROR_MESSAGE = require('../../../lib/rules/require-input-label').ERROR_MESSAGE;
+
+generateRuleTests({
+  name: 'require-input-label',
+
+  config: true,
+
+  good: [
```
bmish: Can you add a test case of `<div>` to make sure unrelated elements are ignored?

lib/helpers/iterate-parents.js
```diff
@@ -0,0 +1,8 @@
+function* iterateParents(path) {
```
rwjblue: I think we can remove this now that https://github.com/glimmerjs/glimmer-vm/pull/1024 landed and is released.

docs/rule/require-input-label.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+```hbs
+<input ...attributes />
+```
+
+### Migration
+
+* the simplest fix is to add an aria-label to the input element. 
+* the most robust fix is to add an associated label element and use CSS to visually hide the label while still making it available to screen readers.
```
rwjblue: I think this suggestion is a bit off. It is assuming that no label previously existed _on purpose_ (e.g. that the label display was unwanted). I think that is much less common than "I just forgot", and the simplest / most semantic solution is to add the label (leaving it displayed)...

lib/helpers/iterate-parents.js
rwjblue: This utility needs to be removed, and update the internals of the rule to use `path.parents` directly

lib/rules/require-input-label.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        // OR:
+        // Parental validation (descriptive elements)
+        let parents = Array.from(path.parents());
+        if (
+          parents.some(parent => parent.node.type === 'ElementNode' && parent.node.tag === 'label')
+        ) {
+          return;
+        }
```
rwjblue: Can you extract this to a new `AstNodeInfo.hasParent(path, filterFunction)`, which would be used here like:

```js
let hasLabelParent = AstNodeInfo.hasParent(path, parent => parent.node.type === 'ElementNode' && parent.node.tag === 'label');
if (hasLabelParent) {
  return;
}
```

test/unit/rules/require-input-label-test.js
```diff
@@ -0,0 +1,56 @@
...
+    '<label><Input />LabelText</label>',
+    '<label><div><input /></div></label>', // technically okay, hopefully no one does this though
+    '<input id="probablyHasLabel" />', // it's likely to have an associated label if it has an id attribute
+    '<input aria-label={{labelText}} />',
+    '<input aria-labelledby="someIdValue" />',
+    '<div></div>',
+    '<input ...attributes/>', // we are unable to correctly determine if this has a label or not, so we have to allow it
+  ],
+
+  bad: [
```
rwjblue: Please add additional "bad" examples for each of the following:

A label wrapping an input, but without label text:

```hbs
<label><input></label>
```

Usage of `{{input}}`:

```hbs
{{input}}
```

```hbs
<div>{{input}}</div>
```

Usage of `<Input/>`:

```hbs
<Input />
```

test/unit/rules/require-input-label-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+const ERROR_MESSAGE = require('../../../lib/rules/require-input-label').ERROR_MESSAGE;
+
+generateRuleTests({
+  name: 'require-input-label',
+
+  config: true,
+
+  good: [
```
rwjblue: Please add additional "good" tests for:

```hbs
<Input ...attributes>
<Input id="foo" />
{{input id=someThing}}
<label>Text here<Input /></label>
```

docs/rule/require-input-label.md
MelSumner: Please feel free to suggest a specific edit.

docs/rule/require-input-label.md
```diff
@@ -0,0 +1,60 @@
+## require-valid-label
+
+Users with assistive technology need `<input>` elements to have associated labels. It's so essential for users, that failure to provide an associated label for an `<input>` element will cause failure of three different WCAG success criteria: [1.3.1, Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html), [3.3.2, Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html), and [4.1.2, Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html). It is also associated with common failure [#68: Failure of Success Criterion 4.1.2 due to a user interface control not having a programmatically determined name](https://www.w3.org/WAI/WCAG21/Techniques/failures/F68).
```
MelSumner:
```suggestion
Users with assistive technology need `<input>` elements to have associated labels. It's so essential for users that failure to provide an associated label for an `<input>` element will cause failure of **three** different WCAG success criteria: [1.3.1, Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html), [3.3.2, Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html), and [4.1.2, Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html). It is also associated with common failure [#68: Failure of Success Criterion 4.1.2 due to a user interface control not having a programmatically determined name](https://www.w3.org/WAI/WCAG21/Techniques/failures/F68).
```

lib/rules/require-input-label.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+// const iterateParents = require('../helpers/iterate-parents');
```
rwjblue: we should remove this comment

MelSumner:
```suggestion
```

PR #1106, opened 2020-02-07
require-input-label: Use `path` argument to iterate over parent elements

Turbo87:
see https://github.com/ember-template-lint/ember-template-lint/pull/1105#issuecomment-583306588 😉


Turbo87:
https://github.com/glimmerjs/glimmer-vm/pull/1024 should make this even easier :)

PR #1107, opened 2020-02-07
Simplify a few rules by using the `path` visitor argument

Turbo87:
see https://github.com/glimmerjs/glimmer-vm/pull/985

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -65,20 +65,12 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
   }
 
   visitor() {
-    let inAttrNode = false;
-
     return {
-      AttrNode: {
-        enter() {
-          inAttrNode = true;
-        },
-        exit() {
-          inAttrNode = false;
-        },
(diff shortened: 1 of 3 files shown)


lib/rules/no-curly-component-invocation.js
```diff
@@ -211,6 +202,14 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
       },
     };
   }
+
+  findAttrParent(path) {
```
bmish: You could extract a `findFirstParentOfType(path, type)` function into a helper file so it could be reused across rules.

Turbo87: I've found a simpler way. Would be awesome if you could have another look 🙏

bmish: LGTM.

rwjblue:
Unfortunately, our @dependabot setup was wrong and we didn't bump minimums in `package.json`. Given certain lockfiles I think this PR would cause issues. Going to bump to max allowed versions of dependencies and fix dependabot config in a separate PR.

PR #1108, opened 2020-02-08
Update `no-bare-strings` rule to allow `&ndash;`

jrjohnson:
The en dash is a slightly smaller typeset version of em dash (mdash).

### DIFF
--- a/lib/rules/no-bare-strings.js
+++ b/lib/rules/no-bare-strings.js
@@ -75,6 +75,7 @@ const DEFAULT_CONFIG = {
     '&bull;', // •
     '&bullet;', // •
     '&mdash;', // —
+    '&ndash;', // –
     '&nbsp;', // non-breaking space
     '&Tab;',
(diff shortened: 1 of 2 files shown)


rwjblue:
Hmm, I wonder if we should just allow all HTML Entities?

@bmish / @Turbo87 - Thoughts?

jrjohnson:
Yes please!

Turbo87:
tbh I think I'd prefer to forbid them all, unless someone can convince me why those should not be translated too

jrjohnson:
How about allowing them all and adding a `no-html-entities` rule as well for anyone who wants to be extra strict about this, or is working in locales that require translation of these?

bmish:
I'd recommend adding an `ignoreHtmlEntities` option to this existing `no-bare-strings` rule.

rwjblue:
@boris-petrov created #1131 for us to continue the discussion

PR #1115, opened 2020-02-17
Add configuration file support for overrides

suchitadoshi1987:
The `overrides` option would enable the following:

- Allows the user to pass an array `glob` patterns
- Enable overriding of rules for a specific set of files
- Allows to change the severity of  rules for a given file glob pattern

It currently contains 3 keys `files, rules, severity`.

`files`: Array of glob patterns
`rules`: An object of specific lint rules that the user wants to override for the specific files
`severity`: Specifies the level of severity for the overrides (1 -> `warning`, 2 -> `error`) (default is 2)

Supports cases like:

```
'use strict';

module.exports = {
  extends: 'recommended',
  overrides: [
    {
      files: ['**/item-description/**/templates/components/**/*.hbs'],
      rules: {
        'no-implicit-this': true,
        'no-bare-strings': true
      },
      severity: 1
    },
    {
      files: ['**/item-main/**/templates/components/**/*.hbs'],
      rules: {
        'no-input-block': true,
      },
    },
  ],
};

```

```
'use strict';

module.exports = {
  extends: 'recommended',
  rules: {
    'no-bare-strings': true,
  },
  overrides: [
    {
      files: ['**/item-description/**/templates/components/**/*.hbs'],
      rules: {
        'no-implicit-this': true,
        'no-bare-strings': false
      },
      severity: 1
    },
  ],
};
```
Fixes: https://github.com/ember-template-lint/ember-template-lint/issues/588


lib/get-config.js
```diff
@@ -302,7 +303,11 @@ module.exports = function(options) {
     processExtends(config, options);
     processIgnores(config);
 
-    validateRules(config, options);
+    validateRules(config.rules, config.loadedRules, options);
+
+    for (const override of config.overrides || []) {
+      validateRules(override.rules, config.loadedRules, options);
+    }
```
rwjblue: I'd generally prefer to refactor to something like:

```js
if (config.overrides) {
  for (const override of config.overrides) {
    validateRules(override.rules, config.loadedRules, options);
  }
}
```

bin/ember-template-lint.js
```diff
@@ -15,23 +15,7 @@ function lintFile(linter, filePath, moduleId) {
...
-      .sync(item, {
-        ignore: ['**/dist/**', '**/tmp/**', '**/node_modules/**'],
-        gitignore: true,
-      })
-      .filter(filePath => filePath.slice(-4) === '.hbs')
-      .forEach(filePath => result.add(filePath));
-  });
-
-  return result;
+  return linter.verify({ source, moduleId, filePath });
```
rwjblue: 👍 - This is good. To date, we have used `moduleId` but that unfortunately strips the extension. In the future we want to use the full file path (because linting within `.js` files should be possible). This sets us up nicely for that...

lib/index.js
```diff
@@ -139,11 +154,25 @@ class Linter {
...
+    return [];
+  }
+
   verify(options) {
     let messages = [];
     let pendingStatus = this.statusForModule('pending', options.moduleId);
     let shouldIgnore = this.statusForModule('ignore', options.moduleId);
 
+    let overrides = JSON.parse(JSON.stringify(this.config.overrides));
+    let overrideRules = this.getRulesForOverrides(overrides);
```
rwjblue: Hmm, I don't think we want to glob the file system for every override. In fact, I don't think we actually should hit `globby` at all in this case.  For very large projects that have a decent number of `overrides`, this would currently result in a non-trivial slow down (due to many many more I/O operations).

Specifically, every time `verify` is called we know the full path that we are linting and there is no reason to run globby again to see if a given full path is included within a given array of globs. Take a look at [micromatch's isMatch](https://github.com/micromatch/micromatch#ismatch) function, which should let us do what we want without additional I/O.

lib/index.js
```diff
@@ -162,7 +191,9 @@ class Linter {
     }
 
     let rules = this.buildRules({
```
rwjblue: I would prefer `buildRules` do the logic to decide the final rule configuration to use (instead of passing in `overrideRules`)

bmish:
Can you document this in the README.md?

suchitadoshi1987:
Closing this PR in favor of https://github.com/ember-template-lint/ember-template-lint/pull/1117

PR #1116, opened 2020-02-18
no-invalid-interactive: add support for the tablist role

Windvis:
This makes it possible to add keydown event handlers to elements with
the `tablist` role which is a requirement if you want to create a fully
keyboard accessible tab component.

reference: https://www.w3.org/TR/wai-aria-practices/#tabpanel


Windvis:
Hmm, seems like this was part of https://github.com/ember-template-lint/ember-template-lint/pull/984 already.

PR #1117, opened 2020-02-19
Add configuration file support for overrides 

suchitadoshi1987:
The `overrides` option would enable the following:

- Allows the user to pass an array `glob` patterns
- Enable overriding of rules for a specific set of files

It currently contains 2 keys `files, rules`.

`files`: Array of glob patterns
`rules`: An object of specific lint rules that the user wants to override for the specific files

Supports cases like:

```
'use strict';

module.exports = {
  extends: 'recommended',
  overrides: [
    {
      files: ['**/item-description/**/templates/components/**/*.hbs'],
      rules: {
        'no-implicit-this': true,
        'no-bare-strings': true
      }
    },
    {
      files: ['**/item-main/**/templates/components/**/*.hbs'],
      rules: {
        'no-input-block': true,
      },
    },
  ],
};

```

```
'use strict';

module.exports = {
  extends: 'recommended',
  rules: {
    'no-bare-strings': true,
  },
  overrides: [
    {
      files: ['**/item-description/**/templates/components/**/*.hbs'],
      rules: {
        'no-implicit-this': false,
        'no-bare-strings': false
      },
    },
  ],
};
```
Fixes: https://github.com/ember-template-lint/ember-template-lint/issues/588


Turbo87:
I generally like the idea, but it seems strange to me that the `severity` is a sort of global thing. IMHO it would be better to define it per-rule as ESLint does it.

rwjblue:
> IMHO it would be better to define it per-rule as ESLint does it.

Agree.

suchitadoshi1987:
@rwjblue @Turbo87  great point and 100% agreed, having per rule severity customization makes more sense. I have updated the PR to support that.

rwjblue:
I’d like to land https://github.com/ember-template-lint/ember-template-lint/pull/1072 first (helps split this PR up a bit also).

README.md
```diff
@@ -42,13 +42,13 @@ var results = linter.verify({ source: template, moduleId: 'template.hbs' });
...
-* `moduleId` - The module path for the file containing the error.
-* `source` - The source that caused the error.
-* `fix` - An object describing how to fix the error.
+- `rule` - The name of the rule that triggered this warning/error.
+- `message` - The message that should be output.
+- `line` - The line on which the error occurred.
+- `column` - The column on which the error occurred.
+- `moduleId` - The module path for the file containing the error.
+- `source` - The source that caused the error.
+- `fix` - An object describing how to fix the error.
```
rwjblue: Seems like you ran prettier on this file, can you revert the changes that are unrelated to your PR here?

I'd love to have a separate PR to make this file formatted how prettier wants it to be, but I want to keep stylistic changes separate from those that introduce functionality...

lib/get-config.js
```diff
@@ -322,7 +323,13 @@ module.exports = function(options) {
...
     processIgnores(config);
 
-    validateRules(config, options);
+    validateRules(config.rules, config.loadedRules, options);
+
+    if (config.overrides) {
+      for (const override of config.overrides) {
+        validateRules(override.rules, config.loadedRules, options);
+      }
+    }
```
rwjblue: Overrides should be able to do _basically_ anything that is done above. Some things that seem missing:

* extends
* pending
* ignore
* specifying plugins
* specifying rule configuration

I think it would _probably_ be fine to prevent `extends` and `plugins` for a first pass, but we'd need to give a nice error if overrides contained anything other than `rules` property for now.

lib/linter.js
```diff
@@ -108,6 +109,14 @@ class Linter {
     return rule;
   }
 
+  findRulesForPath(overrides, filePath) {
```
rwjblue: * please underscore (don't want to accidentally infer public api or anything)
* the name seems slightly off, we aren't finding the rules we are finding the specific configuration for this filePath
* Can you make a utility function (akin to `get-config`) that takes the current config and a filePath, and returns the "final config" for that file? That would replace this helper method, and much of the changes down in `buildRules` with something like `let configuredRules = getConfigForFile(this.config, options.filePath);`,  and we'd use `configuredRules` throughout (instead of `this.config`).

lib/rules/base.js
```diff
@@ -75,6 +76,54 @@ module.exports = class BaseRule {
     return name === this.ruleName || DEPRECATED_RULES.get(name) === this.ruleName;
   }
 
+  _determineConfigForSeverity(config) {
```
rwjblue: Let's make this and the method just below utility methods (not on the base class)

lib/rules/base.js
```diff
@@ -75,6 +76,54 @@ module.exports = class BaseRule {
...
+  processConfigForSeverity(configData) {
+    let severityConfig;
+    // In case of {'no-implicit-this': 'off|warn|error'}
+    if (typeof configData === 'string') {
+      severityConfig = this._determineConfigForSeverity(configData);
+      return severityConfig !== undefined ? severityConfig : configData;
+    }
+
+    // In case of object based config with or without severity information.
+    if (typeof configData === 'object') {
```
rwjblue: I don't know exactly how to make this change compatibly. It is already possible to have a rule that accepts a string, object, array, set, etc (anything really)...

test/fixtures/with-errors/.template-lintrc.js
```diff
@@ -1,5 +1,13 @@
 module.exports = {
   rules: {
-    'no-bare-strings': true
-  }
+    'no-bare-strings': true,
+  },
+  overrides: [
```
rwjblue: I'd prefer to remove the fixture usage, in favor of the (relatively newish) project helpers landed in #1151. Removing the fixtures should be small "one off" PRs though.

lib/get-config.js
suchitadoshi1987: agreed... this first pass is just for the overrides for `rules`. I will add support for others in the next PR

rwjblue: Sounds good, but it is very important that we add an error if you add "random properties" other than `rules` (otherwise we loose the ability to add new properties in the future without it being a breaking change).

lib/linter.js
```diff
@@ -125,7 +125,8 @@ class Linter {
   buildRules(options) {
     let rules = [];
 
-    let configuredRules = this.config.rules;
+    let configuredRules = getRulesConfig(this.config, options.filePath);
+    options.configRules = configuredRules;
```
rwjblue:
```suggestion
    options.config = fileConfig;
```

lib/linter.js
```diff
@@ -125,7 +125,8 @@ class Linter {
   buildRules(options) {
     let rules = [];
 
-    let configuredRules = this.config.rules;
+    let configuredRules = getRulesConfig(this.config, options.filePath);
```
rwjblue:
```suggestion
    let fileConfig = getConfigForFile(this.config, options.filePath);
```

lib/linter.js
```diff
@@ -85,7 +85,7 @@ class Linter {
     let ruleOptions = Object.assign(
       {
         name: ruleName,
-        config: this.config.rules[ruleName],
+        config: options.configRules[ruleName],
```
rwjblue:
```suggestion
        config: options.config.rules[ruleName],
```

lib/get-config.js
```diff
@@ -322,10 +342,28 @@ module.exports = function(options) {
...
 
     config._processed = true;
   }
 
   return config;
-};
+}
+
+module.exports = { getConfig, getRulesConfig };
```
rwjblue:
```suggestion
module.exports = { getProjectConfig, getConfigForFile };
```

lib/get-config.js
```diff
@@ -303,7 +305,25 @@ function validateRules(config, options) {
...
+  let configuredRules = config.rules;
+  let overrides = config.overrides;
+  if (filePath && overrides) {
+    let overridesObj =
+      overrides.find(override => override.files && micromatch.isMatch(filePath, override.files)) ||
+      {};
+    // If there is an override present, then the overridden ruleset takes precedence.
+    return Object.assign({}, configuredRules, overridesObj.rules);
+  }
+  return configuredRules;
```
rwjblue: Let's return the full config object (remember we eventually want to support merging many things, not just rules)

lib/get-config.js
```diff
@@ -322,10 +342,28 @@ module.exports = function(options) {
...
+    if (config.overrides) {
+      for (const override of config.overrides) {
+        // If there are keys in the object which are not yet supported by overrides functionality, throw an error.
+        const overrideKeys = Object.keys(override);
+        const containsValidKeys = overrideKeys.every(item =>
+          SUPPORTED_OVERRIDE_KEYS.includes(item)
+        );
+        if (!containsValidKeys) {
+          throw new Error(
+            'Override object cannot contain anything other than the `files` and `rules` property.'
```
rwjblue:
```suggestion
            'Using `overrides` in `.template-lintrc.js` only supports `files` and `rules` sections. Please update your configuration.'
```

README.md
```diff
@@ -109,9 +109,9 @@ module.exports = {
   extends: 'recommended',
 
   rules: {
-    'no-bare-strings': true
-  }
-}
+    'no-bare-strings': true,
+  },
+};
```
rwjblue: Still need to revert the prettier changes here to the README

test/acceptance/cli-test.js
```diff
@@ -142,7 +142,10 @@ describe('ember-template-lint executable', function() {
           '  1:4  error  Non-translated string used  no-bare-strings',
           '  2:5  error  Non-translated string used  no-bare-strings',
           '',
-          '✖ 2 problems (2 errors, 0 warnings)',
+          path.resolve('./test/fixtures/with-errors/app/templates/components/foo.hbs'),
+          "  1:2  error  Ambiguous path 'fooData' is not allowed. Use '@fooData' if it is a named argument or 'this.fooData' if it is a property on 'this'. If it is a helper or component that has no arguments you must manually add it to the 'no-implicit-this' rule configuration, e.g. 'no-implicit-this': { allow: ['fooData'] }.  no-implicit-this",
+          '',
+          '✖ 3 problems (3 errors, 0 warnings)',
```
rwjblue: This additional error seems odd, is this the one added for override support?

I'd rather keep the tests here focused on the command line itself. This would mean adding some tests here for what the warning output looks like in the command line (both `--json` version and regular), then put the tests for merging `overrides` over in `getConfigForFile` (in the `get-config-test.js`). If we give the right configuration to rules, we know they will work.

lib/get-config.js
```diff
@@ -303,7 +305,25 @@ function validateRules(config, options) {
...
+ */
+function getConfigForFile(config, filePath) {
+  let configuredRules = config.rules;
+  let overrides = config.overrides;
+  if (filePath && overrides) {
+    let overridesObj =
+      overrides.find(override => override.files && micromatch.isMatch(filePath, override.files)) ||
+      {};
+    // If there is an override present, then the overridden ruleset takes precedence.
+    config.rules = Object.assign({}, configuredRules, overridesObj.rules);
```
rwjblue: I worry about mutating here, can we return a new object?

I'm thinking basically:

```suggestion
    config = Object.assign({}, config, {
      rules: Object.assign({}, configuredRules, overridesObj.rules),
    });
```

lib/get-config.js
```diff
@@ -322,10 +342,28 @@ module.exports = function(options) {
...
+          SUPPORTED_OVERRIDE_KEYS.includes(item)
+        );
+        if (!containsValidKeys) {
+          throw new Error(
+            'Using `overrides` in `.template-lintrc.js` only supports `files` and `rules` sections. Please update your configuration.'
+          );
+        }
+        validateRules(override.rules, config.loadedRules, options);
+      }
+    }
```
rwjblue: Let's extract this to a `validateOverrides` utility method. I'd like to keep `getProjectConfig` "simple" by having it just call a bunch of helper methods (makes it so you don't have to think about how all of the things work).

lib/linter.js
```diff
@@ -131,13 +131,14 @@ class Linter {
     let failures = [];
     let rules = [];
 
-    let configuredRules = this.config.rules;
+    let fileConfig = getConfigForFile(this.config, options.filePath);
+    let { loadedRules, rules: configuredRules } = fileConfig;
```
rwjblue: I'd rather avoid this destructuring (yeah, I know, but I can't reason about destructuring with aliases 🤕), would you mind updating the references below to use `fileConfig.rules` or `fileConfig.loadedRules`?

test/unit/get-config-test.js
```diff
@@ -467,9 +467,69 @@ describe('get-config', function() {
...
 
-    let actual = getConfig(config);
+    let actual = getProjectConfig(config);
 
     expect(Object.keys(actual.rules).length).toBeTruthy();
     expect(config).toEqual(cloned);
   });
 });
+
+describe('get-rules-config', function() {
```
rwjblue:
```suggestion
describe('getConfigForFile', function() {
```

rwjblue:
Thanks for working so hard on this!!

PR #1124, opened 2020-02-20
Add `modifier-name-case` rule to `recommended` configuration

bmish:
The rule was added recently in #1078.

[Rule doc](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/modifier-name-case.md).

CC: @mansona


rwjblue:
See https://github.com/ember-template-lint/ember-template-lint/pull/1078#issuecomment-589454545 and https://github.com/ember-cli/ember-resolver/issues/506.

Issue #1125, opened 2020-02-20
rule request: require "as" for block params


lifeart:
given:

```hbs
<MyComponent |a|>
   {{a}}
</MyComponent>
```
no lint issue, no compilation issue, only "unknown scope", component rendered itself, but not block inside

![image](https://user-images.githubusercontent.com/1360552/74983207-a28fd300-5446-11ea-9999-c8e11bb9258c.png)

![image](https://user-images.githubusercontent.com/1360552/74983226-af142b80-5446-11ea-8896-d612ee5c8579.png)

I think we should explictly warn if user has missed ` as ` before  block arguments.

rwjblue:
Conceptually a block param in this case is really just a specially formatted valueless attribute (in the AST). I definitely agree that the example you gave is pretty bad 🤔.

Frankly, I’d consider this a bug (that we don’t actually error during template compilation).

rwjblue:
Seems fine to add a rule (we should also look into making it fail during compilation though). Maybe call it `no-invalid-block-param-definition`?

rwjblue:
Closed by #1126

PR #1126, opened 2020-02-21
Add new rule `no-invalid-block-param-definition`

lifeart:
https://github.com/ember-template-lint/ember-template-lint/issues/1125

### DIFF
--- a/docs/rule/no-invalid-block-param-definition.md
+++ b/docs/rule/no-invalid-block-param-definition.md
@@ -1,0 +1,33 @@
+## no-invalid-block-param-definition
+
+Glimmer parser eats warn about incorrect Angle Component block definition. This behaviour may confuse developers and produce tricky bugs.
+
+### Examples
+
+This rule **forbids** the following:
+
+```hbs
+<MyComponent  |blockParam|>
+    {{blockParam}}
+</MyComponent>
+```
+
+```hbs
+<MyComponent  |blockParam>
+    {{blockParam}}
+</MyComponent>
+```
+
+```hbs
+<MyComponent  blockParam|>
+    {{blockParam}}
+</MyComponent>
+```
+
+This rule **allows** the following:
+
+```hbs
+<MyComponent as |blockParam|>
+    {{blockParam}}
+</MyComponent>
+```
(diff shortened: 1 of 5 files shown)


lib/rules/no-invalid-block-param-definition.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+const Rule = require('./base');
+
+function message(tag, msgs) {
+  return `Unexpected block usage. "<${tag} ... as |param1 param2|>" ${msgs.join(' ')}`;
```
rwjblue: Lets show them the actual change they need to make.

lib/rules/no-invalid-block-param-definition.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+module.exports = class NoInputBlock extends Rule {
```
rwjblue: Classname is wrong

lib/rules/no-invalid-block-param-definition.js
```diff
@@ -0,0 +1,59 @@
...
+    const source = this.sourceForNode(node);
+    const attributes = node.attributes.map(el => this.sourceForNode(el));
+    const modifiers = node.modifiers.map(el => this.sourceForNode(el));
+    const pureNodeSource = [...attributes, ...modifiers].reduce((result, src) => {
+      return result.replace(src, '');
+    }, source);
+
+    const blockStartIndex = pureNodeSource.indexOf('|');
+    const blockEndIndex = blockStartIndex + pureNodeSource.slice(blockStartIndex).indexOf('|');
+    const asKeywordIndex = pureNodeSource.indexOf(' as ');
```
rwjblue: I think it would just be `as `, and please add a "good" test case for

```hbs
<Foo
as |foo|></Foo>
```

docs/rule/no-invalid-block-param-definition.md
```diff
@@ -0,0 +1,33 @@
+## no-invalid-block-param-definition
+
+Glimmer parser eats warn about incorrect Angle Component block definition. This behaviour may confuse developers and produce tricky bugs.
```
rwjblue:
```suggestion
There are a few common failure scenarios when using angle bracket components that use block parameters. This rule attempts to identify these pitfalls.
```

PR #1132, opened 2020-02-21
--fix(require-button-type): Implement fixer

dcyriller:
This is an example for https://github.com/ember-template-lint/ember-template-lint/pull/1025

It is not meant to be merged as is.

The two interesting commits are the last ones: https://github.com/ember-template-lint/ember-template-lint/pull/1132/commits/ea18ddef953540c3201994417efc81eb5adb2121 and ad7f46f7390ab9eb22b3cfd2a970c7249080aa94

And on this line, you'll notice the fixer works: https://github.com/ember-template-lint/ember-template-lint/pull/1132/files#diff-71916b856e5d24a1d54c3bc564bf3b98R215 (according to the CI).

### DIFF
--- a/lib/rules/require-button-type.js
+++ b/lib/rules/require-button-type.js
@@ -4,13 +4,26 @@ const Rule = require('./base');
 
 const ERROR_MESSAGE = 'All `<button>` elements should have a valid `type` attribute';
 
+function addDefaultAttribute(env) {
+  let { builders: b } = env.syntax;
+
+  return {
+    ElementNode(node) {
+      let attribute = b.attr('type', b.text('button'));
+      node.attributes.push(attribute);
+      return node;
+    },
+  };
+}
+
 module.exports = class RequireButtonType extends Rule {
-  logNode({ node, message }) {
+  logNode({ node, message, fixer }) {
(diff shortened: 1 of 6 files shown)


lib/rules/require-button-type.js
```diff
@@ -25,7 +38,7 @@ module.exports = class RequireButtonType extends Rule {
 
         let typeAttribute = attributes.find(it => it.name === 'type');
         if (!typeAttribute) {
-          return this.logNode({ node, message: ERROR_MESSAGE });
+          return this.logNode({ node, message: ERROR_MESSAGE, fixer: addDefaultAttribute });
         }
```
Turbo87: I'm wondering if it would be easier to have a `this.shouldFix` property on the rule and apply the fix directly here, instead of having to use another visitor for it?

rwjblue: Ya, agree. As it stands this fixer seems like it would adding a `type="button"` attribute for **EVERY** element that exists in the source.  It is pretty common to have complex logic to decide if a given node is actually an error to be logged, and that same logic will be needed for the fixer...

PR #1133, opened 2020-02-21
Don't mention `ember-cli-template-lint` in README

dcyriller:


bmish:
Is ember-cli-template-lint considered deprecated at this point?

rwjblue:
Nearly, but I think we'd need to finish https://github.com/ember-cli/ember-cli/pull/9009 before officially marking the repo as deprecated.

Issue #1043, opened 2020-01-06
add ember-cli/ember-cli-app-version to no-curly-component-invocation


st-h:
the no-curly-component-invocation complains about invocations of 
```
{{app-version versionOnly=true}}
```
app-version actually seems to be a helper. 

As this is an addon maintained by ember-cli, I would like to propose to add it to the no-curly-component-invocation rule by default instead of making everybody who uses the addon add this to the configuration.

I don't have any experience with writing (or modifying) linters, but if this is just a matter of adding it to the `BUILT_INS` variable, I am happy to submit a PR.

lifeart:
Implemented here https://github.com/ember-template-lint/ember-template-lint/pull/1136

PR #1136, opened 2020-02-22
Update `no-curly-component-invocation` rule to ignore ember-cli-app-version usage

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/issues/1043

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -129,9 +129,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
           }
 
           if (BUILT_INS.includes(name)) {
-            // {{debugger}}
-            // {{outlet}}
-            // {{yield}}
+            // {{app-version versionOnly=true}}
             return;
           }
 
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -43,7 +43,7 @@ const BUILT_INS = new Set([
   'in-element',
 ]);
 
-const ALWAYS_CURLY = new Set(['yield']);
+const ALWAYS_CURLY = new Set(['yield', 'app-version' /* from ember-cli-app-version */]);
 
 module.exports = class NoCurlyComponentInvocation extends Rule {
   parseConfig(config) {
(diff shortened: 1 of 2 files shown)


lib/rules/no-curly-component-invocation.js
```diff
@@ -43,7 +43,7 @@ const BUILT_INS = [
   'in-element',
 ];
 
-const ALWAYS_CURLY = ['yield'];
+const ALWAYS_CURLY = ['yield', 'app-version' /* from ember-cli-app-version */];
```
rwjblue: I think this belongs in the list of built ins above, not this ALWAYS_CURLY listing.  ALWAYS_CURLY is about things that **must** be invoked in curlies, but this is a helper and can definitely be used in subexpression's etc.

lifeart: yeah, it was my first idea, but rule failing in this case (now it's fixed)

lib/rules/no-curly-component-invocation.js
```diff
@@ -126,6 +128,13 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
             return;
           }
 
+          if (BUILT_INS.includes(name)) {
+            // {{debugger}}
+            // {{outlet}}
+            // {{yield}}
```
Turbo87: this looks incorrect. this code path is only about invocationa with named arguments.

lifeart: ah, you mean examples in comments? will fix it, because it just copy-pasted from namedless example.

and without this block itself test for <https://github.com/ember-template-lint/ember-template-lint/pull/1136/files#diff-37bf2986269b4e055b1de08cc17fa5e1R44> failing

lib/rules/no-curly-component-invocation.js
```diff
@@ -126,6 +128,11 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
             return;
           }
 
+          if (BUILT_INS.includes(name)) {
```
rwjblue, line 131: `BUILT_IN.includes` is not a function (CI failure), likely because this is a set now

```suggestion
          if (BUILT_INS.has(name)) {
```

Turbo87: thought I changed that ... 🤔

Issue #1039, opened 2019-12-30
{{! template-lint-disable }} comment creates no-unused block params bug


Pixelik:
The following, is throwing a `no-unused block params` **template-lint error**, saying that `param` is not being used, which is not true:

```
<MyComponent .... as |param|>

  {{! template-lint-disable }}
  .....
  ....
  <MyOtherComponent .... @param={{param}} />
  .....
  ....
  {{! template-lint-enable }}

</MyComponent>
```

if I remove `{{! template-lint-disable }}` it works. However, I do need to use `{{! template-lint-disable }}` for some other part of my code. 

A temporary solution I found is to place `{{! template-lint-disable }}` _before_ `<MyComponent .... as |param|>`. But I shouldn't have to do that.

rwjblue:
So if I understand the issue correctly, you are saying that the scope tracking stops identifying used/unused block params when it hits `template-lint-disable`, right?

Would you mind submitting a failing test case for the `no-unused` block, so we can figure out a path forward?

rwjblue:
I am generally curious, what does eslint do in this circumstance?

rwjblue:
Closed by https://github.com/ember-template-lint/ember-template-lint/pull/1137

Pixelik:
Thank you for this. Apologies for not replying at the time, I didn't have the time and then i forgot 🙏

PR #1137, opened 2020-02-22
Prevent `no-unused-block-params` from flagging locals that are used within `{{template-lint-disable}}`ed content

lifeart:
ref https://github.com/ember-template-lint/ember-template-lint/issues/1039

tldr: if we disable template-lint, rules was skipped between disable-enable nodes, and we unable to "collect" local scope usages inside rule, we need to go deeper (in middleware).

Since partials almost deprecated, I left partials locals catching inside rule itself.


rwjblue (approved):
Overall, this seems good to me. I've merged master into the branch to kick off a new test run (has been a while, so I want to make sure there aren't things like linting failures before landing).

Issue #813, opened 2019-08-28
Question: Lint a workspace in VSCode?


edeis53:
When using VSCode workspace with multiple folders/apps, template linting doesn't run, however the .js files are linted properly.
eg.
```
My Workspace
     frontend-ember-app
     backend-ember-app
```
     
I've tried configuring the VSCode eslint extension to lint the entire workspace, the eslint server seems to pickup the config, but template linting doesn't run for the .hbs files.

What additional config is required?
```
  "settings": {
    "eslint.autoFixOnSave": true,
    "eslint.workingDirectories": [
      {
        "directory": "/Users/Me/FrontEndApp",
        "changeProcessCWD": true
      },
            {
        "directory": "/Users/Me/BackEndApp",
        "changeProcessCWD": true
      }
    ],
    "eslint.lintTask.enable": true,
    "eslint.lintTask.options": "-c /Users/Me/FrontEndApp/.eslintrc.json --ignore-path /Users/Me/FrontEndApp/.eslintignore ."
  }
```

```
[Info  - 2:10:18 PM] ESLint server is running.
[Info  - 2:10:19 PM] ESLint library loaded from: /Users/Me/FrontEndApp/node_modules/eslint/lib/api.js
```

**Reproduction**
-Init two ember apps.
-Add both app folders to a workspace in VS code.
-Add an empty `<img />` to application.hbs

**Expected Result**
should get a line error within the VScode editor for the line `img tags must have an alt attribute : ember-template-lint` but it doesn't happen.

rwjblue:
Hmm, can you make a demo repo (along with the commands you expect should work, but don't) that I can use to test?

edeis53:
Thanks for your attention @rwjblue .

I have made a reproduction of the issue with step-by-step instructions in the README.md here:
https://github.com/edeis53/ember-vscode-templatelint-issue/blob/master/README.md

Let me know if you need any clarification.

rwjblue:
That reproduction is perfect, thank you!

I haven't stepped through the specific code here but I suspect that the issue is the way we discover the config file. Specifically, we look for configuration ([see here](https://github.com/ember-template-lint/ember-template-lint/blob/333e2eafa634669237c65c5b0ce7c9520bae58f1/lib/get-config.js#L21)) from the current working directory upwards, but in Scenario B (from the README you linked) we actually need to try to look from the _file location_ upwards. AFAICT, that is what ESLint is doing as well.

[...]

PR #1138, opened 2020-02-22
Automatically discover `.template-lintrc.js` when running linter from a project subdirectory (e.g. a monorepo's package, with linting configuration in the project root).

lifeart:
ref: https://github.com/ember-template-lint/ember-template-lint/issues/813

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -20,43 +20,46 @@ const ALLOWED_ERROR_CODES = [
   'QUALIFIED_PATH_RESOLUTION_FAILED',
 ];
 
-function resolveConfigPath(configPath, pluginPath) {
-  let usedPath = configPath
-    ? path.resolve(process.cwd(), configPath)
-    : findUp.sync(CONFIG_FILE_NAME);
+function resolvePlugin(pluginName, fromConfigPath) {
+  // throws exception if not found
+  let pluginPath = resolve.sync(pluginName, { basedir: path.dirname(fromConfigPath) });
 
-  if (pluginPath) {
-    // throws exception if not found
-    pluginPath = resolve.sync(pluginPath, { basedir: path.dirname(usedPath) });
-  }
+  return require(pluginPath); // eslint-disable-line import/no-dynamic-require
+}
 
-  let filePath = pluginPath || usedPath;
-  // Making sure that the filePath exists
-  try {
-    require.resolve(filePath);
-  } catch (e) {
-    // nodejs don't have error code for it and throw:
-    // "The first argument to require.resolve must be a string. Recived null or undefined."
-    if (!filePath) {
-      // tests:
-      // uses an empty set of rules if no .template-lintrc is present
-      // should be able to instantiate without options
-      return {};
-    }
-    let isModuleMissingError = ALLOWED_ERROR_CODES.includes(e.code);
-    if (!isModuleMissingError) {
-      throw e;
-    }
+function resolveProjectConfig(specifiedConfigPath) {
+  let configPath;
 
-    if (configPath) {
+  if (specifiedConfigPath) {
+    configPath = path.resolve(process.cwd(), specifiedConfigPath);
+
+    try {
+      // Making sure that the filePath exists, before requiring it directly this is
+      // needed in order to ensure that we only squelch module missing errors for
+      // the config path itself (and not other files that the config path may
+      // require itself)
+      require.resolve(configPath);
+    } catch (e) {
+      let isModuleMissingError = ALLOWED_ERROR_CODES.includes(e.code);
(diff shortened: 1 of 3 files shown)


rwjblue:
@lifeart - Targets `next` because `find-up` requires newer node?

lib/get-config.js
```diff
@@ -19,7 +22,11 @@ const ALLOWED_ERROR_CODES = [
 ];
 
 function resolveConfigPath(configPath, pluginPath) {
-  let usedPath = path.resolve(process.cwd(), configPath || '.template-lintrc');
+  let usedPath = path.resolve(process.cwd(), configPath || CONFIG_FILE_NAME);
```
rwjblue: I think this might be slightly wrong, if we were provided a config path (e.g. from the bin script) we must use it **exactly** (no find-up). The only time find-up should be used is if we do not have a specified config path.

lib/get-config.js
```diff
@@ -107,7 +114,7 @@ function migrateRulesFromRoot(config, source, options) {
       chalk.yellow(
         [
           'Rule configuration has been moved into a `rules` property.',
-          'Please update your `.template-lintrc.js` file.',
+          `Please update your \`${CONFIG_FILE_NAME}\` file.`,
```
rwjblue: Mind rewriting this into template strings instead of the `[].join(' ')` thing (which was done to make nicer looking multi-line strings)?

lifeart: @rwjblue this is "autofixed" notation (because our eslint rules)

lifeart:
regarding "next" - sorry, little bit confused, should I change it to "master"?

lib/get-config.js
lifeart: fixed, but for me it looks like an "feature" because config path may be just file name, and we able to lookup "upper" non-standard file names

rwjblue: Config path must match a specific on disk file (when not absolute, then it would be relative to CWD). If we want another "config file name" thing, we can add it but that is a separate thing (and I'm not sure that we should 😸 ).

rwjblue:
> regarding "next" - sorry, little bit confused, should I change it to "master"?

No, not neccesarily. I just wanted to understand why it was `next`. Seems like two main reasons:

* Node version support of current find-up version
* Usage of fixturify for tests (which IIRC only exists on next)

Seems totally good to me!

lib/get-config.js
rwjblue: I think this still needs some changes, I'd expect it to look something like:

```js
let usedPath = configPath ? configPath : findUp.sync(CONFIG_FILE_NAME);
```

Basically, if a `configPath` was provided, use it verbatim; otherwise use findUp.

lifeart: image

for such change we have alot of issues

lifeart: @rwjblue we have to rewrite few tests to get new logic support and be sure it won't break any existing apps. current changes lands new feature "gracefully" (without regression) and if we trying to fix our initial logic new "feature" break alot of tests.

Is it possible to land "enhansment" and create new issue about clear resolution logic?
(or I'm just  panicking or doins something worng)

lifeart: @rwjblue 
failing tests:
*  uses an empty set of rules if no .template-lintrc is present
* should be able to instantiate without options
* accepts a fake console implementation

image

lifeart: looks like I found valid fix

```js
  try {
    if (!filePath) {
      // tests:
      // uses an empty set of rules if no .template-lintrc is present
      // should be able to instantiate without options
      return {};
    }
    require.resolve(filePath);
  } catch (e) {
```

lib/get-config.js
```diff
@@ -29,6 +34,12 @@ function resolveConfigPath(configPath, pluginPath) {
   let filePath = pluginPath || usedPath;
   // Making sure that the filePath exists
   try {
+    if (!filePath) {
+      // tests:
+      // uses an empty set of rules if no .template-lintrc is present
+      // should be able to instantiate without options
+      return {};
+    }
```
rwjblue: This case is supposed to be handled in the `catch` block (that is what the `if (configPath) {} else{ return {} }` bit is for

lib/get-config.js
rwjblue: Thanks for pushing up the change, I'll take a look also (maybe I'm misunderstanding something)

rwjblue: The logic sharing makes this a bit annoying, lemme do a small cleanup then I'll merge (I'll push to this branch directly)

lib/get-config.js
lifeart: logic moved to catch, but with notes:

image

PR #1140, opened 2020-02-22
Add markdownlint and fix violations

bmish:
Cleans up docs (especially regarding using consistent headers). Enforces markdownlint via linting/CI scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
   "bin": "./bin/ember-template-lint.js",
   "scripts": {
     "changelog": "lerna-changelog",
-    "lint": "yarn lint:js",
+    "lint": "yarn lint:js && yarn lint:docs",
+    "lint:docs": "markdownlint '**/*.md' -i CHANGELOG.md -i LICENSE.md -i node_modules",
     "lint:js": "eslint . --cache",
     "release": "release-it",
     "test": "jest"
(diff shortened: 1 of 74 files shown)


package.json
```diff
@@ -17,7 +17,8 @@
   "bin": "./bin/ember-template-lint.js",
   "scripts": {
     "changelog": "lerna-changelog",
-    "lint": "yarn lint:js",
+    "lint": "yarn lint:js && yarn lint:docs",
+    "lint:docs": "yarn markdownlint *.md dev/*.md docs/*.md docs/**/*.md -i CHANGELOG.md -i LICENSE.md",
```
rwjblue: For scripts that use bin scripts provided by packages, you do not have to invoke them with `yarn`. Specifically, in this case you would do:

```suggestion
    "lint:docs": "markdownlint *.md dev/*.md docs/*.md docs/**/*.md -i CHANGELOG.md -i LICENSE.md",
```

But also, is there any reason we can't use `markdownlint **/*.md` (e.g. why can't we run against all markdown files)?

rwjblue:
(overall huge 👍 on this)

package.json
bmish: Thanks, fixed both issues!

rwjblue:
Thank you @bmish, this is awesome!

PR #1148, opened 2020-02-22
Use npm-run-all to run multiple scripts.

rwjblue:
This adds a few improvements:

* Avoid's using `&&` in `lint` script (doesn't work properly on all shells / platforms)
* Makes sure that both lint tasks are ran (previously if `lint:js` failed, `lint:docs` would not run)
* Fixes CI to run `lint:docs` (using npm-run-all makes it harder to make the same mistake again)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,6 @@ jobs:
       - name: install dependencies
         run: yarn install --frozen-lockfile
 
-      - run: yarn lint:js
-
       - run: yarn test
 
   nodeX:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -35,8 +35,7 @@ jobs:
       - name: install dependencies
         run: yarn install --frozen-lockfile
 
-      - run: yarn lint:js
-
+      - run: yarn lint
```
bmish: Isn't this duplicating some work since `yarn test` already runs lint?

rwjblue: Good catch! Updated...

(I have another branch coming that makes two CI jobs, one for linting one for tests here and I mixed them up 😩)

PR #1149, opened 2020-02-22
experimental - hbs query

lifeart:
based on https://github.com/estools/esquery (but, it's unconfigurable, so, codebase copypasted)


rwjblue:
If we decide to move forward here, we should probably push this type of traversal down into ember-template-recast.

rwjblue:
Going to close this for now, in order to land we need to add the utility functionality to ember-template-recast (instead of here) and then expose it to rules here for use.

Note: I am a bit worried about the sheer volume of new code here, and I'm not 100% sure what the upside is...

PR #1158, opened 2020-02-24
Introduce rule visitor based fixing support.

rwjblue:
This is an alternative take on https://github.com/ember-template-lint/ember-template-lint/pull/1025.

The idea here is to leverage the rule instance's pre-existing visitor to along with checking the `this.mode` (will either be `fix` or `report`) to mutate the AST _during_ traversal.

This makes it _effectively_ the same as if the rule's visitor was passed to ember-template-recast's `transform`.

TODO:

- [x] Add test harness support for testing fixers
  - [x] Confirm that the fix is idempotent
  - [x] Allow specifying the input and output strings

Closes #1025

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const assert = require('assert');
-
 const Linter = require('../index');
 
 function parseMeta(item) {
@@ -225,6 +224,22 @@ module.exports = function generateRuleTests({
           let actual = linter.verify(options);
 
           assert.deepStrictEqual(actual, []);
+        });
+      }
+
+      if (item.fixedTemplate) {
+        testOrOnly(`\`${item.template}\` -> ${item.fixedTemplate}\``, function() {
+          let options = prepare(item.template, item.config);
+          let result = linter.verifyAndFix(options);
+
+          assert.deepStrictEqual(result.output, item.fixedTemplate);
+        });
+
+        testOrOnly(`${item.fixedTemplate}\` is idempotent`, function() {
+          let options = prepare(item.fixedTemplate, item.config);
+          let result = linter.verifyAndFix(options);
+
+          assert.deepStrictEqual(result.output, item.fixedTemplate);
         });
       }
     });
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -9,16 +9,6 @@ const MAX_AUTOFIX_PASSES = 10;
 const WARNING_SEVERITY = 1;
 const ERROR_SEVERITY = 2;
 
-/**
- * @param {Object} options
- * @param {string} options.source - The source code to fix.
- * @param {Object[]} messages - The lint messages.
- * @returns {string} output - The fixed source.
- */
-function applyFixes(options /* messages */) {
-  return options.source;
-}
-
 function buildErrorMessage(filePath, moduleId, error) {
   let message = {
     fatal: true,
@@ -225,7 +215,7 @@ class Linter {
       shouldVerify = false;
 
       if (iterations < MAX_AUTOFIX_PASSES) {
-        const output = applyFixes(options, messages);
+        const output = this._applyFixes(options, messages);
 
         // let's lint one more time if source was modified
         shouldVerify = output !== currentSource;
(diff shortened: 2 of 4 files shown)


rwjblue:
This is ready for review now. It adds support for testing fixing in the test harness (and sets up a nice little acceptance test to start adding tests for rule public apis) and confirms that simple/basic fixing works.

lib/helpers/rule-test-harness.js
```diff
@@ -67,6 +66,12 @@ function getVerifyOptions(template, meta) {
      '<img alt>',
    ]
  *
+ * @param  {Array}  [fix=[]]             - an array of items that describe an input and expected output for a rules --fix support
+ *  [{
+     input: '<button></button>',
+     output: '<button type="button"></button>',
+   }]
```
Turbo87: I think I would prefer to couple this with the `bad` list that we already have. if an element in `bad` has an `output` key then we would add tests that run `verifyAndFix()`.

rwjblue: Ya, I agree with you now that I’ve done it (and wrote that test plugin). We only attempt to do a fix if you log errors so thing them together makes a lot of sense. When I made the plugin in this PR (in the test suite), I fought with it for a few minutes trying to figure out why it wasn’t running, turns out I was only attempting to mutate and never emitted an error 😩. 

The names there are just a bit off (`template` -> `output`), but I’ll make it work (I might rename `output` to `fixedTemplate` or something).

test/acceptance/rule-test.js
```diff
@@ -0,0 +1,55 @@
...
+        },
+      ],
+
+      name: 'can-fix',
+      config: true,
+
+      fix: [
+        {
+          input: '<MySpecialThing/>',
+          output: '<EvenBettererThing/>',
```
dcyriller: With a few examples, the test files might become quite long quite fast. Should we use Jest snapshots for this?

We could change the tests file structure from:
```
tests
├── unit
    ├── rules
        ├── quotes-test.js
```

to:
```
tests
├── unit
     ├── rules
          ├──  quotes
                 ├── linter-test.js
                 ├── a-collection-of-hbs-snippets-to-fix.hbs
                 ├── jest-config-file
                 └── __snapshots__
```

lib/linter.js
```diff
@@ -242,6 +232,46 @@ class Linter {
...
+        {
+          pending,
+          shouldFix: true,
+          filePath: options.filePath,
+          rawSource: currentSource,
+        },
+        () => {}
+      );
+
+      let { code } = transform(currentSource, () => rule.getVisitor());
```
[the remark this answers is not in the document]
rwjblue: The order is based on the ordering of message output from `.verify`. Basically we do this:

* determine the rule names for all the errors that have `isFixable` set
* loop through those rule names and run the fixer
* each rule runs independently (this ensures that two fixers cannot muck with each others AST)
  * `currentSource` is parsed, then passed to the rules visitor
  * `currentSource` is updated from the AST returned from the visitor (aka it is printed back to source)

> when a transform is applied on a node, should we somehow re-parse / re-lint the modified AST to apply a second fixer to the correct AST?

I'm not 100% sure I understand the question. For each fixer we reparse and lint, then we have a overall loop of "fixing iterations" (IIRC it is set to 10 max). So for a given set of errors we would do:

* run `.verify` to gather the current issues to be fixed
* run fixers (see list just above for how that is done)
* check if the fixers modified the output source
* if they did, and we have looped less than 10 times, start back at the beginning

Basically, ending up with `verify` -> `applyFixers` -> `verify` -> `applyFixers` -> etc.... (up to 10 times). The idea here is that even if two fixers did have an interaction (e.g. they both wanted to operate on the same node) we'd run both again until they "settle" (or error out because they can't agree).

test/acceptance/rule-test.js
rwjblue: I kinda like the input/output in the same file (makes understanding the changes made much easier), I'll look into seeing if we can do `.toMatchInlineSnapshot` though.

lib/helpers/rule-test-harness.js
rwjblue: Updated, thanks for pointing this out!

test/acceptance/rule-test.js
rwjblue: Ya, I have snapshot support ready to go (can't push due to https://www.githubstatus.com/incidents/q07bfjh7jf1t, but will get the PR up soon).

rwjblue: Done in https://github.com/ember-template-lint/ember-template-lint/pull/1174

lib/linter.js
```diff
@@ -241,6 +231,47 @@ class Linter {
...
+      let rule = this._buildRule(
+        ruleName,
+        {
+          pending,
+          shouldFix: true,
+          filePath: options.filePath,
+          rawSource: currentSource,
+          fileConfig: getConfigForFile(this.config, options.filePath),
+        },
+        () => {}
```
dcyriller: This line might be removed 😉

lib/linter.js
```diff
@@ -241,6 +231,47 @@ class Linter {
...
+   * @param {Object} options
+   * @param {string} options.source - The source code to verify.
+   * @param {string} options.filePath
+   * @param {string} options.moduleId
+   * @param {Object[]} results - The lint messages.
+   *
+   * @returns {string} output - The fixed source.
+   */
+  _applyFixes(options, results) {
+    let fixableIssues = results.filter(r => r.isFixable);
```
dcyriller: Do you prefer to have a `isFixable` property or checking against the existence of `fixer` function is enough?

rwjblue: In this implementation, there is no fixer function (in the results listing, or in the rule instances). The idea here is to have _something_ in the results that the `ember-template-lint` command can use to add a bit to our output (like eslint itself has) for):

```
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

dcyriller: The different implementations were a bit mixed up in my mind.

PR #1160, opened 2020-02-25
Add severity support (off, warn, error) to rule configuration (via both `.template-lintrc.js` and inline configuration).

suchitadoshi1987:
Each rule can have its own severity level which can be a string or could be the first element of the array that contains the custom rule configuration.
Supported severity levels are `off`, `warn`, `error`.
You can define a severity level directly on the rule:
Eg:
```
'no-bare-strings': 'warn'`
```
OR
If your rule has a custom configuration, and you want to define the severity level you would need to add the `severity` as the first element of the array.
Eg:

```js
{
   "no-implicit-this": ['warn', { "allow": [ "fooData" ] }
}
```


bmish:
What do you think about exactly matching the [eslint](https://eslint.org/docs/user-guide/configuring) syntax for this? Most developers are familiar with eslint and so I'd prefer if we exactly followed their lead on this.

```
quotes: ["error", "double"]
```

CC: @rwjblue

rwjblue:
Haha ✋ @bmish 😉

rwjblue:
I _think_ that we can do it compatibly. Using these types:

```ts
type SeverityLevelIgnore = 0 | 'off';
type SeverityLevelWarn = 1 | 'warn';
type SeverityLevelError = 2 | 'error';

type SeverityLevel = SeverityLevelIgnore | SeverityLevelWarn | SeverityLevelError;

type RuleSpecificConfiguration = unknown;
type RuleConfiguration = SeverityLevel | [SeverityLevel, RuleSpecificConfiguration];
```

* If we see a rule config value with a `SeverityLevel` (e.g. `'warn'`), we convert that config to a boolean (false for off, and true for warn / error). We then set the severity to the appropriate level (when instantiating the rule).
* If we see a rule config value that is an array, and the first argument is a `SeverityLevel` we pass the RuleSpecificConfiguration _directly_ to the plugin as it's config. We then set the severity to the appropriate level.
* If we see _anything else_ we set the severity level to error (what we do today), and pass through the config verbatim. We also gather a list of rules that are misconfigured, and emit a warning (I'm _pretty_ sure we can whip up a quick codemod-cli project to tell them to run in the warning to auto-fix it for them).

suchitadoshi1987:
Yes, great point @bmish , @rwjblue , this makes sense. I think with my current PR, it does what @rwjblue you are suggesting, but it does it as an object instead of an array. The change should be quick and since it aligns with eslint’s config, nothing like it!

rwjblue:
OK, thinking this through a small bit more, I think we need to tweak that decision tree a tiny bit if we want to allow the numeric severity levels because we definitely have rules that would accept a single config of those values 0, 1, 2.

So we _either_ need to decide to only support the string variants (`'off'`,`'warn'`, `'error'`), or to add some sort of opt-in flag to the config file's top level so we know that when we see `'block-indentation': 2` we know that we are going to use the default config (`'editorconfig'` now IIRC) and treat it as an error (whereas today, the rule would treat that `2` as the indentation level).

lib/rules/base.js
```diff
@@ -80,6 +91,46 @@ module.exports = class BaseRule {
...
+   * Eg:
+   * {'no-implicit-this': 'warn'}
+   * This would set the severity to 1 and return `true`.
+   * { 'no-implicit-this': { "severity": "warn", "config": "allow": [ "fooData" ] } }
+   * This would set the severity of the rule to 1 and return the user defined config back.
+   * { 'no-implicit-this': { "severity": "warn"} }
+   * This would set the severity of the rule to 1 and return true.
+   * @param {*} configData
+   */
+  processConfigForSeverity(configData) {
```
rwjblue: I don't think we need this method to be on the rule itself, I'd like it to be part of `get-config` instead.

Basically, I'd like `get-config.js` utility to return `config.rules` that is in the internal structure that you mentioned in the initial description:

```
rules: {
  'rule-name': {
    severity: 'off' | 'warn' | 'error',
    config: someUserlandValue
  }
}
```

Then when the rule is instantiated (in `lib/linter.js`'s `buildRule`) we pass options that include `severity` and `config` (aka, the same thing we pass today).

This will require changing a few tests that are asserting on the results of `get-config`, but shouldn't be _too_ bad...

lib/get-config.js
```diff
@@ -303,7 +310,60 @@ function validateRules(config, options) {
...
+        config.rules[key] = ruleConfig;
+      }
+    }
+    // array of severity and custom rule config
+    else if (Array.isArray(ruleData)) {
+      if (ruleData.length !== 2) {
+        throw new Error(
+          'Severity and custom rule config are the only two values allowed in the rule config array.'
+        );
+      }
```
rwjblue: I don't think we can throw this error. Rules are allowed to accept an array as their configuration today, adding this error would be a breaking change.

Instead we need to check if it is an array **and** if the first value is one of our known strings (warn, error, off). If the first value is _not_ one of our strings, we need to fallback to current behavior (not error).

lib/get-config.js
```diff
@@ -303,7 +310,60 @@ function validateRules(config, options) {
...
+    let ruleData = config.rules[key];
+    let ruleConfig = {};
+    // In case of {'no-implicit-this': 'off|warn|error'}
+    if (typeof ruleData === 'string') {
+      severityConfig = _determineConfigForSeverity(ruleData);
+      if (severityConfig) {
+        ruleConfig.severity = severityConfig.severity;
+        ruleConfig.config = severityConfig.config;
+        config.rules[key] = ruleConfig;
+      }
```
rwjblue: This `if` needs an `else`
```suggestion
      if (severityConfig) {
        ruleConfig.severity = severityConfig.severity;
        ruleConfig.config = severityConfig.config;
      } else {
        ruleConfig.severity = 2;
        ruleConfig = ruleData;
      }
```

lib/get-config.js
```diff
@@ -303,7 +310,60 @@ function validateRules(config, options) {
...
+    let severityConfig;
+    let ruleData = config.rules[key];
+    let ruleConfig = {};
+    // In case of {'no-implicit-this': 'off|warn|error'}
+    if (typeof ruleData === 'string') {
+      severityConfig = _determineConfigForSeverity(ruleData);
+      if (severityConfig) {
+        ruleConfig.severity = severityConfig.severity;
+        ruleConfig.config = severityConfig.config;
+        config.rules[key] = ruleConfig;
```
rwjblue: This should be set outside of the conditional (not in each branch).

lib/get-config.js
```diff
@@ -303,7 +310,60 @@ function validateRules(config, options) {
...
+      if (ruleData.length !== 2) {
+        throw new Error(
+          'Severity and custom rule config are the only two values allowed in the rule config array.'
+        );
+      }
+      let severity = ruleData[0];
+      severityConfig = _determineConfigForSeverity(severity);
+      if (severityConfig === undefined) {
+        throw new Error('Severity can only be "off", "warn", or "error"');
+      }
```
rwjblue: Same thing as above, arrays are currently valid as rule config. We cannot throw if we see an array like `no-invalid-tagnames: ['div', 'table'],`.

lib/linter.js
```diff
@@ -51,21 +48,22 @@ class Linter {
   }
 
   loadConfig() {
-    this.config = getConfig(this.options);
+    this.config = getProjectConfig(this.options);
   }
 
-  _defaultSeverityForRule(ruleName, pendingStatus) {
+  _determineSeverityForRule(ruleName, pendingStatus, configRule) {
```
rwjblue: FYI - I'd like to see this method go away when we land the overrides support (the logic would move into `getConfigForFile`).

lib/linter.js
```diff
@@ -81,14 +79,14 @@ class Linter {
   _buildRule(ruleName, options, addToResults) {
     let loadedRules = this.config.loadedRules;
     let Rule = loadedRules[ruleName];
-
+    let configRule = this.config.rules[ruleName];
     let ruleOptions = Object.assign(
       {
         name: ruleName,
-        config: this.config.rules[ruleName],
+        config: configRule && configRule.config ? configRule.config : configRule,
```
rwjblue: We should reset this change, the `processRules` method in `get-config.js` should **always** return the new structure (we should be "upgrading" any configs in the older format).

lib/linter.js
```diff
@@ -135,7 +133,8 @@ class Linter {
 
     let loadedRules = this.config.loadedRules;
     for (let ruleName in configuredRules) {
-      if (configuredRules[ruleName] === false) {
+      const configuredRule = configuredRules[ruleName];
+      if (configuredRule === false || (configuredRule && configuredRule.config === false)) {
```
rwjblue: We should not need to check both formats, but we should probably leverage severity here (0 being off, and would replace the check for `false` that we have currently)

test/acceptance/public-api-test.js
```diff
@@ -327,6 +327,68 @@ describe('public api', function() {
       expect(result[0].fatal).toBe(true);
     });
 
+    it('triggers warnings severity is set to warn', function() {
```
rwjblue:
```suggestion
    it('triggers warnings when severity is set to warn', function() {
```

test/unit/rules/invocable-blacklist-test.js
```diff
@@ -5,7 +5,7 @@ const generateRuleTests = require('../../helpers/rule-test-harness');
 generateRuleTests({
   name: 'invocable-blacklist',
 
-  config: ['foo', 'bar'],
+  config: ['error', ['foo', 'bar']],
```
rwjblue: This test should not need to be changed (this indicates that this is a breaking change...) 😸

lib/get-config.js
suchitadoshi1987: From the offline discussion with @rwjblue , our eventual goal is to get devs to follow the standard pattern when arrays are used (first severity and then the custom config), however, this would be a breaking change and we would want to gradually transition the devs towards this pattern.
Will change it fallback to the default implementation

lib/linter.js
suchitadoshi1987: yep!

lib/linter.js
suchitadoshi1987: yup... definitely, this was an oversight... meant to pass in just the standard format, but this got missed while i was working on another part. my bad :\

lib/linter.js
suchitadoshi1987: same as above :(

test/unit/rules/invocable-blacklist-test.js
suchitadoshi1987: 👍

lib/get-config.js
suchitadoshi1987: this was done for the same reason, I wanted to suppress errors caused due to changing the default format while working on another part, need to fix this one

lib/get-config.js
suchitadoshi1987: Or, I can default it to:
```
// defaults to error always.
    let ruleConfig = {
      severity: ERROR_SEVERITY,
      config: ruleData,
    };
```
and override the object if any of the conditions are met

README.md
```diff
@@ -150,6 +150,20 @@ The following properties are allowed in the root of the `.template-lintrc.js` co
...
+Supported severity levels are `off`, `warn`, `error`.
+You can define a severity level directly on the rule:
+Eg: `'no-bare-strings': 'warn'`
+OR
+If your rule has a custom configuration, and you want to define the severity level you would need to add the `severity` as the first element of the array.
+Eg:
+
+```js
+{
+   "no-implicit-this": ['warn', { "allow": [ "fooData" ] }
```
rwjblue:
```suggestion
  "no-implicit-this": ['warn', { "allow": [ "fooData" ] }]
```

lib/get-config.js
```diff
@@ -303,7 +307,56 @@ function validateRules(config, options) {
...
-module.exports = function(options) {
+function _determineConfigForSeverity(config) {
+  switch (config) {
+    case 'off':
+      return { config: false, severity: IGNORE_SEVERITY };
+    case 'warn':
+      return { config: true, severity: WARNING_SEVERITY };
+    case 'error':
+      return { config: true, severity: ERROR_SEVERITY };
+  }
```
rwjblue: Should we deal with `default` here also?

lib/get-config.js
```diff
@@ -303,7 +307,56 @@ function validateRules(config, options) {
...
+
+/**
+ * Sets the severity and config on the rule object.
+ * Eg:
+ * {'no-implicit-this': 'warn'} -> {'no-implicit-this': { severity: 'warn', config: true }}
+ * { 'no-implicit-this': [ 'warn', { allow: [ 'fooData' ] } ] }
+ * would become:
+ * { 'no-implicit-this': { severity: 'warn', config: 'allow': [ 'fooData' ] } }
+ * @param {*} configData
+ */
```
rwjblue: Can you add examples for all of the legacy cases as well? I'm thinking something like these:

```js
{
  'some-rule': 'lol',                   // -> { severity: 2, config: 'lol' }
  'other-thing': ['wat', 'is', 'this'], // { severity: 2, config: ['wat', 'is', 'this'] }
  'hmm-thing-here': { zomg: 'lol' },    // -> { severity: 2, config: { zomg: 'lol' } }
}
```

lib/get-config.js
```diff
@@ -303,7 +307,56 @@ function validateRules(config, options) {
...
+    let ruleConfig = {
+      severity: ruleData === false ? IGNORE_SEVERITY : ERROR_SEVERITY,
+      config: ruleData,
+    };
+    // In case of {'no-implicit-this': 'off|warn|error'}
+    if (typeof ruleData === 'string') {
+      severityConfig = _determineConfigForSeverity(ruleData);
+      if (severityConfig) {
+        ruleConfig.severity = severityConfig.severity;
+        ruleConfig.config = severityConfig.config;
```
rwjblue:
```suggestion
        ruleConfig = severityConfig;
```

lib/get-config.js
```diff
@@ -303,7 +307,56 @@ function validateRules(config, options) {
...
+      }
+    } else if (Array.isArray(ruleData)) {
+      // array of severity and custom rule config
+      let severity = ruleData[0];
+      severityConfig = _determineConfigForSeverity(severity);
+      if (severityConfig) {
+        ruleConfig.severity = severityConfig.severity;
+        ruleConfig.config = ruleData[1] !== undefined ? ruleData[1] : severityConfig.config;
```
rwjblue: What is this ternary for? I think we should just use `ruleData[1]` here, regardless of what it is (when using the severity first syntax, the second array argument is **completely** opaque to us. It can be any value...).

lib/helpers/rule-test-harness.js
```diff
@@ -98,8 +99,9 @@ module.exports = function generateRuleTests({
       if (localConfig === undefined) {
         return;
       }
+      const severity = localConfig === false ? IGNORE_SEVERITY : ERROR_SEVERITY;
 
-      linter.config.rules[name] = localConfig;
+      linter.config.rules[name] = { config: localConfig, severity };
```
rwjblue: We should leverage the same `processRules` to allow us to test both old and new configs.

lib/linter.js
```diff
@@ -81,14 +84,14 @@ class Linter {
   _buildRule(ruleName, options, addToResults) {
     let loadedRules = this.config.loadedRules;
     let Rule = loadedRules[ruleName];
-
+    let configRule = this.config.rules[ruleName];
```
rwjblue:
```suggestion
    let config = this.config.rules[ruleName];
```

lib/linter.js
```diff
@@ -81,14 +84,14 @@ class Linter {
...
     let ruleOptions = Object.assign(
       {
         name: ruleName,
-        config: this.config.rules[ruleName],
+        config: configRule.config,
```
rwjblue:
```suggestion
        config,
```

lib/get-config.js
suchitadoshi1987: the reason being for cases where:
`{rules: {'foo': ['warn']}}`
But since now we are checking for `severity` to bypass building rules, I think its safe to remove

lib/get-config.js
suchitadoshi1987: the reason why I didn't want a `default` value here is that if one of the 3 cases is not found, it returns undefined which helps determine to use user-provided config instead.

rwjblue: makes sense, thank you!

lib/get-config.js
rwjblue: 👍, please remove (so we can make no assumptions about that `rulesData[1]` value)

lib/linter.js
suchitadoshi1987: cant do this since I need to reference the `config` object from `configRule`, so in the above case it would become `config.config` instead

rwjblue:
Will need a rebase now that overrides has landed

lib/linter.js
suchitadoshi1987: need to get this done

lib/config/octane.js
```diff
@@ -3,11 +3,11 @@
...
   extends: ['recommended'],
   rules: {
-    'no-action': true,
+    'no-action': 'error',
     'no-curly-component-invocation': {
-      noImplicitThis: true,
-      requireDash: false,
+      noImplicitThis: 'error',
+      requireDash: 'off',
     },
```
rwjblue: This seems wrong, I think it should be:

```js
'no-curly-component-invocation': ['error', {
  noImplicitThis: true,
  requireDash: false,
}],
```

lib/get-config.js
```diff
@@ -307,20 +311,32 @@ function validateRules(rules, loadedRules, options) {
...
+      let overrideKeys = Object.keys(overrideConfig);
+      let containsValidKeys = overrideKeys.every(item => SUPPORTED_OVERRIDE_KEYS.includes(item));
       if (!containsValidKeys) {
         throw new Error(
           'Using `overrides` in `.template-lintrc.js` only supports `files` and `rules` sections. Please update your configuration.'
         );
       }
-      validateRules(override.rules, config.loadedRules, options);
+      validateRules(overrideConfig.rules, config.loadedRules, options);
+      overrideConfig = processRules(overrideConfig);
```
rwjblue: I don't understand what is happening here. Why are we reassigning the variable (nothing uses it again)? Do we expect `processRules` to just mutate?

lib/get-config.js
```diff
@@ -345,6 +361,50 @@ function getConfigForFile(config, filePath) {
...
+      let severity = ruleData[0];
+      severityConfig = _determineConfigForSeverity(severity);
+      if (severityConfig) {
+        ruleConfig.severity = severityConfig.severity;
+        ruleConfig.config = ruleData[1];
+      }
+    }
+    config.rules[key] = ruleConfig;
+  }
+  return Object.assign({}, config);
```
rwjblue: We return a new object here, but we've already mutated it just above. We should do one thing, or the other, not both...

lib/get-config.js
```diff
@@ -366,11 +426,19 @@ function getProjectConfig(options) {
...
 
-module.exports = { getProjectConfig, getConfigForFile };
+module.exports = {
+  getProjectConfig,
+  getConfigForFile,
+  ERROR_SEVERITY,
+  WARNING_SEVERITY,
+  IGNORE_SEVERITY,
+  processRules,
+};
```
rwjblue: Mind if we sort these like this?

```suggestion
module.exports = {
  ERROR_SEVERITY,
  IGNORE_SEVERITY,
  WARNING_SEVERITY,
  getConfigForFile,
  getProjectConfig,
  processRules,
};
```

test/acceptance/public-api-test.js
```diff
@@ -476,6 +555,62 @@ describe('public api', function() {
       expect(result).toEqual(expected);
     });
 
+    it('Works for older syntax without custom severity', function() {
```
rwjblue: 👏 thank you!

lib/linter.js
```diff
@@ -252,7 +220,7 @@ class Linter {
 
     let currentSource = options.source;
 
-    let pending = this.statusForModule('pending', options.moduleId);
+    let pending = getConfigForFile(this.config, options).pending;
```
rwjblue: We use `getConfigForFile` just below, we should capture it in a variable instead of calling it twice.

```suggestion
    let config = getConfigForFile(this.config, options);
    let pending = config.pending;
```

lib/rules/base.js
```diff
@@ -448,7 +448,10 @@ module.exports = class Base {
       }
 
       try {
+        const { determineRuleConfig } = require('../get-config');
```
rwjblue: Lets hoist this require to the top.

test/unit/base-plugin-test.js
```diff
@@ -408,7 +409,7 @@ describe('base plugin', function() {
     }
 
     function addEvent(event, node, plugin) {
-      events.push([event, getId(node), plugin.config]);
+      events.push([event, getId(node), determineRuleConfig(plugin.config).config]);
```
rwjblue: Can you beef this up to include the severity, then update the assertions below to show both the config and the severity.

lib/linter.js
suchitadoshi1987: :|... oops!

lib/linter.js
```diff
@@ -69,14 +57,24 @@ class Linter {
...
   _buildRule(ruleName, options) {
     let fileConfig = options.fileConfig;
+    let configuredRules = fileConfig.rules[ruleName];
     let Rule = fileConfig.loadedRules[ruleName];
+    let pendingStatus = fileConfig.pendingStatus;
+    let severity = configuredRules.severity;
+
+    if (
+      pendingStatus &&
+      (pendingStatus === true || (pendingStatus.only && pendingStatus.only.indexOf(ruleName) > -1))
```
rwjblue:
```suggestion
      (pendingStatus === true || (pendingStatus.only && pendingStatus.only.includes(ruleName)))
```

lib/rules/base.js
suchitadoshi1987: This is due to the circular dependency of modules that I had to load it dynamically. As discussed, I ll pull this function out and other simpler ones  from `get-config.js` in another PR to be in a separate util file to resolve the circular dependency issues

PR #1166, opened 2020-02-25
Refactor `cli-test.js` to remove usage of fixtures in favor of using fake project

suchitadoshi1987:


test/acceptance/cli-test.js
```diff
@@ -319,40 +382,64 @@ describe('ember-template-lint executable', function() {
...
+
+          let projectWithErrors = new Project();
+          projectWithErrors.chdir();
+          projectWithErrors.setConfig({
+            rules: {
+              'no-bare-strings': true,
+            },
+          });
+
+          let result = run(['.', '--config-path', projectWithErrors.path('.template-lintrc.js')], {
```
rwjblue: Since you are project.chdir'ing you don't need to pass `--config-path` (and the test isn't _specifically_ about passing a custom config path)

test/acceptance/cli-test.js
```diff
@@ -195,13 +236,12 @@ describe('ember-template-lint executable', function() {
 
     describe('with --json param', function() {
       it('should print valid JSON string with errors', function() {
+        setProjectConfigForErrors();
         let result = run(['.', '--json'], {
-          cwd: './test/fixtures/with-errors',
+          cwd: project.path('.'),
```
rwjblue: Seems like `cwd` is _always_ going to be `project.path()`, can we just make that the default (removing the repeated manual `cwd: project.path()` calls all over the place)

test/acceptance/cli-test.js
```diff
@@ -319,40 +382,64 @@ describe('ember-template-lint executable', function() {
       });
 
       describe('given a directory with errors and a lintrc with rules', function() {
-        it('should print properly formatted error messages', function() {
-          let result = run(['.', '--config-path', '../with-errors/.template-lintrc'], {
-            cwd: './test/fixtures/without-errors',
+        it('should print properly formatted error messages', async function() {
+          setProjectConfigWithoutErrors();
+
+          let projectWithErrors = new Project();
```
rwjblue: Don't we have a shared `project` instance from the top-most `beforeEach`?

test/acceptance/cli-test.js
```diff
@@ -319,40 +382,64 @@ describe('ember-template-lint executable', function() {
...
       });
 
       describe('given a directory with errors but a lintrc without any rules', function() {
-        it('should exit without error and any console output', function() {
-          let result = run(['.', '--config-path', '../without-errors/.template-lintrc'], {
-            cwd: './test/fixtures/with-errors',
+        it('should exit without error and any console output', async function() {
+          setProjectConfigForErrors();
+
+          let withoutErrorsProject = new Project();
```
rwjblue: Why not use the shared `project` from the top?

test/acceptance/cli-test.js
suchitadoshi1987: so for this particular case, the `--config-path` is using another project's `.template-lintrc`, and thats what I am trying to mimic here.

test/acceptance/cli-test.js
suchitadoshi1987: my bad, i shouldn't be changing the project dir. `chdir` shouldn't be here at all since its only used to reference the `.template-lintrc.js` in `--config-path`

test/acceptance/cli-test.js
suchitadoshi1987: same as above

test/acceptance/cli-test.js
suchitadoshi1987: 💯

test/acceptance/cli-test.js
rwjblue: I don’t really understand what you mean. This test is just saying “given a project with errors it prints errors and warnings correctly”, this test shouldn’t need to use configPath (it only used it before because it was inconvenient to chdir and back).

PR #1167, opened 2020-02-26
Ensure accessing `this.editorConfig` in a rule instance does not error.

### LINKED ISSUE #1161: Does ember-template-lint v2 support ESLint integration?
Looking at upgrading to v2 and ran into an ESLint issue on the following code:

```typescript
import hbs from "htmlbars-inline-precompile"

hbs`<MyComponent />`
```

which gives the following output when running ESLint:

`
3:1  error  3 error(s): The "path" argument must be of type string. Received type undefined  hbs/check-hbs-template-literals
`

Is the ESLint integration intended to work with v2 yet? I did notice
`warning " > eslint-plugin-hbs@1.0.0" has incorrect peer dependency "ember-template-lint@^1.5.3".` when installing, so perhaps this isn't supported yet.

My `.template-lintrc.js` is:

```javascript
module.exports = {
  extends: ["recommended", "octane", "stylistic"],

  rules: {
    "attribute-indentation": false,
    "eol-last": "always",
    "inline-link-to": true,
    "modifier-name-case": true,
    "no-bare-strings": true,
    "no-implicit-this": { allow: ["rootURL"] },
    "no-invalid-interactive": false,
    "require-valid-alt-text": false,
    "template-length": { max: 275, min: 1 }
  }
}
```

Thanks!

rwjblue:
This ensures that `this.editorConfig` within a rule instance will **always** return an object (even if we don't find an editor config file, if we don't provide a filePath, etc).

Prior to this, any rule that would use `this.editorConfig` would error when consumed through tooling that does not provide `filePath`. This includes `ember-cli-template-lint` and `eslint-plugin-hbs`.

Fixes #1161

### DIFF
--- a/test/unit/base-plugin-test.js
+++ b/test/unit/base-plugin-test.js
@@ -5,16 +5,43 @@ const Rule = require('./../../lib/rules/base');
 const { readdirSync, existsSync, readFileSync } = require('fs');
 const { join, parse: parsePath } = require('path');
 const ruleNames = Object.keys(require('../../lib/rules'));
+const Project = require('../helpers/fake-project');
+const EditorConfigResolver = require('../../lib/get-editor-config');
 
 describe('base plugin', function() {
+  let project, editorConfigResolver;
+  beforeEach(() => {
+    project = Project.defaultSetup();
+
+    editorConfigResolver = new EditorConfigResolver();
+    editorConfigResolver.resolveEditorConfigFiles();
+  });
+
+  afterEach(async () => {
+    await project.dispose();
+  });
+
   function runRules(template, rules) {
     let ast = parse(template);
 
     for (let ruleConfig of rules) {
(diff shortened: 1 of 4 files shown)


test/unit/base-plugin-test.js
```diff
@@ -5,16 +5,47 @@ const Rule = require('./../../lib/rules/base');
...
+      );
+
+      options.configResolver = Object.assign({}, ruleConfig.configResolver, {
+        editorConfig: () => {
+          if (!options.filePath) {
+            return {};
+          }
+
+          return editorConfigResolver.getEditorConfigData(options.filePath);
+        },
```
rwjblue: FYI - Expect future PRs to remove the duplication between this file and `lib/linter.js`. We should have a single canonical mechanism to build a rule instance that ensures the correct arguments are passed.

PR #1173, opened 2020-02-27
Update minimum version of ember-template-recast to 4.1.1.

rwjblue:
Dependabot was not configured to bump minimums (it was just updating the
yarn.lock files). I fix that (I think), and this bumps us to the latest
ember-template-recast.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "chalk": "^3.0.0",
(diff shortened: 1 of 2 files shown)


rwjblue:
🤔 seems like CI didn't kick off for this one, may be related to https://www.githubstatus.com/incidents/q07bfjh7jf1t

PR #1174, opened 2020-02-27
Add test harness features (naming test cases, snapshot support, etc).

rwjblue:
* Add a basic acceptance test for the test harness usage.
* Add support for using snapshots (inline or external) to test harness.

By adding a customized `verifyResults` function you can run your own
assertions against the results. This allows users of `jest` to write
something like:

```js
generateRuleTests({
 name: 'whatever',
 config: true,

  bad: [
   {
     name: 'zomg inline snapshots!!!!',
     template: '<Whatever><Here /></Whatever>',
     verifyResults(results) {
       expect(results).toMatchInlineSnapshot();
     }
   }
 ]
});
```

* Add support for naming test cases (making matching tests massively
  simpler, and making it clearer what a given `bad` or `good` scenario
  is actually testing).

For the You can add a `name` property to any of the `good` or `bad` test cases,
and that name will be used for the test prefixing.

Using something like:

```js
generateRuleTests({
 name: 'whatever',
 config: true,

  bad: [
   {
     name: 'some awesome name',
     ...otherFields
   }
 ]
});
```

You can get test output like:

```
✓ some awesome name: logs errors (7ms)
✓ some awesome name: passes when rule is disabled (1ms)
✓ some awesome name: passes when disabled via inline comment - single rule (2ms)
✓ some awesome name: passes when disabled via long-form inline comment - single rule (1ms)
✓ some awesome name: passes when disabled via inline comment - all rules (1ms)
```

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -181,8 +181,11 @@ module.exports = function generateRuleTests({
           assert.strictEqual(actual.length, 1); // can't have more than one fatal error
           delete actual[0].source; // remove the source (stack trace is not easy to assert)
           assert.deepStrictEqual(actual[0], item.fatal);
+        } else if (item.verifyResults) {
+          item.verifyResults(actual);
         } else {
-          expectedResults = expectedResults.map(parseResult);
+          let results = item.results || [item.result];
+          let expectedResults = results.map(parseResult);
 
           assert.deepStrictEqual(actual, expectedResults);
         }
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -32,6 +32,7 @@ function getVerifyOptions(template, meta) {
  * allows tests to be defined without needing to setup test infrastructure every time
  * @param  {Array}  [bad=[]]            - an array of items that describe the use case that should fail
  *  [{
+     name: 'prevents debugger',
      template: '{{debugger}}',
 
      result: {
@@ -46,6 +47,7 @@ function getVerifyOptions(template, meta) {
  *
  * @param  {Array}  [error=[]]            - an array of items that describe the use case that should error
  *  [{
+     name: 'errors for "sometimes"',
      config: 'sometimes',
      template: 'test',
 
@@ -57,7 +59,7 @@ function getVerifyOptions(template, meta) {
      },
    }]
  *
- * @param  {Array}    [good=[]]          - an array of strings that define templates that should not fail
+ * @param  {Array}    [good=[]]          - an array of items that should not fail
  * [
      '<img alt="hullo">',
      '<img alt={{foo}}>',
@@ -65,6 +67,11 @@ function getVerifyOptions(template, meta) {
      '<img aria-hidden="true">',
      '<img alt="">',
      '<img alt>',
+     {
+       name: 'works with custom config',
+       config: 'customized',
+       template: '<img>',
+     }
    ]
  *
  * @param  {String}   name                - a name to describe which lint rule is being tested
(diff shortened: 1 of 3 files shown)


rwjblue:
> I think the initial goal of the snapshots was decreased verbosity in the tests, but this looks like it's even more verbose than before now.

Hmm, that wasn't my initial goal (though I think that was @dcyriller's point). I _personally_ prefer to keep the assertion value inline with the input value. The reason I want to support using snapshots is so that I don't have to manually change line/column numbers all the time (or manually update all the messages when tweaking verbiage), etc.

However, this implementation absolutely allows for the reduce verbosity, just use something like:

```js
const verifyResults = results => expect(results).toMatchSnapshot();

generateRuleTests({
 name: 'whatever',
 config: true,

  bad: [
   {
     name: 'zomg inline snapshots!!!!',
     template: '<Whatever><Here /></Whatever>',
     verifyResults,
   }
 ]
});
```

And 💥 you've got decreased verbosity 😄. I _personally_ just don't want to do it in this repo, but I think its perfectly valid for consumers of the test harness to want to do it.

rwjblue:
I'll add a test for that too (but I don't want to migrate _this_ test suite to it), to show that it works well and is supported.

rwjblue:
@Turbo87 Just pushed a test showing that external snapshots work just as well.

rwjblue:
FWIW, I think it might be cool to have external snapshots with a structure like:

```
lib
├── rules
     ├── quotes
           ├── index.js
           ├── README.md
           ├── test.js
           └── __snapshots__
```

Where all of the "stuff" for a rule is colocated with it.  I may give that a whirl after this.

lib/helpers/rule-test-harness.js
```diff
@@ -32,6 +32,7 @@ function getVerifyOptions(template, meta) {
  * allows tests to be defined without needing to setup test infrastructure every time
  * @param  {Array}  [bad=[]]            - an array of items that describe the use case that should fail
  *  [{
+     name: 'prevents debugger',
```
dcyriller: A bit late to the party, sorry. but should it be `description`?

dcyriller: Not 100% sure it's relevant ^^

dcyriller:
I need some time to put my ideas together 😄.

From a rule author standpoint, I think the test API might be simplified a bit further. We could write handlebars in a hbs file(s) (instead of inline). This is what I had in mind when I mentioned snapshots.

For instance, for the `quotes` rule, a rule author would write in a `test.js` file:
```js
generateRuleTests(__dirname, { config: 'single' });
generateRuleTests(__dirname, { config: 'double' });
generateRuleTests(__dirname, { config: true });
generateRuleTests(__dirname, { config: 'invalid' });
```

And in a same directory, the rule author would write a few hbs files.
in `mustache.hbs`:
```hbs
{{component "test"}}
{{hello x="test"}}
```
in `element-node.hbs`:
```hbs
<input type='checkbox'>
```

We'd have a file structure akin:
```
lib
├── rules
     ├── quotes
           ├── index.js
           ├── README.md
           └── tests
                 ├── test.js
                 ├── mustache.hbs
                 ├── element-node.hbs
                 └── __snapshots__
```

In this example, you notice that `bad` and `good` have disappeared. Because template-lint output has been captured in snapshots.

@rwjblue / @Turbo87 what do you think?

Turbo87:
I have to admit I'm not a big fan of these one test per file structures. They usually make it quite hard to see all the tested scenarios at a glance. With the proposed structure you will have to potentially open dozens of files to get an idea of the existing scenarios.

dcyriller:
All tested scenarios (input and output) for a given rule are in possibly one snapshot file 🤔

PR #1177, opened 2020-02-27
add command to generate new rule from CLI

MelSumner:
Adds the ability to run `yarn new rule-name` and generate the files necessary for a new rule. 

Message generated in CLI: 
image

It needs tests and alpha-sorting the rules in the files for extra fanciness, but I would like to do that in a separate PR if that's okay.


dev/new-rule/templates/doc.md
```diff
@@ -0,0 +1,53 @@
+# (rule name goes here)
```
bmish: I just updated this template today, can you rebase on the latest master and use my updated version?

dev/new-rule/index.js
```diff
@@ -0,0 +1,130 @@
...
+  creates a functioning/immediately testable rule
+  $ yarn new my-new-rule
+  $ yarn test my-new-rule
+  `);
+}
+
+// Some basics that we'll need
+const pathUtil = __dirname;
+const pathTemplates = path.join(pathUtil, 'templates');
+const pathRoot = path.join(pathUtil, '../..');
```
bmish: nit: use separate parameters for `../..`

dev/new-rule/index.js
```diff
@@ -0,0 +1,130 @@
...
+    )
+  );
+}
+
+function createNewRuleDocFile(newRuleName) {
+  // Create rule doc file
+  let pathDocTemplate = path.join(pathTemplates, 'doc.md');
+  let pathDocNew = path.join(pathRoot, 'docs', 'rule', `${newRuleName}.md`);
+  let orgContent = fs.readFileSync(pathDocTemplate, { encoding: 'utf8' });
+  let placeholderDocTemplate = '(rule name goes here)';
```
bmish: This should be updated to `TODO: rule-name-goes-here` based on the updated rule doc template.

dev/new-rule/index.js
```diff
@@ -0,0 +1,130 @@
...
+function appendDocListFile(newRuleName) {
+  // Append rule + link to rule doc file
+  let pathDocList = path.join(pathRoot, 'docs', 'rules.md');
+  let docListAppend = `\n* [${newRuleName}](rule/${newRuleName}.md)`;
+  fs.appendFileSync(pathDocList, docListAppend);
+  console.log(
+    chalk.cyan(
+      `4. Success! ${chalk.yellow(
+        `${newRuleName}.md`
+      )} was appended to ${pathDocList}: please be sure to alpha sort the list before making a commit!`
```
bmish: I would say `alphabetically` instead of `alpha` for clarity.

.eslintignore
```diff
@@ -0,0 +1 @@
+/dev/new-rule/templates/*
```
dcyriller: As the files derived from these templates will be commited to the code base, it might make sense to make eslint pass for them?

MelSumner: The derived files should pass but these won't have enough information yet to pass.

dev/new-rule/index.js
MelSumner:
```suggestion
  let placeholderDocTemplate = '(TODO: rule name goes here)';
```

dev/new-rule/index.js
```diff
@@ -0,0 +1,130 @@
...
```
MelSumner:
```suggestion
      )} was appended to ${pathDocList}: please be sure to alphabetically sort the list before making a commit!`
```

.eslintignore
dcyriller: `.eslintrc.js` could take another override:
```
  overrides: [
    {
      files: ['dev/new-rule/templates/**/*.js'],
      rules: {
        'node/no-missing-require': 'off',
        'no-unused-vars': 'off',
      },
    },
  ],
```

MelSumner: I had it like that and I was asked to change it to how I do it now; it seems like it's conflicting opinions at this point and I would prefer to either accept a separate PR to do this or something like that. I don't want to keep going in circles if we can avoid it.

dcyriller: Oh, didn’t know that. 👍

PR #1179, opened 2020-02-29
inline-link-to: Implement `fix` mode

Turbo87:
This PR implements the `--fix` mode for the `inline-link-to` rule, based on the implementation in #1158

### DIFF
--- a/lib/rules/inline-link-to.js
+++ b/lib/rules/inline-link-to.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { builders: b } = require('ember-template-recast');
+
 const Rule = require('./base');
 
 const message = 'The inline form of link-to is not allowed. Use the block form instead.';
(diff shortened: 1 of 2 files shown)


test/unit/rules/inline-link-to-test.js
```diff
@@ -16,22 +16,38 @@ generateRuleTests({
...
         message,
         source: "{{link-to 'Link text' 'routeName' one two}}",
         line: 1,
         column: 0,
+        isFixable: true,
+      },
+    },
+    {
+      template: "{{link-to (concat 'Hello' @username) 'routeName' one two}}",
+      fixedTemplate: "{{#link-to 'routeName' one two }}{{concat 'Hello' @username }}{{/link-to}}",
```
rwjblue: I think this specific example could _theoretically_ remove `concat` (since `concat` specifically is needed inside mustaches, but not in the content position).

Seems fine for a later cleanup though...

lib/rules/inline-link-to.js
```diff
@@ -1,20 +1,42 @@
...
-          });
+          if (this.mode === 'fix') {
+            let titleNode = node.params[0];
+
+            let newBody;
+            if (titleNode.type === 'SubExpression') {
+              newBody = b.mustache(titleNode.path, titleNode.params, titleNode.hash);
+            } else if (titleNode.type === 'StringLiteral') {
+              newBody = b.text(titleNode.value);
+            }
```
rwjblue: What about other types (e.g. `{{link-to 1234 'blah.route'}}`), should we add an `else` and log the error?

lib/rules/inline-link-to.js
```diff
@@ -1,20 +1,42 @@
...
+              node.hash,
+              b.blockItself([newBody])
+            );
+          } else {
+            this.log({
+              message,
+              line: node.loc && node.loc.start.line,
+              column: node.loc && node.loc.start.column,
+              source: this.sourceForNode(node),
+              isFixable: true,
```
rwjblue: At the moment it is only fixable if `node.params[0]` is either a `SubExpression` or `StringLiteral`, can you update this to only be true when we can _for sure_ fix it, or add support for the other types of literal?

lib/rules/inline-link-to.js
Turbo87: how do we log the error in this mode? the regular `this.log()` again?

Turbo87: done, I think 😅

lib/rules/inline-link-to.js
Turbo87: done

test/unit/rules/inline-link-to-test.js
Turbo87: yep, probably fine to do this separately. also I'm wondering if it makes sense to do this as a dedicated `no-unnecessary-concat-helper` rule instead.

rwjblue: Ya, sounds good to me!

PR #1184, opened 2020-03-02
.template-lintrc.js: Resolve ember-template-lint-plugin-* packages

dcyriller:
A plugin / package named `ember-template-lint-plugin-mypackage` could be used with the following config:

`.template-lintrc.js`:
```js
{
  plugins: 'mypackage',
  extends: ['mypackage:recommended']
}
```

For now, it is only possible to load it with:
```js
{
  plugins: 'ember-template-lint-plugin-mypackage',
  extends: ['ember-template-lint-plugin-mypackage:recommended']
}
```

If the idea sounds good, I will update the file `docs/plugins.md`.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -29,14 +29,26 @@ function requirePlugin(pluginName, fromConfigPath) {
     return resolve.sync(name, { basedir });
   }
 
+  // attemps to find the plugin by its name
   try {
-    // throws exception if not found
     pluginPath = getPluginPath(pluginName);
-  } catch (error) {
+  } catch (e) {
+    let isModuleMissingError = ALLOWED_ERROR_CODES.includes(e.code);
+    if (!isModuleMissingError) {
+      throw e;
+    }
+
+    // attemps to find the plugin its name prefixed with `ember-template-lint-plugin`
+    // throws exception if not found
     try {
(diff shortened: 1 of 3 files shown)


lib/get-config.js
```diff
@@ -22,8 +22,22 @@ const ALLOWED_ERROR_CODES = [
...
+  try {
+    // throws exception if not found
+    pluginPath = getPluginPath(pluginName);
+  } catch (error) {
+    try {
+      pluginPath = getPluginPath(`ember-template-lint-plugin-${pluginName}`);
+    } catch (_) {
+      throw error;
+    }
+  }
```
rwjblue: I think we'd need to structure this a bit different. The cases I'd like to ensure are handled are:

* we only attempt to resolve with the fuller name for `MODULE_NOT_FOUND` types of errors, not any possible error
* when called with `foo`, if neither `foo` nor `ember-template-lint-plugin-foo` are found the resulting error should mention `foo` (or should mention that we checked for both and found neither)

dcyriller: Indeed, the logic was not quite right. It should be good now :)

I also added a test case to make sure `foo` > `ember-template-lint-plugin-foo`

rwjblue:
Sorry @dcyriller, looks like my changes from #1189 caused a conflict here.

rwjblue:
also, I think only docs are left after rebasing...

PR #1187, opened 2020-03-02
Adds new rule: no-duplicate-id

MelSumner:
Adds a new rule that provides a cursory check for duplicate values in `id` attributes.


lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,39 @@
...
+          return;
+        }
+
+        if (!(node.value.type === 'TextNode')) {
+          return;
+        }
+
+        attrIdStack.push(node.value.chars);
+
+        let hasDuplicates = attrIdStack.some(
```
bmish: It would be more efficient to use an object (which is basically a hashtable) to keep track of which IDs we have seen instead of an array.

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,39 @@
...
+module.exports = class NoDuplicateId extends Rule {
+  visitor() {
+    let attrIdStack = [];
+    return {
+      AttrNode(node) {
+        if (!(node.name === 'id')) {
+          return;
+        }
+
+        if (!(node.value.type === 'TextNode')) {
```
Turbo87:
```suggestion
        if (node.value.type !== 'TextNode') {
```

seems simpler, no?

lib/rules/no-duplicate-id.js
Turbo87: I would recommend `new Set()`, but I'm not yet sure I entirely understand what the `.some()` here does 😅

test/unit/rules/no-duplicate-id-test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  config: true,
+
+  good: [
+    // Unique sibling TextNode IDs
+    '<div id="id-00"></div><div id="id-01"></div>',
+
+    // Ignore MustacheStatements, even duplicates
+    '<div id={{this.divId}}></div>',
+    '<div id={{this.divId}}></div><div id={{this.divId}}></div>',
```
Turbo87: I guess technically we could detect this simple case, but I guess it's fine for now the way it currently works :)

docs/rule/no-duplicate-id.md
```diff
@@ -0,0 +1,31 @@
+# TODO: no-duplicate-id
```
rwjblue: TODO?!?!?!?!?!

MelSumner:
```suggestion
# no-duplicate-id
```

lib/rules/no-duplicate-id.js
rwjblue: Ya, agree. We should swap to `Set` and do something like:

```js
if (seen.has(idValue)) {
  this.log(...)
} else {
  seen.add(idValue);
}
```

josephdsumner:
@MelSumner 
I would love to work on this -- would that be okay with you?

MelSumner:
@josephdsumner that would be really helpful, thanks for checking!

mehulkar:
Is it worth considering an HTML validator as part of the toolchain instead of implementing rules like these? (or is this cheaper/easier to do?)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "eslint.enable": true
+}
```
rwjblue, line 3: Please remove, we shouldn't check in editor specific files.

MelSumner: yup, totally just a mistake. Thanks for catching!

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,150 @@
...
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'ID attribute values must be unique';
+
+// Utility functions
+// -----------------
+
+// Visiting function node filter for AttrNodes: attribute name, node type
+function isValidIdAttrNode(node) {
+  let isIdAttrTag = ['id', '@id'].includes(node.name);
```
MelSumner, line 12:
```suggestion
// consider removing `@id` eventually (or make it a toggle available via config)
  let isIdAttrTag = ['id', '@id'].includes(node.name);
```

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,150 @@
...
+  );
+  return node && isIdAttrTag && isValidnodeType;
+}
+
+// MustacheStatement is StringLiteral
+function isMustacheString(node) {
+  return (
+    node && node.value.type === 'MustacheStatement' && node.value.path.type === 'StringLiteral'
+  );
+}
```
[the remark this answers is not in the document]
MelSumner: I extracted it to an issue - please edit to provide more clarity if you have something in mind already - https://github.com/ember-template-lint/ember-template-lint/issues/1425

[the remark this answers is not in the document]
MelSumner: Also, working through it in the context of this rule specifically, what do you think of this approach?

```js
const ALLOWED_CONFIGS = [
    {
        type: 'ElementNode',
        tagNames: ['id', '@id']
    },
    {
        type: 'BlockStatement',
        tagNames: ['elementId']
    }
]

function isAllowedConfig(nodeToTest) {
    let testNodeType = nodeToTest.type;
    let testNodeTagName = nodeToTest.idTag;

    let allowedConfigForNodeType = 
        ALLOWED_CONFIGS
            .find(node => 
                node.type === testNodeType
            );

    let hasValidTagName = 
        allowedConfigForNodeType
            .tagNames
            .includes(testNodeTagName);

    console.log(hasValidTagName);
}
```

This results in: 

```js
// True
isAllowedConfig({type:"ElementNode",idTag:"id"})
isAllowedConfig({type:"ElementNode",idTag:"@id"})
isAllowedConfig({type:"BlockStatement",idTag:"elementId"});

// False
isAllowedConfig({type:"ElementNode",idTag:"elementId"})
isAllowedConfig({type:"BlockStatement",idTag:"id"})
isAllowedConfig({type:"BlockStatement",idTag:"@id"})
```

MelSumner:
This PR is blocked on #1430

MelSumner:
@Turbo87 @bmish @rwjblue this integrates the previous review comments and integrates the new matches API, should be ready to re-review. 👍🏻

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,139 @@
...
+    // Resolve MustacheStatement value to StringLiteral where possible
+    function getMustacheValue(part, scope) {
+      let refNodeStr = { type: 'MustacheStatement', path: { type: 'StringLiteral' } };
+      let refNodeDyn = { type: 'MustacheStatement', path: { type: 'PathExpression' } };
+      if (NodeMatcher.match(part, refNodeStr)) {
+        return part.path.value;
+      }
+      if (NodeMatcher.match(part, refNodeDyn)) {
+        return scope.sourceForNode(part); // this
+      }
```
rwjblue, line 38: We should think about adding support for special conditionals to make this matching easier.

```js
NodeMatcher.match(part, { 
  type: 'MustacheStatement', 
  path: NodeMatcher.or([ { type: 'StringLiteral' }, { type: 'PathExpression' }])
});
```

^ would be nice here

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,139 @@
...
+          return;
+        }
+
+        let idValue = idHashArg.value.value;
+
+        if (idValueSet.has(idValue)) {
+          this.logNode({ node, message: ERROR_MESSAGE });
+        } else {
+          idValueSet.add(idValue);
+        }
```
rwjblue, line 133: This seems like this is duplicated between `BlockStatement` and here, can we share the code?

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,139 @@
...
+
+    // Resolve MustacheStatement value to StringLiteral where possible
+    function getMustacheValue(part, scope) {
+      let refNodeStr = { type: 'MustacheStatement', path: { type: 'StringLiteral' } };
+      let refNodeDyn = { type: 'MustacheStatement', path: { type: 'PathExpression' } };
+      if (NodeMatcher.match(part, refNodeStr)) {
+        return part.path.value;
+      }
+      if (NodeMatcher.match(part, refNodeDyn)) {
+        return scope.sourceForNode(part); // this
```
MelSumner, line 37:
```suggestion
        return scope.sourceForNode(part);
```

rwjblue (approved):

lib/rules/no-duplicate-id.js
```diff
@@ -0,0 +1,150 @@
...
+
+// Utility functions
+// -----------------
+
+// Visiting function node filter for AttrNodes: attribute name, node type
+function isValidIdAttrNode(node) {
+  let isIdAttrTag = ['id', '@id'].includes(node.name);
+  let isValidnodeType = ['TextNode', 'MustacheStatement', 'ConcatStatement'].includes(
+    node.value.type
+  );
```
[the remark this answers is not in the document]
rwjblue: `<Foo id={{1}} />` is going to be a `MustacheStatement` (as is the other example) 🤦

PR #1195, opened 2020-03-03
Add v2 migration guide

bmish:
The goal is to summarize and make it easier to understand the many breaking changes in the v2 release.

Modeled after the [eslint v6 migration guide](https://eslint.org/docs/user-guide/migrating-to-6.0.0).


rwjblue:
I updated the releases page also.

rwjblue:
Thanks for doing this @bmish!

PR #1198, opened 2020-03-04
Fix ember-template-lint-plugin-* packages resolution

dcyriller:
## Description

I implemented https://github.com/ember-template-lint/ember-template-lint/pull/1184 to be able to load `ember-template-lint-plugin-foo` with this config:

```js
module.exports = {
  plugins: ['foo'],

  extends: ['recommended', 'foo:recommended']
};
```
instead of
```js
module.exports = {
  plugins: ['ember-template-lint-plugin-foo'],

  extends: ['recommended', 'ember-template-lint-plugin-foo:recommended']
};
```

But with that implementation, if `foo` is a valid module, template-lint will try to load `foo` instead of `ember-template-lint-plugin-foo`.

## Fix

This PR proposes to adjust the first implementation. With this PR, template-lint would first try to load `ember-template-lint-plugin-foo` and then, if the first attempt fails, `foo`.

Note that if `ember-template-lint-plugin-foo` is absent and `foo` is not a template-lint plugin, user would get a warning (I added a test for this scenario).

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -33,19 +33,19 @@ function requirePlugin(pluginName, fromConfigPath) {
     return resolve.sync(name, { basedir });
   }
 
-  // attemps to find the plugin by its name
+  // attemps to find the plugin its name prefixed with `ember-template-lint-plugin`
   try {
-    pluginPath = getPluginPath(pluginName);
+    pluginPath = getPluginPath(`ember-template-lint-plugin-${pluginName}`);
   } catch (e) {
     let isModuleMissingError = ALLOWED_ERROR_CODES.includes(e.code);
     if (!isModuleMissingError) {
       throw e;
     }
 
-    // attemps to find the plugin its name prefixed with `ember-template-lint-plugin`
+    // attemps to find the plugin by its name
     // throws exception if not found
     try {
-      pluginPath = getPluginPath(`ember-template-lint-plugin-${pluginName}`);
+      pluginPath = getPluginPath(pluginName);
     } catch (e) {
       let isModuleMissingError = ALLOWED_ERROR_CODES.includes(e.code);
       if (!isModuleMissingError) {
--- a/test/unit/get-config-test.js
+++ b/test/unit/get-config-test.js
@@ -350,7 +350,7 @@ describe('get-config', function() {
       expect(actual.rules['quotes']).toEqual({ config: 'single', severity: 2 });
     });
 
-    it('favors foo plugin over ember-template-lint-plugin-foo', function() {
+    it('favors ember-template-lint-plugin-foo plugin over foo', function() {
       let console = buildFakeConsole();
 
       project.setConfig({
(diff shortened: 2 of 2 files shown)


rwjblue:
Am I correct that your scenario is for `prettier` (which exists in the project) and `ember-template-lint-prettier` where you want specifying `plugins: ['prettier']` to mean `ember-template-lint-plugin-prettier`?

dcyriller:
> Am I correct that your scenario is for prettier (which exists in the project) and ember-template-lint-prettier where you want specifying plugins: ['prettier'] to mean ember-template-lint-plugin-prettier?

It is exactly for that

dcyriller:
> I don't understand the goal here. In general, this doesn't seem like a mistake in the original logic to me. Specifically, we intentionally need to prefer foo over ember-template-lint-plugin-foo (or this is a breaking change?).

With the current logic, we end up loading prettier as a template lint plugin instead of ember-template-lint-plugin.

rwjblue:
I think we have to still prefer `foo` first (to avoid this being a breaking change), but we could make it a bit more succesful during the transitionary phase:

* Add a `keyword` that we look for in `./package.json`, document this in `docs/plugin.md` (and in a future release warn when we don't find the keyword)
* lookup `foo/package` and `ember-template-lint-plugin-foo/package` and search each for the keyword, if one has it and one doesn't use the one that has it
* lookup `foo` and `ember-template-lint-plugin-foo`, check the entry points to see if either have _only_ the keys that we expect. If one "looks like a template lint plugin" use it.
* We couldn't find either thing, throw a nice error saying that we couldn't resolve `foo`.

@dcyriller - What do you think?

rwjblue:
I made a small tweak to the above decision tree:

* if we have to expand `foo` to `ember-template-lint-plugin-foo`, we can _require_ that that thing has the keyword (since this is new functionality)

rwjblue:
@dcyriller - I'm going to revert the original PR for now, I'd like to release the other changes on master (severity feature, yargs, some `--fix`ers for rules) and I don't want to accidentally release the `foo` -> `ember-template-lint-plugin-foo` expansion before we have nailed this down. Sorry for the troubles, but I'm glad you tested and identified the issue quickly...

dcyriller:
Sorry @rwjblue, I didn't have much time this afternoon.

> I think we have to still prefer foo first (to avoid this being a breaking change), but we could make it a bit more succesful during the transitionary phase:

This PR would not be a breaking change anymore since the initial implementation has been reverted before being published 😜.

> Add a keyword that we look for in ./package.json, document this in docs/plugin.md (and in a future release warn when we don't find the keyword)
> ...

Sounds reasonable, I'll have a look in-depth tomorrow. :)

rwjblue:
Friendly ping 😄

dcyriller:
ahah I didn't find the time yet to complete this PR 😸. But it is still on my list! I plan to restart this PR from scratch - taking into account your insights. So feel free to close it if you'd like.

rwjblue:
Seems good, no rush on my end :smile:

PR #938, opened 2019-11-19
Add cli param to disable default ignores

velrest:
This is a second attempt for #659 

Add flag so the by default ignored files
`['**/dist/**', '**/tmp/**', '**/node_modules/**']` and `.gitignore`
can be toggled. This is useful if for instance your editor lint plugin
(vim/ale) copies the file to be linted into /tmp . This prevents the
template from being linted as '**/tmp/**' applies.

Also add `--help` command to list available options.

PR #1200, opened 2020-03-05
Add `--ignore-pattern` and `--no-ignore-pattern` command line flags

kangax:
Based on https://github.com/ember-template-lint/ember-template-lint/pull/938, adds `--disable-ignore-patterns` flag.

I originally went with `--no-ignore-patterns` only to find out (after a bit of hair pulling) that [yargs treats `--no` -prefixed args in a special way](https://github.com/yargs/yargs/blob/master/docs/tricks.md#negating-boolean-arguments) 🤦‍♂The other option would be smth like `--no-ignore-patterns=false` which is a mouthful. I'm already not quite happy with the "disable ignore" double negative but at least there's no need for a value.

Let me know if anything else is needed here. Shall we also add `--ignore-patterns`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Example usage:
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --print-pending
 
 # specify custom ignore pattern `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
-./node_modules/.bin/ember-template-lint /tmp/template.hbs --ignore-pattern "**/foo/**" "**/bar/**"
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --ignore-pattern "**/foo/**" --ignore-pattern "**/bar/**"
 
 # disable ignore pattern entirely
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
--- a/README.md
+++ b/README.md
@@ -81,8 +81,11 @@ Example usage:
 # print list of formated rules for use with `pending` in config file
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --print-pending
 
-# do not ignore `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
-./node_modules/.bin/ember-template-lint /tmp/template.hbs --disable-ignore-patterns
+# specify custom ignore pattern `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --ignore-pattern "**/foo/**" "**/bar/**"
+
+# disable ignore pattern entirely
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
 ```
 
 ### ESLint
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ Example usage:
 
 # print list of formated rules for use with `pending` in config file
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --print-pending
+
+# do not ignore `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --disable-ignore-patterns
 ```
 
 ### ESLint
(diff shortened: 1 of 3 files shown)


rwjblue:
I was thinking that if we did `--ignore-pattern="some-folder/**/*" - -ignore-pattern="node_modules/**/*"`, I _think_ that yargs can be configured here to give us an array, and that array would replace the built ins. We could then add an additional `--no-ignore-patterns` would disable ignore patterns completely. 

What do you think?

test/acceptance/cli-test.js
```diff
@@ -704,6 +737,29 @@ Options:
     });
   }
 
+  function setProjectConfigForErrorsWithIgnoredDir() {
```
rwjblue: Let’s just move this inline into the specific test that uses it (I generally prefer to keep the input and the expectations together such that reviewing the test is easier without jumping around too much), especially since it isn’t used a bunch of times like some of the others.

kangax: Sure thing, unrolled the setup

kangax:
> I was thinking that if we did `--ignore-pattern="some-folder/**/*" - -ignore-pattern="node_modules/**/*"`, I _think_ that yargs can be configured here to give us an array, and that array would replace the built ins. We could then add an additional `--no-ignore-patterns` would disable ignore patterns completely.
> 
> What do you think?

Ok, let me try that. So we'll have either of these scenarios:

- no arguments are given, default ignores are used
- `--ignore-pattern` is given, use that as a list of values
- `--no-ignore-pattern`

kangax:
@rwjblue updated, check it out

README.md
```diff
@@ -80,6 +80,12 @@ Example usage:
 
 # print list of formated rules for use with `pending` in config file
 ./node_modules/.bin/ember-template-lint app/templates/application.hbs --print-pending
+
+# specify custom ignore pattern `['**/dist/**', '**/tmp/**', '**/node_modules/**']` by default
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --ignore-pattern "**/foo/**" "**/bar/**"
```
rwjblue: This confuses me a tad, I think the `**/bar/**` here is going to be the files to lint (not part of the ignore patterns), right?

README.md
```diff
@@ -80,6 +80,12 @@ Example usage:
...
+
+# disable ignore pattern entirely
+./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
```
rwjblue: I know either way works, but mind putting the options before the files to lint?

```suggestion
./node_modules/.bin/ember-template-lint --no-ignore-pattern /tmp/template.hbs 
```

README.md
kangax: Nope, because we specify type of `--ignore-pattern` flag as an array, it allows us to list values separated by space (https://github.com/yargs/yargs/blob/master/docs/api.md#arraykey); I'm guessing it parses until EOL or encountering next flag with `--`

README.md
kangax: hm, wouldn't it be inconsistent with all the other examples?

kangax:
@rwjblue added a test as per your suggestion; also added both "foo" and "bar" dirs to a fake project to make sure both patterns are recognized and ignored

README.md
rwjblue: Gotcha, thank you for the docs link! Seems like we'd need to then use `--` to separate the actual paths to be linted, maybe tweak this example to show that?

rwjblue: Actually, can we set `'greedy-arrays': false` to the `.parserConfiguration` in the `yargs` options? I think this will make it have the behavior that I expected...

See https://github.com/yargs/yargs-parser#greedy-arrays

README.md
rwjblue: Yikes! You are right, lets fix this in a separate PR...

README.md
kangax: Updated and removed a test that uses consecutive args

rwjblue: Awesome, thank you! I pushed https://github.com/ember-template-lint/ember-template-lint/pull/1200/commits/5415760ee1e13bee219172672a60c05206a50aac to fixup this README example now that we require `--ignore-pattern` in front of each.

rwjblue:
Thanks @kangax!

PR #1202, opened 2020-03-05
Update contributing info in README

pablobm:
The readme makes reference to `npm install` and `npm test`, even though the project uses Yarn. Perhaps it would be best to just link to the Contributing Guidelines?


dcyriller:
Thank you :)

PR #1219, opened 2020-03-19
feature/use-concat-for-string-concatenation

betocantu93:
This is a draft PR, just wanted some feedback.

Found a bug in some parts of our codebase where strings were concatenated inside quotes, this caused that CMD + P + Format added newlines `\n`, breaking the UI.

This is the solution I came up with, but not sure if it's valid. Also, I just learned to do this thanks via the ast-workshop, so thank you @Turbo87 !

### DIFF
--- a/docs/rule/use-concat-for-string-concatenation.md
+++ b/docs/rule/use-concat-for-string-concatenation.md
@@ -1,0 +1,24 @@
+# use-concat-for-string-concatenation
+
+This rule requires the usage of `concat` helper to build strings for element attributes, which ideally prevents common formatting errors (i.e. CMD + P + `Format`) with plain concatenation.
+
+## Forbidden
+
+```hbs
+<div class="flex-{{if this.show "100" "50"}}"></div>
+```
+
+## Allowed
+
+```hbs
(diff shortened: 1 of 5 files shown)


docs/rule/use-concat-for-string-concatenation.md
```diff
@@ -0,0 +1,24 @@
+# use-concat-for-string-concatenation
+
+This rule requires the usage of `concat` helper to build strings for element attributes, which ideally prevents common formatting errors (i.e. CMD + P + `Format`) with plain concatenation.
+
+## Forbidden
+
+```hbs
+<div class="flex-{{if this.show "100" "50"}}"></div>
```
Turbo87, line 8: thanks for the PR @betocantu93. 

I have to strongly disagree with this rule though... 😅 

I think it might be better to fix the automatic formatter that you're using instead of making the templates more complex because of the formatter :)

betocantu93: that makes sense, thanks @Turbo87 for the feedback, it works for us for now, until we fix the formatter

Issue #1192, opened 2020-03-03
New Rule: no-unbalanced-curlies


Turbo87:
I recently introduced a small typo like this in one of my projects:

```hbs
{t "foo"}}
```

I guess it would be nice to count the number of opening double curlies and compare it to the number of closing double curlies, to make sure that you're not running into a similar issue.

rwjblue:
@turbo87 - I would expect that to emit a parser error, doesn't it? I guess it may not because the opening was a single curly?

Turbo87:
> I guess it may not because the opening was a single curly?

exactly. it just treated it as regular text. I guess for missing closing curly it would work, but apparently not for the opening curly.

Turbo87:
I guess an alternative and simpler implementation would be to check `chars.includes('}}')` in each `TextNode` 🤔

pablobm:
I'm not sure there's a good solution for this one.

First, here are some scenarios:

```
# Valid syntax. Probably mistakes.
{ x }}
{ x }
{ x 
 x }
 x }}
 x }}}
}
}}
}}}
{

# Syntax errors
{{ x }
{{ x }}}
{{{ x }
{{{ x }}
}}
}}}
}}}}... (any number of closing curlies past one)
```

Initially I thought we could just warn of any unscaped curlies. However I think this only works for `\{{` and `\{{{`. For example, the string `\{` renders verbatim `\{`. As a result, we can't just use the backslash as a general mechanism to mark curlies as "safe", as it were.

An idea would be for a rule to ban any unscaped curlies (except the handlebars ones, that is). This would leave these ways to render curlies:

* Disabling the rule for the affected scope.
* Escaping where possible: `\{{`, but not on `\{` or `\}}`.
* Indirectly as a handlebars expression: `{{ "{" }}`.
* As a property value.

And of course the rule wouldn't be called `no-unbalanced-curlies` any more.

Thoughts?

Turbo87:
@pablobm I think the easiest approach would be to look for `TextNodes` that `chars.include('}}')`. I don't think we need to address all of those cases above, especially since some of them are already covered by syntax errors. We could consider looking for `}}}` too, but since that's already covered by `}}` I'm not sure if it's really necessary.

Turbo87:
Resolved by #1222

PR #1222, opened 2020-03-20
Add new rule: no-unbalanced-curlies

pablobm:
Here's a first approach at solving https://github.com/ember-template-lint/ember-template-lint/issues/1192

Implementation is as suggested in the issue, but this leaves me wondering if there should be a bit more, or if the name of the rule should be different.

If the rule was triggered at any curly, opening or closing, we'd also be able to detect `{ x }`, `{` and `}`. These strike me as unlikely things to find in a template, and have alternatives.

Also note that the list of syntax errors vs. valid text (in `docs/rule/no-unbalanced-curlies.md`) has been corrected from that in the original issue, which was incorrect.

### DIFF
--- a/lib/rules/no-unbalanced-curlies.js
+++ b/lib/rules/no-unbalanced-curlies.js
@@ -1,24 +1,21 @@
 'use strict';
 
 const Rule = require('./base');
-const AstNodeInfo = require('../helpers/ast-node-info');
 
-// TODO Change template to the real error message that you want to report
-const ERROR_MESSAGE = 'Error Message to Report';
+const ERROR_MESSAGE = 'Unbalanced curlies detected';
+const DISALLOWED_CHARS = '}}';
 
 module.exports = class NoUnbalancedCurlies extends Rule {
   visitor() {
     return {
-      // Simple template example: disallowed text present in TextNode
       TextNode(node) {
-        let source = this.sourceForNode(node);
-        let disallowedText = 'DisallowedText';
-        let failingCondition = source.includes(disallowedText);
-        if (failingCondition) {
+        let chars = node.chars;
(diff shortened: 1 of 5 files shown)


lib/rules/no-unbalanced-curlies.js
```diff
@@ -0,0 +1,27 @@
...
+  visitor() {
+    return {
+      TextNode(node) {
+        let chars = node.chars;
+        if (chars.includes(DISALLOWED_CHARS)) {
+          let loc = node.loc;
+          this.log({
+            message: ERROR_MESSAGE,
+            line: loc && loc.start.line,
+            column: loc && loc.start.column + chars.indexOf(DISALLOWED_CHARS),
```
Turbo87, line 18: I'm not sure if this work, if the `chars` contain line breaks 🤔 

could you add a test case for that?

pablobm: Good catch; I added a few cases. I wasn't sure how much to test. I may have overdone it (or underdone?). Incidentally, I took the regexp from `lib/rules/linebreak-style.js`. Anyway, some new questions:

1. My implementation allows for more than one warning, as the logger is in the loop. I couldn't quite see how to write the test for this. Did I miss an example somewhere? Should I bother?
2. Should some code be refactored and DRY'd between `lib/rules/linebreak-style.js` and this?
3. Should `source` (in `this.log()`) be only the affected line, or should it be the whole node (as it is now)?
4. Also reading from `lib/rules/linebreak-style.js`, I see it searches from line breaks within `this.sourceForNode(node)`, while I'm looking in `chars`. What's the difference? Should I do the same?

And probably others I'm missing now!

Issue #1230, opened 2020-03-25
Add ability to lint .html files


MelSumner:
Adding the ability to lint the `.html` files as well as the `.hbs` files. This would allow us to check for the language attribute on the body element, among other things. 

Relevant area of the codebase: https://github.com/ember-template-lint/ember-template-lint/blob/0acaf3513857203c90ae973f76446ecfc61700d1/bin/ember-template-lint.js#L37

Newly generated ember-cli apps use `ember-template-lint .` as part of their `lint:hbs` `package.json` script. This illustrates why making `ember-template-lint` allow `.html` files would be a breaking change (see our [versioning policy](https://github.com/ember-template-lint/ember-template-lint/blob/master/dev/versioning.md) for more details).

General path forward:

1. Allow consumers to **explicitly** pass globs that include/allow for `.html`. At this stage, calling `ember-template-lint **/*.*` would lint against HTML files but `ember-template-lint .` or `ember-template-lint app/templates` (specifying a non-glob path) would not. This can be released without a major version bump.
2. Test to ensure that our linting rules do not flag issues with a newly generated ember-cli app's `index.html`. Fix any issues identified (either with rule changes, or with ember-cli changes).
3. Change the default globbing patterns to allow `.html` by default. This is the breaking change, and requires a major version bump.

Turbo87:
https://github.com/glimmerjs/glimmer-vm/issues/870 might be relevant for this issue 🤔

rwjblue:
@MelSumner - Can you review to confirm if this is fully resolved now?

MelSumner:
this is indeed resolved with #1232 and improved with #1631 . Closing issue.

PR #1232, opened 2020-03-25
Add `.html` to supported file extensions

lifeart:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1230

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -21,15 +21,19 @@ const readFile = promisify(fs.readFile);
 
 const STDIN = '/dev/stdin';
 
+function removeExt(filePath) {
+  return filePath.slice(0, -path.extname(filePath).length);
+}
+
 async function buildLinterOptions(filePath, filename = '', isReadingStdin) {
   if (isReadingStdin) {
     let filePath = filename;
-    let moduleId = filePath.slice(0, -4);
+    let moduleId = removeExt(filePath);
     let source = await getStdin();
 
     return { source, filePath, moduleId };
   } else {
-    let moduleId = filePath.slice(0, -4);
+    let moduleId = removeExt(filePath);
     let source = await readFile(path.resolve(filePath), { encoding: 'utf8' });
 
     return { source, filePath, moduleId };
@@ -51,12 +55,13 @@ async function lintSource(linter, options, shouldFix) {
 
 function expandFileGlobs(filePatterns, ignorePattern) {
   let result = new Set();
+  let supportedExtensions = new Set(['.hbs', '.html', '.handlebars']);
 
   filePatterns.forEach((pattern) => {
-    let isHBS = pattern.slice(-4) === '.hbs';
+    let isSupported = supportedExtensions.has(path.extname(pattern));
     let isLiteralPath = !isValidGlob(pattern) && fs.existsSync(pattern);
 
-    if (isHBS && isLiteralPath) {
+    if (isSupported && isLiteralPath) {
       let isIgnored = !micromatch.isMatch(pattern, ignorePattern);
 
       if (!isIgnored) {
@@ -69,7 +74,7 @@ function expandFileGlobs(filePatterns, ignorePattern) {
     globby
       // `--no-ignore-pattern` results in `ignorePattern === [false]`
       .sync(pattern, ignorePattern[0] === false ? {} : { ignore: ignorePattern, gitignore: true })
-      .filter((filePath) => filePath.slice(-4) === '.hbs')
+      .filter((filePath) => supportedExtensions.has(path.extname(filePath)))
       .forEach((filePath) => result.add(filePath));
   });
 
@@ -178,7 +183,7 @@ function printPending(results, options) {
     }, new Set());
 
     if (failingRules.size > 0) {
-      pendingList.push({ moduleId: filePath.slice(0, -4), only: [...failingRules] });
+      pendingList.push({ moduleId: removeExt(filePath), only: [...failingRules] });
     }
   }
   let pendingListString = JSON.stringify(pendingList, null, 2);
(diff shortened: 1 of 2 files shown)


bin/ember-template-lint.js
```diff
@@ -29,12 +29,13 @@ function lintFile(linter, filePath, toRead, moduleId, shouldFix) {
...
 function expandFileGlobs(positional, ignorePattern) {
   let result = new Set();
+  let supportedExtensions = ['.hbs', '.html'];
 
   positional.forEach((item) => {
     globby
       // `--no-ignore-pattern` results in `ignorePattern === [false]`
       .sync(item, ignorePattern[0] === false ? {} : { ignore: ignorePattern, gitignore: true })
-      .filter((filePath) => filePath.slice(-4) === '.hbs')
+      .filter((filePath) => supportedExtensions.includes(filePath.slice(filePath.lastIndexOf('.'))))
```
rwjblue: Thoughts on using `path.extname` instead of `.lastIndexOf`?

rwjblue:
FWIW, labeled this as breaking per our standard SemVer policy. Since most ember-cli apps are using `ember-template-lint .` they would automatically begin linting additional files that would never have been linted before and therefore _may_ have linting failures.

We’ll need to test this change out against a newly generated ember-cli app to see what is flagged in `app/index.html` and figure out the recommended mitigation policy.

lifeart:
@rwjblue I think we can download fresh app from https://github.com/ember-cli/ember-new-output before each test run, but it need to be discussed about best strategy to implement.

rwjblue:
FYI - I commented over on #1230 with some more actionable steps forward here.

> General path forward:
> 
> 1. Allow consumers to **explicitly** pass globs that include/allow for `.html`. At this stage, calling `ember-template-lint **/*.*` would lint against HTML files but `ember-template-lint .` or `ember-template-lint app/templates` (specifying a non-glob path) would not. This can be released without a major version bump.
> 2. Test to ensure that our linting rules do not flag issues with a newly generated ember-cli app's `index.html`. Fix any issues identified (either with rule changes, or with ember-cli changes).
> 3. Change the default globbing patterns to allow `.html` by default. This is the breaking change, and requires a major version bump.

MelSumner:
@lifeart would you please fix the conflicts? I'm hoping we can remove any blockers and get this in the next major release. Thank you!

bin/ember-template-lint.js
```diff
@@ -45,9 +45,10 @@ function lintSource(linter, options, shouldFix) {
 
 function expandFileGlobs(filePatterns, ignorePattern) {
   let result = new Set();
+  let supportedExtensions = new Set(['.hbs', '.html']);
 
   filePatterns.forEach((pattern) => {
-    let isHBS = pattern.slice(-4) === '.hbs';
+    let isHBS = path.extname(pattern) === '.hbs';
```
Turbo87, line 51: wouldn't we need to check for `.html` here too?

lifeart: good catch! changes landed

bin/ember-template-lint.js
```diff
@@ -45,12 +45,13 @@ function lintSource(linter, options, shouldFix) {
 
 function expandFileGlobs(filePatterns, ignorePattern) {
   let result = new Set();
+  let supportedExtensions = new Set(['.hbs', '.html']);
```
rwjblue, line 48: I think we should technically also support `.handlebars` (though it is super uncommonly used, it is still supported by modern ember-cli apps I think).

lifeart: It's kinda super legacy, do we have issues to add such support?
(land it - no problem, but remove it - may be in future)

rwjblue: Ya, maybe? I don't particularly care to be honest, it costs us very little to support which is why I think we should just mirror whatever ember-cli-htmlbars does by default. 

I also think it might be good to allow this exact list to be configurable in the config file.

lifeart: I would suggest to land ability to configure extensions after we land minimal possibility to lint js/ts files, to have stabilyzed logical path to conditionaly apply linting, based on "tricky" extensions

lifeart: @rwjblue one note to land `.handlebars` - we have to rewrite more code, because parts relay on 3symbols ext
image

rwjblue: What parts rely on 3 chars? We should be using `path.extname`...

lifeart: changes landed https://github.com/ember-template-lint/ember-template-lint/pull/1232/commits/768e9c04fe5ffcf9ff04a1f9060d322f4f44dccc

lib/rules/require-iframe-title.js
```diff
@@ -40,7 +40,7 @@ module.exports = class RequireIframeTitle extends Rule {
           } else if (title.value.type === 'TextNode') {
             let value = title.value.chars.trim();
             let matchIndex = -1;
-            let [, existingNode] = this.knownTitles.filter(([val], index) => {
+            let [, existingNode] = this.knownTitles.find(([val], index) => {
```
rwjblue, line 43: This change seems unrelated to `.html` files (though it seems fine to do, it should just be a separate PR).

lifeart: it was autofixed by `yarn lint:js --fix`

rwjblue: O_o why? Seems like that should have caused linting failures on master then, right?

lifeart: It may be unicorn rules warning fix

lifeart: reverted autofix
image

rwjblue (approved):

Issue #1203, opened 2020-03-06
Rule proposal: <form> method="POST"


raido:
```
{{! bad }}
<form {{on "submit” this.someMethodPreventsDefault}}>

{{! good }}
<form method="POST" {{on "submit” this.someMethodPreventsDefault}}>
```

In case `preventDefault` fails by major crash in application code, default browser action would be performed and can leak sensitive customer data as GET requests.

PR #1238, opened 2020-03-29
Add new rule: `require-form-method`

raido:
This PR adds new rule for form elements to require method attribute with value `POST/post` or `GET/get`.

Closes: https://github.com/ember-template-lint/ember-template-lint/issues/1203

### DIFF
--- a/docs/rule/require-form-method.md
+++ b/docs/rule/require-form-method.md
@@ -1,0 +1,47 @@
+# require-form-method
+
+This rule requires all `<form>` elements to have `method` attribute with `POST`, `GET` or `DIALOG` value.
+
+By default `form` elements without `method` attribute are submitted as `GET` requests.
+In usual applications `submit` event listeners are attached to `form` elements and `event.preventDefault()` is called to avoid form submission.
+
+However in case of failure to prevent default action, form submission as `GET` request can leak sensitive end-user information.
+
+Example uses of `GET` requests:
+
+* non-secure data
+* bookmarking the submission result
+* data search query strings
+
+**Caution** - this rules does not check for `formmethod` attribute on `form` elements themselves.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<form>Hello world!</form>
+<form method="">Hello world!</form>
+<form method="random">Hello world!</form>
+```
+
+This rule **allows** the following:
+
+```hbs
+<form method="post">Hello world!</form>
+<form method="get">Hello world!</form>
+<form method="dialog">Hello world!</form>
+```
+
+## Configuration
+
+The following values are valid configuration:
+
+* boolean - `true` to enable / `false` to disable
+* object -- An object with the following keys:
+  * `allowedMethods` -- An array of allowed form `method` attribute values, default: `['POST', 'GET', 'DIALOG']`
+
+## References
+
+* [MDN - form method attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method)
+* [HTML spec - form method attribute](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-method)
--- a/lib/rules/require-form-method.js
+++ b/lib/rules/require-form-method.js
@@ -1,0 +1,103 @@
+'use strict';
+
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+
+// Form `method` attribute keywords:
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-method
+const VALID_FORM_METHODS = ['POST', 'GET', 'DIALOG'];
+
+const DEFAULT_CONFIG = {
+  allowedMethods: VALID_FORM_METHODS,
+};
+
+module.exports = class RequireFormMethod extends Rule {
+  logNode({ node, message }) {
+    return this.log({
+      message,
+      line: node.loc && node.loc.start.line,
+      column: node.loc && node.loc.start.column,
+      source: this.sourceForNode(node),
+      isFixable: false,
+    });
+  }
+
+  parseConfig(config) {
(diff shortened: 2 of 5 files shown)


rwjblue:
Adding a rule for this seems totally fine, but I'm trying to decide if we should eventually try to get this into the default rule set (after a major bump).

@Turbo87 / @bmish / @dcyriller - Curious what y'all's thoughts are here?

bmish:
I thought there were many situations where `get` forms are completely acceptable, like search results `?query=my-search-term`.

test/unit/rules/require-form-method-post-test.js
```diff
@@ -0,0 +1,78 @@
...
+  config: true,
+
+  good: [
+    '<form method="POST"></form>',
+    '<form method="post"></form>',
+
+    // dynamic values
+    '<form method="{{formMethod}}"></form>',
+    '<form method={{formMethod}}></form>',
+  ],
```
bmish: Need to add `good` test cases to ensure false positives are avoided such as:

* `<div></div>`
* `<div method="get"></div>`

docs/rule/require-form-method-post.md
```diff
@@ -0,0 +1,30 @@
+# require-form-method-post
+
+This rule requires all `<form>` elements to have `method` attribute with `POST` value.
+
+By default `form` elements without `method` attribute are submitted as `GET` requests.
+In usual applications `submit` event listeners are attached to `form` elements and `event.preventDefault()` is called to avoid form submission.
+
+However in case of failure to prevent default action, form submission as `GET` request can leak sensitive end-user information.
```
bmish: This should mention that there are acceptable use cases for `GET` as well: https://www.w3schools.com/tags/att_form_method.asp

Turbo87:
I tend to agree with @bmish. I think converting this to `require-form-method` without enforcing the `post` value would be better, potentially with a `allow-get: false` option if necessary.

raido:
I’ll convert it to generic rule with config option for disabling GET. I thought about it when implementing this as well. But still went ahead with just post to get something out for feedback.

raido:
This is ready for review and has been converted to generic rule with configurability.

lib/rules/require-form-method.js
```diff
@@ -0,0 +1,90 @@
...
+  parseConfig(config) {
+    let configType = typeof config;
+
+    switch (configType) {
+      case 'boolean':
+        return config ? DEFAULT_CONFIG : false;
+      case 'object':
+        if (Array.isArray(config.allowedMethods)) {
+          return {
+            allowedMethods: config.allowedMethods.map((m) => String(m).toUpperCase()),
```
bmish: Can you throw an exception if any unexpected methods are passed in?

docs/rule/require-form-method.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Configuration
+
+The following values are valid configuration:
+
+* boolean - `true` to enable / `false` to disable
+* object -- An object with the following keys:
+  * `allowedMethods` -- An array of allowed form `method` attribute values, default: `['POST', 'GET']`
+
+## References
```
bmish: This is a good, simple link too: https://www.w3schools.com/tags/att_form_method.asp

docs/rule/require-form-method.md
```diff
@@ -0,0 +1,40 @@
+# require-form-method
+
+This rule requires all `<form>` elements to have `method` attribute with `POST` or `GET` value.
+
+By default `form` elements without `method` attribute are submitted as `GET` requests.
+In usual applications `submit` event listeners are attached to `form` elements and `event.preventDefault()` is called to avoid form submission.
+
+However in case of failure to prevent default action, form submission as `GET` request can leak sensitive end-user information.
+Examples of valid uses for `GET` requests are non-secure data, bookmarking the submission result, data search query strings.
```
bmish: I would prefer bullet-point list instead of a sentence here to make it easier to parse the acceptable usages.

raido: Converted to list.

docs/rule/require-form-method.md
raido: Added.

lib/rules/require-form-method.js
raido: Added. I also added "dialog" method type as by HTML spec to defaults list.

lib/rules/require-form-method.js
```diff
@@ -0,0 +1,90 @@
...
+
+        if (tag !== 'form') {
+          return;
+        }
+
+        let typeAttribute = attributes.find((it) => it.name === 'method');
+        if (!typeAttribute) {
+          this.logNode({
+            node,
+            message: ERROR_MESSAGE.replace('%', this.config.allowedMethods),
```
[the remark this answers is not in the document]
raido: Done.

raido:
I need to push a fix after breaking out of the loop. It does not break the switch and throw an exception in case invalid methods are configured as input. Missing conditional there for switch break.

docs/rule/require-form-method.md
```diff
@@ -0,0 +1,47 @@
...
+
+The following values are valid configuration:
+
+* boolean - `true` to enable / `false` to disable
+* object -- An object with the following keys:
+  * `allowedMethods` -- An array of allowed form `method` attribute values, default: `['POST', 'GET', 'DIALOG']`
+
+## References
+
+* [W3 Schools - form method attribute](https://www.w3schools.com/tags/att_form_method.asp)
```
mehulkar, line 46: Could link to mdn instead: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method

raido: I can replace that with MDN.

raido: Done.

dcyriller:
Thank you!

PR #1250, opened 2020-04-05
Update each rule doc to mention what config enables the rule

bmish:
Matches the eslint core plugin rule docs (example: https://eslint.org/docs/rules/no-extra-semi).

This makes it much easier to gain a complete picture of the rule by looking at its rule doc (without depending on the overall README.md / rules.md files).

And adds a test to ensure the docs are correct.

I made the same change in: https://github.com/ember-cli/eslint-plugin-ember/pull/759

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,11 +98,11 @@ ember-template-lint into your normal eslint workflow.
 
 ### Presets
 
-Name   | Description |
-|:-----|:------------|
-| [recommended](lib/config/recommended.js) | enables the recommended rules |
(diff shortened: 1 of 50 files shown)


bmish:
@Turbo87 Yes, that's a good next step. I've actually been thinking about adding the entire rules list directly to README.md so that people can see the list immediately when viewing the project (like pretty much every other lint plugin) instead of having to search for it.

rwjblue:
IMHO, our README needs a significant revamp. Having the rules and their recommended / default values seems good, but we also need to move around loads of other content that seems to clutter things up a bit.

PR #1251, opened 2020-04-06
Pass env from template-recast's transform callback to rules

dcyriller:
That way, we don't need to import ember-template-recast from external plugins to fix the AST.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -232,7 +232,7 @@ class Linter {
         fileConfig,
       });
 
-      let { code } = transform(currentSource, () => rule.getVisitor());
+      let { code } = transform(currentSource, (env) => rule.getVisitor(env));
       currentSource = code;
     }
 
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -81,10 +81,10 @@ module.exports = class Base {
     return config;
   }
 
-  getVisitor() {
+  getVisitor(env) {
     let pluginContext = this;
     let visitor = {};
-    let ruleVisitor = this.visitor();
+    let ruleVisitor = this.visitor(env);
     // We use this structure to advise/unadvise on AST events. The walkers we
     // set up read this structure every time they get an AST event and call the
     // appropriate functions. Handlers get added to the before/after list
--- a/test/unit/base-plugin-test.js
+++ b/test/unit/base-plugin-test.js
@@ -51,7 +51,7 @@ describe('base plugin', function () {
 
       let rule = new Rule(options);
 
-      transform(ast, () => rule.getVisitor());
+      transform(ast, (env) => rule.getVisitor(env));
     }
   }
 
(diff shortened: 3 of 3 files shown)


dcyriller:
It should make https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/issues/77 straightforward

rwjblue:
Released in v2.5.0!

Turbo87:
hmm... I would prefer to reexport the builders instead so that users can write independent functions with them, instead of having to embed them in the visitor method. 🤔

dcyriller:
It would be nice indeed. Do you see a drawback in supporting both? 🤔

Turbo87:
> Do you see a drawback in supporting both?

mostly just that this passing-the-api-to-the-method style can lead to slightly less readable and complicated code since people try to put everything into that one `visitor()` method 😉

rwjblue:
@Turbo87 - The builders are reexported from ember-template-recast, and you can already depend on that (e.g. as a dep or a peerDep) to get them.

Turbo87:
@rwjblue sure, but then you risk not getting the correct version 🤔

rwjblue:
@turbo87 - Ya, agree, however I can't actually tell what you are suggesting we do. :confused:  At the moment, to make external/third party rules you _already_ have to depend on `ember-template-lint` (at least as a peerDep) because your rule has to extend from that. Are you saying that `require('ember-template-lint').builders` should be a thing? If so, that seems perfectly fine to me..

Turbo87:
> Are you saying that require('ember-template-lint').builders should be a thing?

yeah, that was roughly the suggestion. I guess the exact import path is tbd, but that's what I meant by "reexport".

rwjblue:
KK, awesome, now I understand. Will fix.

dcyriller:
> mostly just that this passing-the-api-to-the-method style can lead to slightly less readable and complicated code since people try to put everything into that one visitor() method 😉

Sounds fair to me. I could submit a patch to fix that.

To be clear, I would add: I not only wanted to be able to use `builders` in an external plugin's rule but also template-recast's `parse` and `print` functions.

How do you see the api? Should we go with `require('ember-template-lint').helpers.builders`, `require('ember-template-lint').helpers.parse` and `require('ember-template-lint').helpers.print`?

Turbo87:
maybe we should make `require('ember-template-lint').recast` work? that way we'd have access to all of the necessary things

rwjblue:
How about adding something like this to `lib/index.js`:

```
module.exports.recast = require('ember-template-recast');
```

Turbo87:
yep, that's what I meant by `require('ember-template-lint').recast` :)

rwjblue:
yeah, had typed up the comment but forgot to submit (and GH didn't show me yours) :sob:

PR #1260, opened 2020-04-09
Expose `ember-template-recast` API on main entry point (e.g. `require('ember-template-lint').recast`)

dcyriller:
This is a second take for https://github.com/ember-template-lint/ember-template-lint/pull/1251.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,3 +3,4 @@
 module.exports = require('./linter');
 module.exports.Rule = require('./rules/base');
 module.exports.ASTHelpers = require('./helpers/ast-node-info');
+module.exports.recast = require('ember-template-recast');


lib/linter.js
```diff
@@ -232,7 +232,7 @@ class Linter {
         fileConfig,
       });
 
-      let { code } = transform(currentSource, (env) => rule.getVisitor(env));
+      let { code } = transform(currentSource, () => rule.getVisitor());
```
Turbo87, line 235: technically this is a breaking change, now that it had already been released 😞

dcyriller: Should we keep it or should we release this as a bug fix?

Turbo87: I'll let @rwjblue decide on this :)

rwjblue: I'd just leave it alone, and we can remove for next major. Doesn't really hurt anything to pass the argument...

dcyriller: ya, updated 👍

rwjblue:
Released in https://github.com/ember-template-lint/ember-template-lint/releases/tag/v2.5.1

PR #1263, opened 2020-04-11
Add `--rule` and `--no-config-path` command line options

### LINKED ISSUE #1262: Be able to run one rule at a time
# Motivation

Eslint has this option, if you want to test out a rule you can run `eslint --rule foo:error this_is_a_path`. It would be great if we could run one template lint rule at a time from the cli.

## Solution

```
ember-template-lint --rule foo:error this_is_a_path
```

gabrielcsapo:
# Motivation

There are times where engineers on our team would like to run one rule at a time, just to get a sense of where we are with a migration. eslint has two flags that enable this `--no-eslintrc` and `--rule` cli option. I have added `--rule` and `--no-config-path` that does exactly this https://eslint.org/docs/user-guide/command-line-interface.

## Usage

```sh
# running a single rule without options
./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule 'no-implicit-this:error'
 # running a single rule with options
./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule 'no-implicit-this:["error", { "allow": ["some-helper"] }]'
```

## Screenshots

> Running in a project without a .template-lintrc file local to that project and specifying a rule to run

Screen Shot 2020-04-16 at 1 23 51 AM
> Running in a project without a .template-lintrc file local to that project and specifying a rule to run with warnings as severity

Screen Shot 2020-04-16 at 1 23 57 AM

> Running in a project without a .template-lintrc file local to that project and specifying a rule to run with warnings as severity and specifying additional options

Screen Shot 2020-04-16 at 1 23 32 AM
> Running in a project without a .template-lintrc file local to that project and specifying a rule to run with error as severity and specifying additional options

Screen Shot 2020-04-16 at 1 23 22 AM

fixes #1262

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,9 @@ Example usage:
 
 # running a single rule with options
 ./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
+
+# specify a config object to use instead of what exists locally
+./node_modules/.bin/ember-template-lint --config '{ "rules": { "no-implicit-this": { "severity": 2, "config": true } } }' test/fixtures/no-implicit-this-allow-with-regexp/app/templates
 ```
 
 ### ESLint
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Example usage:
 ./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule 'no-implicit-this:error'
 
 # running a single rule with options
-./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule '{ "rules": { "no-implicit-this": "error" } }'
+./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule 'no-implicit-this:["error", { "allow": ["some-helper"] }]'
 
 # specify a config object to use instead of what exists locally
 ./node_modules/.bin/ember-template-lint --config '{ "rules": { "no-implicit-this": { "severity": 2, "config": true } } }' test/fixtures/no-implicit-this-allow-with-regexp/app/templates
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Example usage:
 ./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:error'
 
 # running a single rule with options
-./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rules": { "no-implicit-this": "error" } }'
 
 # specify a config object to use instead of what exists locally
 ./node_modules/.bin/ember-template-lint --config '{ "rules": { "no-implicit-this": { "severity": 2, "config": true } } }' test/fixtures/no-implicit-this-allow-with-regexp/app/templates
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Example usage:
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
 
 # running a single rule without options
-./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:2'
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:error'
 
 # running a single rule with options
 ./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ Example usage:
 
 # disable ignore pattern entirely
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
+
+# running a single rule without options
+./node_modules/.bin/ember-template-lint app/templates --rule 'no-implicit-this:2'
+
+# running a single rule with options
+./node_modules/.bin/ember-template-lint app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
 ```
 
 ### ESLint
--- a/README.md
+++ b/README.md
@@ -88,10 +88,10 @@ Example usage:
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
 
 # running a single rule without options
-./node_modules/.bin/ember-template-lint app/templates --rule 'no-implicit-this:2'
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:2'
 
 # running a single rule with options
-./node_modules/.bin/ember-template-lint app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
 ```
 
 ### ESLint
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -57,7 +57,7 @@ function parseArgv(_argv) {
       },
       rule: {
         describe:
-          'Specify a rule and it\'s severity to run that rule against required file paths (rule:severity) or can be specified as an json string "{ rule: String, severity: String, config: Object }"',
+          'Specify a rule and it\'s severity to add that rule to loaded rules - (rule:severity) or can be specified as an json string "{ rule: String, severity: String, config: Object }"',
         type: 'string',
       },
       filename: {
@@ -75,6 +75,10 @@ function parseArgv(_argv) {
       },
       verbose: {
         describe: 'Output errors with source description',
+        boolean: true,
+      },
+      'no-templaterc': {
+        describe: 'Does not use the local templaterc, will use a blank templaterc instead',
         boolean: true,
       },
       'print-pending': {
@@ -139,7 +143,11 @@ function run() {
 
   let linter;
   try {
-    linter = new Linter({ configPath: options.configPath, rule: options.rule });
+    linter = new Linter({
+      configPath: options.configPath,
+      rule: options.rule,
+      useConfig: options.templaterc,
+    });
   } catch (e) {
     console.error(e.message);
     process.exitCode = 1;
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -51,6 +51,10 @@ function parseArgv(_argv) {
         default: '.template-lintrc.js',
         type: 'string',
       },
+      config: {
+        describe: 'Define a custom config object - (--config \'{ "rules": { "no-implicit-this": { "severity": 2, "config": true } } }\')',
+        type: 'string',
+      },
       quiet: {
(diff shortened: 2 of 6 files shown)


rwjblue:
Hmm, this is actually quite different from the `--rule` ESLint argument ([docs here](https://eslint.org/docs/user-guide/command-line-interface#rule)):

> This option specifies rules to be used. These rules will be merged with any rules specified with configuration files.

It isn't saying to run _only_ this rule, it is saying to add this rule to the normal config. In order to run only the rule specified you have to do `--no-eslintrc` as well, which IIRC isn't a feature that we have yet.

gabrielcsapo:
Would that option be labeled —no-templaterc ? I can alter the —rule it to be up to spec as the docs day.

gabrielcsapo:
@rwjblue I have updated as per the spec you linked. Thoughts on this implementation?

rwjblue:
So, I'm totally fine with doing these changes, but it _feels_ like we are missing a higher level concept to be able to specify explicit config to be used. Currently we support specifying a config path, and this PR adds support for specifying `--rule` and `--no-templaterc`. But I think what we may be wanting is a `--config` to specify the **explicit** config to be used...

Thoughts?

rwjblue:
Basically I see it this way:

* Use `--rule` to test out what adding a given rule to the config would do
* Use `--config` to specify an **explicit** config to be used
* Use `--configPath` to specify a path to a config file for the **explicit** config to be used

gabrielcsapo:
Updated based on our conversation! Thanks for reviewing @rwjblue

README.md
```diff
@@ -86,6 +86,15 @@ Example usage:
 
 # disable ignore pattern entirely
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
+
+# running a single rule without options
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:2'
+
+# running a single rule with options
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rule": "no-implicit-this", "severity": 2, "config": { "allow": [] }}'
```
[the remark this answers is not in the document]
gabrielcsapo: how would you pass options to a rule?

lib/linter.js
```diff
@@ -34,15 +35,39 @@ class Linter {
 
     this.options = options;
     this.console = options.console || console;
+    // don't use the config on disk, if --no-templaterc or --config is specfied
+    this.useConfig =
+      options.useConfig !== undefined || options.overrideConfig ? options.useConfig : true;
```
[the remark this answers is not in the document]
gabrielcsapo: Going with --no-config-path makes sense, I went with that approach.

bin/ember-template-lint.js
```diff
@@ -134,7 +148,22 @@ function run() {
...
-    linter = new Linter({ configPath: options.configPath });
+    let config;
+
+    if (options.config) {
+      try {
+        config = JSON.parse(options.config);
+      } catch (ex) {
+        console.debug('Could not parse options passed through --config');
+      }
+    }
```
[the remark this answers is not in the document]
gabrielcsapo: Makes sense!

gabrielcsapo:
@rwjblue added all your feedback and more tests! let me know what else I can do!

README.md
```diff
@@ -86,6 +86,15 @@ Example usage:
 
 # disable ignore pattern entirely
 ./node_modules/.bin/ember-template-lint /tmp/template.hbs --no-ignore-pattern
+
+# running a single rule without options
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule 'no-implicit-this:error'
+
+# running a single rule with options
+./node_modules/.bin/ember-template-lint --no-templaterc app/templates --rule '{ "rules": { "no-implicit-this": "error" } }'
```
rwjblue, line 94:
```suggestion
# running a single rule without options
./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule 'no-implicit-this:error'

# running a single rule with options
./node_modules/.bin/ember-template-lint --no-config-path app/templates --rule '{ "rules": { "no-implicit-this": "error" } }'
```

bin/ember-template-lint.js
```diff
@@ -131,10 +146,29 @@ function printPending(results, options) {
...
+      return;
+    }
+  }
 
   let linter;
   try {
-    linter = new Linter({ configPath: options.configPath });
+    if (options['no-config-path'] !== undefined) {
+      options.configPath = false;
+    }
```
rwjblue, line 165: This probably doesn't need to be in the `try`/`catch` block.

lib/get-config.js
```diff
@@ -283,6 +286,51 @@ function processIgnores(config) {
...
+ * ex:
+ * rule:severity
+ * no-implicit-this:["error", { "allow": ["some-helper"] }]
+ */
+function getRuleFromString(rule) {
+  const indexOfSeparator = rule.indexOf(':') + 1;
+
+  // we have to split based on the index of the first instance of the separator because the separator could exist in the second half of the rule
+  const name = rule.substring(0, indexOfSeparator - 1);
+  let severity = rule.substring(indexOfSeparator);
```
rwjblue, line 300: FWIW, I think _technically_ this is not the "severity", it is the config for the rule (which **might** just be `"error"` (the severity) but could be any valid config value e.g. `["error", whateverGoesHere]`).

lib/get-config.js
```diff
@@ -283,6 +286,51 @@ function processIgnores(config) {
...
+    case 'error':
+      _severity = ERROR_SEVERITY;
+      break;
+    case 'warn':
+      _severity = WARNING_SEVERITY;
+      break;
+    case 'off':
+    default:
+      _severity = IGNORE_SEVERITY;
+  }
```
rwjblue, line 325: I think we already have a method that does this. All we should need to do is pass the config to`determineRuleConfig` right?

https://github.com/ember-template-lint/ember-template-lint/blob/0a336f262008346fd55e49f27f4ffa5d79587982/lib/get-config.js#L401-L423

lib/get-config.js
```diff
@@ -283,6 +286,51 @@ function processIgnores(config) {
...
+  let config = true;
+
+  // if we get passed an object that looks like no-implicit-this:["error", { "allow": ["some-helper"] }] we should split it accordingly
+  try {
+    const options = JSON.parse(severity);
+
+    severity = options[0];
+    config = options[1];
+  } catch (ex) {
+    // noop
```
rwjblue, line 310: Is this really a `noop`? Doesn't this mean that the user passed us garbage data? Assuming that is true, we really should throw an error here (just a nicer formatted one).

lib/linter.js
```diff
@@ -36,6 +37,17 @@ class Linter {
...
+
+    // we were passed a rule, add the rule being passed in, to the config.
+    // ex:
+    // rule:severity
+    // no-implicit-this:["error", { "allow": ["some-helper"] }]
+    if (options.rule) {
+      const { name, severity, config } = getRuleFromString(options.rule);
+
+      this.config.rules[name] = { severity, config };
+    }
```
rwjblue, line 49: Can we move this into `loadConfig`? it already uses `this.options` so it has access to all the same info. I'd generally prefer for the contract to be that once `.loadConfig()` is called everything about config loading is done (vs doing somethings in `loadConfig` and other things in this constructor).

gabrielcsapo: Okay makes sense, I moved it to loadConfig

lib/get-config.js
gabrielcsapo: I didn't see this! Resolved this!

lib/get-config.js
gabrielcsapo: We have to check if the user passed json or not, how would we do that without a try catch?

bin/ember-template-lint.js
gabrielcsapo: moved out of the try catch block

lib/get-config.js
gabrielcsapo: What we rather call this?

gabrielcsapo:
@rwjblue updated with a few questions

lib/get-config.js
rwjblue: AFAICT it is the `ruleConfig`, what do you think?

lib/get-config.js
rwjblue: YA, but the issue is that if the user **didn't** pass us JSON it is actually a failure scenario. Right?

rwjblue: AFAICT the only thing we expect _other than_ JSON is three literal strings (`warn`,`error`,`off`) which we can just check for then fallback to `JSON.parse`. We'd still need the try/catch in order to provide an error that is actually helpful (vs the default `JSON.parse` error).

gabrielcsapo: removed this in the most recent update.

lib/get-config.js
gabrielcsapo: makes sense to me, renamed that in the most recent diff.

lib/get-config.js
```diff
@@ -283,6 +286,44 @@ function processIgnores(config) {
...
+    }
+  }
+
+  const { severity, config } = _determineConfigForSeverity(ruleConfig);
+
+  return {
+    name,
+    severity,
+    config,
+  };
```
rwjblue, line 324: Thoughts on doing something like:

```suggestion
if (ruleConfig.startsWith('[')) {
  ruleConfig = JSON.parse(ruleConfig);
}

const config = determineRuleConfig(ruleConfig);

return { name, config };
```

Specifically:

* `determineRuleConfig` already handles dealing with an array or a string and returns the proper `severity`/`config` objects back. This allows us to avoid multiple calls to `_determineConfigForSeverity` (one for detecting, then another for actual processing).
* Avoid destructuring the object just to restructure it. The current destructuring/restructuring will cause errors if we ever decide to change the value returned by `determineRuleConfig` (which was intended as the single place that deals with converting userland config into our internal structure).

lib/linter.js
```diff
@@ -36,13 +37,24 @@ class Linter {
...
     this.config = getProjectConfig(this.options);
+
+    // we were passed a rule, add the rule being passed in, to the config.
+    // ex:
+    // rule:severity
+    // no-implicit-this:["error", { "allow": ["some-helper"] }]
+    if (this.options.rule) {
+      const { name, severity, config } = getRuleFromString(this.options.rule);
+
+      this.config.rules[name] = { severity, config };
```
rwjblue, line 56:
```suggestion
      const { name, config } = getRuleFromString(this.options.rule);

      this.config.rules[name] = config;
```

lib/get-config.js
```diff
@@ -283,6 +286,28 @@ function processIgnores(config) {
...
+ */
+function getRuleFromString(rule) {
+  const indexOfSeparator = rule.indexOf(':') + 1;
+
+  // we have to split based on the index of the first instance of the separator because the separator could exist in the second half of the rule
+  const name = rule.substring(0, indexOfSeparator - 1);
+  let ruleConfig = rule.substring(indexOfSeparator);
+
+  if (ruleConfig.startsWith('[')) {
+    ruleConfig = JSON.parse(ruleConfig);
```
rwjblue, line 303: Thoughts on wrapping this with a try/catch to give a better error message when it is malformed?

Not sure this is "good" per-se, but something like this would maybe seem nicer:

```suggestion
    try {
      ruleConfig = JSON.parse(ruleConfig);
    } catch () {
      throw new Error(`Error parsing specified \`--rule\` config ${rule} as JSON.`);
    }
```

gabrielcsapo: Sounds good to me, I will add a test for this.

PR #1265, opened 2020-04-11
test(cli): Set LC_ALL env var to en_US

dcyriller:
We test --help option in this file with jest snapshots. As this option
is provided by yargs, it has a localized output.

Let's set LC_ALL env var to have a consistent output between various
environments.

### DIFF
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -7,6 +7,7 @@ const Project = require('../helpers/fake-project');
 describe('ember-template-lint executable', function () {
   setupEnvVar('GITHUB_ACTIONS', null);
   setupEnvVar('FORCE_COLOR', '0');
+  setupEnvVar('LC_ALL', 'en_US');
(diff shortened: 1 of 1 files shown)


dcyriller:
CI failure seems to be internal error in GH actions

test/acceptance/cli-test.js
```diff
@@ -13,10 +13,12 @@ describe('ember-template-lint executable', function () {
   beforeEach(function () {
     project = Project.defaultSetup();
     project.chdir();
+    process.env.LC_ALL = 'en_US';
```
rwjblue, line 16: Can you move this up to just below the `describe` and do:

```js
setupEnvVar('LC_ALL', 'en_US');
```

This will ensure that it is properly reset after each test (to either be missing, or the original value).

dcyriller: Didn't see that, nice 👍

PR #1266, opened 2020-04-11
Bump GH actions/checkout to v2

dcyriller:
### Description

I suspect it can fix the CI in the two other PRs I opened a few minutes ago.

It would be great if these dependencies were bumped by dependabot 🤔.

### Links
In https://github.com/actions/checkout, see issue 188 and 23

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


rwjblue:
> It would be great if these dependencies were bumped by dependabot 🤔.

Ya it can be, but actions support is experimental. I just enabled it here though, we'll see what happens 😄

dcyriller:
Oh, I didn't know it was already a thing! I was saying that hoping it for the future 😄

PR #1269, opened 2020-04-13
Extract setup-env-var test helper

dcyriller:
The goal is to reuse it in two different files.


rwjblue:
Seems fine to me, curious why you closed...

dcyriller:
I initially opened it to split https://github.com/ember-template-lint/ember-template-lint/pull/1270 but figured out that it was unecessary and simply used this commit in the other PR 😄

PR #1277, opened 2020-04-14
Use get-stdin dependency to read from stdin

dcyriller:
`process.stdin.fd` is considered as a private NodeJs api. This PR proposes to retire it for `get-stdin`. This would implement a TODO that was in the code.


rwjblue:
Thank you!!!

PR #1282, opened 2020-04-17
Adapt tests for reading from stdin to run on Windows

### LINKED ISSUE #1281: ENOENT: no such file or directory, open 'D:\dev\stdin' on Windows
Try to run the `ember-template-lint` with the `--filename` option on Windows. You'll see the error similar to this one

image

This happens due to the fact that we call the `path.resolve(relativeFilePath)` which, if the `relativeFilePath` is STDIN (`/dev/stdin`) on Windows it'd be resolved as `D:\dev\stdin`.

The fix comes...

stukalin:
Fix #1281

### DIFF
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -2,6 +2,9 @@
 const { readFileSync } = require('fs');
 
 const execa = require('execa');
+const fs = require('fs');
+const path = require('path');
+
 const Project = require('../helpers/fake-project');
 const setupEnvVar = require('../helpers/setup-env-var');
 
@@ -145,15 +148,11 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given no path', function () {
-      // TOFIX on windows
-      if (process.platform === 'win32') {
-        return;
-      }
-
       it('should print errors', async function () {
         setProjectConfigForErrors();
-        let result = await run(['<', 'app/templates/application.hbs'], {
-          shell: true,
+        let result = await run([], {
+          shell: false,
+          input: fs.readFileSync(path.resolve('app/templates/application.hbs')),
         });
 
         expect(result.exitCode).toEqual(1);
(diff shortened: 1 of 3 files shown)


dcyriller:
Do you think you could setup a CI job to test against Windows platform? (Maybe in another PR.) I thought it was already the case, but apparently not. It should nicely highlight the bug.

stukalin:
@dcyriller good idea! I've added the #1284 to cover this!

rwjblue:
OK, now that #1284 has landed I think we can rebase this one and then remove that `TOFIX` comment.

@dcyriller - Anything else needed?

dcyriller:
Thank you @rwjblue!
I have rebased the branch, you can run ‘git pull —force’ to update your local branch @stukalin.

stukalin:
@dcyriller thanks! I am a little bit short of spare time ATM. Will try to get back to that in a day or two

dcyriller:
I've modified a test to reproduce the initial issue you reported :)

test/acceptance/cli-test.js
```diff
@@ -195,7 +195,7 @@ describe('ember-template-lint executable', function () {
         let result = await run(
           ['--filename', 'app/templates/application.hbs', '<', 'app/templates/application.hbs'],
           {
-            shell: false,
+            shell: true,
```
rwjblue, line 198: FWIW, I don't think this specific test _should_ work on Windows. The `<` style redirection is pretty shell dependent AFAIK.

stukalin: Correct! I finally understood how to properly make those tests

stukalin:
@dcyriller ok as far as I can judge your latest commits fix this problem. But I've changed some tests to be platform (shell) independent, so I guess my changes might be merged just in case.

dcyriller:
Nice work @stukalin. I can confirm the changes I introduced in https://github.com/ember-template-lint/ember-template-lint/pull/1297 actually fixed the two bugs we were tracking. I didn't mean to fix them and I don't have a Windows machine at hand to manually test my changes on that platform. But I see now that removing the check `filePath === "/dev/stdin"` fixed the bugs. I'm sorry for that.

It would have been better to reproduce the bugs in tests before fixing them. And you're doing it in this PR, so thank you! On my fork, I have reverted https://github.com/ember-template-lint/ember-template-lint/pull/1297 and applied this PR's changes. [Here](https://github.com/dcyriller/ember-template-lint/pull/5/checks?check_run_id=618228361) you'll see that the CI is red.

This changes look good to me. I have opened https://github.com/ember-template-lint/ember-template-lint/pull/1282 on top of this PR to add shell-specific tests.

test/acceptance/cli-test.js
dcyriller: > The < style redirection is pretty shell dependent AFAIK.

🤦, I missed that.

But I was somehow lucky because `<` style redirection is not supported by PowerShell on Windows. But - and I didn't know it - Windows command prompt seems to support it.

I've not been able to make execa work with PowerShell but execa works by default with cmd.exe on Windows (namely Windows command prompt).

So we could activate that option in the test suite. I gave it a try in https://github.com/ember-template-lint/ember-template-lint/pull/1298

PR #1284, opened 2020-04-18
Added CI support for Windows OS

stukalin:
Addresses #1281 

- Adds CI configuration for Windows OS
- replaces the `execa` `.sync` with the `.node` async version to eliminate differences between Windows and non-Windows platforms
- fixes some tests which do not make sense under Windows and adds the one which does

### DIFF
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -145,17 +145,15 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given no path', function () {
-      // todo doublecheck if we need this guard
-      // you don't have to specify path on Windows OS
-      // if (process.platform === 'win32') {
-      //   return;
-      // }
+      // TOFIX on windows
+      if (process.platform === 'win32') {
+        return;
+      }
 
       it('should print errors', async function () {
         setProjectConfigForErrors();
         let result = await run(['<', 'app/templates/application.hbs'], {
-          // todo rollback if we do need the commented guard
-          shell: process.platform !== 'win32',
+          shell: true,
         });
 
         expect(result.exitCode).toEqual(1);
@@ -235,23 +233,6 @@ describe('ember-template-lint executable', function () {
         setProjectConfigForErrors();
         let result = await run(['/dev/stdin', '<', 'app/templates/application.hbs'], {
           shell: true,
-        });
-
-        expect(result.exitCode).toEqual(1);
-        expect(result.stdout).toBeTruthy();
-        expect(result.stderr).toBeFalsy();
-      });
-    });
-
-    describe('given std input path on Windows OS', function () {
-      if (process.platform !== 'win32') {
-        return;
-      }
-
-      it('should print errors', async function () {
-        setProjectConfigForErrors();
-        let result = await run(['<', 'app/templates/application.hbs'], {
-          shell: false,
         });
 
         expect(result.exitCode).toEqual(1);
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -23,8 +23,8 @@ describe('ember-template-lint executable', function () {
 
   describe('basic usage', function () {
     describe('without any parameters', function () {
-      it('should emit help text', function () {
-        let result = run([]);
+      it('should emit help text', async function () {
+        let result = await run([]);
 
         expect(result.exitCode).toEqual(1);
         expect(result.stderr).toMatchInlineSnapshot(`
@@ -59,8 +59,8 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('with --help', function () {
-      it('should emit help text', function () {
-        let result = run(['--help']);
+      it('should emit help text', async function () {
+        let result = await run(['--help']);
 
         expect(result.exitCode).toEqual(0);
         expect(result.stdout).toMatchInlineSnapshot(`
@@ -95,9 +95,9 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given path to non-existing file', function () {
-      it('should exit without error and any console output', function () {
+      it('should exit without error and any console output', async function () {
         setProjectConfigForErrors();
-        let result = run(['app/templates/application-1.hbs']);
+        let result = await run(['app/templates/application-1.hbs']);
 
         expect(result.exitCode).toEqual(0, 'exits without error');
         expect(result.stdout).toBeFalsy();
@@ -106,9 +106,9 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given path to single file with errors', function () {
-      it('should print errors', function () {
+      it('should print errors', async function () {
         setProjectConfigForErrors();
-        let result = run(['app/templates/application.hbs']);
+        let result = await run(['app/templates/application.hbs']);
 
         expect(result.exitCode).toEqual(1);
         expect(result.stdout).toBeTruthy();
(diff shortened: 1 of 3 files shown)


test/acceptance/cli-test.js
```diff
@@ -853,6 +885,10 @@ describe('ember-template-lint executable', function () {
   function run(args, options = {}) {
     options.reject = false;
     options.cwd = options.cwd || project.path('.');
-    return execa.sync(require.resolve('../../bin/ember-template-lint.js'), args, options);
+
+    // for windows we need to call node <script.js>
+    // however the node method call hangs sometimes on linux, but works smoothly on windows
+    let runFunc = process.platform === 'win32' ? execa.node : execa.sync;
+    return runFunc(require.resolve('../../bin/ember-template-lint.js'), args, options);
```
dcyriller, line 892: Instead of using `execa.node`, can give a try to the following method: `execa('node', [require.resolve('../../bin/ember-template-lint.js'), ...args], options)`?

Indeed, it seems that the `shell` option cannot be used with `execa.node` (see https://github.com/sindresorhus/execa#execanodescriptpath-arguments-options). I wonder if it's not the reason why the method call hangs on linux.

test/acceptance/editors-integration-test.js
```diff
@@ -11,7 +11,11 @@ function setForBasicDebuggerError(project) {
 function run(project, args, options = {}) {
   options.reject = false;
   options.cwd = options.cwd || project.path('.');
-  return execa.sync(require.resolve('../../bin/ember-template-lint.js'), args, options);
+
+  // for windows we need to call node <script.js>
+  // however the node method call hangs sometimes on linux, but works smoothly on windows
+  let runFunc = process.platform === 'win32' ? execa.node : execa.sync;
+  return runFunc(require.resolve('../../bin/ember-template-lint.js'), args, options);
```
dcyriller, line 18: The same comment than https://github.com/ember-template-lint/ember-template-lint/pull/1284/files#r410979081 would apply here.

test/acceptance/editors-integration-test.js
```diff
@@ -31,16 +35,17 @@ describe('editors integration', function () {
...
-        project,
-        ['--json', '--filename', 'template.hbs', '/dev/stdin', '<', 'template.hbs'],
-        {
-          shell: true,
-        }
-      );
+      const args =
+        process.platform === 'win32'
+          ? ['--json', '--filename', 'template.hbs', '<', 'template.hbs']
+          : ['--json', '--filename', 'template.hbs', '/dev/stdin', '<', 'template.hbs'];
```
dcyriller, line 44: It is fine to remove `/dev/stdin` at all from these args. It will allow to remove this ternary.

`ember-template-lint --json --filename template.hbs < template.hbs` is the same command than `ember-template-lint --json --filename template.hbs /dev/stdin < template.hbs` or `ember-template-lint --json --filename template.hbs - < template.hbs`.

Using /dev/stdin in a command is already tested in `cli-test.js`.

test/acceptance/cli-test.js
```diff
@@ -131,9 +131,14 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given no path', function () {
-      it('should print errors', function () {
+      // you don't have to specify path on Windows OS
+      if (process.platform === 'win32') {
```
dcyriller, line 135: I wonder if it is not a bug 😅. Could you push a commit without this guard clause to highlight it in CI?

rwjblue:
Thanks for working on this @stukalin!

test/acceptance/cli-test.js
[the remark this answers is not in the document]
dcyriller: @rwjblue should it be?
```
let arguments = [require.resolve('../../bin/ember-template-lint.js'), ..args];
return execa(process.execPath, arguments, options);
```

test/acceptance/cli-test.js
stukalin: I've added the check. Should pop up during the CI build...

test/acceptance/cli-test.js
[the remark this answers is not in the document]
stukalin: Apparently it should, did it like this.

stukalin:
Arghhh! I knew I should've done the PR from a separate branch!

test/acceptance/cli-test.js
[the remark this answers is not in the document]
rwjblue: Eeck, sorry y'all. Yes.

rwjblue:
Ya, I _think_ you should be able to rebase out those changes that already landed on `master`. Lemme know if you'd like me to do it and force update the branch for you though.

test/acceptance/cli-test.js
stukalin: Yep, it's there. See the CI logs. Instead of offering me to provde the correct arguments it shows that it actually does the linting and gives output.

stukalin:
> Ya, I _think_ you should be able to rebase out those changes that already landed on `master`. Lemme know if you'd like me to do it and force update the branch for you though.

@rwjblue I thought I did this but *suddenly* it all turned out like this. Help would be very appreciated!

rwjblue:
OK, @stukalin I've pushed the rebase (dropping out unrelated changes) for you. I think you will need to do:

```
git pull --force
```

test/acceptance/cli-test.js
```diff
@@ -145,10 +145,17 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given no path', function () {
-      it('should print errors', function () {
+      // todo doublecheck if we need this guard
+      // you don't have to specify path on Windows OS
+      // if (process.platform === 'win32') {
+      //   return;
+      // }
```
rwjblue, line 152: Once we decide what we need here, we'll need to delete or restore this (commenting so we don't forget).

test/acceptance/cli-test.js
```diff
@@ -145,10 +145,17 @@ describe('ember-template-lint executable', function () {
...
+
+      it('should print errors', async function () {
         setProjectConfigForErrors();
-        let result = run(['<', 'app/templates/application.hbs'], {
-          shell: true,
+        let result = await run(['<', 'app/templates/application.hbs'], {
+          // todo rollback if we do need the commented guard
```
rwjblue, line 157: Reminder not to forget this once we've figured out what is needed

dcyriller:
So we have one failure in CI for Windows. `ember-template-lint < template.hbs` is supposed to print help to stderr, whereas on Windows it reports lint errors to stdout. This is a bug, thank you @stukalin for highlighting it 😄.

But I was expecting [this test](https://github.com/ember-template-lint/ember-template-lint/blob/v2.5.2/test/acceptance/cli-test.js#L165-L179) to fail as well. As it covers the initial issue you were reporting (#1282).

These two bugs look pretty close to me, both in term of behavior and in term of code base (basically they would be on two consecutive lines 😄). 

Maybe we can adress both of them in https://github.com/ember-template-lint/ember-template-lint/pull/1282? 

To finish this work done in this PR, I would suggest to:
- add back the guard clause with a comment: "TOFIX on Windows" (resolve [this comment](https://github.com/ember-template-lint/ember-template-lint/pull/1284#discussion_r411672657))
- apply [this suggestion](https://github.com/ember-template-lint/ember-template-lint/pull/1284/files#r411672827)
- remove [this test](https://github.com/ember-template-lint/ember-template-lint/pull/1284/files#diff-bf4e3275754d4a3796eb64dcc5aed295R246-R262) from this PR because it seems to test a bug

@rwjblue do you agree?

rwjblue:
Thank you @stukalin (and @dcyriller)!

PR #1297, opened 2020-04-24
Add unit test on bin/ember-template-lint

dcyriller:
While trying to understand the bug on Windows, I've done this small refactoring. I'm hesitant to submit it, so feel free to discard it if it doesn't worth it.

It adds unit test that we'll be able to make fail easily on Windows.

Best reviewed commit by commit

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -50,6 +50,18 @@ function expandFileGlobs(filePatterns, ignorePattern) {
   return result;
 }
 
+function getFilesToLint(filePatterns, ignorePattern = []) {
+  let files;
+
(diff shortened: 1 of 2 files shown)


bin/ember-template-lint.js
```diff
@@ -20,18 +20,28 @@ async function getSource(filePathToRead) {
   return fs.readFileSync(filePathToRead, { encoding: 'utf8' });
 }
 
-async function lintFile(linter, relativeFilePath, cliOptions) {
-  let resolvedFilePath = path.resolve(relativeFilePath);
-  let filePath = resolvedFilePath === STDIN ? cliOptions.filename || '' : relativeFilePath;
+async function buildLinterOptions(relativeFilePath, cliOptions) {
```
rwjblue: This could be exported for unit testing too, no need yet but seems like a nice extraction...

PR #1298, opened 2020-04-25
Add shell-specific tests for reading from stdin

dcyriller:
https://github.com/ember-template-lint/ember-template-lint/pull/1282 should be settled first

In unix shells, test the following commands (with the option `shell: true` for execa):
```
cat template.hbs | ember-template-lint --filename template.hbs
cat template.hbs | ember-template-lint --filename template.hbs -
ember-template-lint --filename template.hbs < template.hbs
```

In Windows command prompt, test (with the option `shell: true` for execa):
```
node ember-template-lint --filename template.hbs < template.hbs
```

I'd have like to add a test for the command in Windows PowerShell, but AFAICT execa doesn't support passing commands to PowerShell.

With this safety net, I think we'd have catched the issue on Windows (https://github.com/ember-template-lint/ember-template-lint/issues/1281) more easily.

### DIFF
--- a/test/acceptance/stdin-by-platform-test.js
+++ b/test/acceptance/stdin-by-platform-test.js
@@ -1,0 +1,139 @@
+'use strict';
+
+const execa = require('execa');
+const Project = require('../helpers/fake-project');
+const setupEnvVar = require('../helpers/setup-env-var');
+
+describe('ember-template-lint executable', function () {
+  setupEnvVar('GITHUB_ACTIONS', null);
+  setupEnvVar('FORCE_COLOR', '0');
+  setupEnvVar('LC_ALL', 'en_US');
+
+  // Fake project
+  let project;
+  beforeEach(function () {
+    project = Project.defaultSetup();
+    project.setConfig({
+      rules: {
+        'no-bare-strings': true,
+      },
+    });
+    project.write({
+      'template.hbs': '<h2>Here too!!</h2> <div>Bare strings are bad...</div>',
+      components: {
+        'foo.hbs': '{{fooData}}',
+      },
+    });
+    project.chdir();
+  });
+
+  afterEach(async function () {
+    await project.dispose();
+  });
+
+  if (process.platform === 'win32') {
+    describe('in Windows Command Prompt', function () {
+      describe('command: `node ember-template-lint --filename template.hbs < template.hbs`', function () {
+        it('reports errors to stdout', async function () {
+          let result = await execa(
+            process.execPath,
+            [
+              require.resolve('../../bin/ember-template-lint.js'),
+              '--filename',
+              'template.hbs',
+              '<',
+              'template.hbs',
+            ],
+            { shell: true, reject: false, cwd: project.path('.') }
+          );
+
+          expect(result.stdout).toMatchInlineSnapshot(`
+            "template.hbs
+              1:4  error  Non-translated string used  no-bare-strings
+              1:25  error  Non-translated string used  no-bare-strings
+
+            ✖ 2 problems (2 errors, 0 warnings)"
+          `);
+          expect(result.stderr).toBeFalsy();
+        });
+      });
+    });
+  }
+
+  if (process.platform === 'linux' || process.platform === 'darwin') {
+    describe('in unix bash', function () {
+      describe('command: `cat template.hbs | ember-template-lint --filename template.hbs`', function () {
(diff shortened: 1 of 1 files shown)


stukalin:
@dcyriller 

> In Windows command prompt, test (with the option `shell: true` for execa):
> 
> ```
> node ember-template-lint --filename template.hbs < template.hbs
> ```

Actually it works in Windows PowerShell. The problem was that when I run it in `execa` I've got this `ENOENT` error.
image

Although.. now when I looked at it freshly I've noticed that it says that `'C:\Program' is not recognized as an internal or external command,
operable program or batch file.`; maybe that's the answer and I'd simply need to add a couple of quotes somewhere... I'll have another look in a day or so.

rwjblue:
Just merged #1282, ready for a rebase.

test/helpers/set-config.js
```diff
@@ -0,0 +1,63 @@
+function setProjectConfigForErrors(project) {
```
rwjblue, line 1: FWIW, I'd rather not even have these helper methods, and just use `project` methods to setup each test as appropriate. I find the "DRY" ness here actually makes each individual test harder to read and understand.

I know you didn't introduce these helpers, sorry for making extra work. 😩

dcyriller: Would you like to have to have them in `fake-project` module, `project.setProjectConfigForErrors` for instance? Or do you prefer to inline them in every test?

rwjblue:
> Or do you prefer to inline them in every test?

Yeah either inline in the test or inline in the `beforeEach` (if there are a bunch of tests with similar setup steps).

dcyriller:
It makes sense and I agree. I extracted them at first as it seemed it was the chosen path. But following your review, I have applied some macros to inline them in https://github.com/ember-template-lint/ember-template-lint/pull/1302. I split the PRs in two to make them easier to review.

rwjblue:
> Is CI failing due to recent changes in recast? 😄

Ya, fixed (by you :wink: ) in #1303.

test/acceptance/stdin-by-platform-test.js
```diff
@@ -0,0 +1,139 @@
...
+
+            ✖ 2 problems (2 errors, 0 warnings)"
+          `);
+          expect(result.stderr).toBeFalsy();
+        });
+      });
+    });
+  }
+
+  if (process.platform === 'linux' || process.platform === 'darwin') {
```
rwjblue, line 63: Are there other platforms that we are trying to exclude here? If not, maybe this can be an `else` block to the above `if (process.platform === 'win32') {`?

test/acceptance/stdin-by-platform-test.js
```diff
@@ -0,0 +1,139 @@
...
+          expect(result.stdout).toMatchInlineSnapshot(`
+            "template.hbs
+              1:4  error  Non-translated string used  no-bare-strings
+              1:25  error  Non-translated string used  no-bare-strings
+
+            ✖ 2 problems (2 errors, 0 warnings)"
+          `);
+          expect(result.stderr).toBeFalsy();
+        });
+      });
```
rwjblue, line 136: This test seems like _exactly_ the same as the one for `win32`, can we restructure this module so that the tests that run on all platforms are shared then we have the ones that only run on non-win32?

Something like:

```js
describe('ember-template-lint executable', function() {
  // this test

  if (process.platform !== 'win32') {
    describe('posix environments', function() {
      // the non-windows tests
    });
  }
});
```

test/acceptance/stdin-by-platform-test.js
```diff
@@ -0,0 +1,139 @@
...
```
rwjblue, line 89: We can test this one without `shell: true` right? By providing a stdin to the `execa` invocation? And if we do that, that will pass on both windows and posix IIRC. 

Would you mind adding an "all platforms" test for that (in this file)?

test/acceptance/stdin-by-platform-test.js
dcyriller: I've added a commit to consolidate test cases as per your suggestions.

test/acceptance/stdin-by-platform-test.js
dcyriller: It should be done

rwjblue:
Awesome, thank you @dcyriller! I think CI will be good if you rebase?

dcyriller:
oh sure !

PR #1302, opened 2020-04-28
test(cli): Inline test helpers and other minor cleanup

dcyriller:
It should be easier to review commit by commit


rwjblue:
CI was fixed by #1303

PR #1306, opened 2020-04-28
Bug: Fixed invalid URL syntax in documentation reference for no-invalid-link-title rule

MelSumner:
Updating syntax of reference link so it will be a valid URL.


rwjblue:
CI was fixed by #1303

PR #1316, opened 2020-04-29
Add new rule: `no-positional-data-test-selectors`

gabrielcsapo:
# Motivation

[ember-test-selectors](https://github.com/simplabs/ember-test-selectors) is a very popular library that enables better element selectors for testing.

One of the features that had been added to ember-test-selectors over the years was to allow passing a positional argument to curly component invocations as a shorthand (to avoid having to also add a named argument value).

That would look like:

```hbs
{{some-thing data-test-foo}}
```

Internally, that was converted to an `attributeBinding` for `@ember/component`s. Unfortunately, that particular invocation syntax is in conflict with modern Ember Octane templates. For example, in the snippet above `data-test-foo` is actually referring to `this.data-test-foo` (and would be marked as an error by the `no-implicit-this` rule).

Additionally, the nature of these "fake" local properties significantly confuses the codemods that are used to transition an application into Ember Octane (e.g. [ember-no-implicit-this-codemod](https://github.com/ember-codemods/ember-no-implicit-this-codemod) and [ember-angle-brackets-codemod](https://github.com/ember-codemods/ember-angle-brackets-codemod)).

# Design

This rule forbids the following:

```hbs
{{foo-bar data-test-blah}}
{{#foo-bar data-test-blah}}
{{/foo-bar}}
```

And suggests using the following instead:

```hbs
{{foo-bar data-test-blah=true}}
{{#foo-bar data-test-blah=true}}
{{/foo-bar}}
```

Additionally, this rule implements `--fix` support to automatically convert the "bad" examples into "good" ones.

### DIFF
--- a/docs/rule/no-positional-data-test-selectors.md
+++ b/docs/rule/no-positional-data-test-selectors.md
@@ -34,8 +34,8 @@ And suggests using the following instead:
 {{/foo-bar}}
 ```
 
-Additionally, this rule implements `--fix` support to automatically convert the "bad" examples into "good" ones.
+:wrench: The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
 
 ## References
 
-[Not Supported](https://github.com/simplabs/ember-test-selectors/commit/d47f73d76b3ccbc9f0be5df3b897afd08b1636a6)
+- [ember-test-selectors#d47f73d](https://github.com/simplabs/ember-test-selectors/commit/d47f73d76b3ccbc9f0be5df3b897afd08b1636a6)
--- a/docs/rule/no-positional-data-test-selectors.md
+++ b/docs/rule/no-positional-data-test-selectors.md
@@ -1,0 +1,41 @@
+# no-positional-data-test-selectors
+
+## Motivation
+
+[ember-test-selectors](https://github.com/simplabs/ember-test-selectors) is a very popular library that enables better element selectors for testing.
+
+One of the features that had been added to ember-test-selectors over the years was to allow passing a positional argument to curly component invocations as a shorthand (to avoid having to also add a named argument value).
+
+That would look like:
+
+```hbs
+{{some-thing data-test-foo}}
+```
+
+Internally, that was converted to an `attributeBinding` for `@ember/component`s. Unfortunately, that particular invocation syntax is in conflict with modern Ember Octane templates. For example, in the snippet above `data-test-foo` is actually referring to `this.data-test-foo` (and would be marked as an error by the `no-implicit-this` rule).
+
+Additionally, the nature of these "fake" local properties significantly confuses the codemods that are used to transition an application into Ember Octane (e.g. [ember-no-implicit-this-codemod](https://github.com/ember-codemods/ember-no-implicit-this-codemod) and [ember-angle-brackets-codemod](https://github.com/ember-codemods/ember-angle-brackets-codemod)).
+
+## Examples
+
+This rule forbids the following:
+
+```hbs
+{{foo-bar data-test-blah}}
+{{#foo-bar data-test-blah}}
+{{/foo-bar}}
+```
+
+And suggests using the following instead:
+
+```hbs
+{{foo-bar data-test-blah=true}}
+{{#foo-bar data-test-blah=true}}
+{{/foo-bar}}
+```
+
+Additionally, this rule implements `--fix` support to automatically convert the "bad" examples into "good" ones.
+
+## References
+
+[Not Supported](https://github.com/simplabs/ember-test-selectors/commit/d47f73d76b3ccbc9f0be5df3b897afd08b1636a6)
--- a/lib/rules/no-positional-data-test-selectors.js
+++ b/lib/rules/no-positional-data-test-selectors.js
@@ -3,7 +3,7 @@
 const Rule = require('./base');
 const { builders: b } = require('ember-template-recast');
 
-const BUILT_INS = [
+const BUILT_INS = new Set([
   'action',
   'array',
   'component',
@@ -35,7 +35,7 @@ const BUILT_INS = [
   'with',
   '-in-element',
   'in-element',
-];
+]);
 
 module.exports = class NoPositionalDataTestSelectors extends Rule {
   visitor() {
@@ -50,7 +50,7 @@ module.exports = class NoPositionalDataTestSelectors extends Rule {
   }
 
   process(node) {
-    if (BUILT_INS.includes(node.path.original)) {
+    if (BUILT_INS.has(node.path.original)) {
       return;
     }
 
--- a/lib/rules/no-positional-data-test-selectors.js
+++ b/lib/rules/no-positional-data-test-selectors.js
@@ -1,0 +1,84 @@
+'use strict';
+
+const Rule = require('./base');
+const { builders: b } = require('ember-template-recast');
+
+const BUILT_INS = [
+  'action',
+  'array',
+  'component',
+  'concat',
+  'debugger',
+  'each',
+  'each-in',
+  'fn',
+  'get',
+  'hasBlock',
+  'has-block',
+  'has-block-params',
+  'hash',
+  'if',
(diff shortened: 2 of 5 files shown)


gabrielcsapo:
@rwjblue would this be something that would be good to contribute here?

docs/rule/data-attributes-required-default.md
```diff
@@ -0,0 +1,37 @@
+# data-attributes-required-default
+
+:white_check_mark: The `extends: 'recommended'` property in a configuration file enables this rule.
+
+Require the presence of a default value for data attributes.
```
rwjblue, line 5: What does "default" mean here? Do you just mean that they need a value?

gabrielcsapo: correct, so instead of positional params they would be named arguments.

rwjblue:
I don't understand at all :D

In all of the examples you gave they are not arguments, they are attributes. :thinking:

gabrielcsapo:
``` 
<Button data-disabled>Hello</Button>
```

in ABC or valid html parsing this is considered an attribute

Screen Shot 2020-04-30 at 2 53 31 PM

Setting abcs to have a default value also parse correctly by still being an attribute with a value of true.

``` 
<Button data-disabled=true>Hello</Button>
```

Screen Shot 2020-04-30 at 2 53 40 PM

In curly components 

```hbs
{{button
    data-disabled
}}
Hello
{{/button}}
```

This is a positional argument and will be parsed as a param and not a hash attribute.

Screen Shot 2020-04-30 at 2 52 06 PM

When we set a value on this like so:

```hbs
{{button
    data-disabled=true
}}
Hello
{{/button}}
```

Screen Shot 2020-04-30 at 2 52 24 PM

It is parsed as a hash value correctly.

rwjblue:
Sure, all of these things seem true, but what does that have to do with something being wrong with:

```hbs
<Foo data-some-thing />
<div data-whatever />
```

I'm just trying to figure out why you think these examples are bad...

gabrielcsapo:
@rwjblue the spec for data attributes doesn't imply that just because it exists on the html node wouldn't mean that is has a default boolean. This would have the same value as enforcing [_consistent return_](https://eslint.org/docs/rules/consistent-return). If you access the dataset on a node and look for a truthy value for an attribute it will return false. If you were to write `document.querySelector('#bar[data-some-thing'])` that will return true because it is looking for the existence of the property not checking the property has a value.

https://jsfiddle.net/ntz4Lmr9/1/

Screen Shot 2020-04-30 at 6 13 42 PM

docs/rule/data-attributes-required-default.md
rwjblue: > correct, so instead of positional params they would be named arguments.

They are **never** positional params (or named arguments) in these examples. They are attributes.

rwjblue:
As far as I can tell (including from that JSFiddle) using `<div data-foo></div>` is perfectly valid, and `document.querySelector(...).dataset` properly includes the specified `foo` attribute. If code exists that is assuming a non-empty value in `dataset` (just to confirm presence) **that** code should be fixed 😸

rwjblue:
OK, discussed with @gabrielcsapo separately. We are going to re-work this to be explicitly focused on ember-test-selectors's accepting `data-test-*` as positional params for curly invocations (cc @Turbo87).

I'm going to mark all the other comments as resolved here, just to avoid confusion.

rwjblue:
@Turbo87 - Does the rule name seem good to you? I waffled a bit between `no-positional-data-test-selectors` and `no-curly-positional-data-test-selector-argument`... 🤔

docs/rule/no-positional-data-test-selectors.md
```diff
@@ -0,0 +1,27 @@
+# no-positional-data-test-selectors
+
+:white_check_mark: The `extends: 'recommended'` property in a configuration file enables this rule.
```
rwjblue, line 3: I think we can't add to recommended (adding to recommended is a breaking change, we'll have to queue this up for a major bump), can you remove?

```suggestion
```

docs/rule/no-positional-data-test-selectors.md
```diff
@@ -0,0 +1,27 @@
...
+}}
+```
+
+// good
+
+```hbs
+{{foo
+  data-foo={{foo}}
+}}
+```
```
rwjblue, line 23: Can you bring in the content from the PR description here?

lib/config/recommended.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = {
   rules: {
+    'no-positional-data-test-selectors': 'error',
```
rwjblue, line 5:
```suggestion
```

lib/rules/no-positional-data-test-selectors.js
```diff
@@ -0,0 +1,37 @@
...
+              message: `data-test selectors should have a default value. "${pair.original}" should be "${pair.original}=true"`,
+              line: node.loc && node.loc.start.line,
+              column: node.loc && node.loc.start.column,
+              source: this.sourceForNode(node),
+              isFixable: true,
+            });
+          }
+        }
+      });
+    }
```
rwjblue, line 35: A few things;

* We shouldn't need to guard for the presence of `node.params`, every `MustacheStatement` and `BlockStatement` will have a params array
* We should be pushing a `HashPair` (not making the positional param value look like one)

```suggestion
    let testSelectorParamIndex = node.params.findIndex(
      (n) => n.type === "PathExpression" && n.original.startsWith("data-test-")
    );

    if (testSelectorParamIndex === -1) {
      return;
    }
    
    if (this.mode === 'fix') {
      let selectorName = node.params[testSelectorParamIndex].original;

      // remove the item from `params`
      node.params.splice(testSelectorParamIndex, 1);

      // add it as a HashPair
      node.hash.pairs.unshift(b.pair(selectorName, b.boolean(true)));
    } else {
      this.log({
        message: `Passing a `data-test-*` positional param to a curly invocation should be avoided.`,
        line: node.loc && node.loc.start.line,
        column: node.loc && node.loc.start.column,
        source: this.sourceForNode(node),
        isFixable: true,
      });
    }
```

test/unit/rules/no-positional-data-test-selectors-test.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-positional-data-test-selectors',
+  config: true,
+  good: [
```
rwjblue, line 8: Can you add "good" tests for:

```hbs
{{#if data-test-foo}}
{{/if}}
```

```hbs
<div data-test-blah></div>
```

```hbs
<Foo data-test-derp />
```

```hbs
{{something data-test-lol=true}}
```

lib/rules/no-positional-data-test-selectors.js
```diff
@@ -0,0 +1,84 @@
...
+      BlockStatement(node) {
+        this.process(node);
+      },
+    };
+  }
+
+  process(node) {
+    if (BUILT_INS.includes(node.path.original)) {
+      return;
+    }
```
rwjblue, line 55: I'm curious, why do we have to avoid built-ins? Is this to support that `if` case I mentioned adding?

gabrielcsapo: yes @rwjblue

Turbo87:
While I appreciate the work, I'm wondering if this addon-specific rule should really live in the `ember-template-lint` project itself. Should this maybe be extracted to an `ember-template-lint-test-selectors` plugin? or maybe convert this into a codemod directly and not have the lint rule generally 🤔

rwjblue:
Seems good if you wanted to extract that and setup a blueprint for ember-test-selectors to add it to the projects `extends`. IMHO, it is 1) popular enough to be something to land here and 2) implementation agnostic enough to not matter.

PR #1317, opened 2020-04-30
Avoid glob matching when no-glob is present (e.g. globs have already been replaced by shell expansion)

gabrielcsapo:
Our company switched to the latest version of MacOS and ZSH is trying to expand the glob before passing it to the process.

# Example

## With quoted glob

```
$ time ember-template-lint "something/something/**"
$ 2.71s user 2.69s system 83% cpu 6.492 total
```

## Without quoted glob

```
$ time ember-template-lint something/something/**
$ 12.97s user 17.79s system 42% cpu 1:12.74 total
```

## How is discovered it?

I tried to glob a large directory and saw that zsh was trying to pass many arguments, so much so that it was unable to run the command.

```
> ember-template-lint **/*.hbs
> zsh: argument list too long
```

Another issue that was exposed through this was globbing paths that are already resolved, instead of trying to resolve them again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ var results = linter.verify({ source: template, moduleId: 'template.hbs' });
 
 Basic `ember-template-lint` executable is provided, allowing for easy use within i.e. Git pre-commit/push hooks and development of appropriate plugins for text editors.
 
+> If using _ZSH_ wrap all glob patterns in quotes so that it won't be interpreted by the CLI. `./node_modules/.bin/ember-template-lint app/templates/**` (this will expand all paths in app/templates) and `./node_modules/.bin/ember-template-lint "app/templates/**"` (this will pass the glob to ember-template-lint and not interpret the glob).
+
 Example usage:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ var results = linter.verify({ source: template, moduleId: 'template.hbs' });
 
 Basic `ember-template-lint` executable is provided, allowing for easy use within i.e. Git pre-commit/push hooks and development of appropriate plugins for text editors.
 
-> If using _ZSH_ wrap all glob patterns in quotes so that it won't be interpreted by the CLI. `./node_modules/.bin/ember-template-lint app/templates/**` (this will expand all paths in app/templates) and `./node_modules/.bin/ember-template-lint "app/templates/**"` (this will pass the glob to ember-template-lint and not interpret the glob).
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -56,38 +56,40 @@ var results = linter.verify({ source: template, moduleId: 'template.hbs' });
 
 Basic `ember-template-lint` executable is provided, allowing for easy use within i.e. Git pre-commit/push hooks and development of appropriate plugins for text editors.
 
+> If using _ZSH_ wrap all glob patterns in quotes so that it won't be interpreted by the CLI. `./node_modules/.bin/ember-template-lint app/templates/**` (this will expand all paths in app/templates) and `./node_modules/.bin/ember-template-lint "app/templates/**"` (this will pass the glob to ember-template-lint and not interpret the glob).
```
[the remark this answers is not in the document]
gabrielcsapo: Updated the wording for this

bin/ember-template-lint.js
```diff
@@ -45,6 +46,18 @@ function expandFileGlobs(filePatterns, ignorePattern) {
   let result = new Set();
 
   filePatterns.forEach((pattern) => {
+    // if we are passed items that are already globbed before we should not crawl them as directories
+    if (pattern.slice(-4) === '.hbs' && pattern.indexOf('*') === -1) {
```
rwjblue, line 50: I think there is a utility for this :thinking: 

Something something "has magic"....

gabrielcsapo: I used the is-valid-glob module from micromatch.

bin/ember-template-lint.js
```diff
@@ -45,6 +47,18 @@ function expandFileGlobs(filePatterns, ignorePattern) {
...
+    if (pattern.slice(-4) === '.hbs' && !isValidGlob(pattern)) {
+      if (
+        (ignorePattern.length === 0 || !micromatch.isMatch(pattern, ignorePattern)) &&
+        fs.existsSync(pattern)
+      ) {
+        result.add(pattern);
+      }
+
+      return;
+    }
```
rwjblue, line 60: This logic is slightly hard for me to follow, what do you think about:

```suggestion
    let isHBS = pattern.slice(-4) === '.hbs';
    let isLiteralPath = !isValidGlob(pattern) && fs.existsSync(pattern);
    let isIgnored = !micromatch.isMatch(pattern, ignorePattern);
    
    if (isHBS && isLiteralPath) {
      if (!isIgnored) {
        result.add(pattern);
      }
      
      return;
    }
```

rwjblue:
@gabrielcsapo - I just figured out that the changes here don't actually work quite correctly. I'm fixing over in https://github.com/ember-template-lint/ember-template-lint/pull/1554 if you want to read through the explanation and/or review generally...

gabrielcsapo:
Thanks @rwjblue your explanation makes sense! Thanks for the fix forward.

PR #1320, opened 2020-05-02
Add `catchNonexistentRoles` option (default false) to `no-invalid-role` rule

### LINKED ISSUE #1311: Update `no invalid role` rule to check against valid role values
Update the `no-invalid-role` rule to check against the permitted role values listed at https://www.w3.org/WAI/PF/aria/roles and ensure that the rule catches `role` attribute values that are not valid. 

Disallowed:
```html
<div role="accordion"></div>
```
```html
<div role="command interface"></div>
```

Allowed: 
```html
<footer role="contentinfo"></footer>
```
```html
<div role="log"></div>
```

rajasegar:
Fixes #1311

### DIFF
--- a/docs/rule/no-invalid-role.md
+++ b/docs/rule/no-invalid-role.md
@@ -46,6 +46,12 @@ This rule **allows** the following:
 <span role="none"></span>
 ```
 
+## Configuration
+
+* boolean - `true` to enable / `false` to disable
+
+* object - { 'catchNonexistentRoles' : `true|false` } -- Whether invalid role values should be allowed (defaults to `false`)
+
 ## Migration
 
 * If violations are found, remediation should be planned to replace the semantic HTML with the `div` element. Additional CSS will likely be required.
--- a/docs/rule/no-invalid-role.md
+++ b/docs/rule/no-invalid-role.md
@@ -51,6 +51,7 @@ This rule **allows** the following:
 * boolean - `true` to enable / `false` to disable
 
 * object - { 'catchNonexistentRoles' : `true|false` } -- Whether invalid role values should be allowed (defaults to `false`)
+(TODO: change default to true in next major release)
 
 ## Migration
 
--- a/docs/rule/no-invalid-role.md
+++ b/docs/rule/no-invalid-role.md
@@ -8,6 +8,8 @@ Current list of checks:
 
 1. Use of the presentation role for content which should convey semantic information may prevent the user from understanding that content. This rule checks semantic HTML elements for the presence of `role="none"` or `role="presentation"` and compares it to the list of disallowed elements. It should not effect custom elements.
 
+2. Use of invalid role for elements which does not fall under any of the values in this [list](https://www.w3.org/WAI/PF/aria/roles).
+
 ## Examples
 
 This rule **forbids** the following:
@@ -20,6 +22,18 @@ This rule **forbids** the following:
 ```hbs
 <ul role="none">
 </ul>
+```
+
+```hbs
+<div role="accordion"></div>
+```
+
+```hbs
+<div role="command interface"></div>
+```
+
+```hbs
+<div role="COMMAND INTERFACE"></div>
 ```
 
 This rule **allows** the following:
@@ -41,3 +55,4 @@ This rule **allows** the following:
 * [HTML elements reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)
 * [WAI-ARIA presentation(role)](https://www.w3.org/TR/wai-aria/#presentation)
 * [Failure of Success Criterion 1.3.1 due to the use of role presentation on content which conveys semantic information](https://www.w3.org/WAI/WCAG21/Techniques/failures/F92)
+* [The Role Model](https://www.w3.org/WAI/PF/aria/roles)
--- a/lib/rules/no-invalid-role.js
+++ b/lib/rules/no-invalid-role.js
@@ -7,6 +7,10 @@ function createErrorMessageDisallowedRoleForElement(element) {
   return `Use of presentation role on <${element}> detected. Semantic elements should not be used for presentation.`;
 }
 
+function createInvalidRoleErrorMessage(element) {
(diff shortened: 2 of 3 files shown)


lib/rules/no-invalid-role.js
```diff
@@ -112,8 +183,13 @@ module.exports = class NoInvalidRole extends Rule {
 
         const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');
 
-        if (!['presentation', 'none'].includes(roleValue)) {
-          return;
+        if (!VALID_ROLES.includes(roleValue)) {
```
bmish, line 186: I assume roles are not case-sensitive right? So we should check `roleValue.toLowerCase()` right? And make sure to add a test for that.

lib/rules/no-invalid-role.js
```diff
@@ -100,6 +104,73 @@ const DISALLOWED_ELEMENTS = [
   'wbr',
 ];
 
+// List of roles taken from https://www.w3.org/WAI/PF/aria/roles#widget_roles
+// The following is a collated list of Widget, Document Structure and Landmark Roles
+const VALID_ROLES = [
```
bmish, line 109: Just wondering, I noticed the `table` role, as one example, is not in this list. Any idea why? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Row_Role

rajasegar:
Thanks for the review, will make the changes accordingly

lib/rules/no-invalid-role.js
rajasegar: Yes, I tried looking for table role in this [list](https://www.w3.org/WAI/PF/aria/roles) I am not able to find it , don't know why
May be @MelSumner can throw some light here.

rajasegar:
@bmish Addressed your review comments, please check and let me know

bmish:
You might need to rebase on the latest master still. I'm seeing some extra commits show up in the diff.

rajasegar:
@bmish But the files changed still remains at 3 (+1 for the docs). I have encountered some problems while rebasing. that's why I had to merge the local branch again. My bad I always tend to make mistakes while rebasing, can you please help me out how to properly rebase.

lib/rules/no-invalid-role.js
MelSumner: @bmish `table` is not a valid role. See the spec: https://www.w3.org/WAI/PF/aria/roles

rajasegar:
@bmish PR has been rebased on the latest master. Please check and let me know.

lib/rules/no-invalid-role.js
```diff
@@ -116,8 +216,23 @@ module.exports = class NoInvalidRole extends Rule {
           ['presentation', 'none'].includes(roleValue) &&
           ELEMENTS_DISALLOWING_PRESENTATION_OR_NONE_ROLE.includes(node.tag)
         ) {
+          if (ELEMENTS_DISALLOWING_PRESENTATION_OR_NONE_ROLE.includes(node.tag)) {
```
bmish, line 219: This looks like a bad merge conflict resolution. This is part of the existing rule logic that doesn't need to be changed.

lib/rules/no-invalid-role.js
```diff
@@ -116,8 +216,23 @@ module.exports = class NoInvalidRole extends Rule {
...
+            this.log({
+              message: createErrorMessageDisallowedRoleForElement(node.tag),
+              line: node.loc && node.loc.start.line,
+              column: node.loc && node.loc.start.column,
+              source: this.sourceForNode(node),
+            });
+          }
+        }
+
+        if (!this.config.catchInvalidRoles) {
```
bmish, line 229: We want this rule to be prepared for future expansions, so instead of this early return if statement here, we should combine this with the actual relevant if statement below:

```
if (this.config.catchInvalidRoles && !VALID_ROLES.includes(roleValue.toLowerCase())) {
```

test/unit/rules/no-invalid-role-test.js
```diff
@@ -74,5 +74,30 @@ generateRuleTests({
         column: 0,
       },
     },
+
+    {
```
bmish, line 78: Can you also add one valid case showing that an invalid role is allowed when `catchInvalidRoles` is off?

lib/rules/no-invalid-role.js
```diff
@@ -100,7 +106,101 @@ const ELEMENTS_DISALLOWING_PRESENTATION_OR_NONE_ROLE = [
...
+      case 'undefined':
+        return false;
+    }
+
+    let errorMessage = createErrorMessage(
+      this.ruleName,
+      [
+        '  * boolean - `true` to enable / `false` to disable',
+        '  * object -- An object with the following keys:',
+        '    * `catchInvalidRoles` -- Whether invalid role values should be allowed (defaults to `true`)',
```
bmish, line 196: default is `false` actually

rajasegar:
@bmish Addressed the review comments, please check and let me know

rajasegar:
Sure @bmish will make the changes accordingly and update you

rajasegar:
@bmish addressed the latest review comments. Please check.

docs/rule/no-invalid-role.md
```diff
@@ -32,6 +46,12 @@ This rule **allows** the following:
 <span role="none"></span>
 ```
 
+## Configuration
+
+
+* object - { 'catchNonexistentRoles' : `true|false` } -- Whether invalid role values should be allowed (defaults to `false`)
```
bmish, line 53: Can you add a notice at the end of this sentence: `(TODO: change default to true in next major release)` to help us remember?

lib/rules/no-invalid-role.js
```diff
@@ -3,10 +3,16 @@
 const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
 
+const createErrorMessage = require('../helpers/create-error-message');
+
 function createErrorMessageDisallowedRoleForElement(element) {
   return `Use of presentation role on <${element}> detected. Semantic elements should not be used for presentation.`;
 }
 
+function createInvalidRoleErrorMessage(element) {
```
bmish, line 12: Can you rename this to: `createNonexistentRoleErrorMessage`

PR #1322, opened 2020-05-02
Add new rule: `no-nested-landmark`

### LINKED ISSUE #1274: Rule request: disallow nested landmark elements of the same name
A new rule should be added to disallow landmark elements to be nested within landmark elements of the same name.

## List of elements & their corresponding roles
- header (banner)
- main (main)
- aside (complementary)
- form (form, search)
- main (main)
- nav (navigation)
- footer (contentinfo)

### Allowed

```html
<main></main>
```

### Disallowed

```html
<main>
  <main></main>
</main>
```

```html
<main>
  <div role="main"></div>
</main>
```

rajasegar:
Fixes #1274

### DIFF
--- a/docs/rule/no-nested-landmark.md
+++ b/docs/rule/no-nested-landmark.md
@@ -1,53 +1,45 @@
 # no-nested-landmark
 
-TODO: context about the problem goes here
+Landmark elements should not be nested within landmark elements of the same name.
 
-TODO: what the rule does goes here
+## List of elements & their corresponding roles
+
+- header (banner)
+- main (main)
+- aside (complementary)
+- form (form, search)
+- main (main)
+- nav (navigation)
+- footer (contentinfo)
 
 ## Examples
 
 This rule **forbids** the following:
 
 ```hbs
-{{!-- TODO: Example 1  --}}
+<main>
+  <main></main>
+</main>
 ```
 
 ```hbs
-{{!-- TODO: Example 2  --}}
+<div role="main">
+  <main></main>
+</div>
+```
+
+```hbs
+<main>
+  <div role="main"></div>
+</main>
 ```
 
 This rule **allows** the following:
 
 ```hbs
-{{!-- TODO: Example 1  --}}
+<div><main></main></div>
 ```
 
 ```hbs
-{{!-- TODO: Example 2  --}}
+<main></main>
 ```
-
-## Migration
-
-TODO: suggest any fast/automated techniques for fixing violations in a large codebase
-
-* TODO: suggestion on how to fix violations using find-and-replace / regexp
-* TODO: suggestion on how to fix violations using a codemod
-
-## Configuration
-
-TODO: exclude this section if the rule has no extra configuration
-
-* object -- containing the following properties:
-  * string -- `parameterName1` -- TODO: description of parameter including the possible values and default value
-  * boolean -- `parameterName2` -- TODO: description of parameter including the possible values and default value
-
-## Related Rules
-
-* [TODO-related-rule-name1](related-rule-name1.md)
-* [TODO-related-rule-name2](related-rule-name2.md)
-
-## References
-
-* TODO: link to relevant documentation goes here
-* TODO: link to relevant function spec goes here
-* TODO: link to relevant guide goes here
--- a/docs/rule/no-nested-landmark.md
+++ b/docs/rule/no-nested-landmark.md
@@ -23,14 +23,38 @@ This rule **forbids** the following:
 ```
 
 ```hbs
+<main>
+  <div>
+    <main></main>
+  </div>
+</main>
+```
+
+```hbs
 <div role="main">
   <main></main>
 </div>
 ```
 
 ```hbs
+<div role="main">
+  <div>
+    <main></main>
+  </div>
+</div>
+```
+
+```hbs
 <main>
   <div role="main"></div>
+</main>
+```
+
+```hbs
+<main>
+  <div>
+    <div role="main"></div>
+  </div>
 </main>
 ```
 
--- a/docs/rule/no-nested-landmark.md
+++ b/docs/rule/no-nested-landmark.md
@@ -75,3 +75,8 @@ This rule **allows** the following:
   </div>
 </div>
 ```
+
+## References
+
+- [HTML Sectioning Elements](https://www.w3.org/TR/wai-aria-practices-1.1/#html-sectioning-elements)
+- [Landmark Roles](https://www.w3.org/TR/wai-aria-1.1/#landmark_roles)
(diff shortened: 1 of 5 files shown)


docs/rule/no-nested-landmark.md
```diff
@@ -0,0 +1,69 @@
...
+
+This rule **allows** the following:
+
+```hbs
+<div><main></main></div>
+```
+
+```hbs
+<main></main>
+```
```
MelSumner, line 69: This is also allowed: 

```html
<div role="application">
  <div role="document">
    <div role="application"></div>
  </div>
</div>
```

docs/rule/no-nested-landmark.md
```diff
@@ -0,0 +1,77 @@
...
+
+## List of elements & their corresponding roles
+
+- header (banner)
+- main (main)
+- aside (complementary)
+- form (form, search)
+- main (main)
+- nav (navigation)
+- footer (contentinfo)
```
rwjblue, line 13: Can you link (in resources section below) the canonical reference for these?

lib/rules/no-nested-landmark.js
```diff
@@ -0,0 +1,59 @@
...
+
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+function createErrorMessage(element) {
+  return `Nested landmark elements on <${element}> detected. Landmark elements should not be nested within landmark elements of the same name.`;
+}
+
+const LANDMARK_ELEMENTS = ['header', 'main', 'aside', 'form', 'main', 'nav', 'footer'];
+const ROLES = ['banner', 'main', 'complementary', 'form', 'search', 'navigation', 'contentinfo'];
```
rwjblue, line 11: Can you add an inline comment just above each of these lines to link to the spec where they are defined?

lib/rules/no-nested-landmark.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = class NoNestedLandmark extends Rule {
+  visitor() {
+    this._parentLandmarkNode = null;
+
+    let visitor = {
```
rwjblue, line 17: It is vaguely confusing to me when reading a method named `vistor` with an inline variable named `visitor`. Maybe we can call this something else (e.g. `process`/`checkNode`,etc?).

lib/rules/no-nested-landmark.js
```diff
@@ -0,0 +1,59 @@
...
+function createErrorMessage(element) {
+  return `Nested landmark elements on <${element}> detected. Landmark elements should not be nested within landmark elements of the same name.`;
+}
+
+const LANDMARK_ELEMENTS = ['header', 'main', 'aside', 'form', 'main', 'nav', 'footer'];
+const ROLES = ['banner', 'main', 'complementary', 'form', 'search', 'navigation', 'contentinfo'];
+
+module.exports = class NoNestedLandmark extends Rule {
+  visitor() {
+    this._parentLandmarkNode = null;
```
rwjblue, line 15: I think we can remove the manual enter/exit handling in favor of the (newish) support for paths passed to the hooks.

Pseudo code:

```js
return {
  ElementNode(node, path) {
    for (let parent of path.parents()) {
      if (this.isLandmarkElement(parent.node)) {
        this.log(...);
        return;
      }
    }
  },
};
```

For example:

https://github.com/ember-template-lint/ember-template-lint/blob/26fcdea982bb6f1b2050679f9f80ff211c238f33/lib/rules/require-input-label.js#L8-L19

test/unit/rules/no-nested-landmark-test.js
```diff
@@ -0,0 +1,76 @@
...
+  name: 'no-nested-landmark',
+
+  config: true,
+
+  good: [
+    '<div><main></main></div>',
+    '<div role="application"><div role="document"><div role="application"></div></div></div>',
+  ],
+
+  bad: [
```
rwjblue, line 18: Can you add some additional combos of nesting that we want to prevent? All of these tests (at the moment) we are only checking `main` nested inside `main`.

lib/rules/no-nested-landmark.js
rajasegar: The way path.parents() works is that it only capture the immediate parents of the current node not the ancestors I guess. I tried with the following logic:
```js
return {
  ElementNode(node, path) {
    for (let parent of path.parents()) {
      if (this.isLandmarkElement(parent.node)) {
        this.log(...);
        return;
      }
    }
  },
};


```
The above logic works for:
```html
<main>
  <main></main>
</main>
```
But not for:
```html
<main>
  <div>
    <main></main>
  </div>
</main>
```

test/unit/rules/no-nested-landmark-test.js
rajasegar: Since the rule is checking only for landmark elements, it will cover all the elements and roles defined here:
```js
const LANDMARK_ELEMENTS = ['header', 'main', 'aside', 'form', 'main', 'nav', 'footer'];
const ROLES = ['banner', 'main', 'complementary', 'form', 'search', 'navigation', 'contentinfo'];
```
Adding tests for all the items above will be overkill I guess. Nevertheless I tried to capture all the variations for a single element (main) with roles and nesting.
Added for `nav` and `header`, the most commonly used pattern AFAIK. Would be glad to add more if required.
What do you think? @MelSumner

lib/rules/no-nested-landmark.js
rwjblue: > Actually I took the cues from `no-nested-interactive`

Ya, but that was written before we can walk the parents 😄, and should be rewritten / refactored to avoid the error prone manual tracking.

rwjblue: > The way path.parents() works is that it only capture the immediate parent of the current node, not the ancestors I guess.

Nope, it traverses all parents. Using the snippet you mentioned (main > div > main) and using `path.parents()` from the nested `main` _can **definitely**_ see the top level parent `main`. See this AST explorer demo:

https://astexplorer.net/#/gist/9e4b550991b7b62eaa62f847dd09bf93/f050ca2a7b46a78dbae5d650ace1e5ac8fd628a1

test/unit/rules/no-nested-landmark-test.js
rwjblue: Ya, I agree we don't need a test for every combo. I was just looking for at least _one_ test for non-main.

lib/rules/no-nested-landmark.js
rajasegar: Thanks @rwjblue that worked, my bad I missed the condition to check whether the current child node is a landmark element or not, that's why my tests failed. Now after changing to the ElementNode hook, the code looks much cleaner and concise. 👍

rajasegar:
@rwjblue Addressed the review comments, please check and let me know

Issue #1286, opened 2020-04-19
roleValue.trim is not a function


mansona:
I'm getting a bit of a strange error that I can't figure out 🤔 

```
addon/templates/components/es-card-content.hbs
  -:-  error  roleValue.trim is not a function  undefined
```

This is most likely because of the following code:

```html
  <img
    class="card__image"
    src={{@image.src}}
    alt={{if @image.alt @image.alt ""}}
    role={{unless @image.alt "presentation"}}
    width={{@image.width}}
    height={{@image.width}}
  >
```

I figured it was something reasonably easy to recreate so I adding the following tests to the `good` section of `test/unit/rules/no-invalid-role-test.js` expecting it to fail in the same way: 

```
    '<img role={{unless this.altText "presentation"}} alt={{this.altText}}>',
    '<img role={{if this.isPresentation "presentation"}}',
```

but the tests pass 🤔 

Any ideas?

rwjblue:
Yeah, I would expect that to be the reproduction. Do you have the full stack trace?

mansona:
@rwjblue it turns out that code is actually in the `require-valid-alt-text` rule 🤔 

https://github.com/ember-template-lint/ember-template-lint/pull/1290 is a reproduction for the case that I'm experiencing

gabrielcsapo:
I took a shot at fixing this by not validating those alt tags.

rwjblue:
Fixed by https://github.com/ember-template-lint/ember-template-lint/pull/1324

PR #1324, opened 2020-05-03
[fix] require-valid-alt-text parsing mustache statements

gabrielcsapo:
# Motivation

https://github.com/ember-template-lint/ember-template-lint/issues/1286 and instances in our application set the role based on a conditional statement. This will short circuit and not validate those alt tags.


rwjblue:
Thanks @gabrielcsapo!

PR #1327, opened 2020-05-05
Add new rule: [a11y] Ensure duplicate landmark elements have unique labels

### LINKED ISSUE #1272: Ensure duplicate landmark elements have unique labels
If two of the same landmark elements or landmark roles are found in the same template, ensure that they have a unique label (provided by the `aria-label` or `aria-labelledby` attribute). 

## List of elements & their corresponding roles
- header (banner)
- main (main)
- aside (complementary)
- form (form, search)
- main (main)
- nav (navigation)
- footer (contentinfo)

### Allowed

```html
<nav aria-label="primary site navigation"></nav>
<nav aria-label="secondary site navigation within home page"></nav>
```

```html
<nav aria-label="primary site navigation"></nav>
<div role="navigation" aria-label="secondary site navigation within home page"></nav>
```

### Disallowed

```html
<nav></nav>
<nav></nav>
```

```html
<nav></nav>
<div role="navigation"></div>
```

```html
<nav aria-label="site navigation"></nav>
<nav aria-label="site navigation"></nav>
```

rajasegar:
Fixes #1272


lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+  banner: 'header',
+  main: 'main',
+  complementary: 'aside',
+  form: 'form',
+  search: 'form',
+  navigation: 'nav',
+  contentinfo: 'footer',
+};
+
+const LANDMARK_ELEMENTS = ['header', 'main', 'aside', 'form', 'main', 'nav', 'footer'];
```
sharpshark28, line 20: With this change, we're recommending each form require a unique label then? Just ensuring I'm following this correctly. So `<form></form><form></form>` would be invalid for example.

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+
+const ROLE_LANDMARK_MAP = {
+  banner: 'header',
+  main: 'main',
+  complementary: 'aside',
+  form: 'form',
+  search: 'form',
+  navigation: 'nav',
+  contentinfo: 'footer',
+};
```
sharpshark28, line 18: Should we include `region: section,`? https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+        // <nav>
+        const isLandMarkElement = LANDMARK_ELEMENTS.includes(node.tag);
+
+        // <div role="navigation">
+        const isLandMarkRole = LANDMARK_ROLES.includes(roleValue);
+
+        if (isLandMarkElement || isLandMarkRole) {
+          // check for label
+          const labelValue = AstNodeInfo.elementAttributeValue(node, 'aria-label') || '';
+          const isAvailable = this._landmarkHash.find((e) => {
```
sharpshark28, line 51: Not the most clear variable name. Maybe something like `isNotUnique`

lib/rules/no-duplicate-landmark-labels.js
MelSumner: @sharpshark28 yes, that's correct- each form needs a unique label.

docs/rule/no-duplicate-landmark-labels.md
```diff
@@ -0,0 +1,44 @@
...
+
+```hbs
+<nav aria-label="primary site navigation"></nav>
+<nav aria-label="secondary site navigation within home page"></nav>
+```
+
+```hbs
+<nav aria-label="primary site navigation"></nav>
+<div role="navigation" aria-label="secondary site navigation within home page"></div>
+```
```
MelSumner, line 44: An especially important example would be multiple forms on the same page- can you add that as an example, please? Thank you!

lib/rules/no-duplicate-landmark-labels.js
MelSumner: Region is an implicit role that should never be used by authors but let me check on `section` - thank you for sharing the link. It's not the same as the spec so I need to double check a few things.

lib/rules/no-duplicate-landmark-labels.js
rajasegar: May be `isDuplicate` is more direct, I will use that.

rajasegar:
@sharpshark28 @MelSumner Thanks for the review, I have addressed the same in the new commit. Please check and let me know.

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+
+        if (isLandMarkElement || isLandMarkRole) {
+          // check for label
+          const labelValue = AstNodeInfo.elementAttributeValue(node, 'aria-label') || '';
+          const isNotUnique = this._landmarkHash.find((e) => {
+            if (isLandMarkElement) {
+              return e.landmark === node.tag && e.label === labelValue;
+            } else {
+              return e.landmark === ROLE_LANDMARK_MAP[roleValue] && e.label === labelValue;
+            }
```
rwjblue, line 56: It seems like we can simplify this a bit by only caring about the resulting landmark role + label.

_Possibly_ something like:

```js
let landmarkRole = isLandmarkElement ? node.tag : ROLE_LANDMARK_MAP[roleValue];

return e.landmark === landmarkRole && e.label === labelValue;
```

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+    return {
+      ElementNode(node) {
+        const hasRoleAttribute = AstNodeInfo.hasAttribute(node, 'role');
+
+        // Return if it is not a landmark element or doesn't have role
+        if (!(hasRoleAttribute || LANDMARK_ELEMENTS.includes(node.tag))) {
+          return;
+        }
+
+        const roleValue = AstNodeInfo.elementAttributeValue(node, 'role');
```
rwjblue, line 40: The code below doesn't seem to account for `roleValue` to be anything other than a string, but in reality it can be either a string or AST node.

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+
+        // <nav>
+        const isLandMarkElement = LANDMARK_ELEMENTS.includes(node.tag);
+
+        // <div role="navigation">
+        const isLandMarkRole = LANDMARK_ROLES.includes(roleValue);
+
+        if (isLandMarkElement || isLandMarkRole) {
+          // check for label
+          const labelValue = AstNodeInfo.elementAttributeValue(node, 'aria-label') || '';
```
rwjblue, line 50: The code below doesn't seem to account for `labelValue ` to be anything other than a string, but in reality it can be either a string or AST node.

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+function createErrorMessage(element) {
+  return `Duplicate label on <${element}> detected. If two of the same landmark elements or landmark roles are found in the same template, ensure that they have a unique label (provided by the aria-label or aria-labelledby attribute).`;
+}
+
+const ROLE_LANDMARK_MAP = {
```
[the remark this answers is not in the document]
rajasegar: Sure @rwjblue sounds like a good idea to me

lib/rules/no-duplicate-landmark-labels.js
rajasegar: done

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,76 @@
...
+
+const LANDMARK_ELEMENTS = ['header', 'main', 'aside', 'form', 'main', 'nav', 'footer'];
+
+const LANDMARK_ROLES = Object.keys(ROLE_LANDMARK_MAP);
+
+module.exports = class NoDuplicateLandmarkLabels extends Rule {
+  constructor(options) {
+    super(options);
+    this._landmarkHash = [];
+  }
```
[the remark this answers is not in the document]
rajasegar: I tried with `Set` but since the object references change every time, it's not able to catch the duplicate it seems.
```js
  constructor(options) {
    super(options);
    this._landmarkHash = new Set();
  }
  visitor() {
    return {
      ElementNode(node) {
        ...
        if (isLandMarkElement || isLandMarkRole) {
          // check for label
          const label = AstNodeInfo.elementAttributeValue(node, 'aria-label') || '';
          const landmark = isLandMarkElement ? node.tag : ROLE_LANDMARK_MAP[roleValue];
          let current = { landmark, label };
          if (this._landmarkHash.has(current)) {
            this.log({
              message: createErrorMessage(node.tag),
            });
          } else {
            this._landmarkHash.add(current);
          }
...
 ```

File: lib/rules/no-duplicate-landmark-labels.js (hunk 6 above)
Comment (rajasegar) [reply lib/rules/no-duplicate-landmark-labels.js:40]: You mean something like this:
```js
<div role={{landmarkRole}}></div>
```

lib/rules/no-duplicate-landmark-labels.js
rajasegar: is this a good example of what you are talking about?
```hbs
<div role="navigation" aria-label={{myLabel}}></div>
<nav aria-label={{myLabel}}></div>
```

rajasegar: @rwjblue If we are thinking about going that way, I think we have a lot of ground to cover. 
There are so many variations like:
```hbs
<main aria-label={{myLabel}}></main>
<main aria-label="label-{{if myLabel "primary" "secondary"}}></main>
<main aria-label={{@myLabel}}></main>
<main aria-label={{concat "label-" myLabel}}></main>
```
I think the best possible way here, is to convert all the AST Node values to their stringified forms 
so that it will be easy to compare against. Is there any util available already in the codebase for the same?
I took a look at this one 
https://github.com/ember-template-lint/ember-template-lint/blob/cb218e3404bc865434c8dae537aaa4982464fb58/lib/helpers/ast-node-info.js#L110
But it seems not enough to handle other than `TextNode`, for `MustacheStatement` I can get a value like `{{myLabel}}` but for others like `ConcatStatement` we need to handle I guess..

lib/rules/no-duplicate-landmark-labels.js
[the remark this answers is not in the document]
rwjblue: Ya, makes sense, sorry for the troubles.

lib/rules/no-duplicate-landmark-labels.js
rwjblue: Yes, exactly.

lib/rules/no-duplicate-landmark-labels.js
rwjblue: In theory, it should be possible to add support for other types of values but we don't have any good examples of that in the codebase at the moment. Basically, there is no particular reason that we can't tell these are the same value:

```
<main aria-label={{myLabel}}></main>
<main aria-label={{myLabel}}></main>
```

Like you said, we'd need to represent the value independently for each. Perhaps we could use the `sourceForNode` to compare the raw source :thinking:. We'd have to account for this (but taht isn't _too_ hard):

```hbs
<main aria-label={{myLabel}}></main>
{{#let (concat "special-" this.otherThing) as |myLabel|}}
  <main aria-label={{myLabel}}></main>
{{/let}}
```

---

I think for a first pass, we can just bail on checking if the the value is _not_ a TextNode. In my experience they are _very commonly_ literal values, so it makes sense to start with checking them that way, and return early (without logging) when we see a dynamic value. Though I'm super happy to help work out the more complicated thing if you would like to do that instead.

lib/rules/no-duplicate-landmark-labels.js
rwjblue: FWIW, I've tracked removing this "automatic unwrapping" from `ASTNodeInfo.elementAttributeValue` in https://github.com/ember-template-lint/ember-template-lint/issues/1358.

MelSumner:
@rwjblue @rajasegar do we have a clear path forward for next steps on this PR?

rajasegar:
I am sorry I couldn't spend time on this due to some health issues, I will get back to this from tomorrow

MelSumner:
> I am sorry I couldn't spend time on this due to some health issues, I will get back to this from tomorrow

@rajasegar take care of yourself! that's most important.

lib/rules/no-duplicate-landmark-labels.js
rajasegar: Ok, Sound good to me,  hence, for now, I will return straight-away if the role value type is not `TextNode`. And I will address the complicated stuff for dynamic values in a separate PR

rajasegar:
@MelSumner @rwjblue As of now, we are not addressing the dynamic role value scenarios for duplicate label checking, we are only targeting literal values at the moment. We will be addressing the same in a separate PR once the util functions are ready with #1358

lib/rules/no-duplicate-landmark-labels.js
MelSumner: TODO: update this code to reflect that it could be something other than a string

test/unit/rules/no-duplicate-landmark-labels-test.js
```diff
@@ -0,0 +1,72 @@
...
+
+generateRuleTests({
+  name: 'no-duplicate-landmark-labels',
+
+  config: true,
+
+  good: [
+    '<nav aria-label="primary site navigation"></nav><nav aria-label="secondary site navigation within home page"></nav>',
+    '<nav aria-label="primary site navigation"></nav><div role="navigation" aria-label="secondary site navigation within home page"></div>',
+    '<nav aria-label={{siteNavigation}}></nav><nav aria-label={{siteNavigation}}></nav>',
```
MelSumner, line 17:
```suggestion
    '<nav aria-label={{siteNavigation}}></nav><nav aria-label={{siteNavigation}}></nav>',
    // since we can't confirm what the role of the div is, we have to let it pass
    '<nav aria-label="primary site navigation"></nav><div role={{role}} aria-label="secondary site navigation within home page"></div>',
```

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,83 @@
...
+  banner: 'header',
+  main: 'main',
+  complementary: 'aside',
+  form: 'form',
+  search: 'form',
+  navigation: 'nav',
+  contentinfo: 'footer',
+};
+
+const LANDMARK_ELEMENTS = new Set(['header', 'main', 'aside', 'form', 'main', 'nav', 'footer']);
```
hergaiety, line 20:
```suggestion
const LANDMARK_ELEMENTS = new Set(['header', 'main', 'aside', 'form', 'nav', 'footer']);
```

`main` is in this `Set` twice which seems to be a minor mistake. There are no consequences to this mistake because it still evaluates in Javascript to `[ "header", "main", "aside", "form", "nav", "footer" ]` due to JS stripping duplicate values from `Set`'s.

Alternatively, we could rely on Javascript enforcing uniqueness by converting the `ROLE_LANDMARK_MAP`'s values to a `Set` like so:

```js
const LANDMARK_ELEMENTS = new Set(Object.values(ROLE_LANDMARK_MAP));
```

This would better match the pattern on the next line using `Object.keys` as well.

lib/rules/no-duplicate-landmark-labels.js
```diff
@@ -0,0 +1,83 @@
...
+        const hasRoleAttribute = AstNodeInfo.hasAttribute(node, 'role');
+
+        // Return if it is not a landmark element or doesn't have role
+        // or if the role value is not a string literal
+        if (
+          !(
+            hasRoleAttribute ||
+            LANDMARK_ELEMENTS.has(node.tag) ||
+            hasRoleAttribute.value.type === 'TextNode'
+          )
```
hergaiety, line 42:
```suggestion
          (
            hasRoleAttribute ||
            LANDMARK_ELEMENTS.has(node.tag) ||
            hasRoleAttribute.value.type === 'TextNode'
          ) === false
```

Multiline `if` conditionals can sometimes be hard for a developer to read. I think this one is fairly clear, but the `!` is hiding to reverse the entire check. Explicitly checking the evaluation to `=== false` may help avoid developer confusion of how this check works.

rajasegar:
Thanks @MelSumner @hergaiety for the review, addressed the same in the latest commit

MelSumner:
@rajasegar can you resolve the conflicts, and we can re-review for merging. Thank you! :)

rajasegar:
@MelSumner Conflicts resolved and branch updated

rwjblue:
Merged in #1550

PR #1332, opened 2020-05-08
feat: Add no-element-args rule

deanpapastrat:
Today, Ember will throw an error that causes the entire page to fail to render if you pass named arguments to an HTML element, like so:

```hbs
<a @size="mini">My awesome link</a>
```

This PR adds a lint rule that enforces against this usage, and protects people from making this mistake in code that isn't covered by tests.

### DIFF
--- a/docs/rule/no-element-args.md
+++ b/docs/rule/no-element-args.md
@@ -1,0 +1,30 @@
+# no-element-args
+
+:white_check_mark: The `extends: 'recommended'` property in a configuration file enables this rule.
+
+Passing named arguments to HTML elements causes Ember to fail to render your application. Typically, these are added by accident when
+you actually mean to pass an attribute to the element.
+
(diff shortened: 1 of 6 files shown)


chancancode:
Thanks for looking into this and putting this PR together. As I said on the other thread, this is a syntax error, not a lint error (has a well-defined meaning, just may not be what you expect). I think we should just make it a compile-time syntax error in either glimmer-vm or in ember as an ast plugin. The code would be quite similar to what you have in here already, since they are both working against the AST in a similar way.

lifeart:
related issue: https://github.com/ember-template-lint/ember-template-lint/issues/926

lifeart:
one more related pr: https://github.com/ember-template-lint/ember-template-lint/pull/927

deanpapastrat:
Closing in favor of #927

PR #1333, opened 2020-05-09
Ensure `<MyComponent @prop={{can.do}} />` triggers `no-implicit-this` error.

rwjblue:
This scenario recently came up in https://github.com/emberjs/ember.js/issues/18957 and I wanted to confirm that this usage would properly flag `{{can.do}}` as an implicit this fallback.

### DIFF
--- a/test/unit/rules/no-implicit-this-test.js
+++ b/test/unit/rules/no-implicit-this-test.js
@@ -22,6 +22,7 @@ let good = [
   '{{textarea}}',
   '{{yield}}',
   '{{welcome-page}}',
+  '<MyComponent @prop={{can "edit" @model}} />',
   {
     config: { allow: ['book-details'] },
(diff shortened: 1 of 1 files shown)


test/unit/rules/no-implicit-this-test.js
```diff
@@ -94,5 +95,24 @@ generateRuleTests({
...
+      template: '<MyComponent @prop={{can.do}} />',
+      result: {
+        message: message('can.do'),
+        source: 'can.do',
+        line: 1,
+        column: 21,
+      },
+    },
+    {
+      template: '<MyComponent @prop={{can.do}} />',
```
rwjblue, line 108: Note: this is still a failure (even when `can` is explicitly provided as an allowed helper) because it is definitely not a helper invocation.

PR #1335, opened 2020-05-11
Add new rule: no-forbidden-elements

### LINKED ISSUE #747: add new rule: no-forbidden-elements
- disallow use of meta, style, html and script tags in template files. 
- rule should be configurable so teams can add their own disallowed elements

rajasegar:
- Disallow use of meta, style, html and script tags in template files.
- Rule should be configurable so teams can add their own disallowed elements

Fixes #747

### DIFF
--- a/docs/rule/no-forbidden-elements.md
+++ b/docs/rule/no-forbidden-elements.md
@@ -1,8 +1,9 @@
 # no-forbidden-elements
 
-This rule disallows the use of `meta`, `style`, `html` and `script` tags in template files.
+This rule disallows the use of forbidden elements in template files.
 
 The rule is configurable so teams can add their own disallowed elements.
+The default list of forbidden elements are `meta`, `style`, `html` and `script`.
 
 ## Examples
 
@@ -37,6 +38,6 @@ This rule **allows** the following:
 ## Configuration
 
 * boolean - `true` to enable / `false` to disable
-  * array -- an array of strings to forbid
+  * array -- an array of strings to forbid, default: ['meta', 'style', 'html', 'style']
   * object -- An object with the following keys:
     * `forbidden` -- An array of forbidden element names
--- a/docs/rule/no-forbidden-elements.md
+++ b/docs/rule/no-forbidden-elements.md
@@ -21,7 +21,7 @@ This rule **forbids** the following:
 ```
 
 ```hbs
-<html></html>
+<meta charset="utf-8">
 ```
 
 This rule **allows** the following:
--- a/docs/rule/no-forbidden-elements.md
+++ b/docs/rule/no-forbidden-elements.md
@@ -1,0 +1,42 @@
+# no-forbidden-elements
+
+This rule disallows the use of `meta`, `style`, `html` and `script` tags in template files.
+
+The rule is configurable so teams can add their own disallowed elements.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<script></script>
+```
+
+```hbs
+<style></style>
+```
+
+```hbs
+<html></html>
+```
+
+```hbs
+<html></html>
+```
+
+This rule **allows** the following:
+
+```hbs
+<header></header>
+```
+
+```hbs
+<div></div>
+```
+
+## Configuration
+
+* boolean - `true` to enable / `false` to disable
+  * array -- an array of strings to forbid
+  * object -- An object with the following keys:
+    * `forbidden` -- An array of forbidden element names
--- a/lib/rules/no-forbidden-elements.js
+++ b/lib/rules/no-forbidden-elements.js
@@ -1,0 +1,90 @@
+'use strict';
+
+const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+
+function ERROR_MESSAGE_FORBIDDEN_ELEMENTS(element) {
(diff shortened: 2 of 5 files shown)


lib/rules/no-forbidden-elements.js
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    throw new Error(errorMessage);
+  }
+  visitor() {
+    return {
+      ElementNode(node) {
+        let isForbiddenElement =
+          FORBIDDEN_ELEMENTS.includes(node.tag) || this.config.forbidden.includes(node.tag);
+        if (isForbiddenElement) {
```
rwjblue, line 77: We should not hard code `FORBIDDEN_ELEMENTS` like this, `this.config.forbidden` will include the `FORBIDDEN_ELEMENTS` by default (when using `true` for config value) but if configured the explicit list that is provided should be used. 

tldr;

```suggestion
        let isForbiddenElement = this.config.forbidden.includes(node.tag);
        if (isForbiddenElement) {
```

lib/rules/no-forbidden-elements.js
```diff
@@ -0,0 +1,90 @@
...
+
+function ERROR_MESSAGE_FORBIDDEN_ELEMENTS(element) {
+  return `Use of <${element}> detected. Do not use forbidden elements.`;
+}
+
+const FORBIDDEN_ELEMENTS = ['meta', 'style', 'html', 'script'];
+
+const DEFAULT_CONFIG = {
+  forbidden: [],
+};
```
rwjblue, line 14: The `DEFAULT_CONFIG` should be where the list of default forbidden elements live:

```suggestion
const DEFAULT_CONFIG = {
  forbidden: ['meta', 'style', 'html', 'script'],
};
```

test/unit/rules/no-forbidden-elements-test.js
```diff
@@ -0,0 +1,81 @@
...
+
+const ERROR_MESSAGE_FORBIDDEN_ELEMENTS = require('../../../lib/rules/no-forbidden-elements')
+  .ERROR_MESSAGE_FORBIDDEN_ELEMENTS;
+
+generateRuleTests({
+  name: 'no-forbidden-elements',
+
+  config: true,
+
+  good: ['<header></header>', '<div></div>', '<footer></footer>', '<p></p>'],
```
rwjblue, line 13: Can you add some `good` examples that specify custom config to allow `<script>`?

Something like:

```js
{
  config: ['meta'],
  template: '<script></script>',
}
```

test/unit/rules/no-forbidden-elements-test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  bad: [
```
rwjblue, line 15: Can you add `bad` examples where a user configures a custom element? Something like:

```
{
  config: ['Foo'],
  template: '<Foo />',
  result: {
    // error....
  }
}
```

docs/rule/no-forbidden-elements.md
```diff
@@ -0,0 +1,42 @@
+# no-forbidden-elements
+
+This rule disallows the use of `meta`, `style`, `html` and `script` tags in template files.
```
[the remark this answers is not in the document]
rajasegar: Done

lib/rules/no-forbidden-elements.js
```diff
@@ -0,0 +1,90 @@
...
+    } else if (!DEFAULT_CONFIG[key]) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+function sanitizeConfigArray(forbidden = []) {
+  return forbidden.filter((option) => option !== '').sort((a, b) => b.length - a.length);
```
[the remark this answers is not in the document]
rajasegar: Done

rajasegar:
@rwjblue Addressed the review comments, please check and let me know.

test/unit/rules/no-forbidden-elements-test.js
rajasegar: Added

test/unit/rules/no-forbidden-elements-test.js
rajasegar: Done

lib/rules/no-forbidden-elements.js
rajasegar: Done

lib/rules/no-forbidden-elements.js
rajasegar: Done

lib/rules/no-forbidden-elements.js
```diff
@@ -0,0 +1,90 @@
...
+      config
+    );
+
+    throw new Error(errorMessage);
+  }
+  visitor() {
+    return {
+      ElementNode(node) {
+        let isForbiddenElement =
+          this.config.forbidden.includes(node.tag) || this.config.forbidden.includes(node.tag);
```
rwjblue, line 76: This condition seems odd to me, unless I'm mistaken this is checking the same thing twice?

rajasegar: Ah Yes, my bad, I applied your suggestion but failed to check the duplicates. Fixed it now in the latest commit

PR #1338, opened 2020-05-12
Clarify recommended fix for `no-negated-condition` rule

### LINKED ISSUE #1337: Colliding rules no-negated-condition and simple-unless
For the following template `ember-template-lint` throws the error `no-negated-condition`. 
```hbs
{{#if (not this.var)}} <!-- Flip `if ` statement to avoid `if !condition`. (no-negated-condition)-->
  guess not
{{else}}
  what a surprise
{{/if}}
```
If we change the `if` usage to `unless` we in turn get a  `simple-unless` error.
```hbs
{{#unless this.var}}
  guess not
{{else}} <!--Using an {{else}} block with {{unless}} should be avoided. (simple-unless)}}-->
  what a surprise
{{/unless}}
```

What would be the correct approach here in regards of both rules?

bmish:
Fixes #1337. CC: @velrest

### DIFF
--- a/docs/rule/no-negated-condition.md
+++ b/docs/rule/no-negated-condition.md
@@ -9,9 +9,9 @@ It can be hard to reason about negated conditions:
 
 Negated conditions can often be avoided or simplified by:
 
-* Flipping `if (not condition) ... else ...` to `if .... else ...`
-* Replacing `if (not condition)` with `unless`
-* Replacing `unless (not condition)` with `if`
+* Flipping `{{if (not condition)}} {{prop1}} {{else}} {{prop2}} {{/if}}` to `{{if condition}} {{prop2}} {{else}} {{prop1}} {{/if}}`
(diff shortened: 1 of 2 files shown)


docs/rule/no-negated-condition.md
```diff
@@ -9,9 +9,9 @@ It can be hard to reason about negated conditions:
 
 Negated conditions can often be avoided or simplified by:
 
-* Flipping `if (not condition) ... else ...` to `if .... else ...`
-* Replacing `if (not condition)` with `unless`
-* Replacing `unless (not condition)` with `if`
+* Flipping `if (not condition) { block1 } else { block2 }` to `if condition { block2 } else { block1 }`
```
rwjblue, line 12: Can we make these examples use template syntax instead of JS? I find the intermixing of the JS and HBS a bit confusing

bmish: Agreed, fixed!

PR #1340, opened 2020-05-13
Feature: adds the option to provide custom formatters

### LINKED ISSUE #1339: Feature: allow custom reporters
It would be great to allow reporters with the same interface as eslint. https://eslint.org/docs/developer-guide/working-with-custom-formatters

Is this something that would be accepted?

gabrielcsapo:
fixes #1339 

# Motivation

It would be great to allow reporters with the same interface as eslint. https://eslint.org/docs/developer-guide/working-with-custom-formatters

The current PR allows for the following options

```js
ember-template-lint --formatter test . // this will load ember-template-lint-printers-test
```

```js
ember-template-lint --formatter ./custom-formatter.js . // this will load custom-formatter.js relative to the current working directory
```

## Still need to 

- [x] update docs
- [x] add tests


gabrielcsapo:
@rwjblue before I add docs, is this going in the right direction?

gabrielcsapo:
@rwjblue bump on thoughts on direction for this feature?

gabrielcsapo:
@rwjblue thoughts on this?

rwjblue:
Sorry, I've been a bit swamped lately. I'll try to review next week.

gabrielcsapo:
@rwjblue thoughts on this?

package.json
```diff
@@ -54,6 +54,7 @@
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.1.3",
+    "ember-template-lint-printer-test": "link:./test/fixtures/ember-template-lint-printer-test",
```
rwjblue, line 57: This seems odd, why do we need this?

lib/printers/default.js
```diff
@@ -1,15 +1,52 @@
...
+        let PrettyPrinter = require('./pretty');
+        this.delegates.push(new PrettyPrinter(options));
+        break;
+      }
+      default: {
+        const { dir } = path.parse(options.printer);
+
+        // printers can be relative to the current working directory is come from an npm package
+        if (dir === '.') {
+          const customPrinterPath = path.resolve(process.cwd(), options.printer);
```
rwjblue, line 29: We shouldn't use the `process.cwd()` here (we've landed a threaded through working directory properly)

rwjblue: Also I think I'd prefer to have the bin script resolve these things itself, and pass it in

bin/ember-template-lint.js
```diff
@@ -115,6 +115,11 @@ function parseArgv(_argv) {
         boolean: true,
         default: false,
       },
+      printer: {
+        describe: 'Specify formatter to be used in printing output',
+        type: 'string',
+        default: 'pretty',
+      },
```
rwjblue, line 122: Hmm, any reason not to go with what eslint does and call this a formatter?

rwjblue (changes requested):

package.json
gabrielcsapo: This was to test being able to make sure we can load printers from the node_modules directory.

bin/ember-template-lint.js
gabrielcsapo: will update this to be called formatters everywhere it is called printer.

gabrielcsapo: should we rename all the things that we call printer? There is a folder in `/lib` called printers, should this be called formatters?

lib/printers/default.js
gabrielcsapo: will make the changes to do the resolution in the bin script.

bin/ember-template-lint.js
rwjblue: > There is a folder in /lib called printers, should this be called formatters?

Ya, seems good to rename that, but maybe you could make that an "internal only" refactoring change as a separate PR (to keep the "new feature" clearly separate from the "internal renaming")?

gabrielcsapo:
any updates on this @rwjblue ?

bin/ember-template-lint.js
scalvert: It's actually called `--format` in eslint. I think we should match that.

lib/formatters/default.js
```diff
@@ -0,0 +1,64 @@
...
+        } else {
+          // you can also import a printer from a package "ember-template-lint-formatter-${name of package}"
+          try {
+            // eslint-disable-next-line import/no-dynamic-require
+            const CustomPrinter = require(require.resolve(
+              `ember-template-lint-formatter-${options.format}`,
+              {
+                paths: [options.workingDirectory],
+              }
+            ));
```
rwjblue, line 43: We should try to use the formatter package that they explicitly mentioned before falling back to the concatenated version.

lib/formatters/default.js
```diff
@@ -0,0 +1,64 @@
...
+class DefaultPrinter {
+  constructor(options = {}) {
+    this.delegates = [];
+    this.console = options.console || console;
+
+    if (options.json) {
+      options.format = 'json';
+    }
+
+    switch (options.format) {
```
rwjblue, line 14: I would like this option to support multiple formatters at once. Specifically, something like:

```
ember-template-lint . --format pretty --format json --format xunit
```

And have that run all three types. The majority of the code here actually would support that (since we use `this.delegates` and loop over them), but I think this outer switch statement would need to be tweaked slightly (based on `Array.isArray(options.format)`).

lib/formatters/default.js
```diff
@@ -0,0 +1,64 @@
...
+const path = require('path');
+
+class DefaultPrinter {
+  constructor(options = {}) {
+    this.delegates = [];
+    this.console = options.console || console;
+
+    if (options.json) {
+      options.format = 'json';
+    }
```
rwjblue, line 12: I would actually like to differentiate between the `--json` flag and the internal formatters. Specifically, I'd like to diverge the formats and have a new JSON formatter that writes files to disk (instead of emitting stdout).

lib/linter.js
```diff
@@ -382,7 +382,7 @@ class Linter {
       verbose: false,
     };
 
-    let PrettyPrinter = require('./printers/pretty');
+    let PrettyPrinter = require('./formatters/pretty');
```
rwjblue, line 385: We should consider if we still support ember-cli-template-lint as of 3.0.0. I actually think we don't (because ember-cli-template-lint cannot pass a filePath, and we stopped supporting passing a moduleId/moduleName).

If that is the case, then we can probably remove this static method (since it was only supporting ember-cli-template-lint). 

That should be a separate PR though, just thought of this while reviewing.

test/acceptance/cli-test.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
+
 const fs = require('fs');
+const { copy } = require('fs-extra');
```
rwjblue, line 4: Hmm, why do we need this?

test/acceptance/cli-test.js
```diff
@@ -1,7 +1,11 @@
...
 const path = require('path');
 
+const { ensureTodoStorageDir } = require('@ember-template-lint/todo-utils');
```
rwjblue, line 7: ✂️

test/acceptance/cli-test.js
```diff
@@ -1461,6 +1471,99 @@ describe('ember-template-lint executable', function () {
...
+          },
+        });
+
+        let nodeModulesDir = path.join(project.baseDir, 'node_modules');
+        let formatterDir = path.join(nodeModulesDir, 'ember-template-lint-formatter-test');
+
+        await copy(
+          path.resolve(__dirname, '../fixtures/ember-template-lint-formatter-test'),
+          formatterDir
+        );
```
rwjblue, line 1552: Hmm, this is odd. Maybe just write it in the `project.write` above?

test/fixtures/ember-template-lint-formatter-test/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "ember-template-lint-printer-test",
+  "version": "1.0.0",
+  "description": "test printer to show how to package printers and use them."
+}
```
rwjblue, line 5: Should we require a custom keyword? E.g. `ember-template-lint-formatter`?

lib/formatters/default.js
scalvert: Let's do this separately. I can easily add this after this lands.

lib/formatters/default.js
scalvert: Fixed.

lib/formatters/default.js
scalvert: Fixed.

test/acceptance/cli-test.js
scalvert: Don't. Removed.

test/acceptance/cli-test.js
scalvert: Snippety dipped.

test/acceptance/cli-test.js
scalvert: Used `fs.symlinkSync` instead.

test/fixtures/ember-template-lint-formatter-test/package.json
scalvert: We could. Are you suggesting we sniff the keywords?

rwjblue: Lets not...

rwjblue (approved):

PR #1350, opened 2020-05-22
Add eslint-plugin-unicorn with recommended rules

bmish:
Autofixed 99% of violations.

https://github.com/sindresorhus/eslint-plugin-unicorn

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -6,24 +6,31 @@ const chalk = require('chalk');
 const micromatch = require('micromatch');
 const findUp = require('find-up');
 const defaultRules = require('./rules');
-const defaultConfigurations = require('./config/index');
+const defaultConfigurations = require('./config');
 const DEPRECATED_RULES = require('./helpers/deprecated-rules');
 
-const KNOWN_ROOT_PROPERTIES = ['extends', 'rules', 'pending', 'ignore', 'plugins', 'overrides'];
-const SUPPORTED_OVERRIDE_KEYS = ['files', 'rules'];
+const KNOWN_ROOT_PROPERTIES = new Set([
+  'extends',
+  'rules',
+  'pending',
+  'ignore',
+  'plugins',
+  'overrides',
+]);
+const SUPPORTED_OVERRIDE_KEYS = new Set(['files', 'rules']);
 // Severity level definitions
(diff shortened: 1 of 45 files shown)


lib/get-editor-config.js
```diff
@@ -16,8 +16,8 @@ class EditorConfigResolver {
      * comment: ;this is a comment
      */
     this.regex = {
-      section: /^\s*\[(([^#;]|\\#|\\;)+)\]\s*([#;].*)?$/,
-      param: /^\s*([\w.\-_]+)\s*[=:]\s*(.*?)\s*([#;].*)?$/,
+      section: /^\s*\[(([^#;]|\\#|\\;)+)]\s*([#;].*)?$/,
+      param: /^\s*([\w.-]+)\s*[:=]\s*(.*?)\s*([#;].*)?$/,
```
dcyriller, line 20: A _ seems to disappear here, is it okay?

lib/get-config.js
```diff
@@ -296,13 +300,13 @@ function getRuleFromString(rule) {
   const indexOfSeparator = rule.indexOf(':') + 1;
 
   // we have to split based on the index of the first instance of the separator because the separator could exist in the second half of the rule
-  const name = rule.substring(0, indexOfSeparator - 1);
-  let ruleConfig = rule.substring(indexOfSeparator);
+  const name = rule.slice(0, Math.max(0, indexOfSeparator - 1));
```
dcyriller, line 303: For this specific case, I’m not sure if the code looks better using slice instead of substring ^^

lib/get-editor-config.js
bmish: Looks like it's because `\w` already includes `_`:

> The \w metacharacter is used to find a word character. A word character is a character from a-z, A-Z, 0-9, including the _ (underscore) character.

https://www.w3schools.com/jsref/jsref_regexp_wordchar.asp

lib/get-config.js
bmish: That's true. I reverted this change for now.

Note that we can remove `Math.max` if we know the index will always be positive.

bmish:
Hoping to merge this in the next few days to avoid ending up with a lot of merge conflicts.

PR #1353, opened 2020-05-23
Add new rule: `no-named-arguments-with-global-attribute-names`

### LINKED ISSUE #1127: Rule to disallow some arg names that were meant for attribute bindings
A rule to disallow `@class` and `@classNames` and other attribute binding arguments would be good, especially for angle bracket invocations. I am converting some curly component invocations to angle brackets and want to take the conversion to `...attributes` in a separate pass.

jaydgruber:
### What
- adds a new rule `no-named-arguments-with-global-attribute-names`

### Why
- fixes #1127
- based default disallowed named-argument-names off [MDN's global attributes list](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes), with optional additional configuration


lib/rules/no-named-arguments-with-global-attribute-names.js
```diff
@@ -0,0 +1,179 @@
...
+  allow: [],
+  disallow: [],
+};
+
+function makeError(name) {
+  return ERROR_MESSAGE.replace('%', name);
+}
+
+// from the list of global attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
+// excluding `aria-*`, `data-*`, and event handler attributes like `onclick` (handled separately)
```
MelSumner, line 19: Can you shed light on why you excluded the `aria-` attributes?

test/unit/rules/no-named-arguments-with-global-attribute-names-test.js
```diff
@@ -0,0 +1,101 @@
...
+    template: '<MyComponent @class="shaded" />',
+    result: {
+      message: makeError('@class'),
+      source: '<MyComponent @class="shaded" />',
+      line: 1,
+      column: 0,
+    },
+  },
+  {
+    template: '<MyComponent @aria-valuemin={{1}} />',
```
MelSumner, line 30: How is this different from line 12, other than the number inside curly braces?

jaydgruber: The leading `@`

lib/rules/no-named-arguments-with-global-attribute-names.js
jaydgruber: Opted for `.startsWith('aria-') ` to check for them on L124 instead of listing them all out, since I was already doing similar for `data-*`

MelSumner: Ok, that makes sense. 👍

Issue #1180, opened 2020-02-29
Features required for `--fix` support 


rwjblue:
- [x] Ensure file is updated when passing `--fix` (https://github.com/ember-template-lint/ember-template-lint/pull/1183)
- [x] Update the pretty formatter to emit slightly different output when fixable errors are detected (like eslint does) https://github.com/ember-template-lint/ember-template-lint/pull/1185:

```
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

- [x] Add end to end acceptance tests for fixing (that invoke `ember-template-lint` with a template that needs fixes, and confirms the updated output) (https://github.com/ember-template-lint/ember-template-lint/pull/1183)

- [x] Ensure file with BOM is written back to disk with BOM #1190 

- [ ] Write text confirming functionality when reading from STDIN (needs some research, Do we write to STDOUT? What does ALE expect? What does eslint do here?)

dcyriller:
Just did some eslint research, I'm writing the results here for future reference 😄.

`eslint` has the following options: `--format=json`, `--stdin`, `--stdin-filename` (we have similar concepts in ember-template-lint).

Aside from these options, there is `--fix-dry-run`, which is an interesting option and could be ported pretty easily.

There is also a `--fix-type` option to let users specify which kind of fixes should be applied (`problem` / `suggestion` / `layout`).

I also see that with older versions of eslint, Ale would run `yarn eslint --fix the-file.js` ^^.

locks:
Most tasks listed are complete, and there is a separate issue for the last task regarding STDOUT, so I'm closing this one. Thanks for the work!

PR #1355, opened 2020-05-25
Add a test for fixing from stdin (povided a --filename is given)

dcyriller:
This is the command I would use to support fixing templates with Ale. And in my local environement it seems to work.

We don't write to STDOUT but to the file provided with the `--filename` arg.

If this PR is accepted, it would be the last item of this issue checklist: https://github.com/ember-template-lint/ember-template-lint/issues/1180

### DIFF
--- a/test/acceptance/editors-integration-test.js
+++ b/test/acceptance/editors-integration-test.js
@@ -60,5 +60,22 @@ describe('editors integration', function () {
       expect(JSON.parse(result.stdout)).toEqual(expectedOutputData);
       expect(result.stderr).toBeFalsy();
     });
+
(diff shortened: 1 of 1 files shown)


dcyriller:
Do you think we should? It wouldn’t be that complicated to implement but I’m not sure... I could do more research but for instance eslint doesn’t provide such an option.

rwjblue:
So eslint only allows fixers to write to disk?

dcyriller:
Yes, eslint only allows fixers to write to disk.

There is this project that would be a faster wrapper of eslint: `eslint_d`. This project provides  a `--fix-to-stdout` command: https://www.npmjs.com/package/eslint_d#automatic-fixing

PR #1356, opened 2020-05-26
Add new rule: `no-yield-to-default`

### LINKED ISSUE #1352: Rule-request: no yield to default
```hbs
{{yield ... to="default"}} -> {{yield ...}}
{{has-block "default"}} -> {{has-block}}
{{has-block-params "default"}} -> {{has-block-params}}
{{hasBlock "default"}} -> {{hasBlock}}
{{hasBlockParams "default"}} -> {{hasBlockParams}}
```

rajasegar:
Fixes #1352


docs/rule/no-yield-to-default.md
```diff
@@ -0,0 +1,77 @@
+# no-yield-to-default
```
rwjblue, line 1: Still a bunch of `TODO`s in this file.

lib/rules/no-yield-to-default.js
```diff
@@ -0,0 +1,36 @@
...
+
+const ERROR_MESSAGE = 'You should not yield to default';
+
+const DISALLOWED_PATHS = ['yield', 'has-block', 'has-block-params', 'hasBlock', 'hasBlockParams'];
+
+module.exports = class NoYieldToDefault extends Rule {
+  visitor() {
+    return {
+      MustacheStatement(node) {
+        if (DISALLOWED_PATHS.includes(node.path.original)) {
```
rwjblue, line 13: We can't assume that `.path.original` is present. For example with `{{"foo"}}` there is no `.path.original` (I think its `.path.value`?).

lib/rules/no-yield-to-default.js
```diff
@@ -0,0 +1,36 @@
...
+          let isDefaultHashPair = node.hash.pairs.some((p) => {
+            return p.key === 'to' && p.value.value === 'default';
```
rwjblue, line 15: Checking `.value.value` will error if the hash pair value is not a static string.

test/unit/rules/no-yield-to-default-test.js
```diff
@@ -0,0 +1,72 @@
...
+  name: 'no-yield-to-default',
+
+  config: true,
+
+  good: [
+    '{{yield}}',
+    '{{has-block}}',
+    '{{has-block-params}}',
+    '{{hasBlock}}',
+    '{{hasBlockParams}}',
```
rwjblue, line 17: Can you add a few more "good" scenarios for other expected invocations of these keywords (e.g. passing the expected block names)?

test/unit/rules/no-yield-to-default-test.js
```diff
@@ -0,0 +1,72 @@
...
+  ],
+
+  bad: [
```
rwjblue, line 20: Can you add more bad examples for (permutations of each of the disallowed keywords):

```hbs
{{#if (has-block "default")}}

{{/if}}
```

```hbs
{{if (has-block "default") "some-value"}}
```

rajasegar: done

test/unit/rules/no-yield-to-default-test.js
rajasegar: done

lib/rules/no-yield-to-default.js
rajasegar: But we are looking for `{{yield}}` and `{{has-block}}` nodes right, for them the `path.original` is always there.

lib/rules/no-yield-to-default.js
rajasegar: Which means we should look out for dynamic values for named blocks like `to` ?

docs/rule/no-yield-to-default.md
rajasegar: Removed. Anyway the docs require a better description and references I think.

rwjblue: @chancancode - Can you help with some prose/info here?

chancancode: Something like:

> The `yield` keyword can be used for invoking blocks passed into a component. The `to` named argument specifies which of the blocks to yield too. Specifying `{{yield to="default"}}` is unnecessary, as that is the default behavior. Likewise, `{{has-block}}` and `{{has-block-params}}` also defaults to checking the "default" block.

lib/rules/no-yield-to-default.js
chancancode: That is in fact an error, but I think for the purpose of this rule you can just ignore that cause (don't blow up) and let something else handle it. (IIRC, the glimmer compiler already has a syntax error for that.)

docs/rules.md
```diff
@@ -58,6 +58,7 @@
 * [no-whitespace-for-layout](rule/no-whitespace-for-layout.md)
 * [no-whitespace-within-word](rule/no-whitespace-within-word.md)
 * [no-yield-only](rule/no-yield-only.md)
+* [no-yield-to-default](rule/no-yield-to-default.md)
```
bmish, line 61: This file has been deleted, please rebase on the latest master and run `yarn update` to update the README.

bmish:
@rajasegar is this ready?

rajasegar:
No @bmish not yet, @rwjblue had some concerns regarding the use of `.value.value` I am still wondering we should handle this for dynamic values..

bmish:
@rajasegar I see, well heads up that we will have a major release soon, so it's a good time to get any new rules ready and merged.

lib/rules/no-yield-to-default.js
rwjblue: > But we are looking for {{yield}} and {{has-block}} nodes right, for them the path.original is always there.

This is no longer correct, `node.path` is allowed to be roughly "anything". For example, in Ember 3.25+ this is perfectly valid:

```hbs
{{(call-some-helper-that-returns-a-helper) whatever-here="lol"}}
```

PR #1370, opened 2020-06-05
Update `invocable-blacklist` rule to handle nested angle bracket component invocation

bmish:
Fix so that the [rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/invocable-blacklist.md) handles this test case:

* Config: `['nested/foo-bar']`
* Disallows:
  ```hbs
  <Nested::FooBar />
  ```

RFC for this syntax: https://github.com/emberjs/rfcs/pull/457

### DIFF
--- a/docs/rule/invocable-blacklist.md
+++ b/docs/rule/invocable-blacklist.md
@@ -7,19 +7,23 @@ Disallow certain components or helpers from being used. Use case is you bring in
 Given a config of:
 
 ```js
-'invocable-blacklist': ['foo']
+'invocable-blacklist': ['foo-bar']
 ```
 
 This rule **forbids** the following:
 
 ```hbs
-{{foo}}
+{{foo-bar}}
 ```
 
 ```hbs
-{{#foo}}{{/foo}}
+{{#foo-bar}}{{/foo-bar}}
+```
+
+```hbs
+<FooBar />
 ```
 
 ## Configuration
 
-* array of strings - helpers or components to blacklist
+* array of strings - helpers or components to blacklist (using kebab-case names like `nested-scope/component-name`)
--- a/lib/rules/invocable-blacklist.js
+++ b/lib/rules/invocable-blacklist.js
@@ -13,7 +13,14 @@ module.exports = class InvocableBlacklist extends Rule {
         }
         break;
       case 'object':
-        if (Array.isArray(config) && config.length > 0) {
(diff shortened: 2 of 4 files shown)


docs/rule/invocable-blacklist.md
```diff
@@ -7,19 +7,23 @@ Disallow certain components or helpers from being used. Use case is you bring in
...
+```
+
+```hbs
+<FooBar />
 ```
 
 ## Configuration
 
-* array of strings - helpers or components to blacklist
+* array of strings - helpers or components to blacklist (kebab-case names)
```
rwjblue, line 29:
```suggestion
* array of strings - helpers or components to blacklist in the resolvers normalized format (generally kebab-cased)
```

lib/rules/invocable-blacklist.js
```diff
@@ -13,7 +13,14 @@ module.exports = class InvocableBlacklist extends Rule {
         }
         break;
       case 'object':
-        if (Array.isArray(config) && config.length > 0) {
+        if (
+          Array.isArray(config) &&
+          config.length > 0 &&
+          config.every(
+            // Ensure names are passed as kebab-case strings.
+            (name) => typeof name === 'string' && name.length > 0 && dasherize(name) === name
```
rwjblue, line 21: Hmm, can we just convert it for them? Why make it error (or did I misunderstand)?

bmish: Are you suggesting allowing a component name to be passed in either of these two formats?

* `some-scope/component-name` (kebab-case)
* `SomeScope::ComponentName` (angle bracket style)

If you prefer, I could switch to that as the requirement. My original concern was that allowing multiple formats could create inconsistency or confusion (some people pass it one way, other people pass it another way).

rwjblue: Yeah, I definitely see what you mean here. My main concern is that _for the most part_ folks will only think about the angle bracket invocation style (at some point in the future), and its going to end up being error prone in configuration.

I'd like to think about that a bit more, there is definitely nothing wrong with the implementation here. We can follow up and change things to allow either formats later if it turns out to be confusing.

bmish: Sounds good. I'll leave as is for now, but definitely open to adjusting.

PR #1372, opened 2020-06-08
Allow specifying a custom error message in `invocable-blacklist` rule

bmish:
Use case: for a deprecated component, you may want to specify a custom error message that suggests the replacement.

Example config:

```
[
  {
    names: ['component-with-custom-message'],
    message: 'Reason why not to use this component and good alternative.'
  }
]
```

### DIFF
--- a/docs/rule/invocable-blacklist.md
+++ b/docs/rule/invocable-blacklist.md
@@ -1,6 +1,11 @@
 # invocable-blacklist
 
-Disallow certain components or helpers from being used. Use case is you bring in some addon like ember-composable-helpers, but your team deems one or many of the helpers not suitable and wants to guard against their usage.
+Disallow certain components or helpers from being used.
+
+Use cases include:
+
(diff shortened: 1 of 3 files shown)


docs/rule/invocable-blacklist.md
```diff
@@ -26,4 +26,9 @@ This rule **forbids** the following:
 
 ## Configuration
 
+One of these:
+
 * array of strings - helpers or components to blacklist (using kebab-case names like `nested-scope/component-name`)
+* array of objects with the following keys:
+  * `name` - helpers or components to blacklist (using kebab-case names like `nested-scope/component-name`)
```
rwjblue, line 33: Should we allow an array too (so that you could share a single message for a list of names)? I’m thinking that might seem similar to the overrides stuff :thinking:

Also, seems fine to punt for now.

bmish: Good catch, that was actually my original intention, so I have updated each object to take an array of names.

PR #1381, opened 2020-06-15
Save `seen ids` as Set; add Mustache + Concat

josephdsumner:
If merged, this PR will:
- change the data structure that tracks the `already seen` list of `id` values from an `Array` to a `Set`
- expand coverage of evaluated attribute names to include `@id`
- expand coverage of evaluated attribute types to include `MustacheStatement` and `ConcatStatement`

Although the net being cast is larger with `MustacheStatement` and `ConcatStatement` included, the leniency of the rule's logic is intended to stay the same as before (when only `TextNode` nodes were checked). Specifically, an attribute value must be made up entirely of StringLiteral components in order to be added to the `already seen` Set. In cases where at least one component of the attribute value is dynamic/computed/evaluated, that part is assumed to provide whatever 'uniqueness factor' is required for that `id`.

```hbs
{{!-- Reference --}}
<div id="id-00"></div>
{{!-- These would each get marked as duplicates --}}
<div id="id-00"></div>
<div id={{"id-00"}}></div>
<div id={{"id"}}{{"-"}}{{"00"}}></div>
<div id="{{"id"}}-00"></div>
{{!-- These would not be marked as duplicates --}}
<div id={{id-00}}></div>
<div id="{{id}}-00"></div>
<div id={{id}}{{"-"}}{{"00"}}></div>
```

### DIFF
--- a/lib/rules/no-duplicate-id.js
+++ b/lib/rules/no-duplicate-id.js
@@ -4,67 +4,97 @@ const Rule = require('./base');
 
 const ERROR_MESSAGE = 'ID attribute values must be unique';
 
-function isValidIdAttrNode(attrNode) {
-  let isIdAttrTag = ['id', '@id'].includes(attrNode.name);
-  let isValidAttrNodeType = ['TextNode','MustacheStatement','ConcatStatement'].includes(attrNode.value.type);
-  return attrNode && isIdAttrTag && isValidAttrNodeType;
-};
+// Utility functions
+// -----------------
+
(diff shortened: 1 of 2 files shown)


josephdsumner:
@MelSumner 
> We might need to change a line or two of this, but it's good enough to be merged into this branch. Thank you for doing this, it was helpful! 👍

My pleasure! Updated PR incoming to add support for `BlockStatements` per the base branch PR, will make some style updates there as well 👍

PR #1386, opened 2020-06-22
Add new rule: `require-lang-attribute`

MelSumner:
This rule adds template linting support as related to [RFC 635](https://emberjs.github.io/rfcs/0635-ember-new-lang.html). For the linting rule to be effective, the linter will need to be able to lint `.html` files (added in PR #1232).


docs/rule/require-lang-attribute.md
```diff
@@ -0,0 +1,45 @@
...
+* the application's content is in a language that is not English
+* an end-user with a screen reader turned on, whose operating system (OS) is set to a different language, navigates to that page with their screen reader turned on
+* the screen reader would attempt to read the page in the language that is defined by the lang attribute on the page, but the supporting element information ("button", "link", etc) is read out in the language that is set by the operating system.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<html>
```
Turbo87, line 26: what if the `lang` attribute is set dynamically, e.g. by ember-intl?

MelSumner: the lang attribute could be _changed_ dynamically but it should still be set with an initial value, I think. Otherwise you could still run into issues with legal conformance requirements.

test/unit/rules/require-lang-attribute-test.js
```diff
@@ -0,0 +1,25 @@
...
+const ERROR_MESSAGE = require('../../../lib/rules/require-lang-attribute').ERROR_MESSAGE;
+
+generateRuleTests({
+  name: 'require-lang-attribute',
+
+  config: true,
+
+  good: ['<html lang="en"></html>', '<html lang="en-US"></html>'],
+
+  bad: [
```
rwjblue, line 14: Can you add a `bad` example for:

```hbs
<html lang="">
</html>
```

lib/rules/require-lang-attribute.js
```diff
@@ -0,0 +1,25 @@
...
+const Rule = require('./base');
+const AstNodeInfo = require('../helpers/ast-node-info');
+
+const ERROR_MESSAGE = 'The `<html>` element must have a `lang` attribute';
+
+module.exports = class RequireLangAttribute extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        if (node.tag === 'html' && !AstNodeInfo.hasAttribute(node, 'lang')) {
```
rwjblue, line 12: This should probably also check to see that the value isn't a literal empty string.

docs/rule/require-lang-attribute.md
```diff
@@ -0,0 +1,45 @@
...
+
+This rule **allows** the following:
+
+```hbs
+<html lang="en">
+```
+
+```hbs
+<html lang="en-US">
+```
```
rwjblue, line 41: I'd like to see this added to allowed docs:

```hbs
<html lang={{someThingHere}}>
```

lib/rules/require-lang-attribute.js
MelSumner: updated to include a check for this (and added a test)

test/unit/rules/require-lang-attribute-test.js
MelSumner: done 👍

MelSumner:
@rwjblue @bmish this one is ready for review/merge, I think.

docs/rules.md
```diff
@@ -0,0 +1,78 @@
+# Rules
```
bmish, line 1: This file was deleted and should be removed.

test/unit/rules/require-lang-attribute-test.js
```diff
@@ -0,0 +1,34 @@
...
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+const ERROR_MESSAGE = require('../../../lib/rules/require-lang-attribute').ERROR_MESSAGE;
+
+generateRuleTests({
+  name: 'require-lang-attribute',
+
+  config: true,
+
+  good: ['<html lang="en"></html>', '<html lang="en-US"></html>'],
```
bmish, line 12: Can you add this test case?

```
<html lang={{lang}}></html>
```

docs/rule/require-lang-attribute.md
```diff
@@ -0,0 +1,52 @@
...
+This rule **forbids** the following:
+
+```hbs
+<html></html>
+```
+
+```hbs
+<html lang=""></html>
+
+This rule **allows** the following:
```
bmish, line 32: Formatting isn't correct here.

docs/rule/require-lang-attribute.md
```diff
@@ -0,0 +1,52 @@
...
+
+This rule's objective is to ensure that Ember applications achieve [WCAG Success Criterion 3.1.1: Language of Page](https://www.w3.org/WAI/WCAG21/Understanding/language-of-page.html). The state of the `lang` attribute has a usability impact on the experience of users that require screen-reading assistive technology. When the attribute is properly assigned:
+
+> "Both assistive technologies and conventional user agents can render text more accurately when the language of the Web page is identified. Screen readers can load the correct pronunciation rules. Visual browsers can display characters and scripts correctly. Media players can show captions correctly. **As a result, users with disabilities will be better able to understand the content.**"
+>
+> **Source: [WCAG Success Criterion 3.1.1: Intent](https://www.w3.org/WAI/WCAG21/Understanding/language-of-page.html#intent)**
+
+When the language of the page cannot be identified, the integrity of the above information cannot be guaranteed.
+
+Consider the following use case:
```
rwjblue, line 13: Are these bullets listing multiple use cases or a single use case? Basically, should I be reading this list as if AND'ed or OR'ed?

docs/rule/require-lang-attribute.md
```diff
@@ -0,0 +1,53 @@
...
```
MelSumner, line 13:
```suggestion
Consider any of the following use cases:
```

lib/rules/require-lang-attribute.js
```diff
@@ -0,0 +1,36 @@
...
+          this.log({
+            message: ERROR_MESSAGE,
+            line: node.loc && node.loc.start.line,
+            column: node.loc && node.loc.start.column,
+            source: this.sourceForNode(node),
+          });
+        }
+        const hasLangAttribute = AstNodeInfo.hasAttribute(node, 'lang');
+        const langAttributeValue = AstNodeInfo.elementAttributeValue(node, 'lang');
```
rwjblue, line 21: `AstNodeInfo.elementAttributeValue` 😭 

- This will need to be added to #1388's TODOs

PR #1387, opened 2020-06-22
Improve message for `no-implicit-this` to mention using angle bracket invocation

chriskrycho:
The existing message only suggested adding a no-argument helper or component, but in many cases, the best move for a developer may be to switch to an angle-bracket invocation. So: "Why not have both?"


rwjblue:
Thank you!

Issue #1358, opened 2020-05-26
Remove automatic TextNode unwrapping in `AstNodeInfo.elementAttributeValue`


rwjblue:
The fact that `AstNodeInfo.elementAttributeValue` automatically unwraps `TextNode` values makes it **very very** easy to miss the fact that it _might_ return a plain string **OR** an AST node. This leads to common errors where folks treat the value like a string, and then assume they can use string methods on the value (since all of the tests added _happen_ to not use a dynamic value).

Examples issues:

* https://github.com/ember-template-lint/ember-template-lint/issues/1286
* https://github.com/ember-template-lint/ember-template-lint/pull/1327

scalvert:
Does #1388 fully cover this issue, @rwjblue?

Issue #1168, opened 2020-02-26
Tidy up ast-node-info element helpers


MelSumner:
The helpers that only check `node.tag` are not helpful because they are providing less functionality than would be expected by someone using the helpers. 

For example, `isInputElement` only checks `node.tag === 'input'` and doesn't include `{{input}}` or `<Input/>`.

- [ ] Determine if the helper can be made more useful
- [ ] If the helper cannot be made more useful, remove it.
- [ ] Update the locations where the helper has been used, to not use the helper anymore
- [ ] Ensure all tests pass

PR #1388, opened 2020-06-23
Remove automatic `TextNode` unwrapping by `attributeValue` functions in `ast-node-info` helper

josephdsumner:
@MelSumner @rwjblue 

### Motivation / Synopsis:
I have seen a lot of discussion about the desire to update this helper, but I haven't found anything that includes a proposed implementation or detailed design. I'm interested in working on this, so if there is a spec, please let me know.

This PR aims to partially resolve a series of internal issues related to the current structure of the `ast-node-info` helper:

- Primary motivation: progress on #1168
- Directly addresses: issue summarized in #1358
- Known side-effects: partially unblocks #1327

Specifically, it aims to address the divergent return behavior of the `attributeValue` function, and by extension, those of `elementAttributeValue` and `hasAttributeValue`. The current implementation returns a `String` for a `TextNode`, but an `AST Node` otherwise. The auto-unwrap feature is a convenient abstraction for large number of use cases, because Strings/Literals are very common (if not the most common) type of attribute value encountered. However, it leads to substantial amounts of in-rule code duplication for rules that compare sets of attributes (e.g. `no-duplicate-id`, `require-input-label`, `no-duplicate-landmark-label`).

### Summary of Changes:
If merged, this PR (updated per review / to be in alignment with similar breaking changes anticipated as part of v3.0): 
- ~~removes the automatic `TextNode` unwrapping feature from the `attributeValue` family of functions~~
- ~~provides the option for unwrapping `TextNodes` **only** with a new, optional `boolean` parameter called `unwrapTextNode` (defaults to `false`)~~
- ~~adds relevant unit test suites~~
- ~~Updates all instances of affected function calls in the existing ruleset to accommodate the new behavior. This was done with the underlying assumption that the existing rules operate correctly in the 'automatic `TextNode` unwrapping' schema. Those function calls were uniformly updated by setting the `unwrapTextNode` option to `true`, and they may or may not all be relevant.~~
- removes `attributeValue()`, `hasAttributeValue()`, and `elementAttributeValue()` functions from `ast-node-info`
- migrates existing usages of these functions to using `findAttribute()` or the recently added `node-matcher`, as appropriate

### Migration Schema for Existing Rules:
For existing rules, occurrences of the `attributeValue()` family functions are primarily migrated **under the assumption that the (now-removed) automatic `TextNode` unwrapping behavior remains the desired behavior for these rules**.  The migration schema implemented here is considered to be a functional MVP that prioritizes resolving the troublesome behavior / existing issues as opposed to a complete overhaul of `ast-node-info`, feature additions to `node-matcher`, or an integration of the two helpers.

Specifically, the changes herein primarily take the form of a direct 'like-for-like' migration from `elementAttributeValue()` or `hasAttributeValue()` to `findAttribute()`. The primary drawback with this 'direct' approach is that there is an increase in the amount of in-rule code. This is because the rule now performs `TextNode` unwrapping (and associated assertions / pre-requisite checks) by itself. 

By (explicit) example, the following:

```js
const roleAttrValue = AstNodeInfo.elementAttributeValue(node, 'role');
```

would become:

```js
// Assert presence of `role` attribute
let hasRoleAttr = AstNodeInfo.hasAttribute(node, 'role');

let roleAttr, roleAttrValue;

if (hasRoleAttr) {

  // Get `AttrNode` for the `role` attribute
  roleAttr = AstNodeInfo.findAttribute(node, 'role');

  // Assign value of `role` attribute, preserving the existing behavior:
  // - check first for a `TextNode` and unrwap it automatically
  // - otherwise, return the attribute `value` node (some other `AST` type)
  if (roleAttr.value.type === 'TextNode') {
    roleAttrValue = roleAttr.value.chars;
  } else {
    roleAttrValue = roleAttr.value;
  }
}
```

### DIFF
--- a/lib/rules/no-redundant-landmark-role.js
+++ b/lib/rules/no-redundant-landmark-role.js
@@ -56,7 +56,12 @@ module.exports = class NoRedundantLandmarkRole extends Rule {
           return;
         }
 
-        const roleValue = AstNodeInfo.elementAttributeValue(node, 'role') || '';
+        let roleAttrNode = AstNodeInfo.findAttribute(node, 'role');
+        let roleValue;
+        if (roleAttrNode.value.type === 'TextNode') {
+          roleValue = roleAttrNode.value.chars || '';
+        }
+
         const invalidLandmark = LANDMARK_ROLES.find((e) => {
           return e.name === node.tag && e.role === roleValue && !e.allow;
         });
--- a/lib/rules/require-valid-alt-text.js
+++ b/lib/rules/require-valid-alt-text.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const AstNodeInfo = require('../helpers/ast-node-info');
+const NodeMatcher = require('../helpers/node-matcher');
 const Rule = require('./base');
 
 function hasAccessibleChild(node) {
@@ -29,8 +30,12 @@ module.exports = class RequireValidAltText extends Rule {
           return;
         }
 
-        if (AstNodeInfo.hasAttributeValue(node, 'aria-hidden', 'true')) {
-          return;
+        let hasAriaHiddenAttr = AstNodeInfo.hasAttribute(node, 'aria-hidden');
+        if (hasAriaHiddenAttr) {
+          let ariaHiddenAttr = AstNodeInfo.findAttribute(node, 'aria-hidden');
+          if (ariaHiddenAttr.value.type === 'TextNode' && ariaHiddenAttr.value.chars === 'true') {
+            return;
+          }
         }
 
         if (AstNodeInfo.hasAttribute(node, '...attributes')) {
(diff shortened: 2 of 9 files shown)


rwjblue:
Thanks for writing this up @josephdsumner, the PR description was great / well written, thank you!

I'm a bit worried that these changes are _essentially_ a breaking change (because external rule authors use these APIs as well as our internal rules that are updated here).

I'd like to propose a modified direction (to avoid the breaking change, but still mitigate the issues):

* Deprecate the following APIs:
  * `elementAttributeValue`
  * `hasAttributeValue`
  * `attributeValue`
* Migrate to the uses of these methods into using `findAttribute` and properly handling the various types of value nodes that may exist.

This avoids the breakage (by not changing those troublesome APIs), but also guides rule authors (both internal rules and external ones) down a path that is less likely to completely break them.

What do you think?

rwjblue:
@josephdsumner - If you have the time to tackle this before we do 3.0.0 (next couple weeks probably), we could remove the methods completely (instead of adding an additional argument) and change this PR to target `next`...

josephdsumner:
@MelSumner @rwjblue @Turbo87 @bmish 
I have un-drafted this PR and am requesting a re-review. The initial PR description has been updated to reflect the change in direction recommended during review.

lib/rules/no-redundant-landmark-role.js
```diff
@@ -56,14 +56,19 @@ module.exports = class NoRedundantLandmarkRole extends Rule {
...
+        }
+
         const invalidLandmark = LANDMARK_ROLES.find((e) => {
           return e.name === node.tag && e.role === roleValue && !e.allow;
         });
 
         if (invalidLandmark) {
           if (this.mode === 'fix') {
-            let roleAttrNode = AstNodeInfo.findAttribute(node, 'role');
+            // let roleAttrNode = AstNodeInfo.findAttribute(node, 'role');
```
MelSumner, line 71:
```suggestion
```

lib/rules/require-valid-alt-text.js
```diff
@@ -126,7 +157,10 @@ module.exports = class RequireValidAltText extends Rule {
             }
           }
         } else if (isInput) {
-          const isImageInput = AstNodeInfo.hasAttributeValue(node, 'type', 'image');
+          const isImageInput = NodeMatcher.match(node, {
+            attributes: [{ name: 'type', value: { type: 'TextNode', chars: 'image' } }],
+          });
+          // const isImageInput = AstNodeInfo.hasAttributeValue(node, 'type', 'image');
```
MelSumner, line 163:
```suggestion
```

lib/rules/no-redundant-landmark-role.js
rwjblue, line 71: ✂️

rwjblue (approved):
looks good to me, only one small tweak!

lib/rules/require-valid-alt-text.js
josephdsumner: 👍 thank you **so much** for catching that!

rwjblue (approved):

lib/rules/no-redundant-landmark-role.js
josephdsumner: 👍 awesome thanks!

josephdsumner:
@MelSumner @rwjblue 
👍  thanks so much for the reviews! Consolidated the changes -- pretty sure this will be ready to go soon!

PR #1389, opened 2020-06-24
Rename `invocable-blacklist` rule to `no-restricted-invocations` (backwards-compatible)

bmish:
This is done in a backwards-compatible way using our `DEPRECATED_RULES` map to allow the old rule name to continue to be used for now.

The new name is more aligned with the naming ESLint itself uses for many of its [rules](https://eslint.org/docs/rules/) (`no-restricted-imports`, `no-restricted-properties`, `no-restricted-syntax`, etc).

CC: @mongoose700


lib/helpers/deprecated-rules.js
```diff
@@ -4,6 +4,7 @@ let DEPRECATED_RULES = new Map([
   // deprecated in the 1.x cycle, can be removed in 3.x
```
mongoose700, line 4: Since we're on version 2.8, should we add a new comment saying `// deprecated in the 2.x cycle, can be removed in 4.x` before the new addition?

bmish: Added a note.

Issue #1371, opened 2020-06-06
Allow rules to specify and validate their options using JSON Schema 


bmish:
This would match ESLint behavior: https://eslint.org/docs/developer-guide/working-with-rules#options-schemas

This would likely be in a `meta.schema` property on each rule, the same as eslint, and using the same `meta` object that was under consideration in #943. 

There are many benefits:
* No more need to write code inside rules to validate their schemas (currently done today by some of our rules in a `parseConfig()` function)
* Less likely to have bad data passed to rules or rules used incorrectly
* More clear exactly what rule options are supposed to look like

rwjblue:
Seems like a good idea to me!

IMHO, this doesn't really remove the need for `parseConfig` conceptually, but it would make it so that it doesn't _also_ need to validate. Specifically, it is still very useful to be able to normalize any number of valid user-land configurations into a single format that you can rely on internally in the rule (thus avoiding repeated guards or branching logic for different types of user provided config).

PR #1392, opened 2020-06-28
Add configuration options for allowing components as table children in `table-groups` rule

mongoose700:
If you have a custom component with a tag name, you currently need to do `<table {{! template-lint-disable table-groups }}>`. With this change, you can now add those components in the configuration instead.

I also accidentally fixed a bug with how ordering was enforced. It was resetting the `currentAllowedIndex` back to `-1` whenever it encountered a `TextNode`, which includes whitespace, which allowed one of the `good` examples to pass even though it shouldn't have.


docs/rule/table-groups.md
```diff
@@ -54,11 +54,29 @@ This rule **allows** the following:
...
 </table>
 ```
+
+## Configuration
+
+One of these:
+
+* boolean - `true` - Enables the rule to be enforced with no extra allowed components as children of table components.
+* object[] - with the following keys:
+  * `allowed-caption-components` - string[] - components to treat as having the caption tag (using kebab-case names like `nested-scope/component-name`)
```
bmish, line 78: What is the benefit of having 5 options (one of each table group) instead of a single string array `allowed-components` option? Hard for me to tell why it's necessary, and it seems like it just adds extra options for us to maintain.

lib/rules/table-groups.js
```diff
@@ -2,6 +2,13 @@
 
 const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
+const createErrorMessage = require('../helpers/create-error-message');
+const dasherize = require('../helpers/dasherize-component-name');
+
+function isValidComponentName(str) {
```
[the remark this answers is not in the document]
mongoose700: Sure. I think I'll name it `isValidDasherizedComponentOrHelperName`.

test/unit/rules/table-groups-test.js
```diff
@@ -405,5 +471,149 @@ generateRuleTests({
...
+      </table>`,
+        line: 2,
+        column: 6,
+      },
+    },
+  ],
+
+  error: [
+    {
+      // If none of the keys are recognized, they should just use `true`.
```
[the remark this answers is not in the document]
mongoose700: By "they should just use `true`", I mean that we're saying the configuration is invalid (and we throw an error), and they should update their configuration to be `: true`, not that the rule should carry forward with it being `true`. It seems we don't typically make it an error when there are extra ignored keys in the configuration, which I find odd.

docs/rule/table-groups.md
mongoose700: If we only have `allowed-components`, then we don't get the benefits of enforcing ordering. We wouldn't be able to tell that
```
<table>
  <MyTfoot />
  <MyThead />
</table>
```
is invalid.

bmish: I see. The rule documentation doesn't mention anything about enforcing order so I didn't realize it did that. Can you update the doc to mention the ordering enforcement?

test/unit/rules/table-groups-test.js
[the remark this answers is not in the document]
bmish: Got it. We should definitely be stricter about disallowing invalid configs. I filed a ticket https://github.com/ember-template-lint/ember-template-lint/issues/1371 about using JSON Schema to make this easier for us (like ESLint does).

lib/utils/javascript.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+module.exports = {
+  duplicateArrays,
```
bmish, line 4: Can we only copy `flatMap` and `flat` (rename `flatten` to `flat`) for now since those are basically just polyfills of native JavaScript functions? The other utils are custom and would just become deadcode anyway.

lib/utils/javascript.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
bmish, line 1: This file should be under a `helpers` folder since this package calls them helpers not `utils`.

mongoose700:
> > I also accidentally fixed a bug with how ordering was enforced. It was resetting the currentAllowedIndex back to -1 whenever it encountered a TextNode, which includes whitespace, which allowed one of the good examples to pass even though it shouldn't have.
> 
> Would it be easy to extract this into a separate PR for us to merge first? This is a pretty important fix (since ordering enforcement was basically not working before it). Separating it out could help make it more clear that this fix is being made. If you are able to separate it out, I'll merge it right away.

Sure, I can make a separate PR for it. For this PR, should I squash my commits?

bmish:
No need to squash commits.

mongoose700:
Created separate PR here: https://github.com/ember-template-lint/ember-template-lint/pull/1393

lib/helpers/is-dasherized-component-or-helper-name.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const dasherizeComponentName = require('../helpers/dasherize-component-name');
+
+module.exports = function (name) {
+  return typeof name === 'string' && name.length > 0 && dasherizeComponentName(name) === name;
+};
```
bmish, line 7: Can you update the `no-restricted-inocations` rule to use this helper?

PR #1393, opened 2020-06-28
Fix table groups ordering enforcement when whitespace is present in `table-groups` rule

mongoose700:
Previously encountering whitespace would incorrectly reset the `currentAllowedIndex` back to `-1`.


bmish:
Can you update the rule doc too to mention the rule ordering enforcement?

PR #1394, opened 2020-06-28
Add autogenerated rule table for README

### LINKED ISSUE #439: Add eslint plugin esque documentation grid to README.
* list all rules
* separate stylistic from rules from safety/error related rules
* show which are in the recommended config

bmish:
Before this change, it was difficult to find the list of rules in this package, and it was also hard to find what config/preset each rule was a part of. Those are two of the most important pieces of information when visiting a linting plugin's homepage.

After this change, there is a new autogenerated table in README.md with the complete list of rules and emojis to indicate what config/preset each rule is part of.

`yarn update` will automatically update the table (this script is based on a script from eslint-plugin-ember). I have also added a CI test to ensure this script is always run when changes are made.

CC: @mongoose700 

Fixes #439.


scripts/update-rules.js
```diff
@@ -0,0 +1,42 @@
...
+const STYLISTIC = ':dress:';
+
+// Generate rule table contents.
+let rulesTableContent = Object.keys(root)
+  .map((name) => {
+    // Check which configs this rule is part of.
+    const isRecommended = Object.prototype.hasOwnProperty.call(recommendedRulesFile.rules, name);
+    const isOctane = Object.prototype.hasOwnProperty.call(octaneRulesFile.rules, name);
+    const isStylistic = Object.prototype.hasOwnProperty.call(stylisticRulesFile.rules, name);
+    const mark = `${isRecommended ? STAR : ''}${isOctane ? OCTANE : ''}${
```
mongoose700, line 26: I think `[isRecommended ? STAR : '', ...].join('')` would look nicer than this (especially with the way the stylistic ones are on a separate line).

scripts/update-rules.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const recommendedRulesFile = require('../lib/config/recommended');
+const octaneRulesFile = require('../lib/config/octane');
+const stylisticRulesFile = require('../lib/config/stylistic');
+const root = require('../lib/rules');
```
mongoose700, line 8: Could this be `rules` instead of `root`?

scripts/update-rules.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return `| ${mark} | ${link} |`;
+  })
+  .join('\n');
+
+fs.writeFileSync(
+  pathReadme,
+  readmeContent.replace(
+    tablePlaceholder,
+    `<!--RULES_TABLE_START-->\n\n|    | Rule ID |\n|:---|:--------|\n${rulesTableContent}\n<!--RULES_TABLE_END-->`
```
mongoose700, line 40: If we have two `\n` after the table start, we should also have two `\n` before the table end.

scripts/update-rules.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const recommendedRulesFile = require('../lib/config/recommended');
```
mongoose700, line 5: nit: You could do `const { rules: recommendedRules } =` here.

scripts/update-rules.js
bmish: Fixed!

scripts/update-rules.js
bmish: Agreed, much better. Fixed.

scripts/update-rules.js
bmish: Fixed.

scripts/update-rules.js
bmish: Good idea, fixed.

scripts/update-rules.js
```diff
@@ -0,0 +1,44 @@
...
+const readmeContent = fs.readFileSync(pathReadme, 'utf8');
+const tablePlaceholder = /<!--RULES_TABLE_START-->[\S\s]*<!--RULES_TABLE_END-->/;
+
+// Config/preset emojis.
+const EMOJI_STAR = ':white_check_mark:';
+const EMOJI_OCTANE = ':car:';
+const EMOJI_STYLISTIC = ':dress:';
+
+// Generate rule table contents.
+let rulesTableContent = Object.keys(rules)
```
mongoose700, line 20: Is there a reason this isn't `const` (and that it isn't being enforced)?

bmish: Fixed. I also prefer to use const. But here's why it's not enforced inside this package: https://github.com/ember-template-lint/ember-template-lint/pull/830#issuecomment-534148039

rwjblue:
Thank you @bmish!

PR #1400, opened 2020-06-29
Refactor `no-restricted-invocations` rule to significantly improve performance

bmish:
I have a large application with thousands of files. I use this rule to disallow invoking thousands of components.

This refactor results in an **85% decrease in running time** of this rule for my large application. The refactor switches the code from making many complex function calls inside a loop to instead just checking if the denylist arrays includes the component name.

| | Running Time |
| --- | --- |
| Before Refactor | 3 min 7 sec | 
| After Refactor | 27 sec |

CC: @mongoose700

### DIFF
--- a/lib/rules/no-restricted-invocations.js
+++ b/lib/rules/no-restricted-invocations.js
@@ -57,13 +57,22 @@ module.exports = class NoRestrictedInvocations extends Rule {
     let checkDenylist = (node) => {
       let denylist = this.config;
 
+      const name = this._getComponentOrHelperName(node);
+      if (!name) {
+        return;
+      }
+
+      const nameToDisplay = node.type === 'ElementNode' ? `<${node.tag} />` : `{{${name}}}`;
+
       for (let denylistItem of denylist) {
         if (typeof denylistItem === 'object') {
-          for (let name of denylistItem.names) {
-            this._checkNode(node, name, denylistItem.message);
+          if (denylistItem.names.includes(name)) {
+            this._logNode(node, nameToDisplay, denylistItem.message);
           }
         } else {
-          this._checkNode(node, denylistItem);
+          if (denylistItem === name) {
+            this._logNode(node, nameToDisplay, denylistItem.message);
+          }
         }
       }
     };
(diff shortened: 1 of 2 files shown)


lib/rules/no-restricted-invocations.js
```diff
@@ -57,13 +57,22 @@ module.exports = class NoRestrictedInvocations extends Rule {
...
       let denylist = this.config;
 
+      const name = this._getComponentOrHelperName(node);
+      if (!name) {
+        return;
+      }
+
+      const nameToDisplay = node.type === 'ElementNode' ? `<${node.tag} />` : `{{${name}}}`;
+
       for (let denylistItem of denylist) {
```
mongoose700, line 67: We could also get rid of this outer for loop as well if we changed `parseConfig` to return a `Map` mapping invalid component/helper names to error messages instead of leaving it as a list.

lib/rules/no-restricted-invocations.js
```diff
@@ -76,18 +85,19 @@ module.exports = class NoRestrictedInvocations extends Rule {
...
-      if (dasherize(node.tag) === name || node.tag === name) {
-        this._logNode(node, `<${node.tag} />`, message);
-      }
+      // Convert from angle-bracket naming to kebab-case.
+      return dasherize(node.tag);
     } else {
-      if (node.path.original === name || checkForComponentHelper(node, name)) {
-        this._logNode(node, `{{${name}}}`, message);
+      if (node.path.original === 'component' && node.params[0]) {
+        return node.params[0].original;
```
mongoose700, line 98: Not really relevant to these changes, but it would be easy to also check for `(component (component ('disallowed-component')))` which could be used as a (terrible) workaround for this lint rule by making this a `while` loop.

lib/rules/no-restricted-invocations.js
bmish: I actually tested a similar change of converting the `names` array to a set inside `parseConfig`, and that reduced the running time by 1-2 seconds for my very large application, but I decided against making that change for now due to the slight added complexity and negligible performance time benefit (theoretically, it could also cause a slight increase in memory usage). It would be neat if there was a way to pass the config into the rule as a set/map in the first place so we wouldn't have to convert it for every single file.

My experimental change:
```js
return config.map((item) => {
    if (typeof item === 'object') {
        return { names: new Set(item.names), message: item.message };
    } else {
        return item;
    }
});
```

mongoose700: It would increase the complexity of `parseConfig`, but I think that would be more than offset by decreasing the complexity of `checkDenylist`.

Issue #1401, opened 2020-06-30
Plugin tests crash with ember-template-lint@2.9.0


andreyfel:
Test looks like:
```
'use strict';

const generateRuleTests = require('ember-template-lint/lib/helpers/rule-test-harness');
const plugin = require('../../../lib');

generateRuleTests({
  plugins: [plugin],
...
```

The exception is the following:
```
TypeError: configPlugins is not iterable

      at processPlugins (../node_modules/ember-template-lint/lib/get-config.js:161:24)
      at ../node_modules/ember-template-lint/lib/get-config.js:197:32
      at forEachPluginConfiguration (../node_modules/ember-template-lint/lib/get-config.js:98:7)
      at processPlugins (../node_modules/ember-template-lint/lib/get-config.js:195:3)
      at getProjectConfig (../node_modules/ember-template-lint/lib/get-config.js:493:5)
      at Linter.loadConfig (../node_modules/ember-template-lint/lib/linter.js:47:19)
      at new Linter (../node_modules/ember-template-lint/lib/linter.js:39:10)
      at Object.<anonymous> (../node_modules/ember-template-lint/lib/helpers/rule-test-harness.js:141:16)
```

It works with ember-template-lint@2.8.0

andreyfel:
I think it is related to this change:
https://github.com/ember-template-lint/ember-template-lint/commit/0d8ad50336821ef8513a60bcb8efce2f4a77b675#diff-167aef5ef69408a64527334a4ce07ab5
This: `for (let i = 0; i < configPlugins.length; i++) {` changed to `for (let plugin of configPlugins) {`

bmish:
That's strange. From your code, I see that `plugins` is an array, so it should be iterable. If possible, could you add a print statement right in front of this for-of loop so we can see what `configPlugins` looks like?

```js
console.log(`configPlugins: ${JSON.stringify(configPlugins)}`);
```

I tested my own custom plugin rule with v2.9.0 and I don't encounter a crash. I have the same code with `generateRuleTests` as you.

amk221:
Morning 👋 
also seeing this today

amk221:
@bmish
```
configPlugins: ["@zestia/ember-template-lint-plugin"]
configPlugins: undefined
```

PR #1402, opened 2020-07-02
test-harness: Create a linter instance per test

dcyriller:
Instead of mutating the linter instance's config.

Is it okay from a perf standpoint?

This would fix #1401 

I have tested it against https://github.com/ember-template-lint/ember-template-lint-plugin-prettier. The bug is reproduced in CI [there](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/actions/runs/154852105)

Out of time for the moment :)
If someone wants to continue this, feel free.

cc @bmish

PR #1407, opened 2020-07-06
get-config: Do not mutate the configuration's `plugins` key

dcyriller:
This is another take at https://github.com/ember-template-lint/ember-template-lint/pull/1402. It would close https://github.com/ember-template-lint/ember-template-lint/issues/1401.

## Bug cause

While parsing the config, `processPlugins()` function mutates `config.plugins`, but also `config.plugins.configuration.plugins` because of the recursion. Doing so, the `plugins` object gets mutated, which is something we want to avoid. Indeed, the configuration might be used multiple times, as in the test harness (where the bug surfaced).

## Fix description

With this PR, `processPlugins()` would return a `pluginsHash`, instead of mutating `config.plugins` in the function's body.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -196,7 +196,7 @@ function processPlugins(plugins = [], options, checkForCircularReference) {
   return pluginsHash;
 }
 
-function processLoadedRules(config) {
+function processLoadedRules(config, options = {}) {
   let loadedRules;
   if (config.loadedRules) {
     loadedRules = config.loadedRules;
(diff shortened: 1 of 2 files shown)


bmish:
Is it possible to add a test case?

dcyriller:
Sure thing. I have added a first failing test, force-pushed it to have the CI report and pushed the other commits which fix the issue.

rwjblue:
Gonna land this (as I think it is the right direction), we can add the test from #1412 and fix that in a follow up PR...

rwjblue:
Thanks @dcyriller!

PR #1411, opened 2020-07-10
README's rules list: Add  🔧 emoji to fixable rules

dcyriller:
Update `scripts/update-readme.js` to:

- Add 🔧 emoji to fixable rules
- Format README with Prettier

### DIFF
--- a/scripts/update-rules.js
+++ b/scripts/update-rules.js
@@ -2,6 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
+
+const { parse } = require('@babel/parser');
+const { default: traverse } = require('@babel/traverse');
+
 const rules = require('../lib/rules');
 const { rules: recommendedRules } = require('../lib/config/recommended');
 const { rules: octaneRules } = require('../lib/config/octane');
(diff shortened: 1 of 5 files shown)


scripts/update-rules.js
```diff
@@ -15,6 +25,29 @@ const tablePlaceholder = /<!--RULES_TABLE_START-->[\S\s]*<!--RULES_TABLE_END-->/
...
+  const pathRule = path.resolve(__dirname, relativePath);
+  let rule = fs.readFileSync(pathRule, { encoding: 'utf8' });
+
+  let ast = parse(rule, { sourceType: 'module' });
+
+  let isFixable = false;
+
+  traverse(ast, {
+    Identifier(path) {
+      if (path.node.name === 'isFixable') {
```
bmish, line 43: The rule `require-form-method` specifies `isFixable: false` (and the rule is not fixable) so we need to ignore if the value is `false`.

In the future, it would be great if we added a static meta config to rules like eslint rules have, indicating if they are fixable, so we could avoid trying to parse it.

scripts/update-rules.js
```diff
@@ -15,6 +25,29 @@ const tablePlaceholder = /<!--RULES_TABLE_START-->[\S\s]*<!--RULES_TABLE_END-->/
...
+  const pathRule = path.resolve(__dirname, relativePath);
+  let rule = fs.readFileSync(pathRule, { encoding: 'utf8' });
+
+  let ast = parse(rule, { sourceType: 'module' });
+
+  let isFixable = false;
+
+  traverse(ast, {
+    ObjectProperty(path) {
+      if (path.node.key.name === 'isFixable' && path.node.value.value !== false) {
```
bmish, line 43: I would check more types here, especially to avoid crashing it if encounters a `SpreadElement` like `...moreProps`.

```suggestion
      if (
        path.node.key.type === 'Identifier' &&
        path.node.key.name === 'isFixable' &&
        !(path.node.value.type === 'Identifier' && path.node.value.value === false)
      ) {
```

dcyriller:
Thank you!

PR #1413, opened 2020-07-11
Add fixable notice to the rule doc of fixable rules

bmish:
This matches the convention used by ESLint. I made the same change in https://github.com/ember-cli/eslint-plugin-ember/pull/715.

Follow-up to https://github.com/ember-template-lint/ember-template-lint/pull/1411.


dcyriller:
Great addition!

Issue #1425, opened 2020-07-22
extract a more convenient .matches type API


MelSumner:
From this code: 

```js
function isMustacheString(node) {
  return (
    node 
    && node.value.type === 'MustacheStatement' 
    && node.value.path.type === 'StringLiteral'
  );
}
```

It became a little more clear that it would be useful to have a `.matches` `.type` API. 

(Issue based on Based on this comment: https://github.com/ember-template-lint/ember-template-lint/pull/1187#discussion_r444957724)

rwjblue:
I would like to have an API that works like this:

```js
match(node, {
  type: 'MustacheStatement',
  value: {
    path: {
      type: 'StringLiteral',
    }
  }
});
```

Specifically, it avoids having to do gnarly boolean operations in each rule...

josephdsumner:
> I would like to have an API that works like this:
> 
> ```js
> match(node, {
>   type: 'MustacheStatement',
>   value: {
>     path: {
>       type: 'StringLiteral',
>     }
>   }
> });
> ```
> 
> Specifically, it avoids having to do gnarly boolean operations in each rule...

I've gotten started on a WIP for this [here](https://github.com/josephdsumner/ember-template-lint/compare/master...josephdsumner:feature/match-api) with the following structure:
- `lib\helpers\node-matcher`
- `NodeMatcher.match(node, ref)`, where `ref` can be a single reference node to match against OR an array of reference nodes. In the latter case, it returns `true` if the node matches any one of the reference nodes contained in that array.

Probably sensible to make the matcher as a pre-requisite to merging 'no-duplicate-id' #1187 , given how much it comes up.

Turbo87:
I assume we can close this, now that #1430 was merged

PR #1430, opened 2020-07-30
Add `.match` API via new `NodeMatcher` helper

josephdsumner:
Inspired by comments and guidance provided in #1187 and #1425.

If merged, this PR adds a new helper called `node-matcher` at the following location: `lib\helpers\node-matcher.js`

The aim of this helper is to simplify the 'node-checking' procedures used to determine whether or not a visited node has the properties + values requisite for applying the rule's logic. The amount of rule code spent on 'node-checking' is often further increased by the need to support several, rule-relevant AST Node Types. 

The primary usage of this helper is through its `NodeMatcher.match(testNode, ref)` function. It is a wrapper of `jscodeshift.match(testNode, refNode)` -- which returns whether or not the `refNode` is a strict subset of the `testNode`. That wrapper is then extended here to allow the `testNode` to be match-checked against an `Array` of `refNodes` -- returning `true` if any individual `NodeMatcher.match(testNode, refNodes[i])` does.

The following is an example of a before/after `NodeMatcher` refactor. It is taken from a current [WIP] rule: `no-duplicate-id`. A working partially-refactored test spike -- obtained by placing the `feature/no-duplicate-id` branch at the tip of this one -- can be found [here](https://github.com/josephdsumner/ember-template-lint/compare/feature/no-duplicate-ids...josephdsumner:feature/no-duplicate-ids-match-api?expand=1#diff-c48fdcc61be2fe15748c3d5e0b6a1500).

Original Form -- Helpers + Asserting Existence + Specific Traversal
```js
BlockStatement(node) {
  // Helpers
  function isElementIdHashPair(pair) {
    return pair.key === 'elementId';
   }
   function getElementIdHashPair(hashPairs) {
     return hashPairs.find(isElementIdHashPair);
    }
    function isValidElementIdHashPair(hashArg) {
      return hashArg && hashArg.value.type === 'StringLiteral';
    }
  // Validate BlockStatement: elementId="some-StringLiteral"
  let hashPairs = node.hash.pairs;
  let elementIdHashArg = getElementIdHashPair(hashPairs);
  if (!isValidElementIdHashPair(elementIdHashArg)) {
    return;
  }
```
Modified Form -- Match Implementation:
```js
BlockStatement(node) {
  let refPair = { key: 'elementId', value: { type: 'StringLiteral' } };
  let elementIdHashArg = node.hash.pairs.find((testPair) =>
    NodeMatcher.match(testPair, refPair)
  );
  if (!elementIdHashArg) {
    return;
  }
```

### DIFF
--- a/lib/helpers/node-matcher.js
+++ b/lib/helpers/node-matcher.js
@@ -29,6 +29,27 @@ function matchOne(path, filter) {
   return matchNode(path.value, filter);
 }
 
+/**
+ * Utility function that pattern matches a test node against either:
+ * - an individual reference node OR
+ * - an Array of reference nodes
+ *
+ * An individual comparison returns whether or not the reference node is a
+ * strict subset of the test node. Similarly, an Array comparison returns
+ * whether or not any one of its individual reference node elements is a strict
+ * subset of the test node.
+ *
+ * Useful for defining a given rule's 'target nodes' -- that is, the set of
+ * nodes for which it is appropriate to apply the rule's logic. In this context,
+ * the reference(s) can be used as selection criteria that a given visited node
+ * (test node) must satisfy in order to proceed with rule logic execution.
+ *
+ * TODO: Add examples
+ *
+ * @param  {Node} testNode - the node to validate
+ * @param  {Node|Node[]} ref - the reference node(s) to match testNode against
+ * @return {boolean}
+ */
 function match(testNode, ref) {
   if (!testNode || !ref) {
     return;
--- a/lib/helpers/node-matcher.js
+++ b/lib/helpers/node-matcher.js
@@ -29,13 +29,8 @@ function matchOne(path, filter) {
   return matchNode(path.value, filter);
 }
 
-// Wrappers
-function nodeMatchesRef(testNode, refNode) {
-  return testNode && refNode && matchOne(testNode, refNode);
-}
-
 function nodeMatchesValidRef(testNode, refNodes) {
-  return testNode && refNodes && refNodes.some((refNode) => nodeMatchesRef(testNode, refNode));
+  return testNode && refNodes && refNodes.some((refNode) => matchOne(testNode, refNode));
 }
 
 function match(testNode, ref) {
@@ -45,7 +40,7 @@ function match(testNode, ref) {
   if (Array.isArray(ref)) {
     return nodeMatchesValidRef(testNode, ref);
   } else {
-    return nodeMatchesRef(testNode, ref);
+    return matchOne(testNode, ref);
   }
 }
 
--- a/lib/helpers/node-matcher.js
+++ b/lib/helpers/node-matcher.js
@@ -44,11 +44,53 @@ function matchOne(path, filter) {
  * the reference(s) can be used as selection criteria that a given visited node
  * (test node) must satisfy in order to proceed with rule logic execution.
  *
- * TODO: Add examples
- *
  * @param  {Node} testNode - the node to validate
  * @param  {Node|Node[]} ref - the reference node(s) to match testNode against
  * @return {boolean}
+ *
+ * For simple and specific target patterns, a `match` implementation has clearer
+ * and more succinct syntax than its conditional (`if`) counterpart because it
+ * does the following tasks on behalf of the rule:
+ * - asserts the existence of relevant nodes, properties, values, etc.
+ * - compounds the 'AND' logic dictated by a strict multi-comparison matching
+ *
+ * As an example, consider a rule designed to ensure that all `div` elements
+ * with the custom `class` attribute `foo` also have a `role` attribute of
+ * `textbox`. An outline of the rule might look like:
+ *
+ * Example Target:
+ *  `<div class="foo"></div>`
+ *
+ * Example Code Context:
+ *  visitor() {
+ *    return ElementNode(node) {
+ *      if (
+ *        // check node against target here
+ *      ) {
+ *        // execute rule logic against target nodes here
+ *      }
+ *    }
+ *  }
+ *
+ * An implementation using standard AST Node syntax might look like:
+ *      if (
+ *        node.tag === 'div' &&
+ *        node.attributes &&
+ *        node.attributes.find((attributeNode) =>
+ *          attributeNode.name === 'class' &&
+ *          attributeNode.value.chars === 'foo'
+ *        ))
+ *      )
+ *
+ * By comparison, the corresponding `match` implementation might look like:
+ *      if (
+ *        NodeMatcher.match(node, {
+ *          tag: 'div',
+ *          attributes: [ { name: 'class', value: { chars: 'foo' } } ]
+ *        })
+ *      )
+ *
+ * TODO: complex example (multiple supported types of `links`?)
  */
 function match(testNode, ref) {
   if (!testNode || !ref) {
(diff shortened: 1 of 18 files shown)


MelSumner:
@josephdsumner please fix the merge conflicts here. Thank you!

josephdsumner:
> do we really want this additional dependency for that? 🤔

This is a great point, and I don't think we do. I was pointed to `jscodeshift` by @rwjblue  and its implementation in `codemod-cli` and `ember-qunit-codemod` as an example of the target _behavior_ requested in #1187 and #1425. 

If this is a useful api, then I imagine there's probably a much more cost-effective alternative solution out there -- either in the form of our own bespoke match implementation or a smaller package. Opinions and suggestions for either of those would absolutely be welcome.

MelSumner:
> If this is a useful api, then I imagine there's probably a much more cost-effective alternative solution out there -- either in the form of our own bespoke match implementation (example [here](https://github.com/ember-template-lint/ember-template-lint/compare/master...josephdsumner:feature/match-api-direct)), a smaller package, or using something we've already got as a dependency (`ember-template-recast`, etc.). Opinions and suggestions for either of those would absolutely be welcome!

@josephdsumner I think that only adding in what we need may be a more efficient approach. Please update the PR to remove the dependency and add in the match implementation (like the one you linked). Thank you! 👍

josephdsumner:
> @josephdsumner I think that only adding in what we need may be a more efficient approach. Please update the PR to remove the dependency and add in the match implementation (like the one you linked). Thank you! 👍 
 
@MelSumner sounds great! Thanks for the direction on this! I will update the PR branch and re-request reviews when it's ready. 👍

josephdsumner:
@MelSumner think I got this 👍  -- requesting re-review now. Here's a quick summary of the changes:
- swapped dependency implementation to standalone helper
- resolved conflicts
- rebase + squash

lib/helpers/node-matcher.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+// Source: jscodeshift/src/matchNode.js
+const hasOwn = Object.prototype.hasOwnProperty.call.bind(Object.prototype.hasOwnProperty);
+
+function isNode(value) {
+  return typeof value === 'object' && value;
```
rwjblue, line 7: p3/5

```suggestion
  return typeof value === 'object' && value !== null;
```

lib/helpers/node-matcher.js
```diff
@@ -0,0 +1,52 @@
...
+  return matchNode(path.value, filter);
+}
+
+// Wrappers
+function nodeMatchesRef(testNode, refNode) {
+  return testNode && refNode && matchOne(testNode, refNode);
+}
+
+function nodeMatchesValidRef(testNode, refNodes) {
+  return testNode && refNodes && refNodes.some((refNode) => nodeMatchesRef(testNode, refNode));
```
rwjblue, line 38: p4/5

I think we know for sure that `testNode` and `refNodes` are not falsey here, (due to the check in the beginning of `match` for `if (!testNode || !ref) { return; }`), so we can remove the extra defensiveness (and make this method a little bit easier to read).

```suggestion
  return refNodes.some((refNode) => nodeMatchesRef(testNode, refNode));
```

rwjblue: p2/5

That suggestion got me thinking though, maybe we don't even need this method? That `refNodes.some((refNode) => nodeMatchesRef(testNode, refNode))` bit is probably simple enough to include inline in that `Array.isArray(ref)` branch of the `if` in `matches`.

What do you think?

lib/helpers/node-matcher.js
```diff
@@ -0,0 +1,52 @@
...
+  } else {
+    path = { value: path };
+  }
+  return matchNode(path.value, filter);
+}
+
+// Wrappers
+function nodeMatchesRef(testNode, refNode) {
+  return testNode && refNode && matchOne(testNode, refNode);
+}
```
rwjblue, line 35: p3/5

I think we are guaranteed that `testNode`/`refNode` are not falsey, right? Should we drop this func and use `matchOne` throughout?

lib/helpers/node-matcher.js
```diff
@@ -0,0 +1,52 @@
...
+
+function match(testNode, ref) {
+  if (!testNode || !ref) {
+    return;
+  }
+  if (Array.isArray(ref)) {
+    return nodeMatchesValidRef(testNode, ref);
+  } else {
+    return nodeMatchesRef(testNode, ref);
+  }
```
rwjblue, line 49: Thinking we might be able to get away with:

```suggestion
  if (Array.isArray(ref)) {
    return ref.some((refNode) => matchOne(testNode, refNode));
  } else {
    return matchOne(testNode, refNode)
  }
```

What do you think?

lib/helpers/node-matcher.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+module.exports.match = match;
```
rwjblue, line 52: FYI - I had wondered if we need to export this as a `.match` property vs just exporting the match function directly (e.g. `module.exports = match`). I think I've come to the conclusion that this is better, because I want to add better support for "specialized" matchers in the future. I can try to make another issue for that when this lands though (definitely not blocking).

josephdsumner:
> This looks great, thanks for working on it! We should:
> 
> * Fixup the inline comments that I've left
> * Add inline API documentation (e.g. JSDoc style) for this file, with some general usage examples
> * Add documentation in the plugin authoring guides/docs for using this
> * Consider adding to our main export as "public API"

@rwjblue That's great news! I'll ping for a re-review after tackling the four items above. 👍

lib/helpers/node-matcher.js
josephdsumner: So long as the plan is to have a single-export strategy (currently `match`) that does this verification immediately, I see no reason not to get rid of the wrapper for `matchOne` altogether. 👍

josephdsumner: Update: addressed in 8023b15 👍

lib/helpers/node-matcher.js
josephdsumner: I absolutely agree with this approach for consolidating the existing behavior down to `matchOne` as the 'heavy lifting' match function and having `match` absorb both the input existence assertions and the 'check against an `Array` of `refNodes`' control/logic. 

They have been addressed in 8023b15 and a66c952, respectively. 👍

lib/helpers/node-matcher.js
josephdsumner: Absolutely agree. Addressed it directly as part of a66c952 👍

lib/helpers/node-matcher.js
josephdsumner: Absolutely agree -- addressed both parts of that in a66c952 👍

MelSumner:
@josephdsumner the linter is returning two errors, can you fix please?

```
/home/runner/work/ember-template-lint/ember-template-lint/test/unit/helpers/node-matcher-test.js
  3:21  error  There should be at least one empty line between import groups                                     import-helpers/order-imports
  4:19  error  `ember-template-recast` import should occur before import of `../../../lib/helpers/node-matcher`  import-helpers/order-imports
```

josephdsumner:
> @josephdsumner the linter is returning two errors, can you fix please?
> 
> ```
> /home/runner/work/ember-template-lint/ember-template-lint/test/unit/helpers/node-matcher-test.js
>   3:21  error  There should be at least one empty line between import groups                                     import-helpers/order-imports
>   4:19  error  `ember-template-recast` import should occur before import of `../../../lib/helpers/node-matcher`  import-helpers/order-imports
> ```
Yep! Addressed in 78a4af7 👍

rwjblue:
Nice work, thanks @josephdsumner!

PR #1436, opened 2020-08-04
Add note about github actions printer

nadavshatz:
The Github actions behavior is undocumented. I spent many hours trying to debug the JSON printer in CI (Github Actions). 

After reviewing the code I see there's undocumented behavior for a specific printer that is on by default for Github Actions.

I didn't change the default behavior as to not break existing usage, but added the documentation to the Readme for posterity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ Basic `ember-template-lint` executable is provided, allowing for easy use within
 
 > Ensure you wrap all glob patterns in quotes so that it won't be interpreted by the CLI. `yarn ember-template-lint app/templates/**` (this will expand all paths in app/templates) and `yarn ember-template-lint "app/templates/**"` (this will pass the glob to ember-template-lint and not interpret the glob).
 
+**Important** note for those running `ember-template-lint` in Github Actions:
+There is an additional printer always used for Github Actions, if you'd like to disable it set the `DISABLE_GITHUB_ACTIONS_ANNOTATIONS` env var to `true`.
+
 Example usage:
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +60,9 @@ Basic `ember-template-lint` executable is provided, allowing for easy use within
 
 > Ensure you wrap all glob patterns in quotes so that it won't be interpreted by the CLI. `yarn ember-template-lint app/templates/**` (this will expand all paths in app/templates) and `yarn ember-template-lint "app/templates/**"` (this will pass the glob to ember-template-lint and not interpret the glob).
 
+**Important** note for those running `ember-template-lint` in Github Actions:
+There is an additional printer always used for Github Actions, if you'd like to disable it set the `DISABLE_GITHUB_ACTIONS_ANNOTATIONS` env var to `true`.
```
bmish, line 64: Our README is pretty bloated so I think we only need to mention this env variable in one place (in the examples below) and not here.

nadavshatz: I spent a few days debugging this and only found it after digging deep in the code.
The default behavior is unexpected since it overrides the other flags. and works very different than locally.
I didn't wanna change that behavior since it's a major breaking change for those that rely on it.

I do think that this should be mentioned as I added since it's not an expected behavior. I'd consider changing this behavior for the next major release and then I agree it can be reduced in how it's mentioned.

If you insist I'll remove it but I'm telling you as a appreciating user of this library - this was extremely frustrating and time consuming. It isn't even mentioned (in a an easy way to find at least) in issues.

I added this for future people, trying to contribute back and didn't move on once i figured out the issue.

At the end of the day it's your call, let me know.

bmish: Okay, thanks for the context, sounds fine to keep it in both places then.

nadavshatz: Cool! let me know if anything else is needed to get this merged.
Cheers 🍻

Issue #1256, opened 2020-04-07
Rule request: add condition to `require-input-label` rule


MelSumner:
## Context
Currently, the `require-input-label` rule does not check for instances where a valid label has already been provided but another label is associated with the input element through the use of the `aria-labelledby` attribute. 

We could update the `require-input-label` rule or create a separate rule called `no-duplicate-label`. 

## Task
Update the rule so it checks for redundant labels- if the `<input>` element already has an associated `<label>` element, it does not require the `aria-label` or `aria-labelledby` attributes.

### Not okay 
```html
    <label for="input-search">Search</label>
    <input type="text" aria-labelledby="button-search" id="input-search" />
    <button id="button-search" type="button">Search</button>
```

### Okay 
```html
    <input type="text" aria-labelledby="button-search" />
    <button id="button-search" type="button">Search</button>
```

KamiKillertO:
I can have a look at this one

MelSumner:
@KamiKillertO what do you think?

KamiKillertO:
I've had a look but I haven't been able to make it work yet. But after reading the example again I might have an idea 🤔

KamiKillertO:
@MelSumner I've opened a draft PR.

KamiKillertO:
@MelSumner the PR has been merged, should we close this issue too?

MelSumner:
@KamiKillertO good catch! Thank you for doing this work! 👍🏻

PR #1439, opened 2020-08-07
Update `require-input-label` rule to detect duplicate labels

KamiKillertO:
This PR is related to [this issue](https://github.com/ember-template-lint/ember-template-lint/issues/1256) reported by @MelSumner.

The idea is to update the rule `require-input-label` to report errors if an input has multiple labels attached.
The rule will report an error if the input is invoked with more than one attribute that can be used to attach a label.

For example, the following patterns are considered invalid :

```html
<label>
  Input label
  <input type="text" aria-label="Custom label"/>
</label>
```

```html
<input type="text" aria-label="Custom label" aria-labelledby="input-label"/>
```

### DIFF
--- a/lib/rules/require-input-label.js
+++ b/lib/rules/require-input-label.js
@@ -3,7 +3,8 @@
 const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
 
-const ERROR_MESSAGE = 'Input elements require a valid associated label.';
+const ERROR_MESSAGE_NO_LABEL = 'Input elements require a valid associated label.';
+const ERROR_MESSAGE_MULTIPLE_LABEL = 'Input elements should not have multiple valid labels.';
 
 function hasValidLabelParent(path) {
   // Parental validation (descriptive elements)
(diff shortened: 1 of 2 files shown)


test/unit/rules/require-input-label-test.js
```diff
@@ -81,5 +83,41 @@ generateRuleTests({
...
+        source: '<input id="label-input" aria-label="second label">',
+      },
+    },
+    {
+      template: '<input id="label-input" ...attributes>',
+      result: {
+        message: ERROR_MESSAGE_MULTIPLE_LABEL,
+        line: 1,
+        column: 0,
+        source: '<input id="label-input" ...attributes>',
```
MelSumner, line 110: [question] how is this a multiple label issue?

lib/rules/require-input-label.js
```diff
@@ -3,7 +3,8 @@
 const Rule = require('./base');
 const AstNodeInfo = require('../helpers/ast-node-info');
 
-const ERROR_MESSAGE = 'Input elements require a valid associated label.';
+const ERROR_MESSAGE_NO_LABEL = 'Input elements require a valid associated label.';
+const ERROR_MESSAGE_MULTIPLE_LABEL = 'The input element has more than one label associated.';
```
MelSumner, line 7:
```suggestion
const ERROR_MESSAGE_MULTIPLE_LABEL = 'Input elements should not have multiple valid labels.';
```

MelSumner:
Good start! The documentation for the rule will also need to be updated. Thank you!

test/unit/rules/require-input-label-test.js
KamiKillertO: I'm not sure about this one but the current rule does not report an error if `...attributes` is present.
So to stay coherent with the current rule I've decided to report an error if `...attributes` + something else is present.

MelSumner: I think that since we can't determine what's going on if `...attributes` is present, we have to allow it to pass.

rwjblue: @MelSumner and I fixed this one up...

test/unit/rules/require-input-label-test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const { ERROR_MESSAGE } = require('../../../lib/rules/require-input-label');
+const { ERROR_MESSAGE, ERROR_MESSAGE_MULTIPLE_LABEL } = require('../../../lib/rules/require-input-label');
```
MelSumner, line 3:
```suggestion
const { 
  ERROR_MESSAGE,
  ERROR_MESSAGE_MULTIPLE_LABEL
} = require('../../../lib/rules/require-input-label');
```

I think this is what prettier is complaining about @rwjblue

test/unit/rules/require-input-label-test.js
```diff
@@ -1,6 +1,9 @@
 'use strict';
 
-const { ERROR_MESSAGE } = require('../../../lib/rules/require-input-label');
+const { 
+  ERROR_MESSAGE,
+  ERROR_MESSAGE_MULTIPLE_LABEL
+} = require('../../../lib/rules/require-input-label');
```
MelSumner, line 6:
```suggestion
const {
  ERROR_MESSAGE,
  ERROR_MESSAGE_MULTIPLE_LABEL,
} = require('../../../lib/rules/require-input-label');
```

PR #1443, opened 2020-08-10
Add new `a11y` configuration

MelSumner:
Adds a new config file to help keep track of a11y-specific rules.

### DIFF
--- a/dev/versioning.md
+++ b/dev/versioning.md
@@ -55,4 +55,4 @@ There are some files that are intended to be used for special purposes and as su
 
 Exemption list:
 
-* `a11y` - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired. This is excluded from the normal SemVer guarantees as we expect it to always be the "canonical reference" for the A11Y related rules in the codebase.
+* [a11y](../lib/config/a11y.js) - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired. This is excluded from the normal SemVer guarantees as we expect it to always be the "canonical reference" for the A11Y related rules in the codebase.
--- a/dev/versioning.md
+++ b/dev/versioning.md
@@ -51,7 +51,7 @@ A major release is likely to break your lint build.
 
 ## Exemptions
 
-There are some files that are intended to be used for special purposes and as such are exempt from this versioning policy. 
+There are some files that are intended to be used for special purposes and as such are exempt from this versioning policy.
 
 Exemption list:
 
--- a/dev/versioning.md
+++ b/dev/versioning.md
@@ -55,4 +55,4 @@ There are some files that are intended to be used for special purposes and as su
 
 Exemption list:
 
-* `a11y`
+* `a11y` - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired.
--- a/dev/versioning.md
+++ b/dev/versioning.md
@@ -55,4 +55,4 @@ There are some files that are intended to be used for special purposes and as su
 
 Exemption list:
 
-* lib/config/a11y.js
+* `a11y`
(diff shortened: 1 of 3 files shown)


bmish:
What's the use case for this? Since I believe these are all already in or soon-to-be in the `recommended` config.

MelSumner:
@bmish I needed a separate way to track a11y-specific rules and this seemed to be the most straightforward option. It'll be useful in cases where a team wants to understand how many a11y-specific linting rules they are breaking in their app, or for instances where I need to run just the a11y rules on a codebase to get an overall measurement of linting health from that viewpoint. 

Does that make sense?

bmish:
I see. Would this list follow semver (new rules added only in major releases), or would you add a11y rules to it any time they get created so that the list is always complete?

ESLint has another feature that might be useful. ESLint rules can specify a [docs.category](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics-1) property which is then used to group the rules in the [README rules list](https://github.com/ember-template-lint/ember-template-lint#rules) (presumably one of the groups would be a11y). This would be a good improvement regardless (see the [eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember#-rules) rules list for an example).

rwjblue:
> I see. Would this list follow semver (new rules added only in major releases), or would you add a11y rules to it any time they get created so that the list is always complete?

Yep, we definitely need to update the `SemVer` policy to clearly document that we _will_ add new rules to this without a major. Ironically, this was **exactly** the reason that we created the SemVer.md file to begin with 😝.

dev/versioning.md
```diff
@@ -48,3 +48,11 @@ A major release is likely to break your lint build.
...
 * Adding new rules in a config set
 * Making a rule more restrictive
+
+## Exemptions
+
+There are some files that are intended to be used for special purposes and as such are exempt from this versioning policy.
+
+Exemption list:
+
+* lib/config/a11y.js
```
rwjblue, line 58: Let's refer to this in the way that users would (I doubt most folks know that specifying `recommended` maps to this repos `lib/config/recommended.js`). 

```suggestion
* `a11y`
```

rwjblue: Can you also add a small description of _why_ this config is exempted?

dev/versioning.md
```diff
@@ -48,3 +48,11 @@ A major release is likely to break your lint build.
...
 * Adding new rules in a config set
 * Making a rule more restrictive
+
+## Exemptions
+
+There are some files that are intended to be used for special purposes and as such are exempt from this versioning policy.
+
+Exemption list:
+
+* `a11y` - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired.
```
rwjblue, line 58:
```suggestion
* `a11y` - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired. This is excluded from the normal SemVer guarantees as we expect it to always be the "canonical reference" for the A11Y related rules in the codebase.
```

dev/versioning.md
```diff
@@ -48,3 +48,11 @@ A major release is likely to break your lint build.
...
 * Adding new rules in a config set
 * Making a rule more restrictive
+
+## Exemptions
+
+There are some files that are intended to be used for special purposes and as such are exempt from this versioning policy.
+
+Exemption list:
+
+* `a11y` - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired. This is excluded from the normal SemVer guarantees as we expect it to always be the "canonical reference" for the A11Y related rules in the codebase.
```
bmish, line 58:
```suggestion
* [a11y](../lib/config/a11y.js) - this config file was created to help in instances where an a11y-specific linting audit of a codebase is desired. This is excluded from the normal SemVer guarantees as we expect it to always be the "canonical reference" for the A11Y related rules in the codebase.
```

PR #1446, opened 2020-08-10
no-attrs-in-components rule supports octane component template co-location

### LINKED ISSUE #1445: no-attrs-in-components rule has false-negetive result for Octane co-located templates
#What happened?

The `no-attrs-in-components` rule always passes for component templates that use the Octane component template co-location, even if they are actually violating this rule but using `attrs` in a component template.

# What I expected to happen?

The rule should fail.

# Why I think this is happening:

This rules has an `isComponentTemplate` method that checks a given template's file location to determine if it is a component template, since this rule only applies to component templates. This method utilizes the following regex, which does not appear to be updated to account for Octane component template co-location:
https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-attrs-in-components.js#L6

Since this regex does not return a match for co-located component templates, the `isComponentTemplate` method returns false if a component template is co-located (ex `/app/components/foo.hbs`), which early returns and prevents the rule's criteria from being fully evaluated. This means that this rule will never fail for co-located component templates, even if the co-located component template does indeed use `attrs`.

# How I think this could be fixed:

The regex pattern used in this rule seems brittle. 
https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-attrs-in-components.js#L6

It could be changed to something that instead checks for the presence of the string `/components/` in the filepath of each template.

steveszc:
Fixes #1445 

This uses a much simpler Regex for determining if a given template is a component template. 
Now we merely check for the presence of the string `components` in the template's file path.

The existing test cases for this rule do no appear to cover the different component template locations that were previously supported explicitly. It may make sense to add test coverage to this rule to make sure that at least classic `app/templates/components/` and `/app/components/*.hbs` folder formats are working correctly.


lib/rules/no-attrs-in-components.js
```diff
@@ -3,7 +3,7 @@
 const Rule = require('./base');
 
 const componentTemplateRegex = new RegExp(
```
MelSumner, line 5:
```suggestion
const componentTemplateRegex = new RegExp('components');
```

lib/rules/no-attrs-in-components.js
```diff
@@ -3,7 +3,7 @@
...
-  'templates/components|components/.*/template|ui/components|-components/'
+  'components'
```
MelSumner, line 6:
```suggestion
```

lib/rules/no-attrs-in-components.js
```diff
@@ -3,7 +3,7 @@
...
 );
```
MelSumner, line 7:
```suggestion
```

bmish:
Agreed that we need to add a bunch of tests if we're going to change this regex.

MelSumner (approved):
Thank you for updating this rule and adding extra tests!

lifeart:
Hi @steveszc! Thank you for your contribution! Sorry for delay here, do you mind to rebase your PR?

Issue #1448, opened 2020-08-12
New Rule: `no-nested-splattributes`


Turbo87:
We've just had to fix a bug where something roughly like this was used:

```hbs
<div ...attributes>
  <div ...attributes>
    ...
  </div>
</div>
```

A template-lint rule to prevent having an element with `...attributes` nested within another such element would be able to prevent these sorts of issues.

rwjblue:
I'm curious what the actual bug was? Was it just that the attributes were duplicated unexpectedly? 

I ask, because in _general_ the rendering engine should allow / support that (even though I totally agree with you that we should have a linting rule against it).

Turbo87:
> Was it just that the attributes were duplicated unexpectedly?

yep, exactly. we had a change that introduced a new wrapper element and the inner `...attributes` was not removed by the dev.

PR #1449, opened 2020-08-12
Implement `no-nested-splattributes` rule

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1448

### DIFF
--- a/docs/rule/no-nested-splattributes.md
+++ b/docs/rule/no-nested-splattributes.md
@@ -1,0 +1,30 @@
+# no-nested-splattributes
+
+Having `...attributes` on multiple elements nested within each other in a
+component can cause unintended results.
+
+This rule prevent you from running into this issue by disallowing
+`...attributes` if any of the parent elements already has `...attributes` on it.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<div ...attributes>
+  <div ...attributes>
+    ...
+  </div>
+</div>
+```
+
+This rule **allows** the following:
+
+```hbs
+<div ...attributes>...</div>
+<div ...attributes>...</div>
+```
+
+## Migration
+
+- Remove the inner `...attributes` declaration
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -44,6 +44,7 @@ module.exports = {
   'no-negated-condition': require('./no-negated-condition'),
   'no-nested-interactive': require('./no-nested-interactive'),
   'no-nested-landmark': require('./no-nested-landmark'),
+  'no-nested-splattributes': require('./no-nested-splattributes'),
   'no-obsolete-elements': require('./no-obsolete-elements'),
   'no-outlet-outside-routes': require('./no-outlet-outside-routes'),
   'no-partial': require('./no-partial'),
--- a/lib/rules/no-nested-splattributes.js
+++ b/lib/rules/no-nested-splattributes.js
@@ -1,0 +1,38 @@
+'use strict';
+
+const Rule = require('./base');
+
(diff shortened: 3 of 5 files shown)


lib/rules/index.js
```diff
@@ -74,4 +74,5 @@ module.exports = {
   'style-concatenation': require('./style-concatenation'),
   'table-groups': require('./table-groups'),
   'template-length': require('./template-length'),
+  'no-nested-splattributes': require('./no-nested-splattributes'),
```
bmish, line 77: There's supposed to be a rule setup test to ensure this list is alphabetical but I guess that check is missing. I'll follow-up to add that check.

Turbo87: hmm, interesting. I just followed what `yarn new` does :)

bmish: Oh yeah, I'll see if I can fix that command to sort it.

lib/rules/no-nested-splattributes.js
```diff
@@ -0,0 +1,38 @@
...
+            } else {
+              splattributesParent = node;
+            }
+          }
+        },
+        exit(node) {
+          if (splattributesParent === node) {
+            splattributesParent = null;
+          }
+        },
```
rwjblue, line 32: Thoughts on using `path.parents` instead of manual book keeping + enter/exit? I'm sure you thought of it, just curious what led you to choosing this form instead (are we missing some API niceties or something)?

test/unit/rules/no-nested-splattributes-test.js
```diff
@@ -0,0 +1,49 @@
...
+
+generateRuleTests({
+  name: 'no-nested-splattributes',
+
+  config: true,
+
+  good: [
+    '<div ...attributes>...</div>',
+    '<div ...attributes>...</div><div ...attributes>...</div>',
+  ],
```
rwjblue, line 15: Can you add a good entry that doesn't have any `...attributes`. Pretty much doesn't matter (given how the rule works), but doesn't hurt to have either 😸 

```suggestion
  good: [
    '<div ...attributes>...</div>',
    '<div ...attributes>...</div><div ...attributes>...</div>',
    '<div></div>',
  ],
```

docs/rule/no-nested-splattributes.md
```diff
@@ -0,0 +1,30 @@
...
+This rule **allows** the following:
+
+```hbs
+<div ...attributes>...</div>
+<div ...attributes>...</div>
+```
+
+## Migration
+
+- Remove the inner `...attributes` declaration
```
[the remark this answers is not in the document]
Turbo87: Thought about that too, but I think it's probably safer to let the user decide which splattribute he wants to remove. Most likely it will be the inner one, but probably not in 100% of the cases.

lib/rules/no-nested-splattributes.js
Turbo87: yeah, would probably work too, but it felt a bit more efficient this way and not much more verbose than the alternative

docs/rule/no-nested-splattributes.md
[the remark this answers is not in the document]
rwjblue: Ya, makes sense.

test/unit/rules/no-nested-splattributes-test.js
Turbo87: done ✅

docs/rule/no-nested-splattributes.md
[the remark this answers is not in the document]
bmish: I agree that we should not make assumptions about which `...attributes` is the mistaken one (in fact, they could both be legit).

PR #1450, opened 2020-08-12
Implement `no-redundant-fn` rule

Turbo87:

```hbs
<button {{on "click" (fn this.handleClick)}}>Click Me</button>
```

```hbs
<button {{on "click" this.handleClick}}>Click Me</button>
```

```hbs
<button {{on "click" (fn this.handleClick "foo")}}>Click Me</button>
```

### DIFF
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -74,4 +74,5 @@ module.exports = {
   'style-concatenation': require('./style-concatenation'),
   'table-groups': require('./table-groups'),
   'template-length': require('./template-length'),
+  'no-redundant-fn': require('./no-redundant-fn'),
 };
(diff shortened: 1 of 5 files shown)


lib/rules/index.js
```diff
@@ -74,4 +74,5 @@ module.exports = {
   'style-concatenation': require('./style-concatenation'),
   'table-groups': require('./table-groups'),
   'template-length': require('./template-length'),
+  'no-redundant-fn': require('./no-redundant-fn'),
```
rwjblue, line 77: We should sort these alphabetically, but I think @bmish mentioned doing a follow up to fix that (and the `yarn new` script).

bmish: Done: https://github.com/ember-template-lint/ember-template-lint/pull/1455

PR #1452, opened 2020-08-12
Drop support for Node 13

MelSumner:
When I updated the Node version in `package.json` for engines, I got this error: 

```bash
error ember-template-lint@2.10.0: The engine "node" is incompatible with this module. Expected version "13.* || >= 14.*". Got "12.18.1"
error Found incompatible module.
```
I also noticed that we historically seem to do the last two versions of Node, but if we drop support for Node 13 then we'd only support one version of Node. So I set the Node versions in `ci.yml` to use both 13 and 14.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test:
-    name: Node 12.x - ubuntu
+    name: Node 14.x - ubuntu
     runs-on: ubuntu-latest
     timeout-minutes: 10
 
@@ -37,12 +37,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, windows]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [13.x, 14.x]
 
         # excluded because it is the `test` job above
         exclude:
           - os: ubuntu
-            node-version: 12.x
+            node-version: 14.x
 
     steps:
       - uses: actions/checkout@v2
@@ -66,7 +66,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: volta-cli/action@v1
         with:
-          node-version: 12.x
+          node-version: 14.x
 
       - name: install dependencies
         run: yarn install --no-lockfile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test:
-    name: Node 14.x - ubuntu
+    name: Node 12.x - ubuntu
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -37,12 +37,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, windows]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [13.x, 14.x]
```
rwjblue, line 40:
```suggestion
        node-version: [10.x, 12.x, 14.x]
```

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test:
-    name: Node 12.x - ubuntu
+    name: Node 14.x - ubuntu
```
rwjblue, line 15: This isn't changed. It uses the volta config in `package.json`.

README.md
```diff
@@ -26,7 +26,7 @@ To install ember-template-lint
 npm install --save-dev ember-template-lint
 ```
 
-Node.js `10 || >=12` is required.
+Node.js `13 || >=14` is required.
```
rwjblue, line 29: 13 is unsupported. 

```suggestion
Node.js `10 || 12 || >=14` is required.
```

.github/workflows/ci.yml
```diff
@@ -37,12 +37,12 @@ jobs:
...
       matrix:
         os: [ubuntu, windows]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [13.x, 14.x]
 
         # excluded because it is the `test` job above
         exclude:
           - os: ubuntu
-            node-version: 12.x
+            node-version: 14.x
```
rwjblue, line 45: This should be reset (or the `package.json` `volta` key should be updated)

package.json
```diff
@@ -72,7 +72,7 @@
     "release-it-lerna-changelog": "^2.3.0"
   },
   "engines": {
-    "node": "10.* || >= 12.*"
+    "node": "12.* || >= 14.*"
```
rwjblue, line 75:
```suggestion
    "node": "10.* || 12.* || >= 14.*"
```

MelSumner:
> This isn't quite right, we still support 10, 12, and 14.

Ok that makes more sense- I misunderstood it to be that we were dropping support for anything  v.13 and under.

rwjblue:
> Ok that makes more sense- I misunderstood it to be that we were dropping support for anything v.13 and under.

No we support whatever Node itself supports, generally from here https://nodejs.org/en/about/releases/.

PR #1453, opened 2020-08-12
Fix misspelling in `--print-pending` help text.

MelSumner:
fixed spelling error


rwjblue:
Might need to update the fixtures too...

Issue #1231, opened 2020-03-25
{{link-to tagName="button"}} Linting


hergaiety:
My preface for this is that ideally we wouldn't see any `{{link-to` that has a `tagName`, and we are moving towards a world of Angle bracket notation. However, as it currently stands we still support curly bracket notation and I still see real world examples of `{{link-to`'s with custom `tagName`'s.

Most notably, the following template code will miss the https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/require-button-type.md require-button-type rule.

``` html.hbs
{{link-to 'new.path' tagName="button"}}foo{/link-to}}
```

lifeart:
This is absolutely valid code:
```hbs
<LinkTo @route="new.path" @tagName="button">foo</LinkTo>
```

MelSumner:
> This is absolutely valid code:
> 
> ```
> <LinkTo @route="new.path" @tagName="button">foo</LinkTo>
> ```

what do you mean by valid? It's certainly an a11y error. 

If it's an `<a>` element, it should go to a URL. If it's a `button` then it should be some sort of toggle (open a modal, a menu, switch a tab, save, etc.).

lifeart:
Sorry for confusing, I mean it "syntactically" valid. @MelSumner I seen use cases when modal is independent route like `users/edit/ -> users/edit/confirm-delete` and such.

lifeart:
Looks like we can't simply bind "type" html attribute to `{{link-to}}`, we need to write code like
```hbs
{{#link-to "application" tagName="button" type="button" attributeBindings=(array "type")}} {Link-To} {{/link-to}}
```

PR #1459, opened 2020-08-13
Implement `no-link-to-tag-name` rule

Turbo87:
This should resolve https://github.com/ember-template-lint/ember-template-lint/issues/1231

/cc @MelSumner


MelSumner:
@Turbo87 I wonder if we should make a rule that is `no-invalid-tagname` that includes this and `no-input-tagname` and then deprecate `no-input-tagname`- could make it more future flexible. 

What do you think?

Turbo87:
I have to admit that I wasn't aware `no-input-tagname` was a thing. I'm not sure if there would be many advantages from merging the two rules though. `no-invalid-tagname` sounds a bit too generic to me if it only runs for `Input` and `LinkTo`. 🤔

rwjblue:
Probably good to land as separate rules.

PR #1460, opened 2020-08-13
ast-node-info: Remove trivial `isElement()` helpers

Turbo87:
as #1168 suggests, these are too trivial to serve any real purpose and should be replaced by regular `tag` property checks

/cc @MelSumner

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -206,14 +206,6 @@ const helpers = require('ember-template-lint').ASTHelpers;
 
   Returns any attributes on the node with a name that matches `attributeName`.
 
-* `function isImgElement(node): boolean`
-
-  Returns true if this node is an `img` element.
-
-* `function isLinkElement(node): boolean`
-
-  Returns true if this node is a link (`a`) element.
-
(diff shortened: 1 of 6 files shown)


Turbo87:
should we do the same for the helpers that check `node.type`?

rwjblue:
> should we do the same for the helpers that check node.type?

IMHO, yes

rwjblue:
And I want to leverage the work in #1430 to make combining these types of checks together...

josephdsumner:
> And I want to leverage the work in #1430 to make combining these types of checks together...

YES! I was thinking about this parent issue re: where matching should go and the appropriate export strategy. Do we have a coherent vision for what the new moving parts are after the major breaking / 'essentially breaking' changes are made (this one, matches, deprecating the `elementAttributeValue` family, etc.)?

MelSumner:
since this is breaking, should it target the `next` branch @Turbo87 ?

Turbo87:
I wasn't aware that we were using the `next` branch. I'm happy to keep these PRs open until the release or rebase them on `next`. Up to you. :)

PR #1463, opened 2020-08-13
ast-node-info: Remove trivial `isNode()` helpers

Turbo87:
This addresses https://github.com/ember-template-lint/ember-template-lint/pull/1460#issuecomment-673594609


MelSumner:
Same question here- should we be targeting the `next` branch for this PR since it's a breaking change?

Issue #1098, opened 2020-01-29
New rule which forces using @-prefixed attributes in native components


stukalin:
In the https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input we have the following section

> However, there are a few attributes where you must use the @ version.

which tells us to use the `@` for (at the moment of writing) `@type`, `@value` and the `@checked` on the `<Input/>`. Can we add a rule to enforce that?

lifeart:
should rule warn if `{{on 'change' this.change}}` modifier used on `<Input | Textarea> `?

stukalin:
@lifeart could you remind me what's wrong with that `{{on 'change'}}` ?

lifeart:
it's framework component, I beleve developers should use `@onchange={{this.callback}}` to handle changes.

stukalin:
@lifeart sounds like a bit off-the-scope of the proposed rule to me. Not 100% sure though

rwjblue:
Ya, agree with @stukalin. I think adding a rule to help folks use `<Input>` properly makes sense.

PR #1465, opened 2020-08-13
Add new rule `builtin-component-arguments`

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1098

### DIFF
--- a/docs/rule/builtin-component-arguments.md
+++ b/docs/rule/builtin-component-arguments.md
@@ -1,0 +1,51 @@
+# builtin-component-arguments
+
+The builtin `Input` component has several arguments that match attributes
+of the lower-case `input` HTML element. These arguments should be set via e.g.
+`@type`, instead of `type`, but it is easy to forget and can cause subtle
+issues.
+
+This rule warns about `Input` component invocations that use the following attributes instead of arguments:
+
+* `checked`
+* `type`
+* `value`
+
+Please note that this rule currently only warns about these three attributes on
+the `Input` component, but might be extended in the future to also warn about
+other attributes or builtin components.
+
+## Examples
+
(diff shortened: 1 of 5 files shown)


bmish:
Any way to make the name more clear or more consistent with other rule names? `no-erroneous-arguments-on-built-in-components`?

MelSumner:
To be more consistent with the other rules, perhaps you would consider `require-valid-input-args` or similar. What do you think?

lib/rules/builtin-component-arguments.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const Rule = require('./base');
+
+const FORBIDDEN_ATTRIBUTES = {
+  Input: new Set(['checked', 'type', 'value']),
+};
```
rwjblue, line 7: What about `Textarea`?

mongoose700:
I think the rule name should refer to `attributes` rather than `arguments`, to distinguish that it's the ones without the `@` that are invalid.

docs/rule/builtin-component-arguments.md
```diff
@@ -0,0 +1,48 @@
+# builtin-component-arguments
+
+The builtin `Input` component has several arguments that match attributes
+of the lower-case `input` HTML element. These arguments should be set via e.g.
+`@type`, instead of `type`, but it is easy to forget and can cause subtle
+issues.
+
+This rule warns about `Input` component invocations that use `type`, `value` or
+`checked` attributes, instead of arguments.
```
bmish, line 9:
```suggestion
This rule warns about `Input` component invocations that use the following attributes instead of arguments:

* `checked`
* `type`
* `value`
```

PR #1466, opened 2020-08-13
fix: cache pending/ignore config lookups for faster perf/less memory

fivetanley:
This commit improves the lookups of `ignore` and `pending` configs by reducing the amount of iterations required over those config properties.

tl;dr:

- cached the config lookups for `pending` and `ignore` to reduce the amount of times we have to re-iterate the `pending` or `ignore` arrays provided by the config.
- cache string lookups for normalized moduleId paths since
- runtime on codebase with 668 handlebars files went from ~13 seconds to ~9 seconds

This was done in three steps:

First Step

The normalized `moduleId` was cached. Running ember-template-lint on a big codebase (668 template files) took 12.94 seconds to run. I used this command in the app's codebase directory to profile:

```
node --prof ../ember-template-lint/bin/ember-template-lint.js .
```

Next, I generated a graph with the [flamebearer][flamebearer] npm package:

```
node --prof-process --preprocess -j isolate*.log | flamebearer
```

Screen_Shot_2020-08-13_at_9_25_14_AM

Profiling showed that ~19% of the CPU time was spent inside of the `statusForModule` function. Upon further investigation, I noticed that the [normalized moduleId based on the current working directory was being generated in a loop][3], creating a new string for potentially every item in the `pending` or `ignore` config. Our app has no `ignore` config, but does have a lot of `pending` entries.

Moving the `fullPathModuleId` to a variable outside the loop took the runtime from 12.94 seconds to 11.04 seconds, about ~2 seconds of savings on its own!

An important thing to keep in mind for the next two steps: `statusForModule` is called _at least once per rule_ to determine if the rule should be ignored or allowed to fail (or removed from the pending file if now passing). This means that any functions ran or strings created by `statusForModule` are, in the worst case scenario, calculated `numberOfFiles * pendingRuleConfigItems * ignoreConfigItems` times if the loop doesn't find the config for the file early, as it has to search the entire array of `pending` or `ignore` items for every file.

Rather than rejoining the paths, they are instead generated once and stored in a cache object. This reduces the number of strings generated and time generating the same strings over and over. Caching the result of `process.cwd` (since it seems unlikely to change once the program is started) potentially has a nice performance side effect for users on Node 10 (supported until April 2021) as [process.cwd was not cached until Node 12.2](https://github.com/nodejs/node/pull/27224).

Second Step

Lookups for `ignore` and `pending` were separated out into different caches. This was done because while [`statusForModule` will check for a function to run][1] instead of a string, in practice this is only true for `ignore` rules. The reason `ignore` rules need the function check is that they are always a function due to being converted from [strings to functions via the micromatch module][2]. Caching the lookup of `ignore` rules reduced the need to run the functions for every rule/file.

Third Step

Last, but not least: `pending` rules always seem to be a list of `object`s or `string`s (as generated by `--print-pending`) and having a list that potentially has a function in it doesn't seem likely given that the recommendation from ember-template-lint is to copy/paste the output of `--print-pending`.

After re-profiling, the runtime went from 12.94 seconds to 9.04 seconds on my machine. The `get-config` file no longer showed up in the profile.

Screen Shot 2020-08-13 at 12 07 40 PM

[1]: https://github.com/ember-template-lint/ember-template-lint/blob/5937b63bed30380b4bce0f96b19061658a176840/lib/get-config.js#L376-L377
[2]: https://github.com/ember-template-lint/ember-template-lint/blob/5937b63bed30380b4bce0f96b19061658a176840/lib/get-config.js#L280
[3]: https://github.com/ember-template-lint/ember-template-lint/blob/5937b63bed30380b4bce0f96b19061658a176840/lib/get-config.js#L374
[flamebearer]: https://github.com/mapbox/flamebearer

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -364,103 +364,6 @@ function _determineConfigForSeverity(config) {
   }
 }
 
-class ModuleStatusCache {
-  constructor(config, configPath) {
-    this.config = config;
-    this.configPath = configPath || '';
-    this.cache = {
-      pending: {},
-      ignore: {},
-    };
-    this.processCWD = process.cwd();
-  }
-
-  lookupPending(moduleId) {
-    if (!moduleId || !this.config.pending) {
-      return false;
-    }
-    if (!this.cache.pendingLookup) {
-      this.cache.pendingLookup = this._extractPendingCache();
-    }
-    if (!(moduleId in this.cache.pending)) {
-      const fullPathModuleId = path.resolve(this.processCWD, moduleId);
-      this.cache.pending[moduleId] = this.cache.pendingLookup[fullPathModuleId];
-    }
-    return this.cache.pending[moduleId];
-  }
-
-  lookupIgnore(moduleId) {
-    if (!(moduleId in this.cache.ignore)) {
-      const ignores = this.config['ignore'] || [];
-      this.cache.ignore[moduleId] = ignores.find((match) => match(moduleId));
-    }
-    return Boolean(this.cache.ignore[moduleId]);
-  }
-
-  _extractPendingCache() {
-    const list = this.config.pending;
-    const byFullModuleId = {};
-
-    if (!list) {
-      return byFullModuleId;
-    }
-
-    for (const item of list) {
-      if (typeof item === 'string') {
-        const fullPath = this.resolveFullModuleId(item);
-        byFullModuleId[fullPath] = true;
-      } else if (item.moduleId) {
-        const fullPath = this.resolveFullModuleId(item.moduleId);
-        byFullModuleId[fullPath] = item;
-      }
-    }
-
-    return byFullModuleId;
-  }
-
-  resolveFullModuleId(moduleId) {
-    if (!this._baseDirBasedOnConfigPath) {
-      this._baseDirBasedOnConfigPath = path.resolve(this.processCWD, path.dirname(this.configPath));
-    }
-    return path.resolve(this._baseDirBasedOnConfigPath, moduleId);
-  }
-}
-
-let configModuleCacheMap = new WeakMap();
-
-/**
- * Returns the config in conjunction with overrides configuration.
- * @param {*} config
- * @param {*} filePath
- */
-function getConfigForFile(config, options) {
-  if (!configModuleCacheMap.has(config)) {
-    configModuleCacheMap.set(config, new ModuleStatusCache(config, options.configPath));
-  }
-  let moduleStatusCache = configModuleCacheMap.get(config);
-  let filePath = options.filePath;
-  let configuredRules = config.rules;
-  let overrides = config.overrides;
-
-  let fileConfig = Object.assign({}, config, {
-    pendingStatus: moduleStatusCache.lookupPending(options.moduleId),
-    shouldIgnore: moduleStatusCache.lookupIgnore(options.moduleId),
-  });
-
-  if (filePath && overrides) {
-    let overridesRules = {};
-    overrides.forEach((override) => {
-      const isFileMatch = override.files && micromatch.isMatch(filePath, override.files);
-      if (isFileMatch) {
-        overridesRules = Object.assign(overridesRules, override.rules);
-      }
-    });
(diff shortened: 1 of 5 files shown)


lib/get-config.js
```diff
@@ -362,47 +362,85 @@ function _determineConfigForSeverity(config) {
...
+  }
+  resolveFullModuleId(moduleId) {
+    if (!this._baseDirBasedOnConfigPath) {
+      this._baseDirBasedOnConfigPath = path.resolve(process.cwd(), path.dirname(this.configPath));
+    }
+    return path.resolve(this._baseDirBasedOnConfigPath, moduleId);
+  }
 }
+
+let configModuleCacheMap = new WeakMap();
```
fivetanley, line 425: there is probably a better way to do this... not sure where an appropriate place to store instances of the ModuleStatusCache would be.

rwjblue: Let's tweak it so that the `lib/linter.js` instance creates the class you made above directly. That would avoid the need for a cache here in module scope, and naturally tie the two together (they are implicitly tied now, since `config` here is `this.config` over in `lib/linter.js`).

At that point, I would expect `getConfigForFile` utility method here to be removed, in favor of something like:

```js
class Linter {
  // ...snip...
  loadConfig() {
    // ...snip...
    
    this.moduleStatus = new ModuleStatusCache(this.config);
  }

  verify() {
    let results = [];
    let fileConfig = this.moduleStatus.configFor(options);
    // ...snip...
  }
}
```

What do you think?

fivetanley:
@rwjblue followed [your suggestions](https://github.com/ember-template-lint/ember-template-lint/pull/1466#discussion_r473141754) to move the new `ModuleStatusCache` to its own file (inside a new `-private` directory so people hopefully don't try to import it and use it directly) and assigning the instance in the `Linter` instance instead of in a weakmap in the `lib/get-config.js`'s module scope.

I re-ran the numbers on our app, here's the numbers (still an improvement! I didn't expect this to change but it seems good to verify that the refactor is working to improved speed).

I ran master and this PR branch 3 times each:

Before ([1c72839e544b2707a62cdada74c763ac274280cb](https://github.com/ember-template-lint/ember-template-lint/commit/1c72839e544b2707a62cdada74c763ac274280cb): 
```
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
       13.95 real        13.33 user         4.08 sys
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
       14.44 real        13.86 user         4.18 sys
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
       14.61 real        13.92 user         4.20 sys
```

After (this PR):

```
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
        9.80 real        11.90 user         0.94 sys
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
        9.12 real        11.51 user         0.86 sys
$ /usr/bin/time ../ember-template-lint/bin/ember-template-lint.js . --quiet
        9.09 real        11.43 user         0.85 sys
```

test/unit/get-config-test.js
```diff
@@ -613,116 +612,52 @@ describe('determineRuleConfig', function () {
...
+  it('should return an empty object when options.config is set explicitly false', function () {
+    const config = resolveProjectConfig({ config: false });
 
-    expect(actual.rules).toEqual(expected.rules);
+    expect(config).toEqual({});
   });
+});
 
-  it('Merges the overrides rules from multiple overrides with existing rules config', function () {
+describe('getProjectConfig', function () {
```
fivetanley, line 623: This file is a bit noisier in the diff than expected. There were some tests that didn't use `getConfig` underneath the `getConfig` describe block. I moved them to their own describe blocks but they should otherwise be unchanged.

lib/get-config.js
fivetanley: Moved!

rwjblue:
Thank you @fivetanley!

PR #1468, opened 2020-08-14
Use `eslint-plugin-import-helpers` to sort imports alphabetically

Turbo87:
This ensures that the module imports are sorted and grouped a little more consistently.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,15 @@ module.exports = {
   env: {
     node: true,
   },
-  plugins: ['eslint-comments', 'filenames', 'import', 'node', 'prettier', 'unicorn'],
+  plugins: [
+    'eslint-comments',
+    'filenames',
+    'import',
+    'import-helpers',
+    'node',
+    'prettier',
+    'unicorn',
+  ],
   extends: [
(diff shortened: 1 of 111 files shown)


.eslintrc.js
```diff
@@ -92,6 +100,19 @@ module.exports = {
     'import/no-webpack-loader-syntax': 'error',
     'import/unambiguous': 'error',
 
+    'import-helpers/order-imports': [
+      'error',
+      {
+        newlinesBetween: 'always',
+        groups: [
+          '/^(assert|async_hooks|buffer|child_process|cluster|console|constants|crypto|dgram|dns|domain|events|fs|http|http2|https|inspector|module|net|os|path|perf_hooks|process|punycode|querystring|readline|repl|stream|string_decoder|timers|tls|trace_events|tty|url|util|v8|vm|zli)/',
```
bmish, line 108: Where does this list come from?

Turbo87: https://github.com/Tibfib/eslint-plugin-import-helpers/blob/master/docs/rules/order-imports.md#upgrading-from-v014-to-v1

PR #1474, opened 2020-08-17
Add `--fix` support for `link-rel-noopener` rule

fivetanley:


lib/rules/link-rel-noopener.js
```diff
@@ -77,17 +81,47 @@ module.exports = class LinkRelNoopener extends Rule {
...
-          message: this.config.message,
-          line: node.loc && node.loc.start.line,
-          column: node.loc && node.loc.start.column,
-          source: this.sourceForNode(node),
-        });
+        if (this.mode === 'fix') {
+          fix(node, this.config);
+        } else {
+          this.log({
+            isFixable: true,
```
lifeart, line 88: looks like we able to fix it if rel attribute value is a textNode, in other cases, it should be false

fivetanley: @lifeart thanks! great catch. updated

lib/rules/link-rel-noopener.js
```diff
@@ -88,6 +97,41 @@ module.exports = class LinkRelNoopener extends Rule {
...
+  let oldRel = AstNodeInfo.findAttribute(elementNode, 'rel');
+  return !oldRel || AstNodeInfo.isTextNode(oldRel.value);
+}
+
+function fix(node, config) {
+  let oldRel = AstNodeInfo.findAttribute(node, 'rel');
+
+  if (!isFixable(node)) {
+    return;
+  }
```
rwjblue, line 110: Since the `fix` function is only invoked from within a `isFixable` guard we can probably remove this one as a bit redundant?

fivetanley: Ah yep. I was trying to be clever and have `if (this.mode === fix && fix(node, config)) { return } else { this.log(...) }` but prettier/eslint didn't like that style so I backed out of it. This is removed and updated. Thanks for the review!

rwjblue:
Thank you!

PR #1477, opened 2020-08-19
added meta info for `block-indentation` rule

MelSumner:
If this is how we want to add the meta info for each rule, I'll do this for each rule. 

The idea is that adding meta info to each rule will make it easier to break out the documentation in a more automated fashion.

### DIFF
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -82,6 +82,16 @@ function isControlChar(char) {
 }
 
 module.exports = class BlockIndentation extends Rule {
+  static get meta() {
+    return {
+      description: 'requires blocks to be indented and aligned',
+      category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
+      presets: { recommended: true },
+      url:
+        'https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rules/block-indentation.md',
+      fixable: false,
+    };
+  }
   parseConfig(config) {
     let configType = typeof config;
 


rwjblue (approved):
Seems good to me, @bmish what do you think?

lib/rules/block-indentation.js
```diff
@@ -82,6 +82,16 @@ function isControlChar(char) {
...
 
 module.exports = class BlockIndentation extends Rule {
+  static get meta() {
+    return {
+      description: 'requires blocks to be indented and aligned',
+      category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
+      presets: { recommended: true },
+      url:
+        'https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rules/block-indentation.md',
+      fixable: false,
```
bmish, line 92: I would like to exactly match the [eslint rule meta object](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics-1):

```suggestion
      type: 'layout',
      docs: {
        description: 'requires blocks to be indented and aligned',
        category: 'Stylistic Issues', // 'Deprecated Rules', 'Possible Error', 'Best Practices',
        recommended: false,
        url: 'https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rules/block-indentation.md',
      },
      fixable: false,
      schema: [],
```

lib/rules/block-indentation.js
```diff
@@ -82,6 +82,16 @@ function isControlChar(char) {
 }
 
 module.exports = class BlockIndentation extends Rule {
+  static get meta() {
```
Turbo87, line 85: any reason for this being a property method instead of something like `static meta = { ... }`?

rwjblue: I don't think static class fields are supported on our supported Node versions. Could be wrong though...

Issue #989, opened 2019-12-09
Replace whitelist/blacklist references


MelSumner:
Replace instances of whitelist & blacklist with more inclusive alternatives. Suggest "allowed" and "disallowed".

jaydgruber:
I would like to pick this up. I propose using `allowlist` & `denylist` to be consistent w/ Eslint terminology for same. ref: https://eslint.org/docs/user-guide/configuring#eslintignore

I think we could add `'id-denylist': ['error', 'whitelist', 'blacklist'],` to our `.eslintrc` to catch future occurrences. ref: [eslint/id-denylist](https://eslint.org/docs/rules/id-denylist)

PR #1481, opened 2020-08-21
expose `DEFAULT_CONFIG` export from `no-bare-strings` rule

### LINKED ISSUE #116: Support additional whitelist characters along with defaults
Current behavior is to provide a `whitelist` array that replaces the defaults. If a user wants additional characters, they have to copy/paste the default array and add to it.

It would be better to provide a way for merging the defaults with a user provided whitelist.

@rwjblue suggested:

> basically `whitelist` and `additionalWhitelist` or something

jaydgruber:
### What
- exports the default config from `no-bare-strings` rule
- docs the same
- adds a test

### Why
- fixes #116 
  - this asks for `basically whitelist and additionalWhitelist` config
  - elsewhere have found it easier to expose the default config and let the user use it how they want, with user config overriding default
- inspired by [eslint-plugin-ember/avoid-leaking-state-in-ember-objects.md](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/avoid-leaking-state-in-ember-objects.md#configuration)

### DIFF
--- a/docs/rule/no-bare-strings.md
+++ b/docs/rule/no-bare-strings.md
@@ -30,9 +30,24 @@ This rule **allows** the following:
 
 When the config value of `true` is used the following configuration is used:
 
-* `whitelist` - `(),.&+-=*/#%!?:[]{}`
+* `whitelist` - refer to the `DEFAULT_CONFIG.whitelist` property in the [rule](../lib/rules/no-bare-strings.js)
 * `globalAttributes` - `title`, `aria-label`, `aria-placeholder`, `aria-roledescription`, `aria-valuetext`
 * `elementAttributes` - `{ img: ['alt'], input: ['placeholder'] }`
+
+The `DEFAULT_CONFIG` is available as an export. Example use in configuration:
+
+```javascript
+const {
+  DEFAULT_CONFIG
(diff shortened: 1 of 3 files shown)


docs/rule/no-bare-strings.md
```diff
@@ -30,10 +30,25 @@ This rule **allows** the following:
 
 When the config value of `true` is used the following configuration is used:
 
-* `whitelist` - `(),.&+-=*/#%!?:[]{}`
+* `whitelist` - refer to the `DEFAULT_CONFIG.whitelist` property in the [rule](../lib/rules/no-bare-strings.js)
```
MelSumner, line 33:
```suggestion
* `allowedList` - refer to the `DEFAULT_CONFIG.allowedList` property in the [rule](../lib/rules/no-bare-strings.js)
```

While we're making these changes, let's go ahead and improve on the terms (as related to issue #989 )

jaydgruber: I'm happy to pick up #989 (in its entirety!). Any objections to doing that as a separate MR? I feel like renaming config keys might end up being a breaking change, whereas this work would not be.

bmish: Yeah let's do that separately since it's a breaking change.

PR #1487, opened 2020-08-26
BREAKING: replace `whitelist`/`blacklist` references

jaydgruber:
### What
- replaces `whitelist` with `allowlist`
- replaces `blacklist` with `denylist`
- adds both terms to [`eslint/id-denylist`](https://eslint.org/docs/rules/id-denylist) config

### Why
- fixes #989 
- dependent on (should follow) #1481 
- `no-action-modifiers` changes are non-breaking, the rest are breaking changes
- second attempt at this PR. #1485 got merged early & reverting b/c it contained breaking changes that should wait for v3.

PR #1489, opened 2020-08-27
add `allowlist` & `denylist` alternatives to `whitelist` & `blacklist` configs

jaydgruber:
### What
- renames `whitelist` -> `allowlist` in `no-action-modifiers` docs. this is an array config so we can rename it safely
- adds `allowlist` & `denylist` alternatives to `whitelist` & `blacklist` in rules `no-bare-strings` & `simple-unless`.
  - this introduces the new config values in a backwards compatible way
  - if both are provided, the deprecated configs' values always get clobbered by the new ones
- adds test cases for same
- adds both terms to [`eslint/id-denylist`](https://eslint.org/docs/rules/id-denylist) config
  - ignores the `no-bare-strings` & `simple-unless` rules for now

### Why
- addresses #989 in a backwards-compatible way as suggested on #1487

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,7 @@ module.exports = {
     curly: 'error',
     eqeqeq: 'error',
     'func-style': ['error', 'declaration'],
+    'id-denylist': ['error', 'whitelist', 'blacklist'],
     'new-parens': 'error',
     'no-async-promise-executor': 'error',
     'no-console': 'error',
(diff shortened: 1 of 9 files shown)


jaydgruber:
@rwjblue - so what's the cadence here? Wait for this to merge, then open the PR to remove the deprecated configs against `next`?
How often does `next` pull in `master` (or is that something I need to do?)

rwjblue:
This is published in https://github.com/ember-template-lint/ember-template-lint/releases/tag/v2.11.0 for now, I've merged v2.11.0 into `next` so it should be ready to target a follow up PR.

PR #1490, opened 2020-08-27
remove deprecated `whitelist` & `blacklist` configurations

jaydgruber:
### What
- removes `whitelist` & `blacklist` configurations deprecated in #1489 
- removes a few instances that my initial pass w/ [`eslint/id-denylist`](https://eslint.org/docs/rules/id-denylist) did not catch.
  - term was only used descriptively in `.md` file or docstring, not as a property name
  - term was used camelcase `blackList`, where previously we only caught `blacklist`

### Why
- completes #989

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,7 +33,7 @@ module.exports = {
     curly: 'error',
     eqeqeq: 'error',
     'func-style': ['error', 'declaration'],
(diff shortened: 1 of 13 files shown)


rwjblue (approved):

rwjblue:
Thank you @jaydgruber!

Issue #1412, opened 2020-07-11
Processing `overrides` rules is not idempotent


bmish:
### TLDR

The processing of `overrides` rules is not idempotent. The `overrides` rules config is corrupted when creating multiple instances of the Linter class with the same `configPath`.

### My config

My .template-lintrc.js looks like this:

```js
module.exports = {
  extends: ['recommended', 'stylistic'],

  rules: {
    'no-bare-strings': true,
    'require-button-type': false,
  },

  overrides: [
    {
      files: ['**/tests/dummy/**/*.hbs'],
      rules: {
        'no-bare-strings': false,
      },
    },
  ],
};
```

### My use case

In my git precommit hook, I am linting each changed file in my monorepo, with a separate instance of the linter class per file (so that each file can use the .template-lintrc.js for its own package).

```js
const TemplateLinter = require('ember-template-lint');

templates.forEach((templateName) => {
  // Find template lint config file for the package this file is in.
  const configPath = findUp.sync('.template-lintrc.js', { cwd: templateName });
  if (!configPath) {
    // Ignore file if we can't find a config file for it.
    return;
  }

  const linter = new TemplateLinter({ configPath });
  const templateContents = fs.readFileSync(templateName, { encoding: 'utf8' });
  const results = linter.verify({ source: templateContents, filePath: templateName });
});
```

As a workaround to get my git commit hook to work, I can ensure that I only create one linter class per `configPath` (by storing a map of `configPath` to linter instance).

### Observed incorrect behavior

When I run the precommit hook on template files in the dummy test app, I get an error that the rule has an invalid config:

```
  -:-  error  The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of strings to whitelist,  * object -- An object with the following keys:,    * `whitelist` -- An array of whitelisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":0,"config":false}`  undefined
```

The override rule config gets more and more corrupted each additional time the linter class is created using the same `configPath`:

```
You specified `{"severity":2,"config":{"severity":2,"config":{"severity":0,"config":false}}}`  undefined
```

Part of the problem is that the imported config file is being directly mutated.

### Failing test case

Failing `get-config-test.js` test case that you can add to demonstrate this:

```js
it('processing config overrides is idempotent', function () {
[...]

PR #1495, opened 2020-09-02
Determine already processed rule configuration

bobisjan:
fixes ember-template-lint/eslint-plugin-hbs#37


bobisjan:
This is an example of the invalid output, when using ESLint plugin:

```
The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of allowlisted strings,  * object -- An object with the following keys:,    * `allowlist` -- An array of allowlisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":0,"config":false}`
The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of allowlisted strings,  * object -- An object with the following keys:,    * `allowlist` -- An array of allowlisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":2,"config":{"severity":0,"config":false}}`
The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of allowlisted strings,  * object -- An object with the following keys:,    * `allowlist` -- An array of allowlisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":2,"config":{"severity":0,"config":false}}`
The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of allowlisted strings,  * object -- An object with the following keys:,    * `allowlist` -- An array of allowlisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":2,"config":{"severity":2,"config":{"severity":0,"config":false}}}`
The no-bare-strings rule accepts one of the following values.
  * boolean - `true` to enable / `false` to disable,  * array -- an array of allowlisted strings,  * object -- An object with the following keys:,    * `allowlist` -- An array of allowlisted strings,    * `globalAttributes` -- An array of attributes to check on every element,    * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name
You specified `{"severity":2,"config":{"severity":2,"config":{"severity":0,"config":false}}}`
```

rwjblue (commented):
Hmm. The thing is we **explicitly** do not want users to be able to have this format in their configuration files. 

In other words, I _think_ that adding this fallback would allow a user in their config to put:

```js
module.exports = {
  rules: {
    'whatever-rule': {
      config: true,
      severity: 1,
    }
  }
};
```

But that is something we want to ensure **always** throws an error.

rwjblue:
@bobisjan - Have you had a chance to see if we can avoid supporting the thing I mentioned above _and_ fix the original issue you ran into?

bobisjan:
@rwjblue, not yet 🙈, I should have a time during the end of today  🤞.

bmish:
This should fix #1412 right?

bobisjan:
@rwjblue, this should be ready for review 🙌.

lib/get-config.js
```diff
@@ -338,6 +338,8 @@ function validateRules(rules, loadedRules, options) {
 
 function validateOverrides(config, options) {
   if (config.overrides) {
+    config.overrides = config.overrides.map((item) => Object.assign({}, item));
```
rwjblue, line 341: It seems vaguely odd to loop over `config.overides` twice. Maybe we could migrate the `for...of` below to be within this `.map`?

Something like:

```js
    config.overrides = config.overrides.map((overrideConfig) => {
      // If there are keys in the object which are not yet supported by overrides functionality, throw an error.
      let overrideKeys = Object.keys(overrideConfig);
      let containsValidKeys = overrideKeys.every((item) => SUPPORTED_OVERRIDE_KEYS.has(item));
      if (!containsValidKeys) {
        throw new Error(
        );
      }
      // clone a deep copy of the override config to ensure it is not mutated
      let clonedResult = JSON.parse(JSON.stringify(overrideConfig));
	
      // TODO: this should be updated to avoid mutation (mutates for deprecated rules), and the mutation should be moved into `processRules` which is expected to return a new value
      validateRules(clonedResult.rules, config.loadedRules, options);

      clonedResult.rules = processRules(clonedResult);

      return clonedResult;
    });
```

rwjblue (commented):
Looking good, one small tweak/suggestion but otherwise this seems great to me!

lib/get-config.js
rwjblue: Note: I added this TODO in the snippet above, should probably be moved into an issue:

> TODO: this should be updated to avoid mutation (mutates for deprecated rules), and the mutation should be moved into `processRules` which is expected to return a new value

rwjblue (approved):

rwjblue:
Thank you!

PR #1497, opened 2020-09-04
skip validation of html attributes

zelaznik:
The following HTML used to raise an error with the rule 'no-whitespace-within-word', even though the
whitespace is on an HTML attribute.  Now there are guardrails to skip the rule when underneath an AttrNode.

```hbs
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" id="Layer_1" x="0px" y="0px" width="22px" height="22px" viewBox="5.0 -10.0 100.0 135.0" enable-background="new 0 0 100 100" xml:space="preserve">
  <path d="M78.096,87.006c2.966,2.977,2.966,7.793,0,10.769c-2.956,2.968-7.761,2.968-10.727,0L19.68,49.999L67.369,2.224  c2.966-2.967,7.771-2.967,10.727,0c2.966,2.977,2.966,7.793,0,10.77l-36.92,37.006L78.096,87.006z"/>
</svg>
```

### DIFF
--- a/lib/rules/no-whitespace-within-word.js
+++ b/lib/rules/no-whitespace-within-word.js
@@ -105,29 +105,11 @@ function splitTextByEntity(input) {
 module.exports = class NoWhitespaceWithinWord extends Rule {
   visitor() {
     return {
-      Program: {
-        enter() {
-          this.attrNodeCount = 0;
-        },
-        exit() {
-          delete this.attrNodeCount;
-        },
-      },
-
-      AttrNode: {
-        enter() {
-          this.attrNodeCount++;
-        },
-        exit() {
-          this.attrNodeCount--;
-        },
-      },
-
-      TextNode(node) {
-        if (this.attrNodeCount > 0) {
+      TextNode(node, path) {
+        let parents = [...path.parents()];
+        if (parents.find((parent) => parent.node.type === 'AttrNode')) {
           return;
         }
-
         let alternationCount = 0;
         let source = this.sourceForNode(node);
         let characters = splitTextByEntity(source);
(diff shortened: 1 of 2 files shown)


zelaznik:
Here's a PR that fixes the problem.

https://github.com/ember-template-lint/ember-template-lint/pull/1497

(There's not a green build, but when I ran the jest test suite on the master branch I got 8 errors.  The new test that I added has passed, and this PR doesn't break anything that wasn't already failing.)

lib/rules/no-whitespace-within-word.js
```diff
@@ -105,7 +105,29 @@ function splitTextByEntity(input) {
...
+        exit() {
+          this.attrNodeCount--;
+        },
+      },
+
       TextNode(node) {
+        if (this.attrNodeCount > 0) {
+          return;
+        }
+
```
josephdsumner, line 130:
```suggestion
      TextNode(node, path) {
        let parents = [...path.parents()];
        if (parents.find((parent) => parent.node.type === 'AttrNode')) {
          return;
        }
```

josephdsumner (commented):
Added an inline suggestion that matches my initial thoughts above.

rwjblue (approved):

rwjblue:
Thanks @zelaznik and @josephdsumner!

lib/rules/no-whitespace-within-word.js
zelaznik: That looks great!  I had no idea that there was the second "path" argument.

Issue #1417, opened 2020-07-14
Rules override is reversed on extends


alias-mac:
When defining multiple rules to extend from, the overrides seem to be reversed.

For example, when defining the following:
```js
extends: ["octane", "some-plugin:recommended"]
```
we would expect that rules defined in `some-plugin:recommended` to override any rules (with same name/scope) from 'octane'.

Instead we need to define `extends` reversed for those overrides:
```js
extends: ["some-plugin:recommended", "octane"]
```

Created a quick codesandbox with a reproducible example with some random rule:
https://codesandbox.io/s/eager-morse-t819p?file=/.template-lintrc.js

Running `yarn lint:hbs` in that sandbox passes, when it should fail because the `octane` rule should override the previous defined rules.

While this might be considered a breaking change (in case someone was using it already with "overrides" first) it doesn't work like eslint `extends` and thus seems to be confusing/unexpected.

rwjblue:
Thank you for reporting this! IMHO, it is **absolutely** a bug, but I worry that fixing it will _effectively_ be a breaking change so we have to be prepared for that (and release appropriately).

Turbo87:
resolved by https://github.com/ember-template-lint/ember-template-lint/pull/1513

PR #1513, opened 2020-09-11
Fix `extends` overriding order

Turbo87:
Previously the first `extends` was taking precedence over the following. This commit changes the priority, so that the later `extends` can override the rule definitions from the previous.

This should hopefully unblock us from releasing v3.0.0.

Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1417

/cc @rwjblue

### DIFF
--- a/test/fixtures/with-multiple-extends/.template-lintrc.js
+++ b/test/fixtures/with-multiple-extends/.template-lintrc.js
@@ -1,9 +1,4 @@
 module.exports = {
-  plugins: [
-    './plugins/plugin1'
-  ],
-  extends: [
-    'recommended',
-    'plugin1:recommended'
-  ]
+  plugins: ['./plugins/plugin1'],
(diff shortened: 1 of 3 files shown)


test/fixtures/with-multiple-extends/.template-lintrc.js
```diff
@@ -1,9 +1,4 @@
...
-  plugins: [
-    './plugins/plugin1'
-  ],
-  extends: [
-    'recommended',
-    'plugin1:recommended'
-  ]
+  plugins: ['./plugins/plugin1'],
+  extends: ['plugin1:recommended', 'recommended'],
 };
```
Turbo87, line 4: sorry about the formatting changes. prettier felt like it was a good idea 😅

Turbo87:
rebased on `next` to enable us to merge this without breaking `master` :)

rwjblue (approved):

PR #1514, opened 2020-09-11
Ensure `bad` test cases **always** have results.

rwjblue:
Prior to this it was possible to leverage the test harness (either by ourselves, or externally) and make passing tests when no issues were logged by a `bad` test case.

For example:

```js
const generateRuleTests = require('../utils/generate-rule-tests');

generateRuleTests({
  name: 'whatever',

  bad: [
    {
      template: `some template here`,
      verifyResults(results) {
        expect(results).toMatchInlineSnapshot(`Array []`);
      },
    }
  ]
```


scalvert:
Thank you.

PR #1521, opened 2020-09-17
Add new rule `no-aria-hidden-body`

### LINKED ISSUE #1506: New Rule Request: `no-hidden-body`
Requesting a new rule to support [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/TR/WCAG21/#name-role-value) - no-hidden-body. 

The `aria-hidden` attribute should never be present on the `<body>` element, as it hides the entire document from assistive technology.

rajasegar:
Fixes #1506

### DIFF
--- a/docs/rule/no-aria-hidden-body.md
+++ b/docs/rule/no-aria-hidden-body.md
@@ -1,0 +1,29 @@
+# no-aria-hidden-body
+
+The aria-hidden attribute should never be present on the `<body>` element, as it hides the entire document from assistive technology.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<body aria-hidden>
+```
+
+```hbs
+<body aria-hidden="true">
+```
+
+This rule **allows** the following:
+
+```hbs
+<body>
+```
+
+```hbs
+{{!-- TODO: Example 2  --}}
+```
+
+## References
+
+* [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/TR/WCAG21/#name-role-value)
--- a/docs/rule/no-aria-hidden-body.md
+++ b/docs/rule/no-aria-hidden-body.md
@@ -20,10 +20,6 @@ This rule **allows** the following:
 <body>
 ```
 
-```hbs
-{{!-- TODO: Example 2  --}}
-```
-
 ## References
 
 * [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/TR/WCAG21/#name-role-value)
--- a/docs/rule/no-aria-hidden-body.md
+++ b/docs/rule/no-aria-hidden-body.md
@@ -24,4 +24,6 @@ This rule **allows** the following:
 
 ## References
 
+* [Using the aria-hidden attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute)
+* [How Lighthouse identifies hidden body elements](https://web.dev/aria-hidden-body/)
 * [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/TR/WCAG21/#name-role-value)
--- a/docs/rule/no-aria-hidden-body.md
+++ b/docs/rule/no-aria-hidden-body.md
@@ -1,5 +1,7 @@
 # no-aria-hidden-body
 
+:wrench: The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+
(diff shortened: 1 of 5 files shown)


dcyriller:
@rajasegar what do you think of adding a fixer that would remove the tag if present ([how-to here](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/fixer.md))?

rajasegar:
Yes @dcyriller sounds like a good idea to me, let me try

rajasegar:
@MelSumner Can you please review this rule?

rajasegar:
@dcyriller Added the fixer here https://github.com/ember-template-lint/ember-template-lint/pull/1521/commits/1160568fa5e77f46693714053e0e49ed030da074

dcyriller (approved):

docs/rule/no-aria-hidden-body.md
```diff
@@ -0,0 +1,27 @@
...
+<body aria-hidden="true">
+```
+
+This rule **allows** the following:
+
+```hbs
+<body>
+```
+
+## References
```
bmish, line 25: Can you add a reference link that explains `aria-hidden`?

test/unit/rules/no-aria-hidden-body-test.js
```diff
@@ -0,0 +1,37 @@
...
+
+const { ERROR_MESSAGE } = require('../../../lib/rules/no-aria-hidden-body');
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-aria-hidden-body',
+
+  config: true,
+
+  good: ['<body></body>', '<body><h1>Hello world</h1></body>'],
```
bmish, line 11: You need to make sure `aria-hidden` is allowed on non-body elements.

lib/rules/no-aria-hidden-body.js
```diff
@@ -0,0 +1,33 @@
...
+  'The aria-hidden attribute should never be present on the <body> element, as it hides the entire document from assistive technology';
+
+module.exports = class NoAriaHiddenBody extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        let hasAriaHiddenBody =
+          node.tag === 'body' && AstNodeInfo.hasAttribute(node, 'aria-hidden');
+        if (hasAriaHiddenBody) {
+          if (this.mode === 'fix') {
```
bmish, line 16: Could autofixing this violation by deleting the attribute lead to a behavior change? Rule autofixers are generally not supposed to cause behavior changes. Couldn't deleting this attribute be dangerous or require additional manual follow-up to resolve?

docs/rule/no-aria-hidden-body.md
zelaznik: This isn't a WCAG reference, but this blog post introduced me to the value of aria-hidden.

https://heydonworks.com/article/aria-label-is-a-xenophobe

lib/rules/no-aria-hidden-body.js
rajasegar: Screen readers and other assistive technologies don't announce content that's marked as hidden. Applying the aria-hidden="true" attribute to your <body> element hides your entire web page from assistive technology users.
 That's what we are trying to avoid.
https://web.dev/aria-hidden-body/

test/unit/rules/no-aria-hidden-body-test.js
rajasegar: done

docs/rule/no-aria-hidden-body.md
rajasegar: done

lib/rules/no-aria-hidden-body.js
bmish: Yes but the question is whether it's completely safe to autofix this or not, or if the user should manually intervene to fix and verify this.

rajasegar: I assume it's completely safe to autofix, because nobody with the right mind would want to hide the page from the screenreader, @MelSumner Your thoughts?

bmish: I'll merge this but we may want to reconsider this.

rajasegar: Sure @bmish I am fine with it. Thanks a lot 👍

test/unit/rules/no-aria-hidden-body-test.js
seemajune: thank you!

PR #1525, opened 2020-09-19
Add new rule `require-each-key`

### LINKED ISSUE #1520: New rule request: require-each-key
We were having huge performance issues in our projects when rendering table rows.

In our case, a single change to a single table cell was causing **all rows** to rerender, which was blocking the main thread for up to more than a second.

After an investigation I found out that it was a problem with how we use `{{#each}}`.

After we added [`key`](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each#specifying-keys) to all our `each`es, Ember stopped rerendering all rows, but only the ones that actually needed to be updated.

For example, in one of the cases (selecting a table row via clicking a checkbox) resulted in a 10x performance increase (from ~300 ms down to ~30 ms).

This is applicable not only to tables, but to any usage of `{{#each}}`.

I would like to have a rule that will require to always use `key` with `{{#each}}`.

Examples of **valid** code:

```hbs
{{#each this.items key="id" as |item|}}
  {{item.name}}
{{/each}}
```

```hbs
{{#each this.items key="deeply.nested.id" as |item|}}
  {{item.name}}
{{/each}}
```

```hbs
{{#each this.items key="@index" as |item|}}
  {{item.name}}
{{/each}}
```

```hbs
{{#each this.items key="@identity" as |item|}}
  {{item.name}}
{{/each}}
```

Examples of **invalid** code:

```hbs
{{#each this.items as |item|}}
  {{item.name}}
{{/each}}
```

```hbs
{{#each this.items key="@invalid" as |item|}}
  {{item.name}}
{{/each}}
```

rajasegar:
Fixes #1520

### DIFF
--- a/docs/rule/require-each-key.md
+++ b/docs/rule/require-each-key.md
@@ -1,0 +1,54 @@
+# require-each-key
+
+You might be having huge performance issues when rendering large data-sets with `{{#each}}`
+
+In order to avoid that, it is recommended to use `key` with `{{#each}}` because, Ember will not rerendering all items, but only the ones that actually needed to be updated.
+
+This rule will require to always use `key` with `{{#each}}`.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+{{#each this.items as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+```hbs
+{{#each this.items key="@invalid" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+This rule **allows** the following:
+
+```hbs
+{{#each this.items key="id" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+```hbs
+{{#each this.items key="deeply.nested.id" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+```hbs
+{{#each this.items key="@index" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+```hbs
+{{#each this.items key="@identity" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
+## References
+
+* [Specifying Keys](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each#specifying-keys)
+* [The Immutable Pattern in Tracked Properties](https://glimmerjs.com/guides/tracked-properties)
--- a/docs/rule/require-each-key.md
+++ b/docs/rule/require-each-key.md
@@ -1,8 +1,8 @@
 # require-each-key
 
-You might be having huge performance issues when rendering large data-sets with `{{#each}}`
+You might be having performance issues when rendering data-sets with `{{#each}}`
 
-In order to avoid that, it is recommended to use `key` with `{{#each}}` because, Ember will not rerendering all items, but only the ones that actually needed to be updated.
+In order to avoid that, it is recommended to use `key` with `{{#each}}` because, Ember will not be re-rendering all the items, but only the items that actually need to be updated.
 
 This rule will require to always use `key` with `{{#each}}`.
 
--- a/docs/rule/require-each-key.md
+++ b/docs/rule/require-each-key.md
@@ -44,6 +44,12 @@ This rule **forbids** the following:
 {{/each}}
 ```
 
+```hbs
+{{#each this.items key="" as |item|}}
+  {{item.name}}
+{{/each}}
+```
+
 This rule **allows** the following:
 
 ```hbs
--- a/docs/rule/require-each-key.md
+++ b/docs/rule/require-each-key.md
@@ -1,8 +1,30 @@
 # require-each-key
 
-You might be having performance issues when rendering data-sets with `{{#each}}`
+In order to improve rendering speed, Ember will try to reuse the DOM elements where possible. Specifically, if the same item is present in the array both before and after the change, its DOM output will be reused.
 
-In order to avoid that, it is recommended to use `key` with `{{#each}}` because, Ember will not be re-rendering all the items, but only the items that actually need to be updated.
+The key option is used to tell Ember how to determine if the items in the array being iterated over with {{#each}} has changed between renders. By default the item's object identity is used.
+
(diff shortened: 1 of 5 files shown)


rajasegar:
@bmish @EvgenyOrekhov Can you please review this rule?

lib/rules/require-each-key.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = class RequireEachKey extends Rule {
+  visitor() {
+    return {
+      BlockStatement(node) {
+        let isEach = AstNodeInfo.isEach(node);
+        let keyPair = node.hash.pairs.find((p) => p.key === 'key');
+        let keyValue = keyPair && keyPair.value && keyPair.value.value;
+        let isSpecialKey = keyValue && keyValue.startsWith('@');
+        let isValidKey = isSpecialKey ? SPECIAL_KEY_VALUES.has(keyValue) : true;
```
EvgenyOrekhov, line 18: I don't think it's necessary to check for `@`, just `SPECIAL_KEY_VALUES.has(keyValue)` should be enough.

lib/rules/require-each-key.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
+
+const ERROR_MESSAGE = '{{#each}} block requires a valid key value to avoid performance issues';
```
EvgenyOrekhov, line 6: The [documentation](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each) refers to `{{#each}}` as "helper", not "block".

docs/rule/require-each-key.md
```diff
@@ -0,0 +1,54 @@
+# require-each-key
+
+You might be having huge performance issues when rendering large data-sets with `{{#each}}`
```
EvgenyOrekhov, line 3: We were having huge performance issues (up to more than a second blocking time for a single click event) when rendering only 10 table rows. I don't think that counts as "large" :)

docs/rule/require-each-key.md
```diff
@@ -0,0 +1,54 @@
...
+
+In order to avoid that, it is recommended to use `key` with `{{#each}}` because, Ember will not rerendering all items, but only the ones that actually needed to be updated.
```
EvgenyOrekhov, line 5: > because, Ember will not rerendering all items, but only the ones that actually needed to be updated.

Something is wrong with wording and punctuation here.

This sounds better to me:

> so that Ember would not rerender all items, but only the ones that actually need to be updated.

But better to consult with a native English speaker :)

EvgenyOrekhov (commented):
I'm not familiar with ember-template-lint API and project standards, but it looks good.

test/unit/rules/require-each-key-test.js
```diff
@@ -0,0 +1,38 @@
...
+generateRuleTests({
+  name: 'require-each-key',
+
+  config: true,
+
+  good: [
+    '{{#each this.items key="id" as |item|}} {{item.name}} {{/each}}',
+    '{{#each this.items key="deeply.nested.id" as |item|}} {{item.name}} {{/each}}',
+    '{{#each this.items key="@index" as |item|}} {{item.name}} {{/each}}',
+    '{{#each this.items key="@identity" as |item|}} {{item.name}} {{/each}}',
```
bmish, line 15: Need to make sure that this rule only applies to `each` by adding this as a valid case:

```
{{#if foo}}{{/if}}
```

test/unit/rules/require-each-key-test.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      template: '{{#each this.items key="@invalid" as |item|}} {{item.name}} {{/each}}',
+      result: {
+        message: ERROR_MESSAGE,
+        line: 1,
+        column: 0,
+        source: '{{#each this.items key="@invalid" as |item|}} {{item.name}} {{/each}}',
+      },
+    },
```
bmish, line 36: Can you add an invalid test case with `key=""`?

lib/rules/require-each-key.js
rajasegar: Changed block to helper

docs/rule/require-each-key.md
rajasegar: Changed the text accordingly

docs/rule/require-each-key.md
rajasegar: Yes, I agree with you, this needs to have better wording, @MelSumner Can you please help?

test/unit/rules/require-each-key-test.js
rajasegar: done

test/unit/rules/require-each-key-test.js
rajasegar: done

EvgenyOrekhov (approved):
Perhaps the `key=""` example should be added to `require-each-key.md`? Otherwise, LGTM.

docs/rule/require-each-key.md
```diff
@@ -0,0 +1,82 @@
+# require-each-key
+
+In order to improve rendering speed, Ember will try to reuse the DOM elements where possible. Specifically, if the same item is present in the array both before and after the change, its DOM output will be reused.
+
+The key option is used to tell Ember how to determine if the items in the array being iterated over with {{#each}} has changed between renders. By default the item's object identity is used.
+
+This is usually sufficient, so in most cases, the key option is simply not needed. However, in some rare cases, the objects' identities may change even though they represent the same underlying data.
```
EvgenyOrekhov, line 7: I don't have the actual stats, but it seems to me that "in most cases" and "in some rare cases" are exaggerations. Managing data in an immutable fashion is trending, many frameworks/state management libraries require immutability, and even Glimmer docs recommend using "The Immutable Pattern". The current wording might make the problem look small and irrelevant. But for us, the problem was extensive and severe.

bmish (approved):

PR #1532, opened 2020-09-24
Add `s` and `u` elements to `no-obsolete-elements` rule elements list

bmish:
Adding two elements that we missed from: https://dev.w3.org/html5/pf-summary/obsolete.html

### DIFF
--- a/docs/rule/no-obsolete-elements.md
+++ b/docs/rule/no-obsolete-elements.md
@@ -35,9 +35,11 @@ This rule **forbids** the following:
 <plaintext></plaintext>
 <rb></rb>
 <rtc></rtc>
+<s></s>
 <spacer></spacer>
 <strike></strike>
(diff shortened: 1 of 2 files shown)


MelSumner (changes requested):
So that website is well-known for having incorrect information. If you can find references in the spec that indicate that these elements are deprecated, please add the spec reference URL in the doc and ask me to re-review. 

Thank you!

bmish:
@MelSumner good point, I updated to only add two additional elements `s` and `u` that we were missing from the official w3 site: https://dev.w3.org/html5/pf-summary/obsolete.html

rwjblue:
@bmish can you update to target `next` branch?

rwjblue (approved):

MelSumner (approved):

PR #1533, opened 2020-09-24
Add missing `keygen` element to `no-obsolete-elements` rule elements list

bmish:
This is a bug since this element was listed in the rule documentation but not actually included in the rule implementation.

This element can be found in the spec: https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features


MelSumner (approved):
Thank you for adding this!

PR #1541, opened 2020-09-25
Allow nested landmarks of different types in `no-nested-landmark` rule

### LINKED ISSUE #1542: (bug) false positive when landmarks of different types are nested
The following code raises an error with the role `no-nested-landmark`.

```hbs
<footer>
  <nav>
  </nav>
</footer>
```

I'm attaching a pull request that I believe fixes the problem:

https://github.com/ember-template-lint/ember-template-lint/pull/1541

zelaznik:
The rule `no-nested-landmark` raised a false error in our template when a `<nav>` tag was nested inside of a `<footer>`.  When I looked through the code I realized that the rule checks for ANY landmark inside of any other landmark.  This bugfix makes sure that it checks for landmarks of the same type.  I added a helper method `getLandmarkType` that will return either the role of the element, or the tag name if the role is missing.

Fixes #1542.

### DIFF
--- a/test/unit/rules/no-nested-landmark-test.js
+++ b/test/unit/rules/no-nested-landmark-test.js
@@ -14,9 +14,9 @@ generateRuleTests({
     '<div><main></main></div>',
     '<div role="application"><div role="document"><div role="application"></div></div></div>',
     '<footer><nav></nav></footer>', // nested landmarks of different types are okay
-    '<div role="footer"><nav></nav></div>',
+    '<div role="contentinfo"><nav></nav></div>',
     '<footer><div role="navigation"></div></footer>',
-    '<div role="footer"><div role="navigation"></div></div>',
+    '<div role="contentinfo"><div role="navigation"></div></div>',
   ],
 
   bad: [
--- a/test/unit/rules/no-nested-landmark-test.js
+++ b/test/unit/rules/no-nested-landmark-test.js
@@ -13,10 +13,10 @@ generateRuleTests({
   good: [
     '<div><main></main></div>',
     '<div role="application"><div role="document"><div role="application"></div></div></div>',
-    '<footer><nav></nav></footer>', // nested landmarks of different types are okay
-    '<div role="contentinfo"><nav></nav></div>',
-    '<footer><div role="navigation"></div></footer>',
-    '<div role="contentinfo"><div role="navigation"></div></div>',
+    '<header><nav></nav></header>', // nested landmarks of different types are okay
+    '<div role="banner"><nav></nav></div>',
+    '<header><div role="navigation"></div></header>',
+    '<div role="banner"><div role="navigation"></div></div>',
   ],
 
   bad: [
--- a/test/unit/rules/no-nested-landmark-test.js
+++ b/test/unit/rules/no-nested-landmark-test.js
@@ -15,8 +15,8 @@ generateRuleTests({
     '<div role="application"><div role="document"><div role="application"></div></div></div>',
     '<footer><nav></nav></footer>', // nested landmarks of different types are okay
     '<div role="footer"><nav></nav></div>',
-    '<footer><div role="nav"></div></footer>',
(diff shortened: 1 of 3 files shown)


bmish:
The code change looks good, but is there any website/source that can verify that nested landmarks of different types are generally acceptable?

zelaznik:
@bmish There is.  Credit to my coworker @delusioninabox for finding it:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

In the Repeated Landmarks section of the page, they show this pattern:


```html
<header>
  <nav id="main-nav" aria-label="Main">
    <!-- list of links to main website locations -->
  </nav>
</header>

<!-- ... other content here -->

<footer>
  <nav id="footer-nav" aria-label="Main">
    <!-- list of links to main website locations -->
  </nav>
</footer>
```

bmish:
@zelaznik thanks for finding that. Should we make the rule more specific about which landmarks are allowed to be nested inside each other? Or is it really the case that any landmark can be nested inside any other landmark as long as they are different landmarks?

zelaznik:
I'm not a semantic HTML expert, but it looks like any landmark can be nested inside any other landmark.  There are some patterns that are frowned upon but not explicitly a violation:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

> It's not necessary for all links to be contained in a \<nav\> element. \<nav\> is intended only for major block of navigation links; typically the \<footer\> element often has a list of links that don't need to be in a \<nav\> element.

I was also going off of the readme in the `no-nested-landmarks` rule which explicitly uses the phrase "of the same name", which is why I labelled this PR as a bug rather than an enhancement or feature request.

https://github.com/ember-template-lint/ember-template-lint/blob/v2.13.0/docs/rule/no-nested-landmark.md

> Landmark elements should not be nested within landmark elements **_of the same name_**.

zelaznik:
@bmish I just realized a problem with my own pull request.

When I'm checking whether two elements have equivalent roles, the following scenarios would currently be a false negative:

```hbs
<nav>
  <div role="navigation"></div>
</nav>
```

This is because I naively checked whether the tag name matches the aria role.  The tag name "nav" doesn't match the aria-role "navigation."

The description in the readme of the rule provides the appropriate isomorphic map between tag names and aria-roles.

https://github.com/ember-template-lint/ember-template-lint/blob/v2.13.0/docs/rule/no-nested-landmark.md

```
header (banner)
main (main)
aside (complementary)
form (form, search)
main (main)
nav (navigation)
footer (contentinfo)
```

I need to go back and implement that mapping in my code.

zelaznik:
@bmish Now it's all fixed, ready for review again.

test/unit/rules/no-nested-landmark-test.js
```diff
@@ -13,6 +13,10 @@ generateRuleTests({
   good: [
     '<div><main></main></div>',
     '<div role="application"><div role="document"><div role="application"></div></div></div>',
+    '<footer><nav></nav></footer>', // nested landmarks of different types are okay
```
MelSumner, line 16: While this is technically okay, it's not great UX. You'd also need to make sure that a nav in a footer has some sort of accessible name, because if the nav is in the footer there is likely another nav somewhere on the page.

zelaznik: @MelSumner are you saying that nested landmarks in general are not great UX, or are you saying that a nav nested in a footer is not great UX?  If it's the latter, I'll be happy to change the passing examples in the test suite.  If it's the former then I should probably just close the issue and the PR.

MelSumner (approved):
Since it's technically correct, we should allow it for now; however we should figure out whether or not our linting rules should help guide to updated practices for better UX. @rwjblue @bmish @Turbo87 let's have that conversation at some point. 

@zelaznik thank you for doing this work! 👍🏻

Issue #1530, opened 2020-09-23
should "require-input-label" apply to other elements too?


zelaznik:
# Summary:

I'm using axe-core as the gold standard:

https://dequeuniversity.com/rules/axe/3.5/label

It specifically says:

```
Form elements that should have labels
  - Text entry fields, e.g. <input type="text">, <input type="password"> and <textarea>
  - Radio buttons, <input type="radio">
  - Checkboxes, <input type="checkbox">
  - Select menus, <select>
```

I noticed this discrepancy when I was finding some axe errors that didn't raise lint errors.

# Possible implementations:

  - Include the `<textarea>` and `<select>` tags in this rule.
  - Create parallel rules for these components

If you think this is a worthwhile enhancement, let me know.  I'll put a PR together.

zelaznik:
https://github.com/ember-template-lint/ember-template-lint/pull/1549

PR #1549, opened 2020-09-29
Add `<textarea>` and `<select>` support to `require-input-label` rule

zelaznik:
The requirement for input labels doesn't only include the `<input>` tag.  It also applies to `<textarea>` and `<select>` tags.

https://dequeuniversity.com/rules/axe/3.5/label?application=axeAPI

Fixes issue: https://github.com/ember-template-lint/ember-template-lint/issues/1530


lib/rules/require-input-label.js
```diff
@@ -3,8 +3,8 @@
 const AstNodeInfo = require('../helpers/ast-node-info');
 const Rule = require('./base');
 
-const ERROR_MESSAGE_NO_LABEL = 'Input elements require a valid associated label.';
-const ERROR_MESSAGE_MULTIPLE_LABEL = 'Input elements should not have multiple valid labels.';
+const ERROR_MESSAGE_NO_LABEL = 'elements require a valid associated label.';
```
MelSumner, line 6:
```suggestion
const ERROR_MESSAGE_NO_LABEL = 'form elements require a valid associated label.';
```

lib/rules/require-input-label.js
```diff
@@ -3,8 +3,8 @@
...
+const ERROR_MESSAGE_MULTIPLE_LABEL = 'elements should not have multiple valid labels.';
```
MelSumner, line 7:
```suggestion
const ERROR_MESSAGE_MULTIPLE_LABEL = 'form elements should not have multiple labels.';
```

MelSumner (changes requested):
I think we could simplify a little bit by putting clarification in the error message. What do you think?

zelaznik:
I'm okay with that.  I don't have strong feelings about the copy in the error message.  I went ahead and committed your suggestions as well as reverted the commit in question.

MelSumner (approved):
Thank you for improving this linting rule!

zelaznik:
My pleasure

docs/rule/require-input-label.md
```diff
@@ -2,11 +2,11 @@
...
 
-This rule checks to see if the `<input>` element is contained by a label element. If it is not, it checks to see if the `<input>` element has any of these three attributes: `id`, `aria-label`, or `aria-labelledby`. While the `id` element on the `<input>` is not a concrete indicator of the presence of an associated `<label>` element with a `for` attribute, it is a good indicator that one likely exists.
+This rule checks to see if the input element (either `<input>`, `<textarea>`, or `<select>`) element is contained by a label element. If it is not, it checks to see if the input has any of these three attributes: `id`, `aria-label`, or `aria-labelledby`. While the `id` element on the input is not a concrete indicator of the presence of an associated `<label>` element with a `for` attribute, it is a good indicator that one likely exists.
```
bmish, line 5: Can you move the list of elements to an actual list below this paragraph? It's much easier to glance at a list and it will be easier for us to update in the future too.

```
* `<input>`
* `<textarea>`
* `<select>`
```

bmish:
This is a breaking change right? If so, please change the target branch to `next` (from github UI), so we can include it in the major release that is coming soon.

zelaznik:
@bmish I'm new to this repository so I don't know best practices.  Should I rebase from the `next` branch?

zelaznik:
I had a whole host of conflicts when I tried to change the base branch of this PR to `ember-template-lint:next`.  As a result I went back and did an interactive rebase from `ember-template-lint:next`.  As a precaution, I made a [backup](https://github.com/zelaznik/ember-template-lint/tree/__backup__require-labels-on-textarea-and-select-tags) of my original work.

PR #1550, opened 2020-09-30
Adds new rule: `no-duplicate-landmark-elements`

MelSumner:
This is a little bit of a different take on https://github.com/ember-template-lint/ember-template-lint/pull/1327 by @rajasegar 
(I tried to commit to that branch but wasn't able to).


docs/rule/require-landmark-labels.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+```hbs
+<form role="search" aria-label="search"></form>
+<form aria-labelledby="form-title"><div id="form-title">Meaningful Form Title</div></form>
+```
+
+## References
+
+- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
```
MelSumner, line 63:
```suggestion
- [WAI-ARIA specification: Landmark Roles](https://www.w3.org/WAI/PF/aria/roles#landmark_roles)
- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
```

MelSumner:
@rwjblue

- updated the rule name (and all references)
- added it to the a11y.js config file
- added @rajasegar as a co-author (since this will replace #1327 )

I think it's ready to go! 👍🏻

rwjblue (approved):

rrenno:
Hi @MelSumner I was wondering if you could help clear up some confusion we have about this rule. 
For example we have:

```hbs
    {{#if ....}}
        <div role="alert">
            Failed to build
        </div>
    {{else if .....}}
        ...
        <p role="alert">{{message}}</p>
        ...
    {{else}}
        .....
        <div role="alert">
            {{#if explanationMessage}}
            ....
```

The rule will flag the multiple uses of `role="alert"`. The 'solution' seems to be adding `aria-label/aria-labelledby` to each tag, but my question is would the the content of the tags not be sufficient -- e.g the message of the p tag?

Otherwise it seems like we would end up with `<p role="alert" aria-label={{message}}>{{message}}</p>` which is a bit repetitive. Is this just a limitation of the linter or are we doing something fundamentally at odds with a11y? My understanding is a screen reader would correctly pick up the message.

Issue #1551, opened 2020-09-30
Migrate JS API to async/await


rwjblue:
All of our primary JS APIs should be updated to be async. Specifically this would include the following methods:

* `verify`
* `verifyAndFix`

This is a breaking change, and should target the `next` branch for 3.0.

rwjblue:
Closed by #1553

PR #1553, opened 2020-10-01
Convert `Linter.prototype.verify` and `Linter.prototype.verifyAndFix` to be async.

scalvert:
Converts `verify` and `verifyAndFix` APIs to async/await. This allows for the addition, or usage of, other async APIs (specifically, the [new `pending` APIs](https://github.com/ember-template-lint/ember-template-lint-pending-utils) that have been build to enable a more robust `pending` system).

Implements #1551 

---

Note: for consumers of _only_ the command line invocation, this is not breaking (that script was already async), this is only a breaking change for consumers of the JS API that are doing things like:

```js
const TemplateLinter = require('ember-template-lint');
let linter = new TemplateLinter();
let results = linter.verify({ source, filePath: 'app/foo/bar/template.hbs' });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ let linter = new TemplateLinter();
 let template = fs.readFileSync('some/path/to/template.hbs', {
   encoding: 'utf8',
 });
-let results = linter.verify({ source: template, moduleId: 'template.hbs' });
+let results = await linter.verify({ source: template, moduleId: 'template.hbs' });
 ```
 
 `results` will be an array of objects which have the following properties:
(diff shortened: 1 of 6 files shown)


rwjblue (approved):

rwjblue:
Note: for consumers of _only_ the command line invocation, this is not breaking at all (that script was already async), this is only a breaking change for consumers of the JS API that are doing things like:

```js
const TemplateLinter = require('ember-template-lint');
let linter = new TemplateLinter();
let results = linter.verify({ source, filePath: 'app/foo/bar/template.hbs' });
```

PR #1554, opened 2020-10-01
Ensure globbing is only used when using a globlike pattern.

rwjblue:
Within the command line script's `expandFileGlobs` method we were using `is-valid-glob` to determine if a given possible pattern / filename string was in fact a pattern or not. The idea here is that if a possible pattern does not contain any glob related wildcards or matchers **and** it is a literal file (ending with `.hbs`) it should be used directly and we should avoid globbing.

The most common scenario for this is when using built-in editor tooling to lint _just_ the current file. In that scenario we **never** want to waste the overhead of globbing, since we were given an explicit path.

Unfortunately, the previous code used `is-valid-glob` and therefore **always** considered any pattern as "valid" since `isValidGlob` is _basically_ just (see [actual code here](https://github.com/micromatch/is-valid-glob/blob/c18c82999df11133c5946f9a21ef74a0ebf88334/index.js#L3-L11)):

```js
function isValidGlob(pattern) {
  return Boolean(pattern) && (Array.isArray(pattern) || typeof pattern === 'string');
}
```

The package we _should have used_ is [is-glob](https://www.npmjs.com/package/is-glob), which is actually checking the contents of the string to see if it contains globlike matchers.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -11,7 +11,7 @@ const { promisify } = require('util');
 
 const getStdin = require('get-stdin');
 const globby = require('globby');
-const isValidGlob = require('is-valid-glob');
+const isGlob = require('is-glob');
 const micromatch = require('micromatch');
 
 const Linter = require('../lib');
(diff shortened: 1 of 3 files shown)


bin/ember-template-lint.js
```diff
@@ -49,15 +49,24 @@ function lintSource(linter, options, shouldFix) {
...
+function expandFileGlobs(filePatterns, ignorePattern, glob = executeGlobby) {
   let result = new Set();
 
   filePatterns.forEach((pattern) => {
     let isHBS = pattern.slice(-4) === '.hbs';
-    let isLiteralPath = !isValidGlob(pattern) && fs.existsSync(pattern);
+    let isLiteralPath = !isGlob(pattern) && fs.existsSync(pattern);
 
     if (isHBS && isLiteralPath) {
-      let isIgnored = !micromatch.isMatch(pattern, ignorePattern);
```
rwjblue, line 60: This logic was just wrong, this is explicitly matching against the ignore pattern so it should not be negated.

test/unit/bin/expand-file-globs-test.js
```diff
@@ -23,15 +23,14 @@ describe('expandFileGlobs', function () {
       project.write({ 'application.hbs': 'almost empty' });
 
       let files = expandFileGlobs(['application.hbs', 'other.hbs'], []);
-      expect(files.has('application.hbs')).toBe(true);
-      expect(files.has('other.hbs')).toBe(false);
+      expect(files).toEqual(new Set(['application.hbs']));
```
rwjblue, line 26: These assertion changes are making the test a bit stricter (matching on exact expected contents).

test/unit/bin/expand-file-globs-test.js
```diff
@@ -47,6 +46,18 @@ describe('expandFileGlobs', function () {
       expect(files.has('application.hbs')).toBe(true);
     });
 
+    it('does not fallback to globbing if not passed a globlike string', function () {
```
rwjblue, line 49: This test failed prior to the changes (e.g. we were globbing in **all** cases).

PR #1559, opened 2020-10-02
fix: update require-valid-alt-text rule to have same type values #1555

### LINKED ISSUE #1555: Request: update `require-valid-alt-text`
Request to update the `require-valid-alt-text` rule to add an additional check: if the element's `src` is a dynamic value, ensure that the `alt` also has a dynamic value. 

Good: 
```hbs
<img src={{image}}  alt={{imageAlt}} />
```

Bad:
```hbs
<img src={{image}} alt="the team poses as a group at the fall company picnic" />
```

Add these as additional references to the documentation: 
- https://www.w3.org/WAI/WCAG21/Techniques/failures/F20
- https://www.w3.org/WAI/WCAG21/Understanding/name-role-value

rajasegar:
Fixes #1555

### DIFF
--- a/lib/rules/require-valid-alt-text.js
+++ b/lib/rules/require-valid-alt-text.js
@@ -64,6 +64,14 @@ module.exports = class RequireValidAltText extends Rule {
             }
           }
 
+          // src value is dynamic and alt value is not
+          if (srcValue && srcValue.type && altValue && !altValue.type) {
+            this.logNode({
+              message: 'The `alt` attribute value must be of the same type as `src` attribute',
+              node,
+            });
+          }
+
           if (!hasAltAttribute) {
             this.logNode({
               message: 'All `<img>` tags must have an alt attribute',
--- a/test/unit/rules/require-valid-alt-text-test.js
+++ b/test/unit/rules/require-valid-alt-text-test.js
@@ -48,6 +48,7 @@ generateRuleTests({
     '<area aria-labelledby="some-alt">',
     '<area aria-label="some-alt">',
     '<img role={{unless this.altText "presentation"}} alt={{this.altText}}>',
+    '<img src={{image}} alt={{imageAlt}}>',
   ],
 
   bad: [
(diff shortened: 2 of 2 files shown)


lib/rules/require-valid-alt-text.js
```diff
@@ -64,6 +64,14 @@ module.exports = class RequireValidAltText extends Rule {
             }
           }
 
+          // src value is dynamic and alt value is not
+          if (srcValue && srcValue.type && altValue && !altValue.type) {
+            this.logNode({
+              message: 'The `alt` attribute value must be of the same type as `src` attribute',
```
lifeart, line 70: this may be little bit wrong for case like:

```hbs
{{each this.levels as |level index|}}
  <img src={{level.src}} alt={{concat "Photo of car with index" index}} />
{{/each}}
```

rwjblue: Ya, good call out here.

MelSumner:
@rajasegar apologies for the delay- could you please refresh this branch and review the comments on the PR? Thank you!

rajasegar:
@MelSumner Are we clear on how we going to proceed for the case mentioned above by @lifeart 
```hbs
{{#each this.levels as |level index|}}
  <img src={{level.src}} alt="{{concat "Photo of car with index" index}}" />
{{/each}
```
because the value types are different here: MustacheStatement and ConcatStatement
Also we have clashes with existing tests here
https://github.com/ember-template-lint/ember-template-lint/blob/1722f4f5734ec66db2119cd6285408bfa176eef8/test/unit/recommended-config-test.js#L50

faith-or:
@rajasegar -- Didn't realized that this issue was being worked on. Feel free to checkout https://github.com/ember-template-lint/ember-template-lint/pull/1973. 

I think for  https://github.com/ember-template-lint/ember-template-lint/blob/1722f4f5734ec66db2119cd6285408bfa176eef8/test/unit/recommended-config-test.js#L50 

We can update it to: 
```
ensureValid('<img alt={{this.alt}} src={{some-dir/some-thing this.x}}>');
```

What do you think @MelSumner ?

PR #1560, opened 2020-10-02
Updated `is-interactive-element` helper and related rules

MelSumner:
Reading through the HTML spec led me to check our `is-interactive-element` helper, which I found had some inaccuracies and incomplete rules. I've updated the helper and the related rules.

### DIFF
--- a/docs/rule/no-invalid-interactive.md
+++ b/docs/rule/no-invalid-interactive.md
@@ -5,6 +5,12 @@
 Adding interactivity to an element that is not naturally interactive content leads to a very poor experience for
 users of assistive technology (i.e. screen readers). In order to ensure that screen readers can provide useful information to their users, we should add an appropriate `role` attribute when the underlying element would not have made that role obvious.
 
+Some complexities of note:
+
+Both a `tabindex` attribute and an interactive role is required to make a custom element interactive; otherwise it is a non-interactive element and not eligible to have an interaction added.
+
+The `<canvas>` element is not an interactive element, but may contain certain interactive elements as descendants. From the HTML spec: `a` elements, `img` elements with `usemap` attributes, `button` elements, `input` elements whose `type` attribute are in the Checkbox or Radio Button states, `input` elements that are buttons, and `select` elements with a multiple attribute or a display size greater than 1.
+
 ## Examples
 
 This rule **forbids** the following:
@@ -13,19 +19,23 @@ This rule **forbids** the following:
 <div {{action 'foo'}}></div>
 ```
 
+```hbs
+<div role="button" {{action "foo"}}></div> //missing tabindex
+```
+
 This rule **allows** the following:
 
 ```hbs
-<div role="button" {{action "foo"}}></div>
+<div role="button" tabindex="0" {{action "foo"}}></div>
 ```
 
 ## Configuration
 
 The following values are valid configuration (same as the `no-nested-interactive` rule above):
 
-* boolean -- `true` indicates all whitelist test will run, `false` indicates that the rule is disabled.
+* boolean -- `true` indicates all allowed tests will run, `false` indicates that the rule is disabled.
 * object - Containing the following values:
-  * `ignoredTags` - An array of element tag names that should be whitelisted. Default to `[]`.
+  * `ignoredTags` - An array of element tag names that should be allowed. Default to `[]`.
   * `ignoreTabindex` - When `true` tabindex will be ignored. Defaults to `false`.
   * `ignoreUsemapAttribute` - When `true` ignores the `usemap` attribute on `img` and `object` elements. Defaults `false`.
   * `additionalInteractiveTags` - An array of element tag names that should also be considered as interactive. Defaults to `[]`.
@@ -33,3 +43,4 @@ The following values are valid configuration (same as the `no-nested-interactive
 ## References
 
 * [MDN docs/ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)
+* [the canvas element](https://html.spec.whatwg.org/#the-canvas-element)
--- a/lib/helpers/is-interactive-element.js
+++ b/lib/helpers/is-interactive-element.js
@@ -2,9 +2,11 @@
 
 const ast = require('./ast-node-info');
 
+// Spec: https://html.spec.whatwg.org/#interactive-content
 const INTERACTIVE_TAG_NAMES = new Set([
+  'a',
+  'audio',
   'button',
-  'canvas',
   'details',
   'embed',
   'iframe',
@@ -13,10 +15,10 @@ const INTERACTIVE_TAG_NAMES = new Set([
   'label',
   'select',
   'textarea',
+  'video',
 ]);
 
 // Spec: https://www.w3.org/TR/wai-aria/complete#widget_roles
-
 const ARIA_WIDGET_ROLES = new Set([
   'button',
   'checkbox',
@@ -37,10 +39,6 @@ const ARIA_WIDGET_ROLES = new Set([
 
 function isAudioVideoWithControls(node) {
   return (node.tag === 'audio' || node.tag === 'video') && ast.hasAttribute(node, 'controls');
-}
-
-function isHyperLink(node) {
-  return node.tag === 'a' && ast.hasAttribute(node, 'href');
 }
 
 function isHiddenInput(node) {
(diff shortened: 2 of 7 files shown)


docs/rule/no-invalid-interactive.md
```diff
@@ -5,6 +5,12 @@
 Adding interactivity to an element that is not naturally interactive content leads to a very poor experience for
 users of assistive technology (i.e. screen readers). In order to ensure that screen readers can provide useful information to their users, we should add an appropriate `role` attribute when the underlying element would not have made that role obvious.
 
+Some complexities of note:
+
+Both a `tabindex` attribute and an interactive role is required to make a custom element interactive; otherwise it is a non-interactive element and not eligible to have an interaction added.
+
+The `<canvas>` element is not an interactive element, but may contain certain interactive elements as descendants. From the HTML spec: `a` elements, `img` elements with `usemap` attributes, `button` elements, `input` elements whose `type` attribute are in the Checkbox or Radio Button states, `input` elements that are buttons, and `select` elements with a multiple attribute or a display size greater than 1.
```
rwjblue, line 12: The "From the HTML spec" bit should link to the section for each of these elements for easier reading / verification.

docs/rule/no-invalid-interactive.md
```diff
@@ -5,6 +5,12 @@
 Adding interactivity to an element that is not naturally interactive content leads to a very poor experience for
 users of assistive technology (i.e. screen readers). In order to ensure that screen readers can provide useful information to their users, we should add an appropriate `role` attribute when the underlying element would not have made that role obvious.
 
+Some complexities of note:
+
+Both a `tabindex` attribute and an interactive role is required to make a custom element interactive; otherwise it is a non-interactive element and not eligible to have an interaction added.
```
rwjblue, line 10: Link to spec?

lib/rules/no-nested-interactive.js
```diff
@@ -19,6 +19,7 @@
 const isInteractiveElement = require('../helpers/is-interactive-element');
 const parseConfig = require('../helpers/parse-interactive-element-config');
 const Rule = require('./base');
+// const { ERROR_MESSAGE } = require('./no-action');
```
rwjblue, line 22: This is odd, we should either bring it back or delete the commented file.

lib/helpers/is-interactive-element.js
```diff
@@ -2,9 +2,11 @@
 
 const ast = require('./ast-node-info');
 
+// Spec: https://html.spec.whatwg.org/#interactive-content
 const INTERACTIVE_TAG_NAMES = new Set([
+  'a',
```
rwjblue, line 7: This is incorrect. The spec link pasted just above specifically says that `<a>` is only interactive with an `href`:

> a (if the href attribute is present)audio (if the controls attribute is present)buttondetailsembediframeimg (if the usemap attribute is present)input (if the type attribute is not in the Hidden state)labelobject (if the usemap attribute is present)selecttextareavideo (if the controls attribute is present)


Screen Shot 2020-10-03 at 9 25 27 AM

lib/helpers/is-interactive-element.js
```diff
@@ -39,10 +41,6 @@ function isAudioVideoWithControls(node) {
   return (node.tag === 'audio' || node.tag === 'video') && ast.hasAttribute(node, 'controls');
 }
 
-function isHyperLink(node) {
-  return node.tag === 'a' && ast.hasAttribute(node, 'href');
-}
```
rwjblue, line 44: See above, I think this needs to come back.

test/unit/helpers/is-interactive-element-test.js
```diff
@@ -25,10 +25,17 @@ describe('isInteractiveElement', function () {
     });
   }
 
-  let nonInteractive = ['<a></a>', '<input type="hidden">', '<img>', '<div></div>'];
```
rwjblue, line 28: `<a></a>` needs to come back as `nonInteractive`

test/unit/rules/no-nested-interactive-test.js
```diff
@@ -10,26 +12,16 @@ generateRuleTests({
   good: [
     '<button>button</button>',
     '<button>button <strong>!!!</strong></button>',
-    '<a><button>button</button></a>',
```
rwjblue, line 13: This needs to come back.

docs/rule/no-invalid-interactive.md
rwjblue: > The `<canvas>` element is not an interactive element, but may contain certain interactive elements as descendants.

Isn't this true of _basically_ every non-interactive element that is allowed to contain any content? Why is `canvas` called out specifically?

rwjblue: > `input` elements whose `type` attribute are in the Checkbox or Radio Button states

The spec specifically says that any `input` that is **not** `type="hidden"` is interactive.

lib/helpers/is-interactive-element.js
```diff
@@ -13,10 +15,10 @@ const INTERACTIVE_TAG_NAMES = new Set([
   'label',
   'select',
   'textarea',
+  'video',
```
rwjblue, line 18: This seems incorrect. `video` should only be added if `controls` attribute is present. See spec link / image above.

lib/helpers/is-interactive-element.js
```diff
@@ -2,9 +2,11 @@
 
 const ast = require('./ast-node-info');
 
+// Spec: https://html.spec.whatwg.org/#interactive-content
 const INTERACTIVE_TAG_NAMES = new Set([
+  'a',
+  'audio',
```
rwjblue, line 8: This seems incorrect. `audio` should only be added if `controls` attribute is present. See spec link / image above.

rwjblue (changes requested):

lib/rules/no-nested-interactive.js
MelSumner: right! I wasn't sure what it was doing, figured we'd go over it in review.

lib/helpers/is-interactive-element.js
MelSumner: Also, we already have a different rule that enforces this: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/link-href-attributes.md

To assistive technology, an `<a>` is an `<a>`, with or without the `href`.

lib/helpers/is-interactive-element.js
MelSumner: see response above.

test/unit/helpers/is-interactive-element-test.js
MelSumner: see previous comment as to why it was removed.

test/unit/rules/no-nested-interactive-test.js
MelSumner: See above.

lib/helpers/is-interactive-element.js
MelSumner: Again, to AT, `video` means something, with or without controls. Lack of controls is a WCAG violation.

lib/helpers/is-interactive-element.js
rwjblue: > this is one place where the ARIA spec supersedes the HTML spec.

Need supporting evidence here...

test/unit/rules/no-nested-interactive-test.js
```diff
@@ -106,7 +97,7 @@ generateRuleTests({
         '<button><embed type="video/quicktime" src="movie.mov" width="640" height="480"></button>',
 
       result: {
-        message: 'Do not use <embed> inside <button>',
+        message: ERROR_MESSAGE,
```
lifeart, line 100: looks like regression, because error message less meaningful now

lib/helpers/is-interactive-element.js
MelSumner: 1. our own rule that requires href on the a element (and the documentation that rule links to)
2. “In case of conflict, consider users over authors over implementers over specifiers over theoretical purity” from https://www.w3.org/TR/html-design-principles/#priority-of-constituencies

docs/rule/no-invalid-interactive.md
rwjblue: Still open issues here:

- [ ] Change prose RE: `input`; all non-`type="hidden"` inputs are interactive we shouldn't imply otherwise
- [ ] Need to link to the spec in the various spots

MelSumner:
after more review, we've decided to close this PR and write some individual rules that will provide more explicit error messages.

PR #1564, opened 2020-10-07
Add `--working-directory` option.

rwjblue:
This refactors all read/write operations to be relative to the specified `--working-directory`. This working directory is threaded through the various internal/private API methods for configuration, and `process.cwd()` is now **only** used to default the working directory (when the command line option is not specified).

Closes #1204

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -21,7 +21,7 @@ const readFile = promisify(fs.readFile);
 
 const STDIN = '/dev/stdin';
 
-async function buildLinterOptions(filePath, filename = '', isReadingStdin) {
+async function buildLinterOptions(workingDir, filePath, filename = '', isReadingStdin) {
   if (isReadingStdin) {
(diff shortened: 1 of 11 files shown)


bin/ember-template-lint.js
```diff
@@ -134,6 +134,15 @@ function parseArgv(_argv) {
         describe: 'Output errors with source description',
         boolean: true,
       },
+      'working-directory': {
```
scalvert, line 137: My preference would to just name this `cwd` and remove the alias, but I know you disagree, so 🤷‍♂️

scalvert (approved):

lifeart (approved):

bin/ember-template-lint.js
rwjblue: Hehe, yes indeed. I much prefer `--working-directory` over a cryptic shorthand. However, shorthands are **perfect** for aliases. 😃

PR #1566, opened 2020-10-09
Allow `itemprop` use in meta tags in `no-invalid-meta` rule

### LINKED ISSUE #1565: Bug: Rule `no-invalid-meta` falsely reports issues for microdata itemprop meta tags
Per [the spec](https://www.w3.org/TR/microdata/#content-models):

> If a `meta` element has an `itemprop` attribute, the `name`, `http-equiv`, and `charset` attributes must be omitted, and the `content` attribute must be present.

This is related to issue #1275 which addressed the same problem with the `property` attribute. In contrast, this ticket is in reference to the `itemprop` attribute.

```html
<meta itemprop="foobar" content="barfoo">
```

sukima:
According to the [microdata specs][1]:

> If a `meta` element has an `itemprop` attribute, the `name`, `http-equiv`, and `charset` attributes must be omitted, and the `content` attribute must be present.

This allows for `<meta itemprop="foobar" content="baz">`

[1]: https://www.w3.org/TR/microdata/#content-models

Fixes #1565

### DIFF
--- a/lib/rules/no-invalid-meta.js
+++ b/lib/rules/no-invalid-meta.js
@@ -24,22 +24,23 @@ module.exports = class NoInvalidMeta extends Rule {
         const hasContent = AstNodeInfo.hasAttribute(node, 'content');
         const hasName = AstNodeInfo.hasAttribute(node, 'name');
         const hasProperty = AstNodeInfo.hasAttribute(node, 'property');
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
Thank you for fixing this issue! 👍🏻

PR #1594, opened 2020-11-03
Implement autofixer for `no-curly-component-invocation`

lifeart:
snippet: https://astexplorer.net/#/gist/11cac55a6d4fae12c58926f4f2f180d8/0c2f53c9d19bb2eec27f4e9d3c3b750c32a04e52

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -196,10 +196,24 @@ module.exports = function generateRuleTests({
         } else if (item.verifyResults) {
           item.verifyResults(actual);
         } else {
-          let results = item.results || [item.result];
-          let expectedResults = results.map(parseResult);
+          if (!('result' in item) && !('results' in item)) {
+            const hasRequiredProps = 'template' in item && 'fixedTemplate' in item;
+            assert(
+              hasRequiredProps,
+              `\`bad\` test cases without "result" or "results" property supported only for template fixes with shape like { template: string, fixedTemplate: string }, given: ${JSON.stringify(
+                item
+              )}`
+            );
+          }
 
-          assert.deepStrictEqual(actual, expectedResults);
+          // allow to have simple fixers tests, without result / results properties
+          let results = item.results || (item.result ? [item.result] : []);
+
+          if (results.length) {
+            // assert only if given bad results exists
+            let expectedResults = results.map(parseResult);
+            assert.deepStrictEqual(actual, expectedResults);
+          }
         }
       });
 
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -312,6 +312,8 @@ module.exports = class Base {
     let i, ret;
 
     return function () {
+      let maybeRuleHandlerResult;
+
       for (i = 0; i < beforeHandlers.length; i++) {
         ret = beforeHandlers[i].apply(plugin, arguments); // eslint-disable-line prefer-rest-params
         if (ret === REMOVE_HANDLER) {
@@ -321,7 +323,7 @@ module.exports = class Base {
       }
 
       if (ruleHandler && !plugin.isDisabled()) {
-        ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
+        maybeRuleHandlerResult = ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
       }
 
       for (i = 0; i < afterHandlers.length; i++) {
@@ -330,6 +332,10 @@ module.exports = class Base {
           afterHandlers.splice(i, 1);
           i -= 1;
         }
+      }
+
+      if (maybeRuleHandlerResult !== undefined && plugin.mode === 'fix') {
+        return maybeRuleHandlerResult;
       }
     };
   }
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -1,3 +1,5 @@
+const { builders: b } = require('ember-template-recast');
+
 const createErrorMessage = require('../helpers/create-error-message');
 const { transformTagName } = require('../helpers/curly-component-invocation');
 const Rule = require('./base');
@@ -47,14 +49,47 @@ const BUILT_INS = new Set([
 
 const ALWAYS_CURLY = new Set(['yield']);
 
+function toAngleAttrValue(attr) {
+  const node = attr.value;
+  if (node.type === 'SubExpression') {
+    if (node.path.original === 'concat') {
+      // eslint-disable-next-line unicorn/prefer-spread
+      return b.concat(
+        node.params.map((p) => {
+          if (p.type === 'PathExpression') {
+            return b.mustache(p);
+          } else if (p.type === 'StringLiteral') {
+            return b.text(p.original);
+          } else if (p.type === 'SubExpression') {
+            return b.mustache(p.path, p.params, p.hash);
+          } else {
+            return p;
+          }
+        })
+      );
+    } else {
+      return b.mustache(node.path, node.params, node.hash);
+    }
+  }
+  if (node.type === 'StringLiteral') {
+    if (node.original === '') {
+      attr.isValueLess = true;
+      return b.string('');
+    }
+    return b.text(node.original);
+  }
+  return b.mustache(node);
+}
+
 module.exports = class NoCurlyComponentInvocation extends Rule {
   parseConfig(config) {
     return parseConfig(config);
   }
 
-  logNode({ message, node }) {
+  logNode({ message, node, isFixable }) {
     this.log({
       message,
+      isFixable: isFixable || false,
       line: node.loc && node.loc.start.line,
       column: node.loc && node.loc.start.column,
       source: this.sourceForNode(node),
@@ -64,6 +99,30 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
   generateError(name) {
     let angleBracketName = transformTagName(name);
     return `You are using the component {{${name}}} with curly component syntax. You should use <${angleBracketName}> instead. If it is actually a helper you must manually add it to the 'no-curly-component-invocation' rule configuration, e.g. \`'no-curly-component-invocation': { allow: ['${name}'] }\`.`;
+  }
+
+  fixForBlockNode(node) {
+    if (node.program.type !== 'Block') {
+      return;
+    }
+    if (node.program.chained) {
+      return;
+    }
+    if (node.path.type !== 'PathExpression') {
+      return;
+    }
+
+    if (node.params.length) {
+      return;
+    }
+
+    return b.element(transformTagName(node.path.original, this.isLocal(node.path)), {
+      children: node.program.body,
+      blockParams: node.program.blockParams,
+      attrs: node.hash.pairs.map((p) => {
+        return b.attr(`@${p.key}`, toAngleAttrValue(p));
+      }),
+    });
   }
 
   visitor() {
@@ -88,6 +147,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
 
         if (path.original === 'link-to') {
           // {{link-to "bar" "foo"}}
+          // @to-do implement proper fix for link-to node
           this.logNode({ message: this.generateError('link-to'), node });
           return;
         }
@@ -133,8 +193,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
             return;
           }
 
-          // {{foo-bar bar=baz}}
-          this.logNode({ message: this.generateError(name), node });
+          this.logNode({ message: this.generateError(name), node, isFixable: false });
         } else {
           let isExplicitThis = path.this || path.data;
           let isLocal = this.isLocal(path);
@@ -167,7 +226,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
           if (name.includes('-') || name.includes('/')) {
             // {{foo-bar}}
             // {{nested/component}}
-            this.logNode({ message: this.generateError(name), node });
+            this.logNode({ message: this.generateError(name), node, isFixable: false });
             return;
           }
 
@@ -206,7 +265,10 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
 
         // {{#foo}}{{/foo}}
         // {{#foo bar=baz}}{{/foo}}
-        this.logNode({ message: this.generateError(path.original), node });
+        if (this.mode === 'fix') {
+          return this.fixForBlockNode(node);
+        }
+        this.logNode({ message: this.generateError(path.original), node, isFixable: true });
       },
     };
   }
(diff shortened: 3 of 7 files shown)


lib/helpers/rule-test-harness.js
```diff
@@ -189,10 +189,14 @@ module.exports = function generateRuleTests({
         } else if (item.verifyResults) {
           item.verifyResults(actual);
         } else {
-          let results = item.results || [item.result];
-          let expectedResults = results.map(parseResult);
+          // allow to have simple fixers tests, without result / results properties
+          let results = item.results || (item.result ? [item.result] : []);
 
-          assert.deepStrictEqual(actual, expectedResults);
+          if (results.length) {
```
lifeart, line 195: this changes necessary to have simple bad fixer asserts, and fixes issue there we don't have results and result property and we map parseResult over undefined

lib/rules/base.js
```diff
@@ -306,7 +308,7 @@ module.exports = class Base {
       }
 
       if (ruleHandler && !plugin.isDisabled()) {
-        ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
+        maybeRuleHandlerResult = ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
```
lifeart, line 311: this logic required to allow node replacement in fix mode example: (mustache -> element)

lifeart:
@bmish could you take a look?

lifeart:
image

Turbo87:
tbh I'm not a huge fan of this because there is no way to perform this fix in a way that is always 100% correct, even if we had telemetry data, which in this case we don't even seem to use. 

I'm quite scared of the potential of false positives here and I think I'd prefer it if we referred to the corresponding codemod instead.

lifeart:
@Turbo87 I agee this is not 100% workign solution, and in my understanding, we should not consider `--fix` option as silver bullet (as fix option in other linters). It will work for 95% of use cases, and it's quite same results, comparing to codemods. 

This rule have `ignore` options for this rule, also not all emitted errors marked as fixable, `fix` option is optional, and not solve all cases, but allow developers to save time. Also, it allow developers doing "pick" fixes (in editor), per-line, per-rule.

And fixes is mutch easier to run, comparint to codemods (50% of codemods does not working with typescript, some of codemods does not work if optional chaining exists, runtime codemods also not quite stable).

I prefer to have "one more" option for developers, and it shoud be an easy one.

bmish:
What do you think about adding a `fix` rule option (default false) to put this autofixer behind? That would provide developers with the option of using the fixer while mitigating the risks.

Turbo87:
I think what we need is something roughly like https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions, but even then I think it's probably easiest in this specific case to just refer to the codemod.

lifeart:
@bmish https://github.com/ember-template-lint/ember-template-lint/issues/1607 may be related, wondering, could we add kinda global fixers settings, or per-rule, do you have idea how it's may look like?

rwjblue (commented):
I think the only one we can actually fix is block statements, not mustache statements. We can't tell (from the ember-template-lint perspective) if a helper exists for a given mustache invocation (e.g. `{{foo whatever="here"}}` foo could be a helper), and adding a fixer to transform that into `<Foo @whatever="here" />` is not guaranteed to be correct.

We **must** guarantee that any `--fix`'s we include are 100% correct and will not introduce bugs into the consuming application.

lifeart:
@rwjblue I think it should work for MustacheStatement too in case of `{{foo/bar/baz}}` or it's a valid helper name? same for `{{foo.bar.baz name="12"}}` or `{{foo.bar.baz 12}}` - there is no way to have helper with this notation by default.

rwjblue:
> I think it should work for MustacheStatement too in case of {{foo/bar/baz}} or it's a valid helper name?

Yes, it is a valid helper name.

> same for {{foo.bar.baz name="12"}} or {{foo.bar.baz 12}} - there is no way to have helper with this notation by default.

This is no longer true, Ember 3.25 allows helper resolution to be dynamic. For example, this is possible and functional in Ember 3.25:

```hbs
{{#if (this.someHelper "lol")}}
{{/if}}
```

```js
export default class extends Component {
  get someHelper() {
    return Ember.Helper.helper(function(params) {
      return params[0];
    });
  } 
}
```

lifeart:
@rwjblue thank you for review & feedback! Changes addressed, mustache fixers removed

lib/helpers/rule-test-harness.js
rwjblue: I'm not sure we want this 🤔, and at the very least we should tweak it a bit to ensure you can't do:

```js
bad: [
  {
    template: 'whatever',
  }
]
```

Note no results or fixedTemplate? I definitely don't want to allow that...

test/unit/rules/no-curly-component-invocation-test.js
```diff
@@ -275,13 +286,61 @@ generateRuleTests({
...
           source: '{{foo.bar}}',
         },
       ],
     },
   ],
 });
 
+// fixers tests
+
+generateRuleTests({
```
rwjblue, line 299: Can you add more tests to this list to ensure that custom whitespace (within a single hash argument) is preserved?

lifeart: assert for empty results without { template, fixedTemplate } added, tests added

lib/rules/no-curly-component-invocation.js
```diff
@@ -66,6 +83,38 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
...
+    }
+    if (node.path.type !== 'PathExpression') {
+      return;
+    }
+    if (node.path.this === true) {
+      return;
+    }
+    if (node.path.data === true) {
+      return;
+    }
```
rwjblue, line 101: Why do we short circuit in these two scenarios? Seems like we could properly transform both of these:


```hbs
{{#this.foo hi="wave"}}
{{/this.foo}}

{{#@foo lol="huzzah"}}
{{/@foo}}
```

Right? Maybe these guards were used when also fixing plain mustaches?

lib/rules/no-curly-component-invocation.js
```diff
@@ -66,6 +83,38 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
...
+    if (node.params.length) {
+      return;
+    }
+    if (this.isLocal(node.path)) {
+      return;
+    }
```
rwjblue, line 107: Similarly, it seems like we should be able to support this branch as well.

lib/rules/no-curly-component-invocation.js
lifeart: @rwjblue good catch!

is it valid case:

```hbs
{{#let (component "foo") as |my-component|}}<my-component></my-component>{{/let}}
```

?

rwjblue:
> tbh I'm not a huge fan of this because there is no way to perform this fix in a way that is always 100% correct, even if we had telemetry data, which in this case we don't even seem to use.
> 
> I'm quite scared of the potential of false positives here and I think I'd prefer it if we referred to the corresponding codemod instead.

I believe that this concern is mitigated by only transforming `BlockStatements` which we know **for sure** have to be component invocations.

@turbo87 - Does that make sense? Do you agree?

lib/rules/no-curly-component-invocation.js
rwjblue: Yes, this should work properly (though I _generally_ don't see folks use `-` in block params).

lib/rules/no-curly-component-invocation.js
```diff
@@ -47,14 +49,29 @@ const BUILT_INS = new Set([
...
 
+function toAngleAttrValue(attr) {
+  const node = attr.value;
+  if (node.type === 'SubExpression') {
+    return b.mustache(node.path, node.params, node.hash);
+  }
+  if (node.type === 'StringLiteral') {
+    if (node.original === '') {
+      return b.mustache('concat', [b.string('')]);
+    }
```
rwjblue, line 60: Whats this for? I don't think we would normally want to emit a concat statement here (I would expect to do the `b.text(node.original)` just below) 🤔

lib/rules/no-curly-component-invocation.js
```diff
@@ -47,14 +49,29 @@ const BUILT_INS = new Set([
 
 const ALWAYS_CURLY = new Set(['yield']);
 
+function toAngleAttrValue(attr) {
+  const node = attr.value;
+  if (node.type === 'SubExpression') {
+    return b.mustache(node.path, node.params, node.hash);
+  }
```
rwjblue, line 56: Can you check for `concat` here, and handle it specially?

Specifically adding a test for

```hbs
{{#some-thing foo=(concat 'lol-' something ' ' 'wtf')}}
{{/some-thing}}
```

emitting:

```hbs
<SomeThing @foo="lol-{{something}} wtf">
</SomeThing>
```

lib/rules/no-curly-component-invocation.js
lifeart: @rwjblue  it not working for `empty` string

lifeart: @rwjblue https://astexplorer.net/#/gist/e7b5fc128add06ab02974faea2a979cd/b7652f5886bfb214f84bab0721d3b4148fd4d42d

lib/rules/no-curly-component-invocation.js
lifeart: @rwjblue looks like we have printer issue
image

```hbs
<SomeThing @foo={{concat 'lol-' something ' ' 'wtf'}}>
</SomeThing>
```

should be fine for now

lifeart: my bad, 
image works fine

Turbo87:
> @Turbo87 - Does that make sense? Do you agree?

sure, as long as we don't convert ambiguous cases I'm fine with this.

rwjblue:
> sure, as long as we don't convert ambiguous cases I'm fine with this.

Ya, totally agree. See my comment [from here](https://github.com/ember-template-lint/ember-template-lint/pull/1594#pullrequestreview-567456712) which is basically trying to say the same thing:

> We **must** guarantee that any `--fix`'s we include are 100% correct and will not introduce bugs into the consuming application.

lib/rules/no-curly-component-invocation.js
lifeart: @rwjblue fixed by toggling `isValueLess`

lib/rules/no-curly-component-invocation.js
lifeart: * fixed with `isValueLess` toggling

lifeart:
@rwjblue could we land it? or any changes required?

lib/rules/no-curly-component-invocation.js
```diff
@@ -47,14 +49,47 @@ const BUILT_INS = new Set([
 
 const ALWAYS_CURLY = new Set(['yield']);
 
+function toAngleAttrValue(attr) {
+  const node = attr.value;
+  if (node.type === 'SubExpression') {
+    if (node.path.original === 'concat') {
```
rwjblue, line 55: We can't assume that `node.path` is a `PathExpression` here. We need to check that before accessing `.original`:

```suggestion
    if (node.path.type === 'PathExpression' && node.path.original === 'concat') {
```

lib/rules/no-curly-component-invocation.js
```diff
@@ -47,14 +49,47 @@ const BUILT_INS = new Set([
...
+      // eslint-disable-next-line unicorn/prefer-spread
+      return b.concat(
+        node.params.map((p) => {
+          if (p.type === 'PathExpression') {
+            return b.mustache(p);
+          } else if (p.type === 'StringLiteral') {
```
rwjblue, line 61: This should probably handle other literal types. For example:

```hbs
{{#foo-bar 123}}{{/foo-bar}}
{{#foo-bar true}}{{/foo-bar}}
{{#foo-bar false}}{{/foo-bar}}
{{#foo-bar undefined}}{{/foo-bar}}
{{#foo-bar null}}{{/foo-bar}}
```

lib/rules/no-curly-component-invocation.js
```diff
@@ -47,14 +49,47 @@ const BUILT_INS = new Set([
...
+            return b.text(p.original);
+          } else if (p.type === 'SubExpression') {
+            return b.mustache(p.path, p.params, p.hash);
+          } else {
+            return p;
```
rwjblue, line 66: if we hit this `else`, we should avoid fixing

rwjblue (changes requested):

rwjblue:
Sorry for review delay here @lifeart

lib/rules/no-curly-component-invocation.js
lifeart: @rwjblue could you provide `breaking` example?
in my mind else handle cases you described above:

```
{{#foo-bar 123}}{{/foo-bar}}
{{#foo-bar true}}{{/foo-bar}}
{{#foo-bar false}}{{/foo-bar}}
{{#foo-bar undefined}}{{/foo-bar}}
{{#foo-bar null}}{{/foo-bar}}
```

lib/rules/no-curly-component-invocation.js
lifeart: related - https://github.com/ember-template-lint/ember-template-lint/pull/1594#discussion_r579617691

rwjblue: I realized that my comment was slightly off. This conditional is specifically for `(concat` helper. From the context of `concat` I think you are right, all types would be coerced to string anyways.

lib/rules/no-curly-component-invocation.js
```diff
@@ -88,6 +147,7 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
 
         if (path.original === 'link-to') {
           // {{link-to "bar" "foo"}}
+          // @to-do implement proper fix for link-to node
```
rwjblue, line 150: We might want to handle (at least the link-to specific parts) in #1764.

lib/rules/no-curly-component-invocation.js
```diff
@@ -66,6 +101,30 @@ module.exports = class NoCurlyComponentInvocation extends Rule {
...
+    if (node.program.chained) {
+      return;
+    }
+    if (node.path.type !== 'PathExpression') {
+      return;
+    }
+
+    if (node.params.length) {
+      return;
+    }
```
rwjblue, line 117: Since we are returning early in these scenarios, we need to fallback to logNode (because `ember-template-lint . --fix` should still fail when it sees `{{#foo-bar thing}}{{/foo-bar}}`).

rwjblue (commented):
OK, sorry for the run around / back and forth here @lifeart. I think there is only one real issue (making sure to fail when we exit early from the fixer).

lib/rules/no-curly-component-invocation.js
lifeart: updated

lifeart:
@rwjblue no problem, let me know if some other changes required

rwjblue (approved):

rwjblue:
Thank you!

PR #1596, opened 2020-11-05
Add new rule `no-model-argument-in-route-templates`

rwjblue:
Usage of `{{@model}}` in route templates was introduced to simplify the mental model in Ember Octane. Unfortunately, there is a known problem where usage of `{{@model}}` can introduce difficult to diagnose failures when transitioning within the same route (with different params, e.g. `/posts/1` -> `/posts/2`).

You'll notice that when transitioning into `/post/1` initially `{{@model}}` will be the resolved model value. However if you transition from `/posts/1` to `/posts/2`, you will notice that `{{@model}}` updates from that original model value to `undefined` (while the model hook for `/post/2` is still pending) then once that model hook resolves for the second time `{{@model}}` is updated with the final resolved model hook value.

This demonstrates the problem clearly: you **never** expect for `{{@model}}` to have been `undefined` and in a number of cases this will actually cause errors (e.g. if your helper assumed an argument was an object and throw an error).

### DIFF
--- a/docs/rule/no-model-argument-in-route-templates.md
+++ b/docs/rule/no-model-argument-in-route-templates.md
@@ -1,0 +1,82 @@
+# no-model-argument-in-route-templates
+
+:wrench: The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+
+Usage of `{{@model}}` in route templates was introduced to simplify the mental
+model in Ember Octane. Unfortunately, there is a known problem where usage of
+`{{@model}}` can introduce difficult to diagnose failures when transitioning
+within the same route (with different params, e.g. `/posts/1` -> `/posts/2`).
+
+Given this route:
+
+```js
+// app/routes/post.js
+export default class extends Route {
+  model(params) {
+    return this.store.findRecord('post', params.post_id);
+  }
+}
+```
+
+With the following template:
+
+```hbs
+{{! app/templates/post.hbs }}
+{{some-helper @model}}
+```
+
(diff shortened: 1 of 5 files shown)


lib/rules/no-model-argument-in-route-templates.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor(options) {
+    super(options);
+
+    this.isRouteTemplate = isRouteTemplate(options.filePath);
+  }
+
+  visitor() {
+    return {
+      PathExpression(node, path) {
+        if (!this.isRouteTemplate) {
```
bmish, line 20: We should be able to move this check up:

```js
visitor() {
  if (!this.isRouteTemplate) {
    return {};
  }
  
  return { ... };
}
```

docs/rule/no-model-argument-in-route-templates.md
```diff
@@ -0,0 +1,82 @@
...
+{{this.model.foo.bar}}
+```
+
+## Migration
+
+This rule includes a fixer in order to handle the migration for you automatically.
+
+## References
+
+* TODO: link to Ember issue folks can follow to know when this is fixed
```
bmish, line 82: No link available yet?

bmish (approved):

lib/rules/no-model-argument-in-route-templates.js
bmish: Fixed.

Issue #1604, opened 2020-11-10
no-invalid-link-title: False positive for SVG icon content


Turbo87:
https://github.com/rust-lang/crates.io/blob/3af0727b625d3b6c74c8add711adaabb72d2bc0d/app/components/pagination.hbs#L2-L4 is triggering the `no-invalid-link-title` rule, even though the "link text" (the SVG icon) does not match the "title attribute".

/cc @MelSumner

josephdsumner:
@Turbo87 @MelSumner 

I'd be happy to look into this!

I think that the two blocks of whitespace on either side of the SVG are what's flagging the error in this case. They are collected when the helper looks for link text `TextNode` children of the `<nav>` to be compared with the `@title` attribute value. They are trimmed, but there is no filter that removes the empty Strings from the comparison Array. So, what I think you're seeing is essentially the result of `'paragraph text'.includes('')`. In principle, you can confirm this 
 in the app by joining that block of code into a single line -- the error *should* disappear.

I can put in a PR that adds the filter + relevant `good` test for this issue in a moment.

PR #1605, opened 2020-11-10
Fix false positive in `no-invalid-link-title` when using a dynamic value surrounded by whitespace.

josephdsumner:
@Turbo87 @MelSumner 

Related issue: #1604

If merged, this PR updates the helper function that processes the link text. Specifically, it adds a filter to remove empty Strings from the Array for comparison against the `title` attribute. This prevents false-positives generated by whitespace appearing before or after the real (non-TextNode) link text.


MelSumner (approved):
Thanks for catching this & submitting a fix so quickly! 👍

rwjblue (approved):

PR #1606, opened 2020-11-11
Fix false positive with content within `{{#if}}`/`{{else}}` blocks in `no-duplicate-id` rule

### LINKED ISSUE #1511: no-duplicate-id false positive
```hbs
{{#if this.foo}}
  <div id="foobar"></div>
{{else}}
  <span id="foobar"></span>
{{/if}}
```

This gives a warning on the `span` element but it shouldn't. Not sure how easy it is to be fixed though.

MelSumner:
`no-duplicate-id` was returning false positives for content within `{{#if}}`/`{{else}}` blocks. 

If merged, this PR will:
- add a test to allow duplicate ids within control flow helper block statements 
`'{{#if}}<div id="id-00"></div>{{else}}<span id="id-00"></span>{{/if}}',`
- add a helper function to check for control flow helper block statements and ignore them

Fixes #1511

### DIFF
--- a/lib/rules/no-duplicate-id.js
+++ b/lib/rules/no-duplicate-id.js
@@ -1,16 +1,67 @@
 'use strict';
 
+const AstNodeInfo = require('../helpers/ast-node-info');
 const NodeMatcher = require('../helpers/node-matcher');
 const Rule = require('./base');
+
+class ConditionalScope {
+  constructor() {
+    this.frames = [new Set()];
+
+    this.conditionals = [];
+  }
+
+  pushConditional() {
+    this.conditionals.push(new Set());
+  }
+
+  popConditional() {
+    let idsWithinConditional = this.conditionals.pop();
+
+    if (this.conditionals.length > 0) {
+      let parentConditional = this.conditionals[this.conditionals.length - 1];
+      idsWithinConditional.forEach((idValue) => {
+        parentConditional.add(idValue);
+      });
+    } else {
+      this.frames.push(idsWithinConditional);
+    }
+  }
+
+  pushBranch() {
+    this.frames.push(new Set());
+  }
+
+  popBranch() {
+    this.frames.pop();
+  }
+
+  isDuplicateId(idValue) {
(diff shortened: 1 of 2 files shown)


MelSumner:
since we need to also check each branch of the control flow helper block statements, we'll look at how scope uses frames and add additional methods for ids.

rwjblue:
Paired with @MelSumner, I think we have this ready for review.

Turbo87 (approved):
I have a rough idea of what's going on in the code, but I have a feeling a few code comments describing it a bit more couldn't hurt 😅 

other than that, great job! 👍

lib/rules/no-duplicate-id.js
```diff
@@ -1,16 +1,67 @@
 'use strict';
 
+const AstNodeInfo = require('../helpers/ast-node-info');
 const NodeMatcher = require('../helpers/node-matcher');
 const Rule = require('./base');
 
+class ConditionalScope {
+  constructor() {
+    this.frames = [new Set()];
```
scalvert, line 9: Why `frames`?

lib/rules/no-duplicate-id.js
```diff
@@ -1,16 +1,67 @@
...
+  pushConditional() {
+    this.conditionals.push(new Set());
+  }
+
+  popConditional() {
+    let idsWithinConditional = this.conditionals.pop();
+
+    if (this.conditionals.length > 0) {
+      let parentConditional = this.conditionals[this.conditionals.length - 1];
+      idsWithinConditional.forEach((idValue) => {
```
scalvert, line 23: Is `idValue` actually `elementId`? The name feels a bit confusing to me, since it required me to ask you what it represented.

scalvert (approved):
I echo what @Turbo87 said. I loosely follow what's happening in the code (I had thought a bit about how to solve this myself), but adding comments would make sense. Alternatively, you could rename methods to make them a bit more clear what they represent in terms of the functionality, rather than what they're doing.

PR #1577, opened 2020-10-19
WIP of new todo functionality

scalvert:

PR #1611, opened 2020-11-19
Implements new todo functionality (replaces pending)

### LINKED ISSUE #1557: Feature: Deprecate old pending functionality in favor of new todo feature
This library already has a feature that allows for `pending` violations to be identified, by module, and converts them to warnings when considering whether to error or not. 

This [spec](https://github.com/ember-template-lint/ember-template-lint-pending-utils/blob/master/docs/spec.md) proposes a more robust set of functionality when dealing with pending items, starting with renaming pending to **todo**. 

From the **Motivation** section in that spec:

> Linting is a fundamental tool to help ensure the quality of a codebase. Ensuring there are as few linting errors as
> possible (ideally 0), is a useful measure of a baseline of code hygiene.
> 
> It's common to leverage linting not just for syntax adherence, but also to direct developers to employ standardized patterns. As
> such, it's a fairly routine activity to introduce new lint rules into a codebase. This introduction, while necessary, can cause
> unintended friction, such as
> 
> - new lint errors being introduced where they previously didn't exist
> - causing unintended delays to shipping new fixes and features
> - an "all or nothing" approach, where new rules require fixing before rollout.
> 
> Having the ability to identify violations as `todo` would allow for this incremental roll out, while providing tools that will allow maintainers to view the full list of todo violations. Todo violations can include a createdDate, allowing thresholds of time to trigger due date violations. This can assist in prioritizing lint rule violation fixes, and can ensure the todo list doesn't remain stagnant.

The rest of that spec details the intended behavior. This issue will detail the changes needed in `ember-template-lint` in order to implement this behavior as first-party functionality.

CHANGES:

- [x] Detect existing `pending` behavior, and ensure correct branching to support both old `pending` and new `todo`
- [x] Rename `@ember-template-lint/pending-utils` to `@ember-template-lint/todo-utils`
- [x] Leverage [`@ember-template-lint/pending-utils`](https://github.com/ember-template-lint/ember-template-lint-pending-utils) to parse and generate `.lint-todo` dir
- [x] Update docs for the new feature
- [x] Ensure adequate test coverage of expected workflows
- [x] Set `--print-pending` option to `hidden` in help

WORKFLOWS:

- when upgrading to new ember-template-lint version that includes todo:

	```
	# behaves as normal - no todo functionality
	ember-template-lint
	```

- when linting failure exists (severity 2) that has a current todo item, switch linting failure's severity to not trigger a failure
	- todo items should _not_ be included in default output (#1248)
	- todo item counts _should_ be included in counts output

		```shell
		✖ 8 problems (5 errors, 3 warnings, 2 todo)
		```	

		if no problems, no output (as normal)
		if no problems _but_ `--include-todo`, output todo count

		```shell
		✓ 0 problems (0 errors, 0 warnings, 2 todo)
		```			

	- todo items should be included with output _if_ running with the `--include-todo` option

		```shell
		```

- when opting in to todo functionality:

	```shell
	# add new todo items for new errors
	# remove todo items for errors that are no longer present
	# existing errors that have todo items should not be modified

	ember-template-lint . --update-todo

	# or 

	ember-template-lint lib/my-inrepo-addon --update-todo
	```

- when you have fixed a todo item, to update todo list:

	```shell
	# only removes todo items
	ember-template-lint . --fix
	```

- when you have deleted a linted file (foo.hbs), todo files associated with that file become orphaned. If you run linting for a specific file, no output will occur:

	```shell
	ember-template-lint app/templates/foo.hbs
	```

- when you have deleted a linted file (foo.hbs), todo files associated with that file become orphaned. You need to run `--update-todo` to cleanup todo files for that file:

	```shell
	ember-template-lint . --update-todo
	```

- when old pending system (`config.pending`) and new todo system (`.lint-todo`) co-exist, error
- when `--update-todo` ran and old pending system exists (`config.pending`), migrate old to new

renatoi:
Todo functionality is intended to replace the `--print-pending` feature which should have a deprecation warning for now.
Based on spec #1557 and continued the work from PR #1577.

⚠️ this is intended to be a power user feature, ideally for large projects that often have large teams of engineers working on a very large codebase. We do not want this feature to result in people deferring tech debt unnecessarily. For smaller projects, one would expect any and all linting errors would be fixed. 

We'll want to call this out in the README to make it clear we're not advocating hiding your linting errors away.

## Examples

A sneak peek into what this looks like with practical use.

Given a sample project with lint errors, running `ember-template-lint .` predictably results in the expected errors.

Screen Shot 2020-12-03 at 2 12 57 PM

To opt-in to the new TODO functionality, you run with the `--update-todo` option.

Screen Shot 2020-12-03 at 2 13 20 PM

This results in errors being transformed to TODOs. TODOs are not visible by default when running `ember-template-lint`. You can include them as part of the normal output by including the `--include-todo` option.

Screen Shot 2020-12-03 at 2 13 38 PM

Introducing new errors will not automatically convert them to TODOs. You need to run with `--update-todo` again to opt them in.

Screen Shot 2020-12-03 at 2 14 17 PM 

Fixes #1557 

## Related

There is an [eslint formatter](https://github.com/scalvert/eslint-formatter-todo) that mimics this same behavior. 

TODO:

- [x] Add warning if user is using deprecated `pending` feature
- [x] Ensure fixing errors results in removal of existing todos via `--fix`

- Documentation will be added in a subsequent PR

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -49,7 +49,6 @@ module.exports = {
     'no-new-wrappers': 'error',
     'no-octal-escape': 'error',
     'no-return-assign': 'error',
-    'no-return-await': 'error',
     'no-self-compare': 'error',
     'no-sequences': 'error',
     'no-shadow-restricted-names': 'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -49,6 +49,7 @@ module.exports = {
     'no-new-wrappers': 'error',
     'no-octal-escape': 'error',
     'no-return-assign': 'error',
+    'no-return-await': 'error',
     'no-self-compare': 'error',
     'no-sequences': 'error',
     'no-shadow-restricted-names': 'error',
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -164,7 +164,7 @@ function parseArgv(_argv) {
         boolean: true,
       },
       'include-todo': {
-        describe: 'Include todos in the results (todos are hidden by default)',
+        describe: 'Include todos in the results',
         default: false,
         boolean: true,
       },
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -312,7 +312,7 @@ async function run() {
     }
 
     if (!filePaths.has(STDIN)) {
-      fileResults = await linter.processTodos(linterOptions, fileResults);
+      fileResults = await linter.processTodos(linterOptions, fileResults, options.fix);
     }
 
     resultsAccumulator.push(...fileResults);
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -160,10 +160,12 @@ function parseArgv(_argv) {
       },
       'update-todo': {
         describe: 'Update list of linting todos by transforming lint errors to todos',
+        default: false,
         boolean: true,
       },
       'include-todo': {
         describe: 'Include todos in the results (todos are hidden by default)',
+        default: false,
         boolean: true,
       },
       'ignore-pattern': {
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -10,6 +10,7 @@ const path = require('path');
 const { promisify } = require('util');
 
 const { getTodoStorageDirPath } = require('@ember-template-lint/todo-utils');
+const chalk = require('chalk');
 const getStdin = require('get-stdin');
 const globby = require('globby');
 const isGlob = require('is-glob');
@@ -39,22 +40,6 @@ async function buildLinterOptions(workingDir, filePath, filename = '', isReading
     let source = await readFile(resolvedFilePath, { encoding: 'utf8' });
 
     return { source, filePath, moduleId };
-  }
-}
-
-// TODO: inline this function
-async function lintSource(linter, linterOptions, options) {
-  if (options.updateTodo || linter.todosEnabled) {
-    return linter.verifyAndUpdateTodos(linterOptions, options);
-  } else if (options.fix) {
-    let { isFixed, output, messages } = await linter.verifyAndFix(linterOptions);
-    if (isFixed) {
-      fs.writeFileSync(linterOptions.filePath, output);
-    }
-
-    return messages;
-  } else {
-    return linter.verify(linterOptions);
   }
 }
 
@@ -174,7 +159,11 @@ function parseArgv(_argv) {
         boolean: true,
       },
       'update-todo': {
-        describe: 'Update list of linting todos',
+        describe: 'Update list of linting todos by transforming lint errors to todos',
+        boolean: true,
+      },
+      'include-todo': {
+        describe: 'Show list of todos',
         boolean: true,
       },
       'ignore-pattern': {
@@ -230,6 +219,8 @@ function printPending(results, options) {
   if (options.json) {
     console.log(pendingListString);
   } else {
+    console.log(chalk.yellow('WARNING: Print pending is deprecated. Use --update-todo instead.\n'));
+
     console.log(
       'Add the following to your `.template-lintrc.js` file to mark these files as pending.\n\n'
     );
@@ -302,26 +293,39 @@ async function run() {
       filePaths.has(STDIN)
     );
 
-    let messages = await lintSource(linter, linterOptions, options);
+    let fileResults;
 
-    resultsAccumulator.push(...messages);
+    if (options.fix) {
+      let { isFixed, output, messages } = await linter.verifyAndFix(linterOptions);
+      if (isFixed) {
+        fs.writeFileSync(linterOptions.filePath, output);
+      }
+      fileResults = messages;
+    } else {
+      fileResults = await linter.verify(linterOptions);
+    }
+
+    if (options.updateTodo) {
+      await linter.updateTodo(linterOptions, fileResults);
+    }
+
+    if (!filePaths.has(STDIN)) {
+      fileResults = await linter.processTodos(linterOptions, fileResults);
+    }
+
+    resultsAccumulator.push(...fileResults);
   }
 
   let results = processResults(resultsAccumulator);
+
   if (results.errorCount > 0) {
     process.exitCode = 1;
   }
 
-  // TODO: add --update-pending flag
-  // TODO: add new --list-pending feature for new system
-  // TODO: error if using --print-pending and new pending system
-  // TODO: error if we have old config.pending and new pending dir
-  // TODO: migrate to new pending system if using old pending system
-
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount) {
+    if (results.errorCount || results.warningCount || (results.todoCount && options.includeTodo)) {
       let Printer = require('../lib/printers/default');
       let printer = new Printer(options);
       printer.print(results);
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -325,7 +325,7 @@ async function run() {
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount || (results.todoCount && options.includeTodo)) {
+    if (results.errorCount || results.warningCount || (options.includeTodo && results.todoCount)) {
       let Printer = require('../lib/printers/default');
       let printer = new Printer(options);
       printer.print(results);
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -9,6 +9,7 @@ const fs = require('fs');
 const path = require('path');
 const { promisify } = require('util');
 
+const { getTodoStorageDirPath } = require('@ember-template-lint/todo-utils');
 const getStdin = require('get-stdin');
 const globby = require('globby');
 const isGlob = require('is-glob');
@@ -16,8 +17,6 @@ const micromatch = require('micromatch');
 
 const Linter = require('../lib');
 const processResults = require('../lib/helpers/process-results');
-
-const { getTodoStorageDirPath } = require('@ember-template-lint/todo-utils');
 
 const readFile = promisify(fs.readFile);
 
@@ -44,18 +43,18 @@ async function buildLinterOptions(workingDir, filePath, filename = '', isReading
 }
 
 // TODO: inline this function
-async function lintSource(linter, options, shouldFix) {
-  if (options.updateTodos) {
-    return linter.verifyAndUpdateTodos(options);
-  } else if (shouldFix) {
-    let { isFixed, output, messages } = await linter.verifyAndFix(options);
+async function lintSource(linter, linterOptions, options) {
+  if (options.updateTodo || linter.todosEnabled) {
+    return linter.verifyAndUpdateTodos(linterOptions, options);
+  } else if (options.fix) {
+    let { isFixed, output, messages } = await linter.verifyAndFix(linterOptions);
     if (isFixed) {
-      fs.writeFileSync(options.filePath, output);
+      fs.writeFileSync(linterOptions.filePath, output);
     }
 
     return messages;
   } else {
-    return linter.verify(options);
+    return linter.verify(linterOptions);
   }
 }
 
@@ -174,7 +173,7 @@ function parseArgv(_argv) {
           'Print list of formatted rules for use with `pending` in config file (deprecated)',
         boolean: true,
       },
-      'update-todos': {
+      'update-todo': {
         describe: 'Update list of linting todos',
         boolean: true,
       },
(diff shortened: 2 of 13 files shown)


scalvert:
Thanks so much for carrying this over the line, @renatoi!

bin/ember-template-lint.js
```diff
@@ -274,20 +293,39 @@ async function run() {
...
+
   if (results.errorCount > 0) {
     process.exitCode = 1;
   }
 
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount) {
+    if (results.errorCount || results.warningCount || (results.todoCount && options.includeTodo)) {
```
scalvert, line 328: I know it's not a huge thing, but logically it makes sense to reverse this condition, since we want to primarily check if `includeTodo` is `true`, _then_ check if there's a count. Logically, we should be short circuiting if `includeTodo` is `false`.

```suggestion
    if (results.errorCount || results.warningCount || (options.includeTodo || results.todoCount)) {
```

lib/-private/todo-handler.js
```diff
@@ -0,0 +1,101 @@
...
+    [...existingTodos.values()].forEach((todo) => {
+      let result = results.find(
+        (result) =>
+          result.filePath === todo.filePath &&
+          result.rule === todo.ruleId &&
+          result.line === todo.line &&
+          result.column === todo.column
+      );
+
+      // set result to -1 so we don't fail linting for todo violations
```
scalvert, line 61: I think this comment is superfluous now. I think we can remove it.

```suggestion
```

lib/printers/json.js
```diff
@@ -15,8 +15,12 @@ class JsonPrinter {
       let warnings = this.options.quiet
         ? []
         : fileErrors.filter((error) => error.severity === Linter.WARNING_SEVERITY);
+      let todos =
+        this.options.quiet || !this.options.includeTodo
```
scalvert, line 19: So if we have `includeTodo` set to `true` _and_ `quiet`, this is going to remove the todos from output. Do we want that? If you set `includeTodo`, I feel like you're explicitly asking to see the todos.

lib/printers/pretty.js
```diff
@@ -22,30 +22,47 @@ class PrettyPrinter {
     }
 
     let warningCount = this.options.quiet ? 0 : results.warningCount;
-    let count = results.errorCount + warningCount;
+    let todoCount = this.options.quiet || !this.options.includeTodo ? 0 : results.todoCount;
```
scalvert, line 25: So if we have includeTodo set to true and quiet, this is going to remove the todos from output. Do we want that? If you set includeTodo, I feel like you're explicitly asking to see the todos.

bin/ember-template-lint.js
```diff
@@ -165,7 +154,16 @@ function parseArgv(_argv) {
...
+        describe:
+          'Print list of formatted rules for use with `pending` in config file (deprecated)',
+        boolean: true,
+      },
+      'update-todo': {
+        describe: 'Update list of linting todos by transforming lint errors to todos',
+        boolean: true,
+      },
+      'include-todo': {
+        describe: 'Show list of todos',
```
scalvert, line 166:
```suggestion
        describe: 'Include todos in the results (todos are hidden by default)',
```

scalvert (changes requested):

bin/ember-template-lint.js
scalvert: I suppose it won't short circuit anyway, since it's an or...

lib/printers/json.js
renatoi: The same can be said to `quiet` since you're explicitly asking for the output to be quiet, so it's matter of deciding which one has a higher priority. I thought `quiet` should be the overriding setting here due to the existing behavior for warnings: include warnings but if quiet is passed make it output nothing. Thoughts?

scalvert: Ya that makes sense.

lib/-private/todo-handler.js
```diff
@@ -0,0 +1,100 @@
...
+    function buildMessage(result) {
+      return {
+        rule: result.rule,
+        severity: result.severity,
+        moduleId: result.moduleId,
+        message: result.message,
+        line: result.line,
+        column: result.column,
+        source: result.source,
+      };
```
rwjblue, line 79: What is it taht we are not passing forward? Can we use `{ ...result }`?

lib/-private/todo-handler.js
```diff
@@ -0,0 +1,100 @@
...
+      result.severity = TODO_SEVERITY;
+    });
+
+    return results;
+  }
+
+  _buildResult(results) {
+    let builtResults = [];
+
+    function buildMessage(result) {
```
rwjblue, line 70: I don't fully understand the point of this function. Can you explain?

lib/printers/pretty.js
```diff
@@ -12,7 +12,7 @@ class PrettyPrinter {
     for (const filePath of Object.keys(results.files)) {
       let fileResults = results.files[filePath];
       let messages = this.options.quiet
-        ? fileResults.messages.filter((r) => r.severity !== Linter.WARNING_SEVERITY)
+        ? fileResults.messages.filter((r) => r.severity === Linter.ERROR_SEVERITY)
```
rwjblue, line 15: Lets tweak this to be `> Linter.WARNING_SEVERITY`?

lib/printers/pretty.js
rwjblue, line 25: I cannot understand this conditional (using boolean logic && and ternary in the same statement). Can you break it out into separate variables then do the boolean logic on that?

rwjblue (changes requested):

bin/ember-template-lint.js
```diff
@@ -263,6 +263,25 @@ async function run() {
...
   }
 
+  if (
+    fs.existsSync(getTodoStorageDirPath(options.workingDirectory)) &&
+    linter.config.pending.length > 0
+  ) {
+    console.error(
+      'Cannot use the `pending` config option in conjunction with lint todos. Please run with `--update-pending` to migrate to the new todos functionality.'
+    );
+    process.exitCode = 1;
```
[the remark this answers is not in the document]
scalvert: Yes, we have tests for both cases.

https://github.com/ember-template-lint/ember-template-lint/pull/1611/files#diff-08030efb2275758201c5160d222db24848291b8e03c0a415f23e82ccc9ad357fR1433
https://github.com/ember-template-lint/ember-template-lint/pull/1611/files#diff-08030efb2275758201c5160d222db24848291b8e03c0a415f23e82ccc9ad357fR1416

lib/-private/todo-handler.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (itemsToRemoveFromTodos.size > 0) {
+      itemsToRemoveFromTodos.forEach((todo) => {
+        results.push({
+          rule: 'invalid-todo-violation-rule',
+          message: `Todo violation passes \`${todo.ruleId}\` rule. Please run \`--update-todo\` to remove this todo from the todo list.`,
+          filePath: todo.filePath,
+          moduleId: todo.moduleId,
+          severity: 2,
+          isFixable: false,
```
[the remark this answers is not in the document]
scalvert: What does it mean for this to be fixable? Does this require a fixer?

lib/-private/todo-handler.js
scalvert: I honestly can't remember why I did this...

Let me change and we'll see what breaks.

lib/-private/todo-handler.js
scalvert: Similar to the above, I can't remember why I wrote this (seems useless). I'll post back here if I remember.

lib/printers/pretty.js
scalvert: Done.

lib/printers/pretty.js
scalvert: Done.

bin/ember-template-lint.js
```diff
@@ -263,6 +263,25 @@ async function run() {
     return;
   }
 
+  if (
+    fs.existsSync(getTodoStorageDirPath(options.workingDirectory)) &&
+    linter.config.pending.length > 0
```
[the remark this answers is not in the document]
scalvert: Done.

bin/ember-template-lint.js
```diff
@@ -165,7 +154,16 @@ function parseArgv(_argv) {
...
+        describe:
+          'Print list of formatted rules for use with `pending` in config file (deprecated)',
+        boolean: true,
+      },
+      'update-todo': {
+        describe: 'Update list of linting todos by transforming lint errors to todos',
+        boolean: true,
+      },
+      'include-todo': {
+        describe: 'Include todos in the results (todos are hidden by default)',
```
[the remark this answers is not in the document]
scalvert: Done for both options.

.eslintrc.js
```diff
@@ -49,7 +49,6 @@ module.exports = {
     'no-new-wrappers': 'error',
     'no-octal-escape': 'error',
     'no-return-assign': 'error',
-    'no-return-await': 'error',
```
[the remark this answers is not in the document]
scalvert: Unclear. Put it back and fixed the method it was attempting to avoid triggering the lint error for.

lib/-private/todo-handler.js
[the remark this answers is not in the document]
scalvert: Spoke offline about this. We want to set this to `isFixable: true`, then provide a fixer which will delete the todo file to 'clear' the error. 

We should write an acceptance test to ensure this functionality.

- generate failure
- update todo
- fix error
- run linting to generate failure
- ensure error message is generated for outstanding todo files without corresponding messages
- run with `--fix`
- ensure error message is not generated

lib/linter.js
```diff
@@ -45,6 +47,14 @@ class Linter {
     this.editorConfigResolver.resolveEditorConfigFiles();
   }
 
+  get _usesLegacyPending() {
+    return typeof this.config.pending === 'undefined';
+  }
```
rwjblue: I think this getter is not used anymore?

lib/linter.js
```diff
@@ -45,6 +47,14 @@ class Linter {
...
+
+  get todosEnabled() {
+    return this._todoHandler.todosEnabled;
+  }
```
rwjblue: I don't think this is used anymore?

bin/ember-template-lint.js
```diff
@@ -165,7 +154,18 @@ function parseArgv(_argv) {
...
+          'Print list of formatted rules for use with `pending` in config file (deprecated)',
+        boolean: true,
+      },
+      'update-todo': {
+        describe: 'Update list of linting todos by transforming lint errors to todos',
+        default: false,
+        boolean: true,
+      },
+      'include-todo': {
+        describe: 'Include todos in the results (todos are hidden by default)',
```
rwjblue, line 167:
```suggestion
        describe: 'Include todos in the results',
```

we don't have to say the "hidden by default" thing, now that we added `default: false`

scalvert:
Let's add a deprecated warning if the deprecated pending features is used, regardless of whether we're using the new todo feature or not.

rwjblue:
> Let's add a deprecated warning if the deprecated pending features is used

👍

lib/-private/todo-handler.js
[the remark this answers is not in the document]
scalvert: This is implemented.

bin/ember-template-lint.js
scalvert: Removed.

rwjblue:
@scalvert - You added the `breaking` label here, but I'm a tad confused on how it would be breaking. Can you elaborate?

lib/-private/todo-handler.js
```diff
@@ -0,0 +1,94 @@
...
+          filePath: result.filePath,
+          messages: [{ ...result }],
+          errorCount: 1,
+          source: result.source,
+        });
+      }
+    }
+
+    return builtResults;
+  }
```
rwjblue, line 93: We could probably simplify this a bit (avoiding extra array iteration / looping + the simplifying the branching logic) 🤔: 

```suggestion
  _buildResult(results) {
    let builtResults = new Map();

    for (const result of results) {
      let resultForFile = builtResults.get(result.filePath);
      if (resultForFile === undefined) {
        resultForFile = {
          filePath: result.filePath,
          messages: [],
          errorCount: 0,
          source: result.source,
        };
      }
      
      resultForFile.messages.push({ ...result });
      resultForFile.errorCount += 1;
    }

    return builtResults.values();
  }
```

What do you think?

package.json
```diff
@@ -37,6 +37,7 @@
     ]
   },
   "dependencies": {
+    "@ember-template-lint/todo-utils": "^3.1.0",
```
rwjblue, line 40: We should probably update this to latest (3.2.0)

lib/-private/todo-handler.js
scalvert: Sg, except you need to add the result back to the map, no?

package.json
scalvert: Yep, I just published that yesterday so that seems fine. It's only got the due days functionality in it.

lib/-private/todo-handler.js
rwjblue: D'oh! Updated...

scalvert: Updated with this.

package.json
scalvert: Done.

rwjblue (approved):

scalvert (approved):

PR #1620, opened 2020-11-23
Add new rule `no-down-event-binding`

alexlafroscia:
This PR adds a new lint rule that catches usage of "down" events in the `{{on}}` or `{{action}}` modifier and guides developers to use the "up" versions of them instead. This is not _always_ the right thing to do, but is _almost_ always the right thing to do, so I think it's worth having a lint rule for!

This came up thanks to a Tweet from @MelSumner today, where she linked to the WCAG guideline around not using "down" events. This was a major TIL for me, and since I'd bet most developers just guess which even they should be using, it felt like a lint rule could really help push people into the proverbial "pit of success"!

https://twitter.com/melaniersumner/status/1330905297326649344

### DIFF
--- a/lib/rules/no-down-event-binding.js
+++ b/lib/rules/no-down-event-binding.js
@@ -1,0 +1,85 @@
+'use strict';
+
+const { match } = require('../helpers/node-matcher');
+const Rule = require('./base');
+
+const ERROR_MESSAGE = 'Avoid binding to a `down` event; bind to an `up` event instead';
+
+/**
+ * Detects that a Node is an instance of the `{{on}}` modifier with a known event that is being listened to
+ *
+ * @param {object} node
+ * @return {boolean}
+ */
+function isOnModifier(node) {
+  return match(node, { path: { original: 'on' }, params: [{ type: 'StringLiteral' }] });
+}
+
+/**
+ * Detects that a Node is an instance of the `{{action}}` modifier
+ *
+ * @param {object} node
+ * @return {boolean}
+ */
+function isActionModifer(node) {
+  return match(node, { path: { original: 'action' } });
+}
+
+/**
+ * Check if an event name is a "down" event
+ *
+ * @param {string} eventName
+ * @return {boolean}
+ */
+function isDownEvent(eventName) {
+  return eventName.toLowerCase().endsWith('down');
+}
+
+module.exports = class NoDownEventBinding extends Rule {
+  visitor() {
+    return {
+      AttrNode(node) {
+        if (node.name.startsWith('on') && isDownEvent(node.name)) {
+          this.log({
+            message: ERROR_MESSAGE,
+            line: node.loc && node.loc.start.line,
+            column: node.loc && node.loc.start.column,
+            source: this.sourceForNode(node),
+          });
+        }
+      },
+
+      ElementModifierStatement(node) {
+        if (isOnModifier(node)) {
+          const eventNameNode = node.params[0];
+
+          if (isDownEvent(eventNameNode.value)) {
+            this.log({
+              message: ERROR_MESSAGE,
+              line: eventNameNode.loc && eventNameNode.loc.start.line,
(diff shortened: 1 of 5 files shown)


alexlafroscia:
By the way, I just wanted to say that the experience of adding new rules here is _so good_. I love that the contribution docs walk through exactly how to generate the rule, run tests, update the README, etc. It really made creating this a piece of cake!

lib/rules/no-down-event-binding.js
```diff
@@ -0,0 +1,74 @@
...
+  return eventName.toLowerCase().endsWith('down');
+}
+
+module.exports = class NoDownEventBinding extends Rule {
+  visitor() {
+    return {
+      ElementModifierStatement(node) {
+        if (isOnModifier(node)) {
+          const eventNameNode = node.params[0];
+
```
lifeart, line 44: Could we have fixer for this rule? 
Looks like it’s a simple one

alexlafroscia: It would definitely be easy to write! My only concern is the idea of auto-fixing something that is really different behavior. I don't know what I would want to fix this on save and not notice that things had changed, in case you really *do* need to bind to a "down" event..

test/unit/rules/no-down-event-binding-test.js
```diff
@@ -0,0 +1,46 @@
...
+  name: 'no-down-event-binding',
+
+  config: true,
+
+  good: [
+    "<div {{on 'keyup' this.doSomething}}></div>",
+    "<div {{action this.doSomething on='keyup'}}></div>",
+  ],
+
+  bad: [
```
bmish, line 16: What about the `onkeydown` DOM event?

https://www.w3schools.com/jsref/event_onkeydown.asp

bmish, line 16: What about components?

```hbs
<MyComponent @keyDown={{@close}}>
```

```hbs
{{my-component keyDown=foo}}
```

lib/rules/no-down-event-binding.js
bmish: Autofixers should not change behavior. Changing the event used could be a significant change in behavior.

test/unit/rules/no-down-event-binding-test.js
alexlafroscia: Hmm, that's an interesting idea. That could handle "classic component" arguments that automatically become event listeners, but that's increasingly irrelevant for Glimmer components that don't use that behavior. 

If we think about the mental model going forward and assume these arguments do not automatically become event handlers, would we want to lint random component arguments? It would probably be safe to assume that an argument called `keyDown` maps to such an event listener, but would that be assuming too much from the linter's perspective?

alexlafroscia: Yeah, that would be caught by this -- the tests aren't necessarily exhaustive. Would you prefer I add test cases for that event specifically? This rule disallows any `__down` event.

bmish: I don't see any test case with pure HTML. Can you add this?

```html
<input type="text" onkeydown="myFunction()">
```

bmish: I'd be fine excluding components as the default behavior. If someone wanted to follow-up later, they could add a rule option `catchComponents` (default false) to be more aggressive. 

For now, could you add these as valid test cases then?

alexlafroscia: Oh! Great point. I'll add a test for that!

alexlafroscia: Sure, I'll add those tests, too!

bmish (approved):

alexlafroscia:
Heads-up: I rebased one more time. I realized that when I updated the behavior and tests based on the feedback from @bmish, I forgot to update the documentation file! That's been fixed now.

Issue #1315, opened 2020-04-29
Plan for v3.0 Release


MelSumner:
The last major release was [V2](https://github.com/ember-template-lint/ember-template-lint/releases/tag/v2.0.0) in 2020-02-22.

List of breaking changes to include:

* [x] [Drop Node 13 support](https://github.com/ember-template-lint/ember-template-lint/issues/1444)
* [x] Remove old items from `lib/helpers/deprecated-rules.js`
* [x] Include support for linting .html files (#1232 )
* [x] Update `catchNonexistentRoles` option to default to `true` in [no-invalid-role](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-role.md)
* [x] Update `noImplicitThis` option default to true and `requireDash` option default to false in `no-curly-component-invocation` rule (#1642)
* [x] Update rule to default to `strict` behavior in [link-rel-noopener](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/link-rel-noopener.md)
* [x] New stylistic rules (#1638) (see current list of rules [here](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/stylistic.js)):
   * [modifier-name-case](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/modifier-name-case.md)
   * [no-whitespace-for-layout](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-whitespace-for-layout.md)
   * [no-whitespace-within-word](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-whitespace-within-word.md)
* [x] Promote all `octane` rules to the `recommended` config (and delete the `octane` config) since all supported Ember LTS versions are Ember Octane now (3.16, 3.20, 3.24) (#1639)
* [x] New recommended rules (#1643) (see current list of rules [here](https://github.com/ember-template-lint/ember-template-lint#rules)):
   * [builtin-component-arguments](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/builtin-component-arguments.md)
   * [deprecated-each-syntax](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/deprecated-each-syntax.md)
   * [deprecated-inline-view-helper](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/deprecated-inline-view-helper.md)
[...]

PR #1647, opened 2020-12-10
Enable `no-model-argument-in-route-templates` as a `recommended` rule

bmish:
Part of V3 release #1315.

https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-model-argument-in-route-templates.md


knownasilya:
Sounds like a bug in Ember.. Something that can't be fixed there?

rwjblue:
@knownasilya this was reverted in #1655

PR #1652, opened 2020-12-15
Add `no-mut-helper` rule

### LINKED ISSUE #1646: Rule Request: `no-mut-helper`
### Reasons to not use [the mut helper](https://api.emberjs.com/ember/3.16/classes/Ember.Templates.helpers/methods/mut?anchor=mut):

1. The mut helper is non-intuitive to use, see, teach, and learn since it can either be a getter or a setter based on the context in which it’s used. Example:
```
{{#let (mut this.foo) as |foo|}}
  <!-- When used like this, it's a getter -->
  {{foo}}

  <!-- When used like this, it's a setter -->
  <button {{action foo 123}}>Update Foo</button>
{{/let}}
```
2. The mut helper is usually only used as a pure setter, in which case there are other helpers that are only setters that should be used instead of mut (e.g. [ember-simple-set-helper](https://github.com/pzuraq/ember-simple-set-helper))

3. The mut helper can re-introduce 2 way data binding into glimmer components on named arguments where a child can change a parent’s data, which goes against the Data Down Actions Up principle, goes against glimmer components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/). Example:
```
<input 
  type="checkbox" 
  checked={{@checked}} 
  {{on "change" (fn (mut @checked) (not @checked))}} 
/>
```

### Migration Strategy
When used as a pure setter only, `mut` could be replaced by ember-simple-set-helper:
```
// before:
<MyComponent
    ...
    @onEscape={{action (mut this.isConfirmTransitionModalOpen) false}}
/>
```
```
// after:
<MyComponent
    ...
    @onEscape={{ember-simple-set-helper$set this “isConfirmTransitionModalOpen” false}}
/>
```

When used as a pure getter only, `mut` could be removed:
```
// before:
<MyComponent
      ...
      @toggled={{mut this.showAggregatedLine}}
/>
```
```
// after:
<MyComponent
      ...
      @toggled={{this.showAggregatedLine}}
/>
```

When used as a getter and setter, `mut` could be replaced with a different namespace for the property and a dedicated action function to set the property: (Note: the other option would be to pull in the `pick` helper from [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers) and use it like [this](https://github.com/pzuraq/ember-set-helper#picking-values-with-ember-composable-helpers). (Note: Another option could be to use [ember-box](https://github.com/pzuraq/ember-box)).
```
// before:
{{#let (mut this.foo) as |foo|}}
  {{foo}}
  <input onchange={{action foo value=”target.value”}} />
{{/let}}
```
```
// after HBS:
  {{this.foo}}
  <input {{on “change” this.updateFoo}} />
```
```
// after JS:
@tracked
foo;

@action
updateFoo(evt) {
  this.foo = evt.target.value;
  // or set(this, ‘foo’, evt.target.value); for legacy Ember code
}
```

When `mut` is being passed into a built-in classic component that uses 2 way data binding, mut could be removed:
```
// before:
<Input
            ...
            @value={{mut this.showcaseName}}
/>
```
```
// after:
<Input
            ...
            @value={{this.showcaseName}}
/>
```

### References
- [Blog post from @pzuraq on mut helper vs set helper](https://www.pzuraq.com/on-mut-and-2-way-binding/)
- [RFC to deprecate mut helper](https://github.com/emberjs/rfcs/issues/538)
- [Blog post from @chriskrycho on mut helper and set helper](https://v5.chriskrycho.com/journal/mut-and-set-and-auto-tracking-in-ember-octane/) 
- [Repo for ember-set-helper](https://github.com/pzuraq/ember-set-helper)
- [Repo for ember-simple-set-helper](https://github.com/pzuraq/ember-simple-set-helper) 
- [no-extra-mut-helper-argument template lint rule (further evidence that mut is non-intuitive)](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-extra-mut-helper-argument.md) 
- [Ember-box repo](https://github.com/pzuraq/ember-box)

collinadams:
Fixes #1646
no-mut-helper md preview 1
no-mut-helper md preview 2
[no-mut-helper md preview mov.mov.zip](https://github.com/ember-template-lint/ember-template-lint/files/5699070/no-mut-helper.md.preview.mov.mov.zip)

### DIFF
--- a/docs/rule/no-mut-helper.md
+++ b/docs/rule/no-mut-helper.md
@@ -16,8 +16,8 @@ Example:
 {{/let}}
 ```
 
-The need for the [no-extra-mut-helper-argument](no-extra-mut-helper-argument.md) rule is further evidence that `mut` has a non-intuitive signature and frequently gets misused.\
-2. The mut helper is usually only used as a pure setter, in which case there are other template helpers that are pure setters that could be used instead of mut (e.g. [ember-set-helper](https://github.com/pzuraq/ember-set-helper)).\
+The need for the [no-extra-mut-helper-argument](no-extra-mut-helper-argument.md) rule is further evidence that `mut` has a non-intuitive signature and frequently gets misused.
+2. The mut helper is usually only used as a pure setter, in which case there are other template helpers that are pure setters that could be used instead of mut (e.g. [ember-set-helper](https://github.com/pzuraq/ember-set-helper)).
 3. The mut helper can re-introduce 2 way data binding into Glimmer Components on named arguments where a child can change a parent’s data, which goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
 
 Example:
@@ -121,7 +121,7 @@ setProfileDescription({ target: { value } }) {
 
 ## Migration
 
-1. When used as a pure setter only, `mut` could be replaced by a JS action or [ember-set-helper](https://github.com/pzuraq/ember-set-helper):
+1. When used as a pure setter only, `mut` could be replaced by a JS action ("Option 1" below) or [ember-set-helper](https://github.com/pzuraq/ember-set-helper) ("Option 2" below):
 
 Before:
 
--- a/docs/rule/no-mut-helper.md
+++ b/docs/rule/no-mut-helper.md
@@ -1,0 +1,244 @@
+# no-mut-helper
+
+## Reasons to not use [the `mut` helper](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each?anchor=mut)
+
+1. The mut helper is non-intuitive to use, see, teach, and learn since it can either be a getter or a setter based on the context in which it’s used.
+
+Example:
+
+```hbs
+{{#let (mut this.foo) as |foo|}}
+  <!-- When used like this, it's a getter -->
+  {{foo}}
+
+  <!-- When used like this, it's a setter -->
+  <button {{action foo 123}}>Update Foo</button>
+{{/let}}
+```
+
+The need for the [no-extra-mut-helper-argument](no-extra-mut-helper-argument.md) rule is further evidence that `mut` has a non-intuitive signature and frequently gets misused.\
+2. The mut helper is usually only used as a pure setter, in which case there are other template helpers that are pure setters that could be used instead of mut (e.g. [ember-set-helper](https://github.com/pzuraq/ember-set-helper)).\
+3. The mut helper can re-introduce 2 way data binding into Glimmer Components on named arguments where a child can change a parent’s data, which goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
+
+Example:
+
+```hbs
+<input
+  type="checkbox"
+  checked={{@checked}}
+  {{on "change" (fn (mut @checked) (not @checked))}}
+/>
+```
+
+## What this rule does
+
+This rule forbids any use of the `mut` helper, both as a getter and a setter, in any context. It also
+surfaces possible alternatives in the lint violation message to help guide engineers to resolving
+the lint violations.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<MyComponent @isDropdownOpen={{mut this.isDropdownOpen}}/>
+```
+
+```hbs
+<button {{action (mut this.isDropdownOpen) false}}>Close Dropdown</button>
+```
+
+```hbs
+<MyComponent @closeDropdown={{action (mut this.isDropdownOpen) false}}/>
+```
+
+```hbs
+<MyComponent onclick={{fn (mut this.isDropdownOpen) false}}/>
+```
+
+```hbs
+<Input onchange={{action (mut this.profile.description) value="target.value"}}/>
+```
+
+```hbs
+{{my-component click=(action (mut this.isDropdownOpen) false)}}
+```
+
+```hbs
+{{my-component value=(mut this.profile.description)}}
+```
+
+This rule **allows** the following:
+
+```hbs
+<MyComponent @isDropdownOpen={{this.isDropdownOpen}}/>
+```
+
+```hbs
+<button {{action (ember-set-helper$set this "isDropdownOpen" false)}}>Close Dropdown</button>
+```
+
+```hbs
+<button {{on "click" (ember-set-helper$set this "isDropdownOpen" false)}}>Close Dropdown</button>
+```
+
+```hbs
+<button {{on "click" (fn this.setIsDropdownOpen false)}}>Close Dropdown</button>
+```
+
+```hbs
+<button {{action "setIsDropdownOpen" false}}>Close Dropdown</button>
+```
+
+```hbs
+<MyComponent @closeDropdown={{action this.setIsDropdownOpen false}}/>
+```
+
+```hbs
+<MyComponent onclick={{fn this.setIsDropdownOpen false}}/>
+```
+
+```hbs
+<Input onchange={{action this.setProfileDescription}}/>
+```
+
+Coupled with a corresponding JS action to set:
+
+```js
+@action
+setProfileDescription({ target: { value } }) {
+  set(this, 'profile.description', value);
+}
+```
+
+```hbs
+{{my-component click=(action (ember-set-helper$set this "setIsDropdownOpen" false)}}
+```
+
+```hbs
+{{my-component value=this.profile.description}}
+```
+
+## Migration
+
+1. When used as a pure setter only, `mut` could be replaced by a JS action or [ember-set-helper](https://github.com/pzuraq/ember-set-helper):
+
+Before:
+
+```hbs
+<MyComponent
+  @closeDropdown={{action (mut this.setIsDropdownOpen) false}}
+/>
+```
+
+After (Option 1 HBS):
+
+```hbs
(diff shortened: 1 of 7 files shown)


lib/rules/no-mut-helper.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  visitor() {
+    return {
+      MustacheStatement(node) {
+        this._checkForMutHelper(node);
+      },
+
+      BlockStatement(node) {
```
lifeart, line 38: loks like `mut` could not be path in block statement

```hbs
{{#mut ... }}

{{/mut}}}
```

test/unit/rules/no-mut-helper-test.js
```diff
@@ -0,0 +1,189 @@
+'use strict';
+
+const { message, generateMessageWithAlternative } = require('../../../lib/rules/no-mut-helper');
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+const setterAlternative = 'ember-simple-set-helper$set';
```
lifeart, line 6: `$set` ?

why do we need to mention `ember-simple-set-helper` in error?

lib/rules/no-mut-helper.js
```diff
@@ -0,0 +1,70 @@
...
+      },
+    };
+  }
+
+  _checkForMutHelper(node) {
+    if (node.path.original === 'mut') {
+      if (this.config.setterAlternative) {
+        this.log({
+          message: generateMessageWithAlternative(this.config.setterAlternative),
+          line: node.loc && node.loc.start.line,
```
lifeart, line 53: should we have more `pointing` issue (especially in editors), using `path` higlighting?

`node.loc` -> `node.path.loc`

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+setIsDropdownOpen(isDropdownOpen) {
+  set(this, 'isDropdownOpen', isDropdownOpen);
+}
+```
+
+After (Option 2):
+
+```hbs
+<MyComponent
+  @closeDropdown={{ember-set-helper$set this "isDropdownOpen" false}}
```
lifeart, line 155: `$set` looks redundant

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+This rule **allows** the following:
+
+```hbs
+<MyComponent @isDropdownOpen={{this.isDropdownOpen}}/>
+```
+
+```hbs
+<button {{action (ember-set-helper$set this "isDropdownOpen" false)}}>Close Dropdown</button>
```
chriskrycho, line 78: We should probably show the standard (not Batman/Wall Street) syntax here, since very few apps and addons are using that syntax.

lib/rules/no-mut-helper.js
collinadams: ah, good point Alex. this was accidental copy paste residue from the similar `no-unbound` rule.

fixed and thanks for the suggestion.

test/unit/rules/no-mut-helper-test.js
collinadams: good point, Alex. as Chris mentioned in his comment below, most Ember apps use standard syntax, not [Batman/Wall Street sytnax](https://github.com/rwjblue/ember-holy-futuristic-template-namespacing-batman) (where the helper name is prefixed with the addon name of the dependency).

Fixed - I have update the Pull Request to only use `{{set}}` and omit the addon name prefix of the dependency.

Thanks for the suggestion.

docs/rule/no-mut-helper.md
collinadams: fair point, Alex. I have updated the Pull Request to no longer use the addon name prefix when invoking the `{{set}}` helper.

fixed - thanks for the suggestion.

docs/rule/no-mut-helper.md
collinadams: thanks for reviewing, Chris. ya, good point, agreed. fixed throughout the Pull Request, thanks for the suggestion.

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+{{my-component click=(action (set this "setIsDropdownOpen" false)}}
+```
+
+```hbs
+{{my-component value=this.profile.description}}
+```
+
+## Migration
+
+1. When used as a pure setter only, `mut` could be replaced by a JS action or [ember-set-helper](https://github.com/pzuraq/ember-set-helper):
```
rwjblue, line 124:
```suggestion
1. When used as a pure setter only, `mut` could be replaced by a JS action ("Option 1" below) or [ember-set-helper](https://github.com/pzuraq/ember-set-helper) ("Option 2" below):
```

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+```hbs
+<Input
+  @value={{this.profile.description}}
+/>
+```
+
+## Configuration
+
+* object -- containing the following properties:
+  * string -- `setterAlternative` -- Optional: String name of a helper that could replace mut as a setter. If configured, the lint violation error message will include this as a possible alternative for resolving the lint violation.
```
rwjblue, line 232: I know we don't normally do this (though I think we should start adding it), but would you mind adding a small example config just below?

I'm thinking something like:

---

## Configuration

* object -- containing the following properties:
  * string -- `setterAlternative` -- Optional: String name of a helper that could replace mut as a setter. If configured, the lint violation error message will include this as a possible alternative for resolving the lint violation.

Example:

```js
// .template-lintrc.js

module.exports = {
  rules: {
    'no-mut-helper': {
      setterAlternative: 'what goes here?!?!?!?!',
    }
  }
};
```

rwjblue: Also, doesn't this need to accept `boolean` too?

lib/rules/no-mut-helper.js
```diff
@@ -0,0 +1,68 @@
...
+      },
+
+      SubExpression(node) {
+        this._checkForMutHelper(node);
+      },
+    };
+  }
+
+  _checkForMutHelper(node) {
+    if (node.path.original === 'mut') {
```
rwjblue, line 47: Let's check the `type` of `node.path` to be certain (future versions of Glimmer allow subexpressions to be nested in `path`).

```suggestion
    if (node.path.type === 'PathExpression' && node.path.original === 'mut') {
```

rwjblue (changes requested):

rwjblue:
@collinadams - Thanks for working on this!

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+
+Example:
+
+```hbs
+{{#let (mut this.foo) as |foo|}}
+  <!-- When used like this, it's a getter -->
+  {{foo}}
+
+  <!-- When used like this, it's a setter -->
+  <button {{action foo 123}}>Update Foo</button>
```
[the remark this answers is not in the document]
collinadams: hm, my understanding was that it would only be a setter if it was wrapped by the `action` or `fn` helpers?

docs/rule/no-mut-helper.md
collinadams: thanks for the suggestion, Rob.

Agreed - i added a config example.

The config only expects an optional `string` (no `boolean`), which would be the name of helper that can be added to give engineers an extra hint on how to resolve the error, if their application defines or depends on an hbs helper alternative (like ember-set-helper$set) that could be used instead to resolve the violation without having to write a new JS action. If the config or the config.setterAlternative is undefined, then the default error message will be shown, which suggests engineers replace `mut` with a JS action. If the config accidentally passes a non-string value for config.setterAlternative, this is considered a mis-configuration and the rule throws an error to help the engineer fix the config.

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,244 @@
...
+{{/let}}
+```
+
+The need for the [no-extra-mut-helper-argument](no-extra-mut-helper-argument.md) rule is further evidence that `mut` has a non-intuitive signature and frequently gets misused.\
+2. The mut helper is usually only used as a pure setter, in which case there are other template helpers that are pure setters that could be used instead of mut (e.g. [ember-set-helper](https://github.com/pzuraq/ember-set-helper)).\
+3. The mut helper can re-introduce 2 way data binding into Glimmer Components on named arguments where a child can change a parent’s data, which goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
```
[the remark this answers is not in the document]
collinadams: ah, fair point. i broke it out into a top level ordered list (for general issues vs glimmer component specific issues) that contained unordered lists underneath.

hope that satisfies your request here, but happy to tweak it again if you want something different. 🙂

docs/rule/no-mut-helper.md
```diff
@@ -0,0 +1,261 @@
...
+Example:
+
+```js
+// .template-lintrc.js
+
+module.exports = {
+  rules: {
+    'no-mut-helper': {
+      setterAlternative: '{{set}}',
+    }
```
rwjblue, line 246: Sorry, I just realized my suggestion was slightly wrong:

```suggestion
    'no-mut-helper': ['error', {
      setterAlternative: '{{set}}',
    }]
```

collinadams: haha ah, right. 🤦‍♂️

rwjblue (approved):

rwjblue:
Thanks @collinadams!

Issue #787, opened 2019-07-25
add new rule: no-obscure-hbs


efx:
Originally opened in https://github.com/ember-template-lint/ember-cli-template-lint/issues/437
I have some handlebars syntax in my application that "works", but is "wrong".
[Twiddle](https://ember-twiddle.com/07c76e285823a864ecfd1a59dc754705?openFiles=templates.application.hbs%2C)
```hbs
<p>some user name {{this.list.[0].name}}</p>
```

locks and rwjblue proposed adding a template rule to warn/error before any deprecations. ( see [ember-cli discord](https://discordapp.com/channels/480462759797063690/486548111221719040) on 2/06/19)

rwjblue:
Thanks for reporting! I think we should probably make rules for the individual "issues" that we have, otherwise adding new scenarios will be a "breaking change".

PR #1654, opened 2020-12-16
#787 Rgoygov/no-obscure-array-access

GoygovRustam:


lib/rules/no-obscure-array-access.js
```diff
@@ -0,0 +1,28 @@
...
+  return `Obscure expressions are prohibited. ${expression}. Please use Ember's get helper instead. i.e {{get list '0'}} `;
+}
+
+module.exports = class NoObscureArrayAccess extends Rule {
+  visitor() {
+    return {
+      PathExpression(node) {
+        // Regex test for any word character plus digit. i.e.  "test.1"
+        let regex = /\w+.\d/;
+        if (node.original && node.original.match(regex)) {
```
lifeart, line 15: could we add fixer for it?

`this.a[1][2][3]` -> `get this.a '1.2.3.'`

lifeart: example: https://astexplorer.net/#/gist/5edec4c14fb7c3554278b8a32b1e3297/ad0139632c9557c891b9d99c56bb51de24caa350

.gitignore
```diff
@@ -16,3 +16,4 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
+.vscode
```
rwjblue, line 19: I try to avoid adding editor specific files to the repo, also if you add a global gitignore for this in your local system setup it will be a bit less hassle for you in **all** the repos you work in. 😃 

https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files?algolia-query=gitignore#configuring-ignored-files-for-all-repositories-on-your-computer

docs/rule/no-obscure-array-access.md
```diff
@@ -0,0 +1,39 @@
+# no-obscure-array-access
+
+Using obscrure expressions `{{list.[1].name}}` is descouraged and might be deprecated soon
```
rwjblue, line 3:
```suggestion
Using obscure expressions `{{list.[1].name}}` is discouraged and might be deprecated soon.
```

lib/rules/no-obscure-array-access.js
rwjblue: Ya, we can definitely make this fixable.

rwjblue, line 15: Instead of checking original here, I'd prefer we look at `path.parts` and see if any are numbers.

Something like:

```js
let hasObscureAccess = path.parts.some(part => Number.isInteger(Number(part)))
if (hasObscureAccess) {
  this.log(...);
}
``

rwjblue (changes requested):

PR #1657, opened 2020-12-20
Add example script for counting lint violation disable directive comments

bmish:
This adds an example shell script to our documentation for generating a list of rules and the number of times disable directive comments are used to disable each of them. This can be useful for identifying the largest sources of tech debt in a codebase.

Example output:

```
DISABLED ESLINT RULES:
 40 ember/closure-actions
 35 ember/no-jquery
 33 ember/no-observers
  5 complexity
  2 no-console
DISABLED EMBER-TEMPLATE-LINT RULES:
 21 no-bare-strings
  4 require-valid-alt-text
  4 no-inline-styles
  3 table-groups
DISABLED STYLELINT RULES:
 14 declaration-no-important
  2 length-zero-no-unit
```

A follow-up to consider is adding this functionality as a CLI option on the ember-template-lint command (i.e. `ember-template-lint --count-disable-directives`).

### DIFF
--- a/docs/count-lint-violations.sh
+++ b/docs/count-lint-violations.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+
+# Outputs a sorted list of lint violations by rule.
+# Inspired by this: https://macr.ae/article/counting-eslint-disabled
+
+# Example output:
+#
+# DISABLED ESLINT RULES:
+#  40 ember/closure-actions
+#  35 ember/no-jquery
+#  33 ember/no-observers
+#   5 complexity
+#   2 no-console
+# DISABLED EMBER-TEMPLATE-LINT RULES:
+#  21 no-bare-strings
+#   4 require-valid-alt-text
+#   4 no-inline-styles
+#   3 table-groups
+# DISABLED STYLELINT RULES:
+#  14 declaration-no-important
+#   2 length-zero-no-unit
+
+if [ "$#" -eq 0 ]; then
+    echo "You must enter at least one command line argument with the path to search."
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


docs/count-lint-violations.sh
```diff
@@ -0,0 +1,64 @@
...
+# These regexps are used to remove all characters and whitespace surrounding the rule names in the disable directive comment.
+
+# /* eslint-disable-line no-console */
+# /* eslint-disable-line no-console -- explanation of why it is disabled */
+# // eslint-disable-line no-console, no-empty
+REGEXP_C_STYLE_COMMENT_SHELL='^.*\/[*\/] +| +\*\/|, *| +--.+| +$'
+
+# {{! template-lint-disable no-bare-strings no-negated-condition }}
+# {{!-- template-lint-disable no-bare-strings --}}
+REGEXP_HBS_STYLE_COMMENT_SHELL='^.*{{!-{0,2} *| *-{0,2}}}.*$'
```
mongoose700, line 37: This is only going to trigger if there's exactly one space. Is that a requirement for the comment to work?

rwjblue (commented):
@bmish - Any objections to making this a command line option for our normal script? I think it would fairly straightforward to incorporate that into the base rule class (where we process the configuration comments) so that it would emit failures for each configuration comment. These failures would then print to the console just like normal failures (including line, column, and filename).

bmish:
@rwjblue agreed that that would be ideal. I would like to follow-up with that sometime, so I'm going to start with the documentation for now.

docs/count-lint-violations.sh
bmish: I'll tweak this to allow multiple spaces and add tests when I convert this to a real script.

Issue #1624, opened 2020-11-27
ember-template-lint in Github actions in PR makes annotation show twice


MrRobz:
I have set-up github action so that it runs `ember-template-lint .` for every pull-request to the repo.

But what I see is the annotations are getting duplicated (shown twice) in github PR files page. Which I do not want.

Eg:
![image](https://user-images.githubusercontent.com/21201812/100475238-a7399400-3108-11eb-988e-bf51e4b72312.png)

"Interaction added to non-interactive element" error has been annotated twice as well as "Do not use action as..." error.

You can view the sample PR I created here: https://github.com/Robin-Thomas-577/ember-template-lint-sample-repo/pull/1/files

Interestingly, if passed `DISABLE_GITHUB_ACTIONS_ANNOTATIONS=true`only one annotation is shown. But I thought DISABLE_GITHUB_ACTIONS_ANNOTATIONS was to disable annotation altogether! 

**package versions used**
ember-cli: 3.22.0
node: 12.19.0
ember-template-lint: 2.14.0

Turbo87:
I've noticed that same thing, and I've been wondering what might be causing this. My best guess is that the GitHub team started to parse the command line output for things that look like ESLint reports and automatically converts them to warnings now too, but I haven't found any proof for that yet.

Turbo87:
https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md looks like my guess was indeed correct. the `setup-node` action automatically sets this up and since our output formatting these days matches that of ESLint we are now getting picked up automatically, in addition to what we're reporting explicitly.

PR #1668, opened 2020-12-28
Remove explicit GitHub Actions reporter

Turbo87:
The `setup-node` action on GitHub Actions automatically sets up a "[problem matcher](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md)" for the ESLint output format and since our output these days matches the ESLint format we don't need this explicit reporter anymore.

Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1624


Turbo87:
I was worried about the volta action not supporting these problem matchers, but it turns out @rwjblue took care of that already: https://github.com/volta-cli/action/pull/42 ❤️

rwjblue (approved):

PR #1669, opened 2020-12-29
Re-roll yarn.lock from scratch

bmish:


Turbo87:
btw we could switch from dependabot to renovatebot and have this kind of PR automated 😉

bmish:
@Turbo87 I hope Dependabot adds regular lock.file recreation capability too. I have used Renovatebot too and it's nice, but now it just seems simpler to stick with the in-house GitHub solution (Dependabot)...

PR #1671, opened 2020-12-30
Add rule: `no-this-in-template-only-components`

dwickern:
After running [ember-no-implicit-this-codemod](https://github.com/ember-codemods/ember-no-implicit-this-codemod) I was left with a bunch of template-only components referencing `{{this.prop}}`. I put together this rule to help me finish the conversion. Maybe it would be useful to someone else.

Related to https://github.com/emberjs/ember.js/issues/18132

### DIFF
--- a/lib/rules/no-this-in-template-only-components.js
+++ b/lib/rules/no-this-in-template-only-components.js
@@ -37,7 +37,7 @@ function hasContext(hbsFilePath) {
       return fs.existsSync(jsFilePath);
     }
   }
-  // what other types of files?
+  // TODO: handle pods layout
   return false;
 }
 
--- a/lib/rules/no-this-in-template-only-components.js
+++ b/lib/rules/no-this-in-template-only-components.js
@@ -12,7 +12,7 @@ function message(original, fixed) {
   );
 }
 
-function hasContext(hbsFilePath) {
+function isTemplateOnlyComponent(hbsFilePath) {
   const paths = hbsFilePath.split(path.sep);
   if (paths[0] === 'app' || paths[0] === 'addon') {
     if (paths[1] === 'templates') {
@@ -25,20 +25,20 @@ function hasContext(hbsFilePath) {
           ...paths.slice(3, -1),
           `${moduleName}.js`
         );
-        return fs.existsSync(jsFilePath);
+        return !fs.existsSync(jsFilePath);
       } else {
         // route template always has a context
-        return true;
+        return false;
       }
     } else if (paths[1] === 'components') {
       // co-located structure: /app/components/foo.hbs => /app/components/foo.js
       const moduleName = path.basename(hbsFilePath, '.hbs');
       const jsFilePath = path.join(path.dirname(hbsFilePath), `${moduleName}.js`);
-      return fs.existsSync(jsFilePath);
+      return !fs.existsSync(jsFilePath);
     }
   }
   // TODO: handle pods layout
-  return false;
+  return true;
 }
 
 function isFixable(original) {
@@ -51,7 +51,7 @@ function isFixable(original) {
 
 module.exports = class NoThisInTemplateOnlyComponents extends Rule {
   visitor() {
-    if (hasContext(this._filePath)) {
+    if (!isTemplateOnlyComponent(this._filePath)) {
       // template is allowed to use `this`
       return null;
     }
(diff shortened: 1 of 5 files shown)


lib/rules/no-this-in-template-only-components.js
```diff
@@ -0,0 +1,80 @@
...
+      }
+    } else if (paths[1] === 'components') {
+      // co-located structure: /app/components/foo.hbs => /app/components/foo.js
+      const moduleName = path.basename(hbsFilePath, '.hbs');
+      const jsFilePath = path.join(path.dirname(hbsFilePath), `${moduleName}.js`);
+      return fs.existsSync(jsFilePath);
+    }
+  }
+  // what other types of files?
+  return false;
```
rwjblue, line 41: Mind making a TODO comment here for "pods" layout?

lib/rules/no-this-in-template-only-components.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function isFixable(original) {
+  if (original === 'this.elementId') {
+    // elementId usage in a template-only component can't be auto-fixed
+    return false;
```
rwjblue, line 47: Can you extract this check to check for any of the "built in" properties? 

I think we can figure this out by looking through 

https://github.com/emberjs/ember.js/blob/2bb644532d407d4b4700a81f510c5aed4caf5d84/packages/%40ember/-internals/glimmer/lib/component.ts#L647-L652

And figuring out all of the properties defined on those mixins. I _think_ this is the full list:

```
childViews
concatenatedProperties
attributeBindings
element
elementId
tagName
classNames
classNameBindings
target
actionContextObject
action
actionContext
```

lib/rules/no-this-in-template-only-components.js
```diff
@@ -0,0 +1,80 @@
...
+      // template is allowed to use `this`
+      return null;
+    }
+
+    return {
+      PathExpression(path) {
+        if (path.this) {
+          const fixed = path.original.replace(/^this\./, '@');
+          if (this.mode === 'fix' && isFixable(path.original)) {
+            path.original = fixed;
```
rwjblue, line 64: Hmm, I thought we also had to set `path.data = true`?

lib/rules/no-this-in-template-only-components.js
```diff
@@ -0,0 +1,80 @@
...
+const Rule = require('./base');
+
+function message(original, fixed) {
+  return (
+    `Usage of 'this' in path '${original}' is not allowed in a template-only component. ` +
+    `Use '${fixed}' if it is a named argument or create a component.js for this component.`
+  );
+}
+
+function hasContext(hbsFilePath) {
```
rwjblue, line 15: Mind renaming this to `isTemplateOnlyComponent`? I could totally imagine extracting this into a shared utility in the future...

lib/rules/no-this-in-template-only-components.js
```diff
@@ -0,0 +1,80 @@
...
+    if (paths[1] === 'templates') {
+      if (paths[2] === 'components') {
+        // classic structure: /app/templates/components/foo.hbs => /app/components/foo.js
+        const moduleName = path.basename(hbsFilePath, '.hbs');
+        const jsFilePath = path.join(
+          paths[0],
+          'components',
+          ...paths.slice(3, -1),
+          `${moduleName}.js`
+        );
```
rwjblue, line 27: We need to ensure that alternate file types are valid. Can you:

* Add an option to allow `validComponentExtensions: ['.lol', '.js', '.ts']` (and search through that list)
* Default the list of valid extensions to `.js` and `.ts`

rwjblue (commented):
Thanks for working on this!

rwjblue:
@dwickern - 👋 friendly ping 😸

dwickern:
Sorry for the wait and thanks for the thorough review!

lib/rules/no-this-in-template-only-components.js
dwickern: I added those plus `parentView`

lib/rules/no-this-in-template-only-components.js
dwickern: Originally I was setting `path.data = true; path.this = false` but it still worked without those

rwjblue (approved):

Issue #710, opened 2019-05-04
[Rule Idea] Enforce Splattributes


Turbo87:
This rule should check that exactly one element in the template has `...attributes` attached to it.

Alternatively we could make it configurable to only error for multiple `...attributes`, but allow zero and one instances of it.

rwjblue:
Multiple instances of `...attributes` is perfectly valid, any reason you think it should be forbidden?

Turbo87:
seems very confusing to me 🤔

what happens in that case? do they all get the same attributes?

CvX:
Yeah, the attributes are applied to all elements with `...attributes`.

Here's the mention in the RFC: "It can also be used on multiple elements in the same template." (https://emberjs.github.io/rfcs/0311-angle-bracket-invocation.html#html-attributes)

Turbo87:
hmm okay, guess then the rule should only warn if there are no splattributes at all

lifeart:
possible cases to not warn:

- if inside template is only `{{` components / helpers
- if inside template there is no HTML tags

should we recommend any ordering for splattibutes (before / after defined attributes)?

rwjblue:
> should we recommend any ordering for splattibutes (before / after defined attributes)?

Ya, we can recommend (I'd prefer to have it at the end by default) but I don't think we can enforce because precise placement is actually quite important.

lifeart:
I think this rule should work if component has internal "wrapper" node, or component and "wrapped node inside" (only in this case), other cases looks so complex and our linting may confuse developers.

rule apply exmples:

```hbs
<div ...attributes>
 ... some content
</div>
```

```hbs
<LooksLikeComponentNode>
   <div ...attributes></div>
</LooksLikeComponentNode>
```

```hbs
<LooksLikeComponentNode ...attributes />
```
unsupported cases:

```hbs
<div></div>
<div></div>
```

```hbs
<LooksLikeComponentNode>
<div></div>
<div></div>
</LooksLikeComponentNode>
```

```hbs
<LooksLikeComponentNode>
some text
</LooksLikeComponentNode>
```

PR #1672, opened 2020-12-31
Implement `require-splattributes` rule

Turbo87:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/710

---

# require-splattributes

Splattributes (`...attributes`) make it possible to use attributes on component
invocations (e.g. `<SomeComponent class="blue">`). Forgetting to add
`...attributes` however makes it impossible to apply attributes like `class` to
a component.

This rule warns about templates that don't have `...attributes` in them.

Please note that this rule is only useful for Glimmer components or tagless
(`tagName: ''`) classic components, because regular classic components have
this functionality built into the root element, which is not part of their
templates.

## Examples

```hbs
<div>
  component content
</div>
```

```hbs
<SomeOtherComponent>
  component content
</SomeOtherComponent>
```

```hbs
<div ...attributes>
  component content
</div>
```

```hbs
<div class="foo">
  <SomeOtherComponent ...attributes />
</div>
```

## Migration

* Add `...attributes` on at least one element or component invocation in the template (usually the root element)
* Use `{{! template-lint-disable require-splattributes }}` where you explicitly don't want or need `...attributes`

## Related Rules

* [no-nested-splattributes](no-nested-splattributes.md)
* [splat-attributes-only](splat-attributes-only.md)

## References

* [Splattributes](https://guides.emberjs.com/release/components/component-arguments-and-html-attributes/#toc_html-attributes) in the Ember.js guides

### DIFF
--- a/docs/rule/require-splattributes.md
+++ b/docs/rule/require-splattributes.md
@@ -1,0 +1,80 @@
+# require-splattributes
+
+Splattributes (`...attributes`) make it possible to use attributes on component
+invocations (e.g. `<SomeComponent class="blue">`). Forgetting to add
+`...attributes` however makes it impossible to apply attributes like `class` to
+a component.
+
+This rule warns about templates that don't have `...attributes` in them.
+
+Please note that this rule is only useful for Glimmer components or tagless
+(`tagName: ''`) classic components, because regular classic components have
+this functionality built into the root element, which is not part of their
+templates.
+
+This rule also should not be used foo route/controller templates, because those
+don't support `...attributes`. Instead of unconditionally enabling this rule in
+your `.template-lintrc.js` file, you might want to consider using
+[`overrides`](../overrides.md) to only enable it for component templates (see
+below).
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<div>
+  component content
+</div>
+```
+
+```hbs
+<SomeOtherComponent>
+  component content
+</SomeOtherComponent>
+```
+
+This rule **allows** the following:
+
+```hbs
+<div ...attributes>
+  component content
+</div>
+```
+
+```hbs
+<div class="foo">
+  <SomeOtherComponent ...attributes />
+</div>
+```
+
+## Example configuration
+
+```javascript
+module.exports = {
+  extends: 'recommended',
+  rules: {
+    // ...
+  },
+  overrides: [
+    {
+      files: ['app/components/**/*.hbs'],
+      rules: { 'require-splattributes': 'error' },
+    },
+  ],
+};
+```
+
+## Migration
+
+* Add `...attributes` on at least one element or component invocation in the template (usually the root element)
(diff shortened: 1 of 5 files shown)


lib/rules/require-splattributes.js
```diff
@@ -0,0 +1,47 @@
...
+              this.report({
+                message: 'The root element in this template should use `...attributes`',
+                node: elementNodes[0],
+              });
+            } else {
+              this.report({
+                message: 'At least one element in this template should use `...attributes`',
+                node,
+              });
+            }
```
rwjblue, line 32: Isn't this pretty odd for route templates? Where would you put `...attributes` there? 

Is your thought here that you'd only use this rule on the components subdirectory?

Turbo87: indeed, I didn't think about this when I implemented it and I was only thinking about components 😅 

since our config supports `overrides` I suppose it would be possible though to only enable it for component files 🤔 

I suppose you would want me to mention that in the rule docs though?

rwjblue: > I suppose you would want me to mention that in the rule docs though?

Ya, exactly.

Turbo87: done :)

docs/rule/require-splattributes.md
```diff
@@ -0,0 +1,62 @@
...
+
+Please note that this rule is only useful for Glimmer components or tagless
+(`tagName: ''`) classic components, because regular classic components have
+this functionality built into the root element, which is not part of their
+templates.
+
+This rule also should not be used foo route/controller templates, because those
+don't support `...attributes`. Instead of unconditionally enabling this rule in
+your `.template-lintrc.js` file, you might want to consider using
+[`overrides`](../overrides.md) to only enable it for component templates.
```
rwjblue, line 18: Maybe make another section for "recommended configuration" and include an example of how to apply it (e.g. the syntax folks should use in `.template-lintrc.js`)?

Turbo87: sorry for the long delay here. I've added an "Example configuration" section.

docs/rule/require-splattributes.md
```diff
@@ -0,0 +1,80 @@
...
+a component.
+
+This rule warns about templates that don't have `...attributes` in them.
+
+Please note that this rule is only useful for Glimmer components or tagless
+(`tagName: ''`) classic components, because regular classic components have
+this functionality built into the root element, which is not part of their
+templates.
+
+This rule also should not be used foo route/controller templates, because those
```
rwjblue, line 15:
```suggestion
This rule also should not be used for route/controller templates, because those
```

MelSumner, line 15:
```suggestion
This rule also should not be used for route/controller templates, because those
```

rwjblue (approved):

docs/rule/require-splattributes.md
Turbo87: lol, thanks 😅

PR #1675, opened 2021-01-04
Adding ability to set days to decay to warn/error

scalvert:
Implements [due days spec](https://github.com/ember-template-lint/ember-template-lint-todo-utils/blob/master/docs/todo-due-dates.md).


test/helpers/fake-project.js
```diff
@@ -87,6 +87,16 @@ module.exports = class FakeProject extends FixturifyProject {
     this.writeSync();
   }
 
+  writeLintTodo(daysToDecay) {
```
rwjblue, line 90: This seems like an odd function name, it isn't about writing a todo its about writing lintTodoConfig right?

rwjblue (changes requested):

lib/helpers/get-package-json.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+module.exports = function getPackageJson(basePath = process.cwd()) {
```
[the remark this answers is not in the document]
scalvert: Right, `workingDirectory`. 👍

lib/helpers/get-package-json.js
```diff
@@ -0,0 +1,19 @@
...
+  } catch (error) {
+    if (error.code === 'ENOENT') {
+      throw new Error(
+        `The ${path.resolve(basePath)} directory does not contain a package.json file.`
+      );
+    }
+  }
+
+  return _package;
+};
```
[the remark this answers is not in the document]
scalvert: It's fine to move the retrieval of the package.json to that lib. I don't necessarily agree with the parsing of the env/command line values as those are distinctly different based on specific implementation (eslint uses env vars, ember-template-lint uses options). It is, of course, possible to have a custom function passed in that returns the provided options, whether env var or option. I'll think on that a bit.

lib/helpers/get-package-json.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+module.exports = function getPackageJson(basePath = process.cwd()) {
+  let _package = {};
+  const packageJsonPath = path.join(path.resolve(basePath), 'package.json');
+
+  try {
+    _package = fs.readJsonSync(packageJsonPath);
```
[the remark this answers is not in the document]
scalvert: Sure, can do.

test/helpers/fake-project.js
scalvert: `lintTodo` is the name of the key in the package.json file. So, it is writing a `lintTodo`. 🤷‍♂️

lib/helpers/get-todo-config.js
```diff
@@ -0,0 +1,37 @@
+const getPackageJson = require('./get-package-json');
+
+module.exports = function getDaysToDecay(options, pkg = getPackageJson()) {
```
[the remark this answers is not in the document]
scalvert: Maybe. As per my comment below, there's enough different between the two that it made sense to have separate implementations, but there's enough in common that I could figure out a way to extract the relevant info from the consuming project.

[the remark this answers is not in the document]
scalvert: Your example above is really what it would look like for eslint, which uses env vars. Let's not mix the two, or other readers will be confused :)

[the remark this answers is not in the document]
scalvert: Done.

lib/helpers/get-package-json.js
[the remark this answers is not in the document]
scalvert: Done!

lib/helpers/get-package-json.js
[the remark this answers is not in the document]
scalvert: Done.

lib/helpers/get-package-json.js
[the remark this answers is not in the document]
scalvert: Done, but we decided resolve-package-path wasn't needed (we're assuming top-level execution of the command).

test/helpers/fake-project.js
scalvert: Updated.

test/acceptance/cli-test.js
```diff
@@ -1644,6 +1657,668 @@ describe('ember-template-lint executable', function () {
...
+        expect(result.exitCode).toEqual(1);
+        expect(result.stdout).toMatchInlineSnapshot(`
+          "app/templates/application.hbs
+            1:5  error  Non-translated string used  no-bare-strings
+
+          ✖ 1 problems (1 errors, 0 warnings)"
+        `);
+      });
+
+      it('should set todo to error if errorDate has expired via env var', async function () {
```
renatoi, line 2222: In this test you're testing it via CLI options, not env vars.

scalvert: Fixed.

bin/ember-template-lint.js
```diff
@@ -232,6 +240,20 @@ function printPending(results, options) {
   }
 }
 
+function getTodoConfigFromOptions(options) {
+  let todoConfig = {};
+
+  if (Number.isInteger(options.todoDaysToWarn)) {
```
rwjblue, line 246: When do we expect it _not_ to be a number? If it is just when it isn't passed, then I think you don't need the manual checks/gaurds (since yargs will "do the right thing").

So assuming `options` just has `undefined` for these props, couldn't you do:

```js
function getTodoConfigFromOptions(options) {
  return {
    warn: options.todoDaysToWarn,
    error: options.todoDaysToError,
  };
}
```

That _seems_ to give us the same thing? Or is this a case where you don't want the props on `options` at all if the options weren't specified?

bin/ember-template-lint.js
```diff
@@ -232,6 +240,20 @@ function printPending(results, options) {
   }
 }
 
+function getTodoConfigFromOptions(options) {
```
rwjblue, line 243: Would you mind renaming this (sorry even longer :sob: ) to `getTodoConfigFromCommandLineOptions`?

lib/-private/todo-handler.js
```diff
@@ -64,7 +70,11 @@ module.exports = class TodoHandler {
           result.column === todo.column
       );
 
-      result.severity = TODO_SEVERITY;
+      if (todo.warnDate instanceof Date && today > todo.warnDate) {
+        result.severity = WARNING_SEVERITY;
+      } else {
+        result.severity = TODO_SEVERITY;
+      }
```
rwjblue, line 77: I'd like to add a utility function for this over in the todo utils. 

```js
result.severity = getTodoSeverity(todo);
```

^ would allow both the eslint formatter and ember-template-lint to use the same shared func and allow you to remove the additional early return just above this area (because `getTodoSeverity(todo)` would return `ERROR_SEVERITY` when the `todo.errorDate` has already passed)

lib/linter.js
```diff
@@ -262,8 +262,8 @@ class Linter {
     return currentSource;
   }
 
-  async updateTodo(linterOptions, results) {
-    await this._todoHandler.update(linterOptions.filePath, results);
+  async updateTodo(linterOptions, results, daysToDecay) {
+    await this._todoHandler.update(linterOptions.filePath, results, daysToDecay);
```
rwjblue, line 266: Can we call this more generally `todoConfig`? Its _possible_ that we add more config values other than the decay days itself...

package.json
```diff
@@ -37,10 +37,11 @@
     ]
   },
   "dependencies": {
-    "@ember-template-lint/todo-utils": "^3.2.0",
+    "@ember-template-lint/todo-utils": "^4.1.0",
     "chalk": "^4.0.0",
     "ember-template-recast": "^5.0.1",
     "find-up": "^5.0.0",
+    "fs-extra": "^9.0.1",
```
rwjblue, line 44: Hmm, where is this used?

lib/-private/todo-handler.js
```diff
@@ -55,7 +55,13 @@ module.exports = class TodoHandler {
       }
     }
 
+    const today = new Date();
+
     [...existingTodos.values()].forEach((todo) => {
+      if (todo.errorDate instanceof Date && today > todo.errorDate) {
```
rwjblue, line 61: Should we normalize these dates somehow? I _think_ `new Date()` is going to be for the current date and time (including timezone?).

It would be bad/odd if things pass in CI because it runs in GMT but locally I'm running in EST/PST...

lib/-private/todo-handler.js
```diff
@@ -64,7 +70,11 @@ module.exports = class TodoHandler {
           result.column === todo.column
       );
 
-      result.severity = TODO_SEVERITY;
+      if (todo.warnDate instanceof Date && today > todo.warnDate) {
```
rwjblue, line 73: Same thing RE: timezones/time of day...

bin/ember-template-lint.js
scalvert: That's a good point, and we _could_ do that, but to your later point yes, I don't want to pass anything if they're not present. This means that we won't inadvertently set the values to undefined, which is what would happen.

bin/ember-template-lint.js
scalvert: OMG that's how I had it, but it felt so C#-y that I had to rename. I'm happy to put it back! Will harken me back to my C# days!

lib/-private/todo-handler.js
scalvert: And you'd evaluate the date from the passed in todo, and pass back the right severity? Seems like a good idea.

bin/ember-template-lint.js
rwjblue: Seems good to me, dealers choice here 😸

lib/-private/todo-handler.js
scalvert: Also, seems severe... :)

lib/linter.js
scalvert: Ah yes, this was a spot I missed. I'll update.

package.json
scalvert: I think it was added during the time before we extracted the package.json/todo config code. I'll remove, since it's not used.

lib/-private/todo-handler.js
rwjblue: > And you'd evaluate the date from the passed in todo, and pass back the right severity?

Yeah exactly. This will help us ensure that our two clients stay in sync over time (especially worried WRT the timezone thing, and encapsulating that into the utils repo helps centralize the logic).

lib/-private/todo-handler.js
scalvert: Ya, we should be setting the hour/min/second to 0 I believe.

lib/-private/todo-handler.js
scalvert: Same as above.

lib/-private/todo-handler.js
rwjblue: seems good, but I think that will still have the GMT / PDT issue won't it?

scalvert: I think this means we want to set `createdDate` to that in addition, in todo-utils.

rwjblue: Either way, I think if we encapsulate the logic in the utils we don't have to perfectly nail this one

lib/-private/todo-handler.js
scalvert: Confirm.

lib/-private/todo-handler.js
scalvert: Yep, agree. Will update that repo.

scalvert: I think we need to store the dates in UTC, and do the local conversion and comparison from that. Timezone comparisons are particularly finicky, so I'll see what I can come up with.

bin/ember-template-lint.js

lib/-private/todo-handler.js
scalvert: Done.

lib/linter.js
scalvert: Done.

package.json
scalvert: Done.

lib/-private/todo-handler.js
scalvert: Done.

lib/-private/todo-handler.js
scalvert: Done.

lib/-private/todo-handler.js
```diff
@@ -56,6 +57,12 @@ module.exports = class TodoHandler {
     }
 
     [...existingTodos.values()].forEach((todo) => {
+      let severity = getSeverity(todo);
+
+      if (severity === ERROR_SEVERITY) {
+        return;
+      }
```
rwjblue, line 64: We _probably_ don't need this short circuit here (since the `result.severity = severity` would just set it back to `ERROR_SEVERITY`), but seems good to avoid the wasted work finding the associated `result` item. :thumbsup:

rwjblue (approved):

lib/-private/todo-handler.js
scalvert: Ya I mainly wanted to early return to avoid unnecessary work.

PR #1683, opened 2021-01-12
Update error message for resolved todos

MelSumner:
If merged, this PR will provide a small change to the message that shows in the CLI, based on using the feature and finding that the documented command was not the same as the CLI message.

### DIFF
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -29,7 +29,7 @@ ember-template-lint . --include-todo
 If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--update-todo`
 
 ```bash
-ember-template-lint . --update-todo
+ember-template-lint . --fix
 ```
 
 ### Configuring Due Dates
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -26,7 +26,7 @@ If you want to see TODOs as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--update-todo`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--fix`
 
 ```bash
 ember-template-lint . --fix
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -26,10 +26,10 @@ If you want to see TODOs as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--update-todo`
 
 ```bash
-ember-template-lint . --fix
+ember-template-lint . --update-todo
 ```
 
 ### Configuring Due Dates
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -46,7 +46,7 @@ module.exports = class TodoHandler {
         itemsToRemoveFromTodos.forEach((todo) => {
           results.push({
             rule: 'invalid-todo-violation-rule',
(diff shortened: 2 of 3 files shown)


docs/todos.md
```diff
@@ -26,10 +26,10 @@ If you want to see TODOs as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--update-todo`
 
 ```bash
-ember-template-lint . --fix
+ember-template-lint . --update-todo
```
rwjblue, line 32: The original content is correct, though perhaps the prose leading up to this is a bit confusing.

This is specifically talking about the following:

* you have a pre-existing issue in a template, that you had previously ran `--update-todo` for
* you **fix** that issue (no longer requiring the todo file)
* at this point, running `ember-template-lint` on that file will error (because there is a stale todo that needs to be cleaned up)
* running `ember-template-lint --fix` will do that cleanup for you

rwjblue (changes requested):

docs/todos.md
MelSumner: @rwjblue 2 concerns: 
1. see the screenshot that I posted - it explicitly instructs the user to run `--update-todo` to remove the stale todo, hence the PR. I don't know how much more proof I could provide here. 

2. This is perhaps orthogonal to this PR but related in principle: `--fix` is already used to apply auto fixes other things; I had planned to raise this concern even before I discovered that the documentation was incorrect. What happens if I don't want it to also auto-fix everything else?

rwjblue: You _can_ use `--update-todo` that is definitely fine. But `--fix` is a much more preferred here generally, using `--update-todo` will force **every** current violation to be migrated to a todo whereas `--fix` will only ever remove a todo that is no longer needed.

Also, it should *always* be safe to run `--fix`. If it is not in some circumstance, please open an issue dedicated to that scenario.

rwjblue:
The thing to fix here is not the docs (though it would probably be fine to suggest that either `--update-todo` or `--fix` can fix that stale todo for you) but we should fix the error message that we add when there is stale todo so that we recommend `--fix` instead.

rwjblue:
That would be done here:

https://github.com/ember-template-lint/ember-template-lint/blob/dbc87799a57b23c1dfdb7af68442ae875d1b4365/lib/-private/todo-handler.js#L49

MelSumner:
ok, I'll do that.

docs/todos.md
scalvert: We should definitely update the error message. That was written before we implement the auto-fix capability. 

Also was going to mention what Rob said about the usage of `--fix`. Why would you ever not want `--fix` to fix all the things that can be fixed? I also believe, as @rwjblue said, it should always be safe to run this.

I think the action item from this is to update the error message.

scalvert:
Correct, the error message just needs to be updated.

scalvert:
@MelSumner are you going to take care of updating the error message, or would you like me to?

docs/todos.md
MelSumner, line 32:
```suggestion
ember-template-lint . --fix
```

docs/todos.md
```diff
@@ -26,10 +26,10 @@ If you want to see TODOs as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--update-todo`
```
MelSumner, line 29:
```suggestion
If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding TODO file. You can remove this file by running `--fix`
```

MelSumner:
@scalvert @rwjblue updated.

rwjblue (commented):
Looks good, can you also update the PR description + title?

MelSumner:
Sure, apologies for the delay, juggling a few things this afternoon.

scalvert (approved):

scalvert:
Thanks!

PR #1687, opened 2021-01-12
Respect extends order and rules overrides

### LINKED ISSUE #1417: Rules override is reversed on extends

### LINKED ISSUE #1644: Unable to turn rules off in `next` branch
I tried using the `next` branch (for the upcoming V3 release #1315) and it seems that my configuration file now fails to turn off rules, here's an example:

This fails to turn the eol-last rule off (violations are reported from this rule still):

```js
// .template-lintrc.js
module.exports = {
  extends: ['stylistic'],
  rules: {
    'eol-last': false,
  },
};
```

Note that this successfully turns the eol-last rule on:

```js
// .template-lintrc.js
module.exports = {
  extends: [],
  rules: {
    'eol-last': ['error', 'editorconfig'],
  },
};
```

Note that the behavior is correct on master.

renatoi:
Added tests cases that cover these two cases and fixes the issues reported on both:

-  Rules override is reversed on extends. Fixes #1417 
- Unable to turn rules off in `next` branch. Fixes #1644

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -253,6 +253,7 @@ function processExtends(config, options) {
   let logger = options.console || console;
 
   let extendedList = normalizeExtends(config, options);
+  let extendedRules = {};
 
   if (extendedList) {
     for (const extendName of extendedList) {
@@ -268,7 +269,7 @@ function processExtends(config, options) {
           delete configuration.loadedConfigurations;
 
           if (configuration.rules) {
-            config.rules = Object.assign({}, config.rules, configuration.rules);
+            extendedRules = Object.assign({}, extendedRules, configuration.rules);
           } else {
             logger.log(chalk.yellow(`Missing rules for extends: ${extendName}`));
           }
@@ -279,6 +280,8 @@ function processExtends(config, options) {
 
       delete config.extends;
     }
+
+    config.rules = Object.assign({}, extendedRules, config.rules);
   }
 }
 
--- a/test/unit/get-config-test.js
+++ b/test/unit/get-config-test.js
@@ -674,7 +674,7 @@ describe('getProjectConfig', function () {
     expect(reprocessedConfig.rules).toEqual(expected);
   });
 
-  it('rules override extends', function () {
+  it('merges rules from plugins with rules from config', function () {
     let config = {
       plugins: [
         {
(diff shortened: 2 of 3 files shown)


test/unit/get-config-test.js
```diff
@@ -673,6 +673,78 @@ describe('getProjectConfig', function () {
     let reprocessedConfig = getProjectConfig(project.baseDir, { config });
     expect(reprocessedConfig.rules).toEqual(expected);
   });
+
+  it('rules override extends', function () {
```
rwjblue, line 677: Can you add another test like this one, that uses overrides by filePath? It **should** work the same with your fixes, but I want to make sure it stays under test too...

lib/get-config.js
```diff
@@ -268,7 +269,7 @@ function processExtends(config, options) {
           delete configuration.loadedConfigurations;
 
           if (configuration.rules) {
-            config.rules = Object.assign({}, config.rules, configuration.rules);
+            extendedRules = Object.assign({}, extendedRules, configuration.rules);
```
scalvert, line 272: Am I missing something? Why don't we just do

```suggestion
            extendedRules = Object.assign({}, configuration.rules);
```

scalvert (approved):

lib/get-config.js
scalvert: I did miss something. It's looping.

test/unit/get-config-test.js
rwjblue: Also lets avoid the word "override" here (since we have an actual concept in this domain that is `overrides`)

renatoi: Done. I added the new test cases in `modules-status-cache-test.js` because that's where the config tests for a specific file lives, but I had to add a `getProjectConfig()` first so that we cover all cases (`extends` vs `rules` vs `overrides`).

rwjblue (approved):

PR #1688, opened 2021-01-13
Adds output feedback when creating TODOs

### LINKED ISSUE #1685: TODO feature: `todo-days-to-warn` and `todo-days-to-error` commands lack CLI messages
Expected: 
1. When I use the `--todo-days-to-warn` or `--todo-days-to-error` CLI commands correctly, I should receive a confirmation message in my CLI that the days have been set accordingly. 
2. When I use the `--todo-days-to-warn` or `--todo-days-to-error` CLI commands incorrectly, I should receive an error message in my CLI, telling me what I have done incorrectly.

Actual: 
Whatever happens, happens silently. There is no confirmation either way.

scalvert:
Resolves #1685

When using `--todo-days-to-warn` or `--todo-days-to-error` without `--update-todo`, it prints the following error:

```shell
You must use `--update-todo` when using any of `--todo-days-to-warn` or `--todo-days-to-error`.
```

When new todos are created:

```shell
✔ 1 todos created
```

When new todo created with `warn` days:

```shell
✔ 1 todos created (warn after 10 days)
```

When new todo created with `error` days:

```shell
✔ 1 todos created (error after 10 days)
```

When new todo created with `warn` and `error` days:

```shell
✔ 1 todos created (warn after 5 and error after 10 days)
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,7 +70,7 @@ module.exports = {
     'prefer-rest-params': 'error',
     'prefer-spread': 'error',
     'prefer-template': 'error',
-    quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: false }], // Disallow unnecessary template literals.
+    quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: true }],
     radix: 'error',
     'require-atomic-updates': 'error',
     'require-await': 'error',
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -345,10 +345,11 @@ async function run() {
         getTodoConfigFromCommandLineOptions(options)
       );
 
-      let [added] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
+      let [added, removed] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
 
       todoInfo = {
         added,
+        removed,
         todoConfig,
       };
     }
@@ -369,12 +370,12 @@ async function run() {
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (
-      results.errorCount ||
-      results.warningCount ||
-      (options.includeTodo && results.todoCount) ||
-      (options.updateTodo && todoInfo.added)
-    ) {
+    let hasErrors = results.errorCount > 0;
+    let hasWarnings = results.warningCount > 0;
+    let hasTodos = options.includeTodo && results.todoCount;
+    let hasUpdatedTodos = options.updateTodo && (todoInfo.added || todoInfo.removed);
+
+    if (hasErrors || hasWarnings || hasTodos || hasUpdatedTodos) {
       let Printer = require('../lib/printers/default');
       let printer = new Printer(options);
       printer.print(results, todoInfo);
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -258,6 +258,7 @@ async function run() {
   let options = parseArgv(process.argv.slice(2));
   let positional = options._;
   let config;
+  let todoInfo;
 
   if (options.config) {
     try {
@@ -307,6 +308,14 @@ async function run() {
     return;
   }
 
+  if ((options.todoDaysToWarn || options.todoDaysToError) && !options.updateTodo) {
+    console.error(
+      'You must use `--update-todo` when using any of `--todo-days-to-warn` or `--todo-days-to-error`.'
+    );
+    process.exitCode = 1;
+    return;
+  }
+
   let filePaths = getFilesToLint(options.workingDirectory, positional, options.ignorePattern);
 
   let resultsAccumulator = [];
@@ -331,11 +340,17 @@ async function run() {
     }
 
     if (options.updateTodo) {
-      await linter.updateTodo(
-        linterOptions,
-        fileResults,
-        getTodoConfig(options.workingDirectory, getTodoConfigFromCommandLineOptions(options))
+      let todoConfig = getTodoConfig(
+        options.workingDirectory,
+        getTodoConfigFromCommandLineOptions(options)
       );
+
+      let [added] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
+
+      todoInfo = {
+        added,
+        todoConfig,
+      };
     }
 
     if (!filePaths.has(STDIN)) {
@@ -354,10 +369,15 @@ async function run() {
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount || (options.includeTodo && results.todoCount)) {
+    if (
+      results.errorCount ||
+      results.warningCount ||
+      (options.includeTodo && results.todoCount) ||
+      (options.updateTodo && todoInfo.added)
+    ) {
       let Printer = require('../lib/printers/default');
       let printer = new Printer(options);
-      printer.print(results);
+      printer.print(results, todoInfo);
     }
   }
 }
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -373,7 +373,9 @@ async function run() {
     let hasErrors = results.errorCount > 0;
     let hasWarnings = results.warningCount > 0;
     let hasTodos = options.includeTodo && results.todoCount;
-    let hasUpdatedTodos = options.updateTodo && (todoInfo.added || todoInfo.removed);
+    let hasUpdatedTodos =
+      options.updateTodo &&
+      (Number.isInteger(todoInfo.added) || Number.isInteger(todoInfo.removed));
 
     if (hasErrors || hasWarnings || hasTodos || hasUpdatedTodos) {
       let Printer = require('../lib/printers/default');
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -310,7 +310,7 @@ async function run() {
 
   if ((options.todoDaysToWarn || options.todoDaysToError) && !options.updateTodo) {
     console.error(
-      'You must use `--update-todo` when using any of `--todo-days-to-warn` or `--todo-days-to-error`.'
+      'Using `--todo-days-to-warn` or `--todo-days-to-error` is only valid when the `--update-todo` option is being used.'
     );
     process.exitCode = 1;
     return;
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -17,10 +17,10 @@ module.exports = class TodoHandler {
     this._processedResults = [];
   }
 
-  async update(filePath, results, todoConfig) {
+  update(filePath, results, todoConfig) {
     this._processedResults = this._buildResult(results);
 
-    await writeTodos(this._workingDir, this._processedResults, filePath, todoConfig);
+    return writeTodos(this._workingDir, this._processedResults, filePath, todoConfig);
   }
 
   async processResults(filePath, results, shouldFix) {
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -17,10 +17,17 @@ module.exports = class TodoHandler {
     this._processedResults = [];
   }
 
-  update(filePath, results, todoConfig) {
+  async update(filePath, results, todoConfig) {
     this._processedResults = this._buildResult(results);
 
-    return writeTodos(this._workingDir, this._processedResults, filePath, todoConfig);
+    let todoBatchCounts = await writeTodos(
+      this._workingDir,
+      this._processedResults,
+      filePath,
+      todoConfig
+    );
+
+    return todoBatchCounts;
   }
 
   async processResults(filePath, results, shouldFix) {
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -262,8 +262,8 @@ class Linter {
     return currentSource;
   }
 
-  async updateTodo(linterOptions, results, todoConfig) {
-    await this._todoHandler.update(linterOptions.filePath, results, todoConfig);
+  updateTodo(linterOptions, results, todoConfig) {
+    return this._todoHandler.update(linterOptions.filePath, results, todoConfig);
   }
 
   processTodos(linterOptions, results, shouldFix) {
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -262,12 +262,24 @@ class Linter {
     return currentSource;
   }
 
-  updateTodo(linterOptions, results, todoConfig) {
-    return this._todoHandler.update(linterOptions.filePath, results, todoConfig);
+  async updateTodo(linterOptions, results, todoConfig) {
+    let todoBatchCounts = await this._todoHandler.update(
+      linterOptions.filePath,
+      results,
+      todoConfig
+    );
+
+    return todoBatchCounts;
   }
 
-  processTodos(linterOptions, results, shouldFix) {
-    return this._todoHandler.processResults(linterOptions.filePath, results, shouldFix);
+  async processTodos(linterOptions, results, shouldFix) {
+    let fileResults = await this._todoHandler.processResults(
+      linterOptions.filePath,
+      results,
+      shouldFix
+    );
+
+    return fileResults;
   }
 
   /**
--- a/lib/printers/pretty.js
+++ b/lib/printers/pretty.js
@@ -57,7 +57,7 @@ class PrettyPrinter {
     }
 
     if (this.options.updateTodo && todoInfo) {
-      let todoSummary = `✔ ${todoInfo.added} TODOs created `;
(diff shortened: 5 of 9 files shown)


.eslintrc.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
     // Optional eslint rules:
     'array-callback-return': 'error',
     'block-scoped-var': 'error',
-    complexity: 'error',
+    complexity: ['error', 25],
```
scalvert, line 32: Grrrrrr

.eslintrc.js
```diff
@@ -72,7 +72,7 @@ module.exports = {
     'prefer-rest-params': 'error',
     'prefer-spread': 'error',
     'prefer-template': 'error',
-    quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: false }], // Disallow unnecessary template literals.
+    quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: true }],
```
scalvert, line 75: I'm not sure why this was set to `false`, as this seemed to cause issues with inline snapshots 🤷

MelSumner (approved):
Thank you for doing this, the error message is helpful. I think the linter is complaining about line 13 in `lib/rules/no-negated-condition.js` but once the tests pass it looks good to me.

lib/linter.js
```diff
@@ -262,8 +262,8 @@ class Linter {
     return currentSource;
   }
 
-  async updateTodo(linterOptions, results, todoConfig) {
-    await this._todoHandler.update(linterOptions.filePath, results, todoConfig);
+  updateTodo(linterOptions, results, todoConfig) {
+    return this._todoHandler.update(linterOptions.filePath, results, todoConfig);
```
rwjblue, line 266: I'd like to revert these changes and keep this declared as an async function. It makes it much easier to spot on quick review that it is returning a promise.

bin/ember-template-lint.js
```diff
@@ -307,6 +308,14 @@ async function run() {
     return;
   }
 
+  if ((options.todoDaysToWarn || options.todoDaysToError) && !options.updateTodo) {
+    console.error(
+      'You must use `--update-todo` when using any of `--todo-days-to-warn` or `--todo-days-to-error`.'
+    );
```
rwjblue, line 314: Lets tweak it slightly to:

> Using `--todo-days-to-warn` or `--todo-days-to-error` is only valid when the `--update-todo` option is being used.

bin/ember-template-lint.js
```diff
@@ -354,10 +369,15 @@ async function run() {
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount || (options.includeTodo && results.todoCount)) {
+    if (
+      results.errorCount ||
+      results.warningCount ||
+      (options.includeTodo && results.todoCount) ||
+      (options.updateTodo && todoInfo.added)
+    ) {
```
rwjblue, line 377: This is getting a tad complicated to reason about. Maybe break this down into some local variables?

Something like:

```js
let hasErrors = results.errorCount > 0;
let hasTodoInfo = options.includeTodo && results.todoCount > 0;
let hasUpdatedTodos = options.updateTodo && todoInfo.added > 0;

if (hasErrors || hasWarnings || hasTodoInfo || hasUpdatedTodos) {
```

FWIW, I think this may also allow you to remove the `.eslintrc.js` change...

lib/-private/todo-handler.js
```diff
@@ -17,10 +17,10 @@ module.exports = class TodoHandler {
     this._processedResults = [];
   }
 
-  async update(filePath, results, todoConfig) {
+  update(filePath, results, todoConfig) {
     this._processedResults = this._buildResult(results);
 
-    await writeTodos(this._workingDir, this._processedResults, filePath, todoConfig);
+    return writeTodos(this._workingDir, this._processedResults, filePath, todoConfig);
```
rwjblue, line 23: I'd prefer to keep this as an explicit async function. I realize that we are just returning a promise here, but declaring it as an async function makes it much easier to grok this file without being intimately familiar with the `@ember-template-lint/todo-utils` libraries API (e.g. does `writeTodos` return a promise?).

bin/ember-template-lint.js
```diff
@@ -331,11 +340,17 @@ async function run() {
...
-      await linter.updateTodo(
-        linterOptions,
-        fileResults,
-        getTodoConfig(options.workingDirectory, getTodoConfigFromCommandLineOptions(options))
+      let todoConfig = getTodoConfig(
+        options.workingDirectory,
+        getTodoConfigFromCommandLineOptions(options)
       );
+
+      let [added] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
```
rwjblue, line 348: We should probably also keep track of removals, right? In theory someone could use `--update-todo` instead of `--fix` to clear a prior todo that is no longer valid.

bin/ember-template-lint.js
rwjblue: Oh, also, I think we also need to track removals that are **done** by `--fix` so there is an additional case to account for here.

lib/linter.js
rwjblue: It would basically be `return await` I guess?

lib/printers/pretty.js
```diff
@@ -55,6 +55,24 @@ class PrettyPrinter {
...
+      if (todoInfo.todoConfig) {
+        let todoConfig = todoInfo.todoConfig;
+
+        if (todoConfig.warn && todoConfig.error) {
+          todoSummary += `(warn after ${todoConfig.warn} and error after ${todoConfig.error} days)`;
+        } else if (todoConfig.warn) {
+          todoSummary += `(warn after ${todoConfig.warn} days)`;
+        } else if (todoConfig.error) {
+          todoSummary += `(error after ${todoConfig.error} days)`;
+        }
```
rwjblue, line 71: I think I'd prefer to avoid the "and" between the two (when both error and warn are present), replacing it either with a comma or semicolon.

I think that may also allow us to simplify the logic slightly, but not sure (building end user helpful prose is always slightly annoying 😝 ).

What do you think?

lib/printers/pretty.js
```diff
@@ -55,6 +55,24 @@ class PrettyPrinter {
         this.console.log(chalk.red(chalk.bold(fixableSummary)));
       }
     }
+
+    if (this.options.updateTodo && todoInfo) {
```
rwjblue, line 59: Unless I'm mistaken, this probably should be `todoInfo.added > 0`. Otherwise, we would print the message `0 todos created`.

Is that what we want? (I'm not sure)

test/acceptance/cli-test.js
```diff
@@ -1606,6 +1619,91 @@ describe('ember-template-lint executable', function () {
         expect(result.stdout).toEqual('');
       });
 
+      it('with --update-todo, outputs todos created summary', async function () {
```
rwjblue, line 1622: Can you add a similar test when there are no todos needed?

lib/-private/todo-handler.js
rwjblue: After thinking on it, I think this might have been forced on you by linting configuration. If that is the case, we should absolutely change that config. Using an `async` without an `await` is perfectly fine/valid IMHO and we shouldn't prevent it.

lib/linter.js
rwjblue: Mentioned above, if this was driven by linting we should fix the linting.

rwjblue (commented):
Thanks for pushing this forward @scalvert!

lib/linter.js
scalvert: It was driven by linting. One thing to note is that `processTodos` below also returns a promise, but the signature matches this change. So, either I think we do both one way, or the other. I'm assuming you'd prefer to `return await` to better signify a promise being returned, correct. 

Also important to note, if we used TS we'd be able to mark the return type and avoid this back and forth :)

bin/ember-template-lint.js
scalvert: Y u no suggest?

bin/ember-template-lint.js
scalvert: > FWIW, I think this may also allow you to remove the .eslintrc.js change...

It does not, as that was triggered in `pretty.js` too, unfortunately.

scalvert: > Oh, also, I think we also need to track removals that are done by --fix so there is an additional case to account for here.

`--update-todo` and `--fix` are mutually exclusive, so I don't think that's the case. I do think we should error when these flags are used in conjunction.

lib/-private/todo-handler.js
scalvert: Correct. I can change the config.

bin/ember-template-lint.js
scalvert: Yep. I wasn't sure how far we wanted to go in reporting this stuff. I'm happy to add it.

lib/printers/pretty.js
scalvert: Yep. This building was done simply to accommodate the `and`, so removing it will for sure simplify.

lib/printers/pretty.js
scalvert: Well if you explicitly use `--update-todo` and get no output, even if no todos are created, is that what we want/correct? I would say that outputting explicitly `0 Todos created` seems appropriate to me in this case. WDYT? 

If you agree, I should add a test case for this.

test/acceptance/cli-test.js
scalvert: You mean add a test without the `--update-todo` option, and that results in the message _not_ being displayed?

bin/ember-template-lint.js
```diff
@@ -307,6 +308,14 @@ async function run() {
     return;
   }
 
+  if ((options.todoDaysToWarn || options.todoDaysToError) && !options.updateTodo) {
+    console.error(
+      'You must use `--update-todo` when using any of `--todo-days-to-warn` or `--todo-days-to-error`.'
```
MelSumner, line 313:
```suggestion
      'Using `--todo-days-to-warn` or `--todo-days-to-error` is only valid when the `--update-todo` option is being used.'
```

lib/linter.js
rwjblue: We removed the no-return-await bit in https://github.com/ember-template-lint/ember-template-lint/pull/1689 so I think we can restructure this back to an async function now.

bin/ember-template-lint.js
rwjblue: lol, :trollface: because it would fail your tests

bin/ember-template-lint.js
rwjblue: Sure, that seems fine, but that wasn't what I meant.

I'm saying we output in stdout whenever we mutate todos, which means `--fix` will also need to be able to trigger some content here (e.g. `2 todos removed`).

lib/printers/pretty.js
rwjblue: Ya, that seems good to me (to basically **always** print something about todos whenever `--update-todo` is used).

test/acceptance/cli-test.js
rwjblue: Yes, kinda. Though in the other comment thread I agree with you that we should always print some message about todos created/removed whenever `--update-todo` is used. So in that case, the new test should basically just confirm that `--update-todo` with no prior failures results in a message like `0 todos created`.

lib/linter.js
scalvert: Done.

bin/ember-template-lint.js

bin/ember-template-lint.js
scalvert: Oh sorry I did misinterpret what you were saying. We can additionally track removals for `--fix`.

lib/-private/todo-handler.js
scalvert: Done and this is fixed.

test/acceptance/cli-test.js
scalvert: Agree.

lib/printers/pretty.js
scalvert: 👍

test/acceptance/cli-test.js
scalvert: Done.

bin/ember-template-lint.js
```diff
@@ -354,10 +370,15 @@ async function run() {
   if (options.printPending) {
     return printPending(results, options);
   } else {
-    if (results.errorCount || results.warningCount || (options.includeTodo && results.todoCount)) {
+    let hasErrors = results.errorCount > 0;
+    let hasWarnings = results.warningCount > 0;
+    let hasTodos = options.includeTodo && results.todoCount;
+    let hasUpdatedTodos = options.updateTodo && (todoInfo.added || todoInfo.removed);
```
rwjblue, line 376: I think this needs to be updated for `--fix` support. Because we don't want to allow `--fix --update-todo` together. 

More directly: `todoInfo.removed` may have values **regardless** of `options.updateTodo`.

rwjblue: I wonder if we can do:

```suggestion
    let hasUpdatedTodos = todoInfo.added > 0 || todoInfo.removed > 0;
```

We only expect `added` / `removed` to be set when `--update-todo` is set, so we can leverage that and simplify the boolean condition maybe... 🤔

test/acceptance/cli-test.js
```diff
@@ -1626,7 +1724,8 @@ describe('ember-template-lint executable', function () {
           "app/templates/application.hbs
             1:5  todo  Non-translated string used  no-bare-strings
 
-          ✖ 1 problems (0 errors, 0 warnings, 1 todos)"
+          ✖ 1 problems (0 errors, 0 warnings, 1 todos)
+          ✔ 1 todos created "
```
rwjblue, line 1728: :sob: what do we do about this pluralization? Just ignore it?

I think we've talked about it before...

bin/ember-template-lint.js

lib/printers/pretty.js
scalvert: Done.

bin/ember-template-lint.js
scalvert: Again, we don't want to compare to 0 - we want 0 states to be output, but yes, we can remove the `options.updateTodo` portion.

scalvert: I'm actually waffling a bit on whether we should have output with `--fix`. I think it makes sense to me to report output with an explicit action like `--update-todo`. With `--fix` I feel like this is a side-effect of the fixing, in that we're just being good citizens with cleaning up ghost files. I'd be inclined to leave it out. WDYT?

test/acceptance/cli-test.js
scalvert: I think so. I considered it, but wanted to keep it consistent with the other output's lack of pluralization.

scalvert: I'd prefer it was consistent than 'right'. If we want to fix all the pluralization, I'd be game (in a separate PR), but consistency > correctness (though one could argue that consistency _is_ correct!).

bin/ember-template-lint.js
scalvert: See my comment [here](https://github.com/ember-template-lint/ember-template-lint/pull/1688#discussion_r556815114) about supporting `--fix` and LMKWYT.

bin/ember-template-lint.js
```diff
@@ -307,6 +308,14 @@ async function run() {
     return;
   }
 
+  if ((options.todoDaysToWarn || options.todoDaysToError) && !options.updateTodo) {
+    console.error(
+      'Using `--todo-days-to-warn` or `--todo-days-to-error` is only valid when the `--update-todo` option is being used.'
+    );
+    process.exitCode = 1;
+    return;
+  }
```
rwjblue, line 317: This is great, do we need to write _basically_ the same thing as this for using `--update-todo --fix`?

bin/ember-template-lint.js
rwjblue: >  since we only set todoInfo to an object in that condition

I think we ought to initialize it regardless (makes it simpler logic wise, since we don't have to guard all over the place).

> I'm actually waffling a bit on whether we should have output with --fix. I think it makes sense to me to report output with an explicit action like --update-todo. With --fix I feel like this is a side-effect of the fixing, in that we're just being good citizens with cleaning up ghost files. I'd be inclined to leave it out. WDYT?

We should mirror what we do for `--fix`ed rules (non-todo), do we tell folks that we fixed/mutated a bunch of things? I _think_ a `--fix` run is completely silent (it emits the same output as `ember-template-lint .`, the remaining issues).

tldr; we can leave `--fix` info out for now 😸

PR #1692, opened 2021-01-14
Update README.md

MelSumner:
Removed `:dress:` from `no-whitespace-for-layout` and `no-whitespace-within-word` since these are A11y rules, not stylistic preferences.


bmish:
I was thinking of the `stylistic` config as representing any rules related to code formatting / spacing which I thought would include these two rules.

MelSumner:
@bmish I was thinking that if a rule was related to a potential legal liability, we should perhaps not call it stylistic. What do you think?

bmish:
What do you mean "potential legal liability"? The `stylistic` config has a number of rules related to whitespace / indentation already.

MelSumner:
@bmish the two rules in this PR, `no-whitespace-for-layout` and `no-whitespace-within-word` are related to WCAG success criteria (for digital accessibility). It's the accessibility part that has the potential for legal liability, not the stylistic part. Does that make more sense?

MelSumner:
Ok. I 'll close this PR then.

PR #1697, opened 2021-01-17
Allow rule visitor to be an async function (remove `env` argument)

lifeart:
tldr: this feature allow us to have async `visitor` function for template-lint plugin

why: async visitor allow us implement side indegrations inside template-lint plugin (language server integration, a11y tooling integration, side process integration)

for example:

```js
class Plugin extends Base {
  async visitor(env) {
    const project = await languageServer.getProject(env.filePath);
    const registry = project.getReferences();
    return {
      ElementNode(node) {
         const componentName = toComponentName(node.tag);
         const usages = registry[componentName] || [];
          if (usages.length === 0) {
             this.logError({
                msg: `Unused component ${componentName}`
              })
          }
      }
    }
  }
}
```

```js
class Plugin extends Base {
  async visitor(env) {
    const component = await a11yTemplateLinter.renderComponent(env.filePath);
    const a11yIssues = await component.getErrors();
    return {
      Template() {
        if (a11yIssues.length) {
          this.logError({
            msg: `A11Y error ${env.filePath}`
          })
        }
      }
    }
  }
}
```

```js
class Plugin extends Base {
  async visitor(env) {
    const project = await languageServer.getProject(env.filePath);
    const relatedFiles = await project.relatedFiles(env.filePath);
    const testFile = relatedFiles.find(file => file.includes('/tests/'));
    if (!testFile) {
        return;
    }
    const testModuleName = getModuleNameForTest(testFile);
    const testResults = await emberCli.runTests({ filter: testModuleName });
    const coverage = await testResults.getCoverageResult(env.filePath);
    return {
      ElementNode(node) {
        if (!coverage.hasCoverageForLoc(node.loc)) {
          this.logError({
            msg: `Is not covered by tests`
          })
        }
      }
    }
  }
}
```

cc @rwjblue, @MelSumner

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -208,7 +208,7 @@ class Linter {
       shouldVerify = false;
 
       if (iterations < MAX_AUTOFIX_PASSES) {
-        const output = this._applyFixes(options, messages);
+        const output = await this._applyFixes(options, messages);
 
         // let's lint one more time if source was modified
         shouldVerify = output !== currentSource;
@@ -231,9 +231,9 @@ class Linter {
    * @param {string} options.moduleId
    * @param {Object[]} results - The lint messages.
    *
-   * @returns {string} output - The fixed source.
+   * @returns {Promise<string>} output - The fixed source.
    */
-  _applyFixes(options, results) {
+  async _applyFixes(options, results) {
     let fixableIssues = results.filter((r) => r.isFixable);
 
     // nothing to do, bail out
@@ -255,7 +255,8 @@ class Linter {
         fileConfig,
       });
 
-      let { code } = transform(currentSource, (env) => rule.getVisitor(env));
+      let visitor = await rule.getVisitor();
+      let { code } = transform(currentSource, () => visitor);
       currentSource = code;
     }
 
@@ -294,7 +295,6 @@ class Linter {
    * @returns {Object[]} results - The lint results.
    */
   // temporary eslint-disable while we're implementing the new pending feature (which requires async/await)
-  // eslint-disable-next-line require-await
   async verify(options) {
     let results = [];
     let fileConfig = this._moduleStatusCache.getConfigForFile(options);
@@ -333,7 +333,8 @@ class Linter {
 
     for (let rule of rules) {
       try {
-        transform(templateAST, () => rule.getVisitor());
+        let visitor = await rule.getVisitor();
+        transform(templateAST, () => visitor);
         results.push(...rule.results);
       } catch (error) {
         let message = buildErrorMessage(options.filePath, options.moduleId, error);
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -83,10 +83,10 @@ module.exports = class Base {
     return config;
   }
 
-  getVisitor(env) {
+  async getVisitor() {
     let pluginContext = this;
     let visitor = {};
-    let ruleVisitor = this.visitor(env);
+    let ruleVisitor = await this.visitor();
     // We use this structure to advise/unadvise on AST events. The walkers we
     // set up read this structure every time they get an AST event and call the
     // appropriate functions. Handlers get added to the before/after list
--- a/test/acceptance/rule-test.js
+++ b/test/acceptance/rule-test.js
@@ -95,6 +95,53 @@ describe('rule public api', function () {
     });
   });
 
+  describe('async rule visitor support', function () {
+    generateRuleTests({
+      plugins: [
+        {
+          name: 'test',
+          rules: {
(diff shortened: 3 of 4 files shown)


lib/linter.js
```diff
@@ -333,7 +333,8 @@ class Linter {
 
     for (let rule of rules) {
       try {
-        transform(templateAST, () => rule.getVisitor());
+        let visitor = await rule.getVisitor();
+        transform(templateAST, () => visitor);
```
lifeart, line 337: there is no `env` initially, and I removed it in other places,
getting 1 test error - 
image

lifeart:
created related template-recast mr https://github.com/ember-template-lint/ember-template-recast/pull/458
// cc @rwjblue

lib/linter.js
lifeart: https://github.com/ember-template-lint/ember-template-recast/pull/458

rwjblue: As far as I can tell, the only reason you might want `env` arg is to access `env.syntax`, but you can get that much easier from `require(‘ember-template-lint’).syntax` (or something like that)

lifeart: @rwjblue in general (after template recast changes), I think it’s a good idea, because in env we have code reference and module name, and we could resolve some async stuff, based on it

rwjblue: But we don't need `env` to get either of those things, both are already on the plugin instance directly

lifeart: @rwjblue thank you! MR updated, now all green

lib/rules/base.js
```diff
@@ -83,10 +83,10 @@ module.exports = class Base {
     return config;
   }
 
-  getVisitor(env) {
+  async getVisitor(env) {
```
rwjblue, line 86: Should remove `env` argument completely

```suggestion
  async getVisitor() {
```

lib/rules/base.js
```diff
@@ -83,10 +83,10 @@ module.exports = class Base {
...
     let pluginContext = this;
     let visitor = {};
-    let ruleVisitor = this.visitor(env);
+    let ruleVisitor = await Promise.resolve(this.visitor(env));
```
rwjblue, line 89: No need for `Promise.resolve` here, `await {}` (e.g. await on a non-promise) works just fine

test/acceptance/rule-test.js
```diff
@@ -95,6 +95,53 @@ describe('rule public api', function () {
...
 
+  describe('async rule visitor support', function () {
+    generateRuleTests({
+      plugins: [
+        {
+          name: 'test',
+          rules: {
+            'no-promise': class extends Rule {
+              async visitor() {
+                await new Promise((resolve) => setTimeout(resolve, 1000));
```
rwjblue, line 106: Can we dial this timeout down? I'd rather the test didn't take 1s (literally any async should have failed before)...

```suggestion
                await new Promise((resolve) => setTimeout(resolve, 10));
```

test/acceptance/rule-test.js
```diff
@@ -95,6 +95,53 @@ describe('rule public api', function () {
     });
   });
 
+  describe('async rule visitor support', function () {
+    generateRuleTests({
+      plugins: [
+        {
+          name: 'test',
+          rules: {
+            'no-promise': class extends Rule {
```
rwjblue, line 104: Seems odd to call this "no promise", maybe `rule-with-async-visitor-hook`?

rwjblue (changes requested):

rwjblue (commented):
FYI - I just pushed a fix for incorrect documentation emoji to `next` (it was failing your CI), might need to rebase now.

docs/rule/modifier-name-case.md
```diff
@@ -1,6 +1,6 @@
 # modifier-name-case
 
-:dress: The `extends: 'stylistic'` property in a configuration file enables this rule.
+:nail_care: The `extends: 'stylistic'` property in a configuration file enables this rule.
```
rwjblue, line 3: These were fixed on next directly a minute or two ago, I think you should be able to rebase and drop these doc changes

lifeart: @rwjblue done

rwjblue (approved):

rwjblue:
Leaving open for a little bit so @scalvert has a chance to review.

scalvert:
This is a really great addition, @lifeart. Thanks for doing this.

Side note: I think 75% of this PR was adding async/await keywords... 😄

scalvert (approved):

PR #1706, opened 2021-01-22
Fixes todo creation messaging to include correct counts

scalvert:
When working on a bulk todo creation script, @MelSumner and I discovered an issue with the todo creation output where the values output were incorrect for todos created/removed. This was due to the fact that we were generating the output via a loop, and that loop only used the last value from the iteration to provide the counts to output. In addition, we also discovered an unnecessary amount of work processing the todo config, which we moved outside the loop.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -258,7 +258,14 @@ async function run() {
   let options = parseArgv(process.argv.slice(2));
   let positional = options._;
   let config;
-  let todoInfo;
+  let todoInfo = {
+    added: 0,
+    removed: 0,
+    todoConfig: options.updateTodo
+      ? getTodoConfig(options.workingDirectory, getTodoConfigFromCommandLineOptions(options))
+      : {},
+  };
+  let todosToAdd = false;
 
   if (options.config) {
     try {
@@ -340,18 +347,16 @@ async function run() {
     }
 
     if (options.updateTodo) {
-      let todoConfig = getTodoConfig(
-        options.workingDirectory,
-        getTodoConfigFromCommandLineOptions(options)
+      let [added, removed] = await linter.updateTodo(
+        linterOptions,
+        fileResults,
+        todoInfo.todoConfig
       );
 
-      let [added, removed] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
+      todosToAdd = true;
 
-      todoInfo = {
-        added,
-        removed,
-        todoConfig,
-      };
+      todoInfo.added += added;
+      todoInfo.removed += removed;
     }
 
     if (!filePaths.has(STDIN)) {
@@ -373,9 +378,7 @@ async function run() {
     let hasErrors = results.errorCount > 0;
     let hasWarnings = results.warningCount > 0;
     let hasTodos = options.includeTodo && results.todoCount;
-    let hasUpdatedTodos =
-      options.updateTodo &&
-      (Number.isInteger(todoInfo.added) || Number.isInteger(todoInfo.removed));
+    let hasUpdatedTodos = options.updateTodo && todosToAdd;
 
     if (hasErrors || hasWarnings || hasTodos || hasUpdatedTodos) {
       let Printer = require('../lib/printers/default');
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
👍

bin/ember-template-lint.js
```diff
@@ -258,7 +258,14 @@ async function run() {
...
   let positional = options._;
   let config;
-  let todoInfo;
+  let todoInfo = {
+    added: 0,
+    removed: 0,
+    todoConfig: options.updateTodo
+      ? getTodoConfig(options.workingDirectory, getTodoConfigFromCommandLineOptions(options))
+      : {},
+  };
```
rwjblue, line 267: It's _probably_ always fine to include the todoConfig

bin/ember-template-lint.js
```diff
@@ -340,18 +347,16 @@ async function run() {
...
-        options.workingDirectory,
-        getTodoConfigFromCommandLineOptions(options)
+      let [added, removed] = await linter.updateTodo(
+        linterOptions,
+        fileResults,
+        todoInfo.todoConfig
       );
 
-      let [added, removed] = await linter.updateTodo(linterOptions, fileResults, todoConfig);
+      todosToAdd = true;
```
rwjblue, line 356: This doesn't seem needed, can't it just look at todoInfo.added down below?

rwjblue: Also, why is it `true`? Seems like it's possible to call `linter.updateTodo` and have that return `[0,0]` (nothing added or removed).

Unless I'm missing something, it seems that using the numbers directly is better.

bin/ember-template-lint.js
```diff
@@ -373,9 +378,7 @@ async function run() {
     let hasErrors = results.errorCount > 0;
     let hasWarnings = results.warningCount > 0;
     let hasTodos = options.includeTodo && results.todoCount;
-    let hasUpdatedTodos =
-      options.updateTodo &&
-      (Number.isInteger(todoInfo.added) || Number.isInteger(todoInfo.removed));
+    let hasUpdatedTodos = options.updateTodo && todosToAdd;
```
rwjblue, line 381:
```js
    let hasUpdatedTodos = options.updateTodo && (todoInfo.added > 0 || todoInfo.removed > 0);
```

Like that ^

bin/ember-template-lint.js
scalvert: We don't want to look at added because 0 is a valid value as we allow you to output `0 todos created...`. Using the flag is a cleaner way to determine if we should have todo output at all.

bin/ember-template-lint.js
scalvert: Is this an actionable comment?

bin/ember-template-lint.js
scalvert: Again, read above. We want to allow output whether you have created/removed or not, hence the boolean.

bin/ember-template-lint.js
scalvert: I assume you mean

```suggestion
    todoConfig: getTodoConfig(options.workingDirectory, getTodoConfigFromCommandLineOptions(options)),
  };
```

rwjblue: ya exactly

bin/ember-template-lint.js
rwjblue: If that is the case, then why do we need the boolean at all?

bin/ember-template-lint.js
rwjblue: OK, but then we should just do:

```suggestion
    let hasUpdatedTodos = options.updateTodo;
```

Because with the boolean (as far as I can tell) it is impossible for the `todosToAdd` to be a different value than `options.updateTodo`...

scalvert: Agree. Much easier.

bin/ember-template-lint.js
scalvert: Agree and done.

bin/ember-template-lint.js
scalvert: We don't...

:)

rwjblue (approved):
Looks good to me. Good job on the test too, it's really great to have high level end user expectation tests for this stuff.

PR #1707, opened 2021-01-24
no-this-in-template-only-components: improve app structure detection

dwickern:
Improves `isTemplateOnlyComponent` to support linting some more file structures:
- adds support for pods layout, using a simple heuristic so we don't need to read the app's `podModulePrefix`
- adds support for linting an addon's dummy app, by recursively detecting the structure inside `tests/dummy/app/`
- adds support for `` hbs`...` `` literals which fixes https://github.com/ember-template-lint/eslint-plugin-hbs/issues/41

### DIFF
--- a/test/unit/rules/no-this-in-template-only-components-test.js
+++ b/test/unit/rules/no-this-in-template-only-components-test.js
@@ -8,6 +8,10 @@ generateRuleTests({
 
   config: true,
 
+  meta: {
+    filePath: 'app/components/foo.hbs',
+  },
+
   good: [
     '{{welcome-page}}',
     '<WelcomePage />',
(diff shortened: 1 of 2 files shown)


test/unit/rules/no-this-in-template-only-components-test.js
```diff
@@ -61,6 +71,8 @@ generateRuleTests({
     {
       template: '{{input id=(concat this.elementId "-username")}}',
       result: {
+        filePath: 'app/components/foo.hbs',
```
lifeart, line 74: filePath may be absolute fs path

rwjblue: Ya, but I think we can't really fix this until workingDir lands.

lifeart:
Hi @dwickern, do you have time to rebase it?

PR #1710, opened 2021-01-25
Adds `octane` config back as an alias of `recommended`

scalvert:
Adds back the `octane` config, but as a passthrough to the `recommended` config. This is a change that's aimed to provide backwards compatibility to plugin authors that may have extended the original `octane` config.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -4,6 +4,7 @@ module.exports = {
   '0-8-recommended': require('./0-8-recommended'),
   '1-x-recommended': require('./1-x-recommended'),
   a11y: require('./a11y'),
+  octane: require('./octane'),
   recommended: require('./recommended'),
   stylistic: require('./stylistic'),
 };
--- a/lib/config/octane.js
+++ b/lib/config/octane.js
@@ -1,0 +1,5 @@
+'use strict';
+
+module.exports = {
+  extends: 'recommended',
+};


bmish:
Since the `octane` config is being removed as part of the v3 major release, we don't actually need to maintain backwards compatibility, and I think it's best that consumers cleanup their config to remove references to `octane`.

eslint-plugin-ember also removed the `octane` config as part of its last major release without maintaining backwards compatibility and there haven't been any complaints.

scalvert:
This is a change that @rwjblue and I just discussed, so I'll let him weight in on it.

rwjblue (approved):

rwjblue:
@bmish the goal here is to ensure that plugin authors are able to support both v2 and v3 of ember-template-lint for at least one version. We will deprecate specifying `octane` as a preset within the 3.x series, so folks will do that cleanup.

> eslint-plugin-ember also removed the octane config as part of its last major release without maintaining backwards compatibility and there haven't been any complaints.

This is _slightly_ wrong as a corollary. It would be akin to `eslint` making a change making it more difficult to support both eslint@6 and eslint@7 at the same time. For example, eslint-plugin-ember maintained support for multiple versions of eslint during the eslint@5 -> eslint@6 and eslint@6 -> eslint@7 timeframes.

bmish:
That's fair! Sounds good.

PR #1714, opened 2021-01-26
Update recommended config to include 'no-down-event-binding'

MelSumner:
Adds 'no-down-event-binding' rule to recommended config.


MelSumner:
@rwjblue @scalvert if we could put this in the next beta it would be useful. tysm!

scalvert (approved):

rwjblue (approved):

PR #1716, opened 2021-01-27
Add `no-dynamic-subexpression-invocations`.

### LINKED ISSUE #1705: New rule: require `fn` for named arguments with params/hash
We should write a rule, that will detect using a local (block param), named argument, or `this.` path expression with either positional or named arguments as a named argument in a component invocation. 

The documentation for that rule should make it clear that the rule is invalid when using Ember 3.25 or higher (because all of these patterns are now allowed and fully supported), but can help catch issues for older Ember versions (including 3.20 and 3.24, the currently supported LTS versions).

When using Ember < 3.25, if you have the following:

```hbs
<SomeComponent @whatever={{@anything someAgs}}>
</SomeCompoent>
```

This will **always** be an error (and a very obtuse one at that).

rwjblue:
When using Ember versions prior to 3.25 the usage of dynamic invocations for helpers and modifiers did not work. Unfortunately, some versions of Ember silently ignored additional arguments (3.16) wherease others throw a very bizarre error (3.20) when you attempt to invoke what the rendering engine knows as a dynamic value as if it were a helper.

For example, in versions of Ember prior to 3.25 the helper invocation here is impossible:

```hbs
{{#if (this.someHelper)}}
  Hi!
{{/if}}
```

This rule helps applications using Ember versions prior to 3.25 avoid these types of invocation.

Fixes #1705

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ Each rule has emojis denoting:
 | :white_check_mark:         | [no-duplicate-attributes](./docs/rule/no-duplicate-attributes.md)                           |
 |                            | [no-duplicate-id](./docs/rule/no-duplicate-id.md)                                           |
 |                            | [no-duplicate-landmark-elements](./docs/rule/no-duplicate-landmark-elements.md)             |
+|                            | [no-dynamic-subexpression-invocations](./docs/rule/no-dynamic-subexpression-invocations.md) |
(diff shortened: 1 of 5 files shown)


pzuraq (approved):

scalvert (approved):

scalvert:
LGTM

PR #1717, opened 2021-01-28
Allow `meta` in `<head>` contexts

### LINKED ISSUE #1713: update 'no-forbidden-elements' for 3.0
Since we are linting .html files as of v3.0, we should consider updating the `no-forbidden-elements` rule to allow the `<meta>` tag in a `head.hbs` file, since this is how you add meta info via `ember-cli-head`. 

so in `app/templates/head.hbs` (I'm not sure if we want to restrict it that closely), we should allow something like

```hbs
{{#if this.model.url}}
  <link rel="canonical" href={{if this.model.canonicalUrl this.model.canonicalUrl this.model.url}}>
  <meta property="og:url" content={{this.model.url}}>
{{/if}}
```

Alternatively, we could also remove the `<meta>` tag from the `no-forbidden-elements` list. 

Thoughts?

MelSumner:
Resolves #1713 

[x] Part 1: allows meta tags if nested inside of a `<head>` element (a necessary addition once .html files are also linted). 
[x] Part 2: allows meta tags if in a `head.hbs` file to accommodate `ember-cli-head`

### DIFF
--- a/docs/rule/no-forbidden-elements.md
+++ b/docs/rule/no-forbidden-elements.md
@@ -34,6 +34,12 @@ This rule **allows** the following:
 ```hbs
 <div></div>
 ```
+```hbs
+<head>
+  <meta charset="utf-8">
+</head>
+```
+
 
 ## Configuration
 
--- a/docs/rule/no-forbidden-elements.md
+++ b/docs/rule/no-forbidden-elements.md
@@ -34,17 +34,17 @@ This rule **allows** the following:
 ```hbs
 <div></div>
 ```
+
 ```hbs
 <head>
   <meta charset="utf-8">
 </head>
 ```
 
-
 ## Configuration
 
 * boolean - `true` to enable / `false` to disable
-  * array -- an array of strings to forbid, default: ['meta', 'style', 'html', 'style']
+  * array -- an array of strings to forbid, default: ['meta', 'style', 'html', 'script']
   * object -- An object with the following keys:
     * `forbidden` -- An array of forbidden element names
 
--- a/lib/rules/no-forbidden-elements.js
+++ b/lib/rules/no-forbidden-elements.js
@@ -92,10 +92,7 @@ module.exports = class NoForbiddenElements extends Rule {
   visitor() {
     return {
       ElementNode(node, path) {
-        if (hasHeadParent(path)) {
-          return;
-        }
-        if (this.isInHeadHbsFile(this, node)) {
+        if (hasHeadParent(path) || this.isInHeadHbsFile(this, node)) {
           return;
         }
 
--- a/lib/rules/no-forbidden-elements.js
+++ b/lib/rules/no-forbidden-elements.js
@@ -39,7 +39,7 @@ function hasHeadParent(path) {
     type: 'ElementNode',
     tag: 'head',
   };
-  let hasHeadElementInParentPath = parents.find((parent) =>
+  let hasHeadElementInParentPath = parents.some((parent) =>
     NodeMatcher.match(parent.node, refParentNode)
   );
   return Boolean(hasHeadElementInParentPath);
@@ -89,7 +89,7 @@ module.exports = class NoForbiddenElements extends Rule {
   visitor() {
     return {
       ElementNode(node, path) {
-        if (hasHeadParent(path) || this.isInHeadHbsFile(this, node)) {
+        if ((node.tag === 'meta' && hasHeadParent(path)) || this.isInHeadHbsFile(this, node)) {
           return;
         }
 
--- a/lib/rules/no-forbidden-elements.js
+++ b/lib/rules/no-forbidden-elements.js
@@ -82,14 +82,14 @@ module.exports = class NoForbiddenElements extends Rule {
     throw new Error(errorMessage);
   }
 
-  isInHeadHbsFile(scope, node) {
-    return scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta';
+  isInHeadHbsFile(scope) {
+    return scope._filePath === 'app/templates/head.hbs';
   }
 
   visitor() {
     return {
       ElementNode(node, path) {
-        if ((node.tag === 'meta' && hasHeadParent(path)) || this.isInHeadHbsFile(this, node)) {
+        if (node.tag === 'meta' && (hasHeadParent(path) || this.isInHeadHbsFile(this))) {
           return;
         }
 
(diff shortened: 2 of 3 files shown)


docs/rule/no-forbidden-elements.md
```diff
@@ -34,6 +34,12 @@ This rule **allows** the following:
 ```hbs
 <div></div>
 ```
+```hbs
```
rwjblue, line 37: Would you mind adding a new line between the blocks?

lib/rules/no-forbidden-elements.js
```diff
@@ -32,6 +32,16 @@ function sanitizeConfigArray(forbidden = []) {
   return forbidden.filter(Boolean).sort((a, b) => b.length - a.length);
 }
 
+function hasHeadParent(path) {
+  let parents = [...path.parents()];
+  let headElementInParentPath = parents.find(
+    (parent) => parent.node.type === 'ElementNode' && parent.node.tag === 'head'
```
rwjblue, line 38: Might be a good candidate for the matches API

lib/rules/no-forbidden-elements.js
```diff
@@ -32,6 +32,16 @@ function sanitizeConfigArray(forbidden = []) {
...
+  );
+  if (headElementInParentPath) {
+    return true;
+  }
+  return false;
```
rwjblue, line 43: This could be simplified a bit:

```suggestion
  return Boolean(headElementInParentPath);
```

lib/rules/no-forbidden-elements.js
```diff
@@ -69,10 +79,26 @@ module.exports = class NoForbiddenElements extends Rule {
...
   visitor() {
     return {
-      ElementNode(node) {
+      ElementNode(node, path) {
+        if (hasHeadParent(path)) {
+          return;
+        }
+        if (this.isInHeadHbsFile(this, node)) {
+          return;
+        }
```
rwjblue, line 98: Hmm. These exceptions only apply for `meta`, not for all forbidden elements. 

Can you add some test cases for:

* other forbidden elements within `<head'>` that still flags an issue?
* other forbidden elements within a filePath for head.hbs

rwjblue (changes requested):

lib/rules/no-forbidden-elements.js
```diff
@@ -32,6 +33,17 @@ function sanitizeConfigArray(forbidden = []) {
   return forbidden.filter(Boolean).sort((a, b) => b.length - a.length);
 }
 
+function hasHeadParent(path) {
+  let parents = [...path.parents()];
+  let refParentNode = {
+    type: 'ElementNode',
+    tag: 'head',
+  };
+  let hasHeadElementInParentPath = parents.find((parent) =>
```
scalvert, line 42: Would using `some` here make more sense?

lib/rules/no-forbidden-elements.js
```diff
@@ -69,10 +81,26 @@ module.exports = class NoForbiddenElements extends Rule {
 
     throw new Error(errorMessage);
   }
+
+  isInHeadHbsFile(scope, node) {
+    if (scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta') {
+      return true;
+    }
+    return false;
```
scalvert, line 89: This expression can be simplified.

```suggestion
    return scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta';
```

lib/rules/no-forbidden-elements.js
scalvert: If these two conditions stay, they can also be combined.

```suggestion
      ElementNode(node, path) {
        if (hasHeadParent(path) || this.isInHeadHbsFile(this, node)) {
          return;
        }
```

scalvert (changes requested):

lib/rules/no-forbidden-elements.js
```diff
@@ -69,10 +81,20 @@ module.exports = class NoForbiddenElements extends Rule {
...
+    return scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta';
+  }
+
   visitor() {
     return {
-      ElementNode(node) {
+      ElementNode(node, path) {
+        if (hasHeadParent(path) || this.isInHeadHbsFile(this, node)) {
+          return;
+        }
```
rwjblue, line 94: I think this is still an issue. This short circuit returns for `head.hbs` (or if the parent includes `head` element), but there are some forbidden element types that we should catch that wouldn't be caught by this logic.

Can you add the following "bad" test cases:


* With the `app/templates/head.hbs` filePath:

```hbs
<html></html>
```

* In any filePath other than `head.hbs`:

```hbs
<head>
  <html></html>
</head>
```

rwjblue (changes requested):

lib/rules/no-forbidden-elements.js
MelSumner: ok, I'll work on it some more :)

lib/rules/no-forbidden-elements.js
MelSumner: good call, changed it.

lib/rules/no-forbidden-elements.js
```diff
@@ -69,10 +81,20 @@ module.exports = class NoForbiddenElements extends Rule {
...
+
+  isInHeadHbsFile(scope, node) {
+    return scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta';
+  }
+
   visitor() {
     return {
-      ElementNode(node) {
+      ElementNode(node, path) {
+        if ((node.tag === 'meta' && hasHeadParent(path)) || this.isInHeadHbsFile(this, node)) {
```
rwjblue, line 92:
```suggestion
        if (node.tag === 'meta' && (hasHeadParent(path) || this.isInHeadHbsFile(this))) {
```

lib/rules/no-forbidden-elements.js
```diff
@@ -69,10 +81,20 @@ module.exports = class NoForbiddenElements extends Rule {
 
     throw new Error(errorMessage);
   }
+
+  isInHeadHbsFile(scope, node) {
+    return scope._filePath === 'app/templates/head.hbs' && node.tag === 'meta';
+  }
```
rwjblue, line 87:
```suggestion
  isInHeadHbsFile(scope) {
    return scope._filePath === 'app/templates/head.hbs';
  }
```

rwjblue (approved):

PR #1719, opened 2021-01-28
Fixes todo creation/removal when using scoped configurations.

scalvert:
Using todos with the happy path workflow (`ember-template-lint . --update-todo`) with a non-overridden config works in a very straightforward manner. When a config override is provided using one of

- `--no-config-path`
- `--no-inline-config`
- `--rule`
- `--config`

this caused a situation where existing todo files would be incorrectly removed. 

This PR fixes this issue by ensuring that when a config is overridden, the todo batch modifications are done in a non-destructive way by operating only within the scope of the configured rules.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -254,13 +254,12 @@ function getTodoConfigFromCommandLineOptions(options) {
   return todoConfig;
 }
 
-function isOverridingConfig(options) {
-  return (
+function _isOverridingConfig(options) {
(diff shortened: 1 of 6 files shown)


scalvert:
@MelSumner this fixes the bug with your custom todo creation script!

rwjblue (approved):

MelSumner (approved):
Thank you for your work to figure this out! Phew!

PR #1720, opened 2021-01-29
Allow comments in HTML files

### LINKED ISSUE #1718: Fatal error when an HTML comment is inside an HTML file in 3.0.0
The following code causes the fatal error:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>title</title>
  </head>
  <body>
    <!-- FOO -->
    <p>Hello World</p>
  </body>
</html>
```

Error:

```
  1:0  error  Closing tag `body` (on line 10) without an open tag.  undefined

✖ 1 problems (1 errors, 0 warnings)
error Command failed with exit code 1.
```

Removing `<!-- FOO -->` fixes it.

This shouldn't happen.

renatoi:
Fixes #1718: Fatal error when an HTML comment is inside an HTML file in 3.0.0

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -13,6 +13,7 @@ const {
 const EditorConfigResolver = require('./get-editor-config');
 
 const MAX_AUTOFIX_PASSES = 10;
+const DOCTYPE_REGEX = /\n?<!doctype[^>]+>\n?/i;
 
 function buildErrorMessage(filePath, moduleId, error) {
   let message = {
@@ -195,7 +196,12 @@ class Linter {
     let hasBOM = originalSource.charCodeAt(0) === 0xfeff;
     let byteOrderMarkPrefix = hasBOM ? '\uFEFF' : '';
 
+    let doctypeMatch = originalSource.match(DOCTYPE_REGEX);
+    let hasDoctype = doctypeMatch !== null;
+
     let currentSource = hasBOM ? originalSource.slice(1) : originalSource;
+    currentSource = hasDoctype ? currentSource.replace(doctypeMatch, '') : currentSource;
+
     let shouldVerify = true;
     let iterations = 0;
     let messages = [];
@@ -217,9 +223,12 @@ class Linter {
       }
     } while (shouldVerify);
 
+    let output = hasDoctype ? doctypeMatch[0] + currentSource : currentSource;
+    output = hasBOM ? byteOrderMarkPrefix + output : output;
+
     return {
       isFixed: currentSource !== originalSource,
-      output: hasBOM ? byteOrderMarkPrefix + currentSource : currentSource,
+      output,
       messages,
     };
   }
@@ -314,7 +323,7 @@ class Linter {
 
     let hasBOM = options.source.charCodeAt(0) === 0xfeff;
     let source = hasBOM ? options.source.slice(1) : options.source;
-    source = source.replace(/<!doctype[^>]+>\n?/i, '');
+    source = source.replace(DOCTYPE_REGEX, '');
 
     let templateAST;
 
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -314,9 +314,7 @@ class Linter {
 
     let hasBOM = options.source.charCodeAt(0) === 0xfeff;
     let source = hasBOM ? options.source.slice(1) : options.source;
-
-    // Strips doctype from the source if any
-    source = source.replace(/<!doctype[^>]+>\n?/gi, '');
+    source = source.replace(/<!doctype[^>]+>\n?/i, '');
 
     let templateAST;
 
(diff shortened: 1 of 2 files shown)


lib/linter.js
```diff
@@ -308,6 +308,9 @@ class Linter {
     let hasBOM = options.source.charCodeAt(0) === 0xfeff;
     let source = hasBOM ? options.source.slice(1) : options.source;
 
+    // Strips doctype from the source if any
```
scalvert, line 311: I'm not sure this comment adds any value, as it's clear from the code below what it's doing. Can we remove it?

test/acceptance/public-api-test.js
```diff
@@ -1475,5 +1475,36 @@ describe('public api', function () {
       expect(result.output).toEqual('\uFEFF<button type="button">LOL, Click me!</button>');
       expect(result.isFixed).toEqual(true);
     });
+
+    it("[.html] files with comments won't cause errors", async function () {
+      // reset config to default value
```
scalvert, line 1480: I don't think this comment serves any purpose. Can we remove it?

lib/linter.js
```diff
@@ -308,6 +308,9 @@ class Linter {
...
+    source = source.replace(/<!doctype[^>]+>\n?/gi, '');
```
scalvert, line 312: I don't believe we need the 'g' flag on this regex, as you're just replacing a single occurrence. You only need the 'g' flag if you're planning on looping/replacing multiple occurrences.

```js
let pattern = /a/g;
console.log(pattern.test('a')); // true
console.log(pattern.test('a')); // false
```

scalvert (changes requested):

test/acceptance/public-api-test.js
scalvert: I still see this comment 👀

scalvert (changes requested):

test/acceptance/public-api-test.js
renatoi: Oops, my bad. Removed now.

scalvert (approved):

scalvert:
@rwjblue are there any other test cases we want to add to ensure this works?

lib/linter.js
```diff
@@ -307,6 +307,7 @@ class Linter {
 
     let hasBOM = options.source.charCodeAt(0) === 0xfeff;
     let source = hasBOM ? options.source.slice(1) : options.source;
+    source = source.replace(/<!doctype[^>]+>\n?/i, '');
```
rwjblue, line 310: We need to keep track of the original doctypes, so that we can put it back for fixed templates

test/acceptance/public-api-test.js
```diff
@@ -1475,5 +1474,35 @@ describe('public api', function () {
       expect(result.output).toEqual('\uFEFF<button type="button">LOL, Click me!</button>');
       expect(result.isFixed).toEqual(true);
     });
+
+    it("[.html] files with comments won't cause errors", async function () {
```
rwjblue, line 1478: Can you tweak this to mention DOCTYPE? The issue isn't really with comments, its due to doctype...

rwjblue (changes requested):
We need some new test cases:

* Using various casing for `<DOCTYPE`
* Using `<DOCTYPE` at the start of a file that has things that can be `--fix`ed, and confirming that the update template still includes the `<DOCTYPE`

test/acceptance/public-api-test.js
MelSumner, line 1478:
```suggestion
    it("DOCTYPE in [.html] files won't cause errors", async function () {
```

lib/linter.js
renatoi: Done. Please verify.

test/acceptance/public-api-test.js
```diff
@@ -1475,5 +1474,149 @@ describe('public api', function () {
...
+            filePath: templatePath,
+            moduleId: templatePath.slice(0, -4),
+          });
+
+          expect(results).toEqual([]);
+        }
+      }
+    });
+
+    it('[.html] files with DOCTYPE are fixable and updated output includes original DOCTYPE if source includes it', async function () {
```
MelSumner, line 1536: I don't think this example will work, because the button in an .html file won't trip the `require-button-type` rule (if that's what you were going for). The `no-invalid-meta` rule could work but it doesn't have an automatic fix available.

renatoi: It works because thats the same config set in the `describe` which had the `require-button-type` set, an earlier test was already testing this. What I am testing here is if the updated output contains the original doctype. The test pass so it shows it works.

PR #1722, opened 2021-01-30
No invalid roles fatal error fix

### LINKED ISSUE #1721: Fatal error when a role has a mustache statement in 3.0.0
The following template creates a fatal error:

```hbs
<div role`={{if this.inModal "dialog" "contentinfo" }}></div>
```

Error:

```
  -:-  error  Cannot read property 'toLowerCase' of undefined  undefined
```

When a mustache statement is the value of the attribute, it should be ignored.

renatoi:
Fixes #1721: Fatal error when a role has a mustache statement in 3.0.0


rwjblue (approved):
Thanks!

PR #1726, opened 2021-02-02
Auto-writing todoConfig when one doesn't exist

scalvert:
While it's possible to use the new todo feature without a `daysToDecay` config, it's not the recommended approach.

In order to ensure that each invocation uses at least the default configuration, this PR adds a call to `ensureTodoConfig`, which writes a default config of

```json
{
  "lintTodo": {
    "daysToDecay": {
      "warn": 30,
      "error": 60
    }
  }
}
```

It writes this default when using `--update-todo` the first time, if a prior `.lint-todo` directory doesn't exist, _and_ if a prior `lintTodo` config doesn't exist.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -9,7 +9,11 @@ const fs = require('fs');
 const path = require('path');
 const { promisify } = require('util');
 
-const { getTodoStorageDirPath, getTodoConfig } = require('@ember-template-lint/todo-utils');
+const {
+  ensureTodoConfig,
+  getTodoStorageDirPath,
+  getTodoConfig,
+} = require('@ember-template-lint/todo-utils');
 const chalk = require('chalk');
 const getStdin = require('get-stdin');
 const globby = require('globby');
@@ -331,6 +335,10 @@ async function run() {
     );
     process.exitCode = 1;
     return;
+  }
+
+  if (options.updateTodo) {
+    ensureTodoConfig(options.workingDirectory);
   }
 
   let filePaths = getFilesToLint(options.workingDirectory, positional, options.ignorePattern);
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     ]
   },
   "dependencies": {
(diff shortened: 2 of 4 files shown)


bin/ember-template-lint.js
```diff
@@ -333,6 +337,10 @@ async function run() {
     return;
   }
 
+  if (options.updateTodo) {
+    ensureTodoConfig(options.workingDirectory);
+  }
```
rwjblue, line 342: Do we want to do this speculatively (as opposed to doing it only when we need to read / write)?

MelSumner (approved):
I still think we should encourage users to do a bit more than generic defaults but I am glad that we'll at least offer something by default. 👍

renatoi (approved):

bin/ember-template-lint.js
scalvert: Say more...

scalvert: I thought it made more sense to only do it during the write, since if ppl don't use todos, why mutate their package.json? I'm not 100% sure that's what you meant...

scalvert:
@MelSumner @rwjblue do you think we should have some output that indicates we wrote a todo config to your package.json? It may be jarring to see your file changes in git when you didn't expect it.

MelSumner:
> @MelSumner @rwjblue do you think we should have some output that indicates we wrote a todo config to your package.json? It may be jarring to see your file changes in git when you didn't expect it.

I'm always a fan of telling the user what's happened so I think it's a great idea. 👍

rwjblue:
> @MelSumner @rwjblue do you think we should have some output that indicates we wrote a todo config to your package.json?

Ya, seems good to me.

scalvert:
We've decided to take a less invasive approach here. We're no longer mutating the package.json of the project. We're instead using https://github.com/ember-template-lint/ember-template-lint-todo-utils/pull/88. I'm going to close this in favor of that change.

PR #1727, opened 2021-02-02
Fix `no-dynamic-subexpression-invocations` handling of `{{1}}`

rwjblue:
Previously, we incorrectly assumed that all `MustacheStatement`/`SubExpression` nodes's `.path` property was itself a `PathExpression`. This was incorrect, there are a number of other node types that `.path` can be in those circumstances.

### DIFF
--- a/test/unit/rules/no-dynamic-subexpression-invocations-test.js
+++ b/test/unit/rules/no-dynamic-subexpression-invocations-test.js
@@ -13,6 +13,18 @@ generateRuleTests({
     '{{something here="goes"}}',
     '<button onclick={{fn something "here"}}></button>',
     '{{@thing "somearg"}}',
+    '<Foo @bar="asdf" />',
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
👍

test/unit/rules/no-dynamic-subexpression-invocations-test.js
```diff
@@ -13,6 +13,18 @@ generateRuleTests({
     '{{something here="goes"}}',
     '<button onclick={{fn something "here"}}></button>',
     '{{@thing "somearg"}}',
+    '<Foo @bar="asdf" />',
```
MelSumner, line 16: Does any of this need to go into the docs for this rule?

rwjblue: I don't think so (these aren't actually related to what the rule is trying to check), it is just to ensure we don't regress again...

Issue #1728, opened 2021-02-03
no-this-in-template-only-components - does not support absolute filePath's


lifeart:
related: https://github.com/lifeart/ember-language-server/issues/199

giving:

```
filePath: 'c:\\Users\\lifeart\\Documents\\repos\\brn\\frontend\\app\\components\\exercise-type\\template.hbs',
moduleId: 'c:\\Users\\lifeart\\Documents\\repos\\brn\\frontend\\app\\components\\exercise-type\\template.hbs',
```

paths from
```js
const paths = path.normalize(hbsFilePath).split(path.sep);
```

![image](https://user-images.githubusercontent.com/1360552/106714496-fa4e2c80-660c-11eb-9bcf-b2de540f9ad4.png)

and detecting logic relay on relative file paths, instead of absoulte

![image](https://user-images.githubusercontent.com/1360552/106714570-1356dd80-660d-11eb-90db-dffb244373f9.png)

In my understanding, 

template-lint has `workingDir` option (cwd by default), and it should be used to get local path from filePath

Turbo87:
> template-lint has `workingDir` option (cwd by default), and it should be used to get local path from filePath

that would assume that template-lint is always run from the root folder of an app/addon. wouldn't it be better to look for the `package.json` file and use its folder as the app-root?

boris-petrov:
Not sure it's exactly the same, but I had [another similar issue with another rule](https://github.com/ember-template-lint/ember-template-lint/issues/1648) - just wanted to mention that if there is any connection.

lifeart:
yes, likely

Issue #1346, opened 2020-05-20
filePath or moduleId (or both)?


jgwhite:
The README suggests using the linter like this:

```js
var results = linter.verify({ source: template, moduleId: 'template.hbs' });
```

And indeed, [this is how ember-language-server uses it](https://github.com/ember-tooling/ember-language-server/blob/db75e45744e8da8023564385a68ca0f2dc0468d9/src/template-linter.ts#L44-L47):

```js
const errors = linter.verify({
  source,
  moduleId: textDocument.uri
});
```

However, [`bin/ember-template-lint.js` seems to pass `filePath` *and* `moduleId`](https://github.com/ember-template-lint/ember-template-lint/blob/706eac0674cc04a210171c21d95ebb628acc23e3/bin/ember-template-lint.js#L29).

Digging deeper, it also seems that `getConfigForFile` [only takes `overrides` into account if `filePath` is provided](https://github.com/ember-template-lint/ember-template-lint/blob/706eac0674cc04a210171c21d95ebb628acc23e3/lib/get-config.js#L416).

I also notice [this tell-tale TODO](https://github.com/ember-template-lint/ember-template-lint/blame/639db0d160b6a1878f325ed0c77f57a411ccaaed/lib/linter.js#L292):

```js
// TODO: deprecate moduleId and moduleName
```

Which is the correct option to pass? `moduleId`? `filePath`? Both?

rwjblue:
tldr; `filePath` (when representing the _real on disk_ path) is the new hawtness, `moduleId` is the old stinker.

Certain features will not work when not using `filePath` (e.g. `overrides`), but since some rules may have been relying on the presence of `moduleId` the bin script still passes both `moduleId` and `filePath`. 

As mentioned in that TODO comment, we need to deprecate using `moduleId`. This can be done by:

* emitting a warning when the linter is only passed a `moduleId` and not a `filePath`
* emitting a warning when a rule accesses `moduleId` off its constructor args (likely via a getter with a deprecation)

Both of those should emit only one warning for the duration of the run (e.g. it shouldn't emit a console warning for every template / instantiation done).

jgwhite:
Perfect, thanks @rwjblue!

rwjblue:
Reopening to track getting things updated a bit (we at least need to update the `README.md` docs, even if we delay the actual deprecation).

PR #1729, opened 2021-02-03
Make `this.filePath` and `this.workingDir` available to plugins

lifeart:
related issue: https://github.com/ember-template-lint/ember-template-lint/issues/1728
related conversation: https://discord.com/channels/480462759797063690/666416704418611233/806549043445760070

# things to figure out: 

looks like we need somehow resolve `relative` path (from working dir), but, node.relative working with current `cwd` and it may be different to working dir`

for `no-this-in-template-only` components rule we need mocked project (to test it propertly), but current test-harness does not allow us to perfrom `beforeEach`, `beforeAll`, `afterEach`, `afterAll` callbacks to setup/destroy mock project.


bin/ember-template-lint.js
```diff
@@ -33,7 +33,7 @@ async function buildLinterOptions(workingDir, filePath, filename = '', isReading
     let resolvedFilePath = path.resolve(workingDir, filePath);
     let source = await readFile(resolvedFilePath, { encoding: 'utf8' });
 
-    return { source, filePath, moduleId };
+    return { source, filePath, moduleId, workingDir };
```
rwjblue, line 36: We shouldn't add this here, the linter already has the `workingDir`:

https://github.com/ember-template-lint/ember-template-lint/blob/4be6bf9f0421cca14e8ac9d10fc154cdeeb1c7c1/bin/ember-template-lint.js#L239-L245

rwjblue (changes requested):
Instead of changing the bin script, this should update `Linter` here 

https://github.com/ember-template-lint/ember-template-lint/blob/4be6bf9f0421cca14e8ac9d10fc154cdeeb1c7c1/lib/linter.js#L91-L99

lib/rules/base.js
```diff
@@ -65,6 +68,26 @@ module.exports = class Base {
     this.source = options.rawSource.match(reLines);
   }
 
+  get filePath() {
+    return this._filePath;
+  }
```
rwjblue, line 73: 🤔 why not change the class field to be `this.filePath = options.filePath` instead of adding a getter?

lib/rules/base.js
```diff
@@ -65,6 +68,26 @@ module.exports = class Base {
...
+
+  get resolvedFilePath() {
+    return path.resolve(this.workingDir, this.filePath);
+  }
+
+  get moduleName() {
+    return this._moduleName;
+  }
```
rwjblue, line 81: I don't think we should make `moduleName` "public"

lib/rules/base.js
```diff
@@ -65,6 +68,26 @@ module.exports = class Base {
...
+
+  get rawSource() {
+    return this._rawSource;
+  }
+
+  get workingDir() {
+    return this._workingDir;
+  }
```
rwjblue, line 89: I'd rather use `this.workingDir = options.workingDir` than make a specific getter for it, unless you have a particular reason for wanting it to be a getter...

lib/rules/base.js
```diff
@@ -65,6 +68,26 @@ module.exports = class Base {
     this.source = options.rawSource.match(reLines);
   }
 
+  get filePath() {
+    return this._filePath;
+  }
+
+  get resolvedFilePath() {
+    return path.resolve(this.workingDir, this.filePath);
+  }
```
rwjblue, line 77: Thoughts on naming this `absoluteFilePath`?

lib/helpers/rule-test-harness.js
```diff
@@ -91,15 +103,24 @@ module.exports = function generateRuleTests({
...
-          ? item.meta
-          : defaultMeta !== undefined
-          ? JSON.parse(JSON.stringify(defaultMeta))
-          : {};
-      meta.filePath = meta.filePath || 'layout.hbs';
-      meta.moduleId = meta.moduleId || 'layout';
+      let meta = metaToObject(item);
+      meta.filePath = meta.filePath || ConfigDefaults.filePath;
+      meta.moduleId = meta.moduleId || ConfigDefaults.moduleId;
+      meta.workingDir = meta.workingDir || ConfigDefaults.workingDir;
```
rwjblue, line 122: I'd rather these use an `in` check instead of falseyness

```suggestion
      meta.filePath = 'filePath' in meta ? meta.filePath : ConfigDefaults.filePath;
      meta.moduleId = 'moduleId' in meta ? meta.moduleId : ConfigDefaults.moduleId;
      meta.workingDir = 'workingDir' in meta ? meta.workingDir : ConfigDefaults.workingDir;
```

lib/helpers/rule-test-harness.js
```diff
@@ -91,15 +103,24 @@ module.exports = function generateRuleTests({
 
     let linter;
 
+    function metaToObject(item) {
+      const fallbackMeta = JSON.parse(JSON.stringify(defaultMeta || {}));
+      if (typeof item !== 'object' || item === null) {
+        return fallbackMeta;
+      }
+      if (typeof item.meta === 'object') {
+        return item.meta;
```
rwjblue, line 112: This should always return a new "clone", so we don't mutate the original shared `item.meta`

lib/helpers/rule-test-harness.js
```diff
@@ -303,3 +322,5 @@ module.exports = function generateRuleTests({
     });
   });
 };
+
+module.exports.ConfigDefaults = ConfigDefaults;
```
rwjblue, line 326: Why do we need to expose these?

test/unit/base-plugin-test.js
```diff
@@ -115,7 +152,24 @@ describe('base plugin', function () {
...
+    });
+    it('can access workingDir', function () {
+      class AwesomeRule extends Rule {
+        visitor() {
+          expect(this.workingDir).toBe('foo');
+        }
+      }
 
+      runRules('foo', [plugin(AwesomeRule, 'awesome-rule', { workingDir: 'foo' })]);
+    });
```
rwjblue, line 172:
```suggestion
    });

```

test/unit/base-plugin-test.js
```diff
@@ -115,7 +152,24 @@ describe('base plugin', function () {
...
     });
+    it('can access rawSource', function () {
+      class AwesomeRule extends Rule {
+        visitor() {
+          expect(this.rawSource).toBe('foo');
+        }
+      }
+
+      runRules('foo', [plugin(AwesomeRule, 'awesome-rule', true)]);
+    });
```
rwjblue, line 163:
```suggestion
    });

```

test/unit/base-plugin-test.js
```diff
@@ -115,7 +152,24 @@ describe('base plugin', function () {
 
       runRules('foo', [plugin(AwesomeRule, 'awesome-rule', true)]);
     });
+    it('can access rawSource', function () {
```
rwjblue, line 155:
```suggestion

    it('can access rawSource', function () {
```

test/unit/base-plugin-test.js
```diff
@@ -106,6 +117,32 @@ describe('base plugin', function () {
...
+          expect(this.moduleName).toBe('foo/bar');
+        }
+      }
+
+      runRules('foo', [
+        plugin(AwesomeRule, 'awesome-rule', {
+          moduleName: 'foo/bar',
+        }),
+      ]);
+    });
```
rwjblue, line 145:
```suggestion
    });

```

test/unit/base-plugin-test.js
```diff
@@ -106,6 +117,32 @@ describe('base plugin', function () {
...
+          expect(this.filePath).toBe('foo.hbs');
+        }
+      }
+
+      runRules('foo', [
+        plugin(AwesomeRule, 'awesome-rule', {
+          filePath: 'foo.hbs',
+        }),
+      ]);
+    });
```
rwjblue, line 132:
```suggestion
    });

```

lib/rules/base.js
lifeart: lol, I'm looking forward to support deprecation warns once accessed xD

lib/rules/base.js
lifeart: @rwjblue in my understanding, `filePath` already absolute one, nope?

lib/rules/base.js
lifeart: and we already have `_filePath` (keeping backward compability)

lib/rules/base.js
lifeart: for now, removed from public surface

lib/helpers/rule-test-harness.js
lifeart: lol, this change cost us dozens of failed tests

lifeart: image

lib/rules/base.js
rwjblue: Ya, but the deprecation would be for `_filePath` in this case since conceptually `filePath` is the main go forward / correct value to use (as opposed to `moduleId`/`moduleName`).

Let's do this:

```js
// preserve `this._filePath` compatibility for now, even though it is
// private API (future versions may deprecate)
this._filePath = this.filePath = options.filePath;
```

Then use a private symbol + getters for `_moduleId` + `_moduleName`, (you are right that this makes future deprecation fairly painless).

```js
const MODULE_ID = Symbol('_moduleId');
const MODULE_NAME = Symbol('_moduleName');

class Rule {
  constructor(options) {
    // ...snip...
    this[MODULE_ID] = options.moduleId;
    this[MODULE_NAME] = options.moduleName;
  }

  // ...snip...

  get _moduleId() {
    // TODO: deprecate accessing _moduleId
    return this[MODULE_ID];
  }

  get _moduleName() {
    // TODO: deprecate accessing _moduleName
    return this[MODULE_NAME];
  }
}
```

lib/rules/base.js
rwjblue: If `filePath` is already absolute, what is the point of the `path.resolve` here? Do you mean to find the relative file path instead (e.g. from `/Users/rwjblue/src/app-name/app/templates/foo.hbs` to `app/templates/foo.hbs`)?

lib/helpers/rule-test-harness.js
rwjblue: FWIW, this `in` check makes it possible for us to deprecate specifying / accessing `moduleId`

lib/helpers/rule-test-harness.js
rwjblue: nvm I see where it is used, sorry for confusion

lib/rules/base.js
```diff
@@ -57,7 +57,8 @@ module.exports = class Base {
     this.config = this.parseConfig(options.config);
     this._configStack = [this.config];
 
-    this._filePath = options.filePath;
+    this.workingDir = options.workingDir;
+    this.filePath = options.filePath;
```
rwjblue, line 61:
```suggestion
    // TODO: add a deprecation for accessing _filePath
    this.filePath = this._filePath = options.filePath;
```

test/unit/base-plugin-test.js
```diff
@@ -28,19 +29,29 @@ describe('base plugin', function () {
...
         ruleConfig
       );
 
+      let configKeys = Object.keys(config || {});
+
+      for (let key of configKeys) {
+        if (key in options) {
+          options[key] = config[key];
+        }
+      }
```
rwjblue, line 53: Why do this instead of adding to the `Object.assign` just above?

lib/rules/base.js
rwjblue: NOTE: I think the private symbol thing could be done in a separate PR (it does seem unrelated to this workingDir work)

lib/rules/base.js
lifeart: done

lib/rules/base.js
lifeart: @rwjblue yes, it's my goal, but I don't know how to get it working (see `things to figure out` in mr description, because of cwd)

yeah, I suppose filePath  should be absoulte.

https://github.com/ember-template-lint/ember-template-lint/issues/1346#issuecomment-631720981
image

test/unit/base-plugin-test.js
lifeart: here is a mess with `config, options` meaning,

`ruleConfig` in assign supposed to be `rawConfig`, but it's not true

image

image

`ruleConfig` here has output shape as `{ Rule, name, config }`, and it should not be assigned in `Object.assign` as is.

I'm fixing override isssues in tests, by one more `for let`, but this part of code should be completely rewritten, but, looks like it's out of scope for now.

lib/rules/base.js
rwjblue: We can confirm (and add a `relativeFilePath` or something) in a separate PR, since it isn't _really_ related to this anymore (now that you removed it).

test/unit/base-plugin-test.js
rwjblue: YA, I was just thinking that the last entry in that `Object.assign` should be `ruleConfig.config` not `ruleConfig` (e.g. I think using `ruleConfig` is a bug?).

lifeart: @rwjblue correct, I was trying to "fix it", but list of tests failing, don't want to mess fixes in current scope

lib/rules/base.js
lifeart: agree, `resolvedFilePath` removed from changes

test/unit/base-plugin-test.js
rwjblue: @lifeart - Mind making another issue so we don't forget to come back and fix?

lib/helpers/rule-test-harness.js
rwjblue: This still needs to be addressed.

test/unit/base-plugin-test.js
rwjblue: I still don't quite understand why this code is needed for this PR. 

* What tests fail without this extra assignment? 
* Why do we only allow `configKeys` that are already in `options`?

lifeart: @rwjblue without this changes `workingDir`, `_moduleName`, `filePath` is not configurable inside tests (previously we don't have tests for it)

commenting lines fail 3 tests (fom this MR)

image

image


image

lifeart: if I just assign `config` from `ruleConfig`, instead of `ruleConfig` object itself, we will get this count of error on tests:

image
and `failed` scope quite scary, like

`  ● base plugin › scopes instructions › it doesn't call a disabled rule's visitor handlers`
`  ● base plugin › scopes instructions › enabling a disabled-by-default rule actually enables it`
`  ● base plugin › scopes instructions › handles descendant instruction of in-element tree instruction`

and such
image

rwjblue: OK, finally dug into this. We were both wrong 😸 

---

`config` in this context is the actual rules configuration (e.g. `true`/`false` or any other options it takes in the config file) and is **not** related to construction arguments at all. See my changes in 73d4d35 (and the commit message explanation).

rwjblue:
OK, I've pushed a few changes here that fix things up. Please take a look at them when you have a chance @lifeart, but I think this is ready to land (once CI passes).

rwjblue (approved):

PR #1731, opened 2021-02-03
Add `require-has-block-helper` lint rule

patocallaghan:
Follow up to introducing the `hasBlock` and `hasBlockParams` deprecations in https://github.com/emberjs/ember.js/pull/19374/

Adds a new lint rule, `require-has-block-helper` (open to suggestions on the name) to warn about usage of `hasBlock` and `hasBlockParams`. It was also suggested we create a `fixer` for this lint rule but I haven't got around to that yet. I can either introduce it in a follow up PR or if you don't mind waiting another day or two, I can introduce into this PR.

/cc @pzuraq @rwjblue

### DIFF
--- a/docs/rule/require-has-block-helper.md
+++ b/docs/rule/require-has-block-helper.md
@@ -1,6 +1,6 @@
 # require-has-block-helper
 
-In Ember 3.25 the properties `hasBlock` and `hasBlockParams` were deprecated. Their replacement is to use `has-block` and `has-block-params` helpers instead.
+In Ember 3.26 the properties `hasBlock` and `hasBlockParams` were deprecated. Their replacement is to use `has-block` and `has-block-params` helpers instead.
 
 This rule prevents the usage of `hasBlock` and `hasBlockParams` and suggests using `has-block` or `has-block-params` instead.
 
--- a/docs/rule/require-has-block-helper.md
+++ b/docs/rule/require-has-block-helper.md
@@ -1,0 +1,54 @@
+# require-has-block-helper
+
+In Ember 3.25 the properties `hasBlock` and `hasBlockParams` were deprecated. Their replacement is to use `has-block` and `has-block-params` helpers instead.
+
+This rule prevents the usage of `hasBlock` and `hasBlockParams` and suggests using `has-block` or `has-block-params` instead.
+
+For more information about this deprecation you can view the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0689-deprecate-has-block.md) or its entry on the [Deprecations page](https://deprecations.emberjs.com/v3.x/#toc_has-block-and-has-block-params).
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+{{hasBlock}}
+{{#if hasBlock}}
+
+{{/if}}
+```
+
+```hbs
+{{hasBlockParams}}
+{{#if hasBlockParams}}
+
+{{/if}}
+```
+
+This rule **allows** the following:
+
+```hbs
+{{has-block}}
+{{#if (has-block)}}
+
+{{/if}}
+```
+
+```hbs
+{{has-block-params}}
+{{#if (has-block-params)}}
+
+{{/if}}
+```
+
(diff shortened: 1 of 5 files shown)


docs/rule/require-has-block-helper.md
```diff
@@ -0,0 +1,54 @@
+# require-has-block-helper
+
+In Ember 3.25 the properties `hasBlock` and `hasBlockParams` were deprecated. Their replacement is to use `has-block` and `has-block-params` helpers instead.
```
rwjblue, line 3:
```suggestion
In Ember 3.26 the properties `hasBlock` and `hasBlockParams` were deprecated. Their replacement is to use `has-block` and `has-block-params` helpers instead.
```

lib/rules/index.js
```diff
@@ -93,4 +93,5 @@ module.exports = {
   'style-concatenation': require('./style-concatenation'),
   'table-groups': require('./table-groups'),
   'template-length': require('./template-length'),
+  'require-has-block-helper': require('./require-has-block-helper'),
```
rwjblue, line 96: I think you need to alphabetize this one (the generator always adds it to the bottom, but then we fail a test if they aren't alphabetized).

rwjblue (commented):
Overall looks great! Happy to land and iterate on the fixer separately (once CI is happy here).

RE: the fixer, https://github.com/ember-template-lint/ember-template-lint/commit/32c8d58eb00919fd5af5740cd2200ad4f561fba9 is a good example

docs/rule/require-has-block-helper.md
```diff
@@ -0,0 +1,54 @@
...
+{{hasBlockParams}}
+{{#if hasBlockParams}}
+
+{{/if}}
+```
+
+This rule **allows** the following:
+
+```hbs
+{{has-block}}
```
lifeart, line 29: is it make sence to aloow plain `{{has-block}}` usage? (I suppose it should be only inside expressions)
cc @rwjblue

lifeart: same for `has-block-params`, and wondering, should we require block name  `(has-block "default")` ? (imho - yes)

patocallaghan: So according to the [deprecation guide](https://github.com/emberjs/rfcs/blob/master/text/0689-deprecate-has-block.md#deprecation-guide) in the RFC `{{has-block}}` and `{{has-block-params}}` are both valid.

> You may optionally pass a name to (has-block), the name of the block to check. The name corresponding to the block that {{hasBlock}} represents is "default". Calling (has-block) without any arguments is equivalent to calling (has-block "default").

lifeart: @patocallaghan, yes, but i'm trying to figure out real use-cases for exact usage of `{{has-block}}`, `{{has-block-params}}` (outside expressions), and for this 1% I suppose easier to disable rule.

regarding requred `default` - ember trying to be more explicit and straightforward, in `3.25` we have named blocks enabled by default, and think it's good approach to requre block name in this helpers.

patocallaghan: Sure fair point.

@rwjblue I know the RFC called out supporting not passing "default" but do you have any thoughts should we be recommending the use of `{{has-block "default'}}`? If that is the case, should the fixer I'm going to work on migrate things to use the expanded `{{has-block "default"}}` syntax rather than just `{{has-block}}`?

rwjblue: > If that is the case, should the fixer I'm going to work on migrate things to use the expanded {{has-block "default"}} syntax rather than just {{has-block}}?

The fixer should use whatever the `hasBlock`/`hasBlockParams` call was passed (so if it was passing a positional param string, the fixer should preserve that).

rwjblue: >  should we require block name (has-block "default") ? (imho - yes)

I don't think we should. 

See https://github.com/ember-template-lint/ember-template-lint/issues/1352.

PR #1736, opened 2021-02-04
Add fixer for require-has-block-helper rule

patocallaghan:
Follow up to #1731 

Adds fixer for require-has-block-helper rule

Some examples of the fixer. See tests for more.

|Before|After|
|-|-|
|`{{hasBlock}}`|`{{has-block}}`|
|`{{hasBlockParams}}`|`{{has-block-params}}`|
|`{{if hasBlockParams "true" "false"}}`|`{{if (has-block-params) "true" "false"}}`|
|`{{if (hasBlockParams "inverse") "true" "false"}}`|`{{if (has-block-params "inverse") "true" "false"}}`|
|`{{#if hasBlock}}{{/if}}`|`{{#if (has-block))}}{{/if}}`|

/cc @rwjblue @pzuraq


lib/rules/require-has-block-helper.js
```diff
@@ -14,14 +16,31 @@ function getErrorMessage(name) {
...
-            line: node.loc && node.loc.start.line,
-            column: node.loc && node.loc.start.column,
-            source: this.sourceForNode(node),
-          });
+      PathExpression(node, path) {
+        if (this.mode === 'fix') {
+          if (TRANSFORMATIONS[node.original]) {
+            let parent = path.parent;
+            if (
+              parent &&
```
rwjblue, line 24: When is `parent` not truthy? I think it always has to be... 🤔

lib/rules/require-has-block-helper.js
```diff
@@ -14,14 +16,31 @@ function getErrorMessage(name) {
...
+              (parent.node.type === 'BlockStatement' ||
+                (['MustacheStatement', 'SubExpression'].includes(parent.node.type) &&
+                  parent.node.path.original !== node.original))
+            ) {
```
rwjblue, line 28: Let's break this conditional down into a few variables, to help simplify the mental model.

```suggestion
           let isBlockStatement = parent.node.type === 'BlockStatement';
           let isImplicitInvocation = ['MustacheStatement', 'SubExpression'].includes(parent.node.type) && parent.node.path.original !== node.original);
                  
            if (isBlockStatement || isImplicitInvocation) {
```

rwjblue (commented):
Looks good, one minor suggestion / tweak and should be good to go...

lib/rules/require-has-block-helper.js
patocallaghan: Yep, you're right 👍

lib/rules/require-has-block-helper.js
patocallaghan: Thanks, reads much better 👍

rwjblue (approved):
Thank you!!

PR #1737, opened 2021-02-05
Moves todo tests to their own file

scalvert:
This PR moves the todo tests from cli-tests to their own file. It also (either unfortunately or fortunately) includes an upgrade of `@ember-template-lint/todo-utils`, which introduced a change in behavior when retrieving the `todoConfig`. The new version provides a default config when one is not found. The reason this was necessary to bundle up with these other, arguably unrelated, changes was because the floating dependency test would fail when encountering that new version.

This dependency upgrade also replaces #1726.


test/helpers/run.js
```diff
@@ -0,0 +1,20 @@
+const execa = require('execa');
+
+let project;
+
+function setProject(_project = undefined) {
```
scalvert, line 5: This is a bit odd to have to do, but it's because of line 11. Happy to change if you can think of a better way to do this without having to pass in the project for every run call.

test/helpers/run.js
```diff
@@ -0,0 +1,20 @@
...
+  project = _project;
+}
+
+function run(args, options = {}) {
+  options.reject = false;
+  options.cwd = options.cwd || project.path('.');
```
rwjblue, line 11: Lets just do `options.cwd || process.cwd()`. Both the `cli-test.js` and the new file call `process.chdir(project.path())` in their setup (and reset in their teardown).

scalvert: OK that worked. Thanks.

PR #1739, opened 2021-02-06
Fixes issue where todos were deleted from a different engine

scalvert:
Fixes an issue reported in https://github.com/scalvert/eslint-formatter-todo/issues/80.

While this wouldn't directly plague this lib due to the extra file-based scoping we do here, it ensures that we're still applying the same scope filtering, and ensures this with a test.

### DIFF
--- a/test/acceptance/todo-cli-test.js
+++ b/test/acceptance/todo-cli-test.js
@@ -5,6 +5,7 @@ const {
   todoStorageDirExists,
   readTodos,
   getTodoStorageDirPath,
+  writeTodos,
 } = require('@ember-template-lint/todo-utils');
 const { differenceInDays, subDays } = require('date-fns');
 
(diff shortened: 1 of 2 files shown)


rwjblue (commented):
LGTM once #1737 lands and this is rebased

rwjblue (approved):

test/acceptance/todo-cli-test.js
```diff
@@ -160,6 +161,52 @@ describe('todo usage', () => {
       expect(todoStorageDirExists(project.baseDir)).toEqual(true);
     });
 
+    it.only('does not remove todos from another engine', async function () {
```
rwjblue, line 164: `.only` 🥩 🔪

rwjblue (changes requested):

test/acceptance/todo-cli-test.js
rwjblue: hopefully this would have failed CI, if not can you file an issue?

scalvert: It should have failed linting, as that rule is for sure enabled.

rwjblue (approved):

Issue #1625, opened 2020-11-28
no-invalid-interactive triggers on <summary> tag


wuron:
Adding `{{on}}` modifier to `<summary>` tag triggers `no-invalid-interactive` rule. This seems incorrect, given that summary is a clickable element that has implicit ARIA role "button".

zelaznik:
@wuron I'm not part of the core team, but I just saw this issue on the GitHub page, and I think you're right.

Before the rule `no-invalid-interactive` existed, I wrote my own custom rule that does the more or less the same thing.  This is a hopefully exhaustive list of CSS selectors that match to interactive elements.  I copied these selectors from a course on accessibility from Frontend Masters.

https://frontendmasters.com/courses/web-accessibility/exercise-2-solution/
https://github.com/jkup/focusable/blob/master/index.js

```js
const VALID_INTERACTIVE_SELECTORS = [
  'a[href]',
  'area[href]',
  'input:not([disabled]):not([type="hidden"])',
  'select:not([disabled])',
  'textarea:not([disabled])',
  'button:not([disabled])',
  'iframe',
  'object',
  'embed',
  '[tabindex="0"]',
  '[contenteditable]',
  'audio[controls]',
  'video[controls]',
  'summary',
  '[tabindex^="0"]',
  '[tabindex^="1"]',
  '[tabindex^="2"]',
  '[tabindex^="3"]',
  '[tabindex^="4"]',
  '[tabindex^="5"]',
  '[tabindex^="6"]',
  '[tabindex^="7"]',
  '[tabindex^="8"]',
  '[tabindex^="9"]',
];
```

I'm looking through the helper function that determines if an element is interactive.

https://github.com/ember-template-lint/ember-template-lint/blob/v3.0.0-beta.3/lib/helpers/is-interactive-element.js

We're missing `<summary>`, `<area>`.  I think we're also missing tags such as `<p contenteditable></p>`.  Such a tag is an affordance theory disaster, but it is technically valid.

I can submit a PR later today that updates the helper function `is-interactive-element`.

zelaznik:
I tabled the issues that I raised that were unrelated to `<details>` and `<summary>`, but here's the PR that fixes this bug:

https://github.com/ember-template-lint/ember-template-lint/pull/1743

MelSumner:
Here's the link to the spec for details, outlining that it is an interactive element: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element

[...]

PR #1743, opened 2021-02-07
"is-interactive-element" should match to <summary>, not <details>

zelaznik:
# Bug:

- When using the following pattern, the `is-interactive-element` marks the wrong element as interactive.

    ```hbs
    <details> {{!-- <details> is NOT reachable by keyboard --}}
       <summary> {{!-- <summary> is the focusable element --}}
        Foo
      </summary>
      Bar
    </details>
    ```

- I modified the `is-interactive-element` helper so that it skips the `<details>` tag and marks the `<summary>` tag as interactive.  I also had to modify the unit tests surrounding nested interactive elements.

### DIFF
--- a/lib/helpers/is-interactive-element.js
+++ b/lib/helpers/is-interactive-element.js
@@ -5,7 +5,7 @@ const ast = require('./ast-node-info');
 const INTERACTIVE_TAG_NAMES = new Set([
   'button',
   'canvas',
-  'details',
+  'summary',
   'embed',
   'iframe',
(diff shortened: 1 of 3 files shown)


zelaznik:
Fixes the following issue:  https://github.com/ember-template-lint/ember-template-lint/issues/1625

zelaznik:
Looks like `<details>` is in fact an interactive element, as Melanie Sumner pointed out in her review.

Issue #1732, opened 2021-02-04
rule request: no-capital-arguments


lifeart:
Ref: https://discord.com/channels/480462759797063690/483601670685720591/806661156936482876

https://github.com/emberjs/rfcs/blob/68812bf2d439c6bb77ad491e0159b371b68c5c35/text/0276-named-args.md#reserved-names

![5262824E-7B8F-4D8E-8229-D4B96BC7E362](https://user-images.githubusercontent.com/1360552/106854582-a0ad3700-66cc-11eb-967b-362a88786c13.jpeg)

tldr: @Foo, not working, and emitting reserved word error

`{{#let @NavItem as |NavItem|}}` results in the template compiler saying `Assertion Failed: '@NavItem' is reserved.` but `{{#let this.args.NavItem as |NavItem|}}` works

cc @locks

rwjblue:
Ya this is a good idea, though I **do** think we should relax that rule on the Ember side.

rwjblue:
We discussed this in a recent Ember core team meeting, and we are going to relax this restriction.

lifeart:
@rwjblue I'm wondering, could we lend rule asap (because it's not working in current embers), and later we could disable it (once changes landed), also, it's good case for https://github.com/ember-template-lint/ember-template-lint/issues/1741

PR #1759, opened 2021-02-16
Add new rule: `no-capital-arguments`

lifeart:
resolves: https://github.com/ember-template-lint/ember-template-lint/issues/1732

### DIFF
--- a/lib/rules/no-capital-arguments.js
+++ b/lib/rules/no-capital-arguments.js
@@ -1,0 +1,53 @@
+'use strict';
+
+const Rule = require('./base');
+
+const ERROR_MESSAGE_CAPITAL = 'Capital argument names is not supported';
+function ERROR_MESSAGE_RESERVED(name) {
+  return `${name} is reserved argument name, try to use another`;
+}
+
+const AllowedPrefix = /[a-z]/;
+const RESERVED = new Set(['@arguments', '@args', '@block', '@else']);
+
+module.exports = class NoCapitalArguments extends Rule {
+  isReserved(name) {
+    return RESERVED.has(name);
+  }
+  visitor() {
+    return {
+      PathExpression(node) {
+        if (node.data) {
+          let part = node.parts[0] || '';
+          let firstChar = part.charAt(0);
+          let isReserved = this.isReserved(`@${part}`);
+          if (!AllowedPrefix.test(firstChar) || isReserved) {
+            this.log({
+              message: isReserved ? ERROR_MESSAGE_RESERVED(`@${part}`) : ERROR_MESSAGE_CAPITAL,
+              line: node.loc && node.loc.start.line,
+              column: node.loc && node.loc.start.column + 1,
+              source: part,
+            });
+          }
+        }
+      },
+      AttrNode(node) {
+        if (node.name.startsWith('@')) {
+          let firstChar = node.name.charAt(1);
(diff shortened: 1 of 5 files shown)


test/unit/rules/no-capital-arguments-test.js
```diff
@@ -0,0 +1,37 @@
...
+  name: 'no-capital-arguments',
+
+  config: true,
+
+  good: ['<Foo @name="bar" />'],
+
+  bad: [
+    {
+      template: '<Foo @Name="bar" />',
+      fixedTemplate: '<Foo @name="bar" />',
```
lifeart, line 17: fixers should be safe, because capital arguments is not working :)

rwjblue: No, the fixer is not guaranteed to be safe. The fixer as implemented assumes that the only possible way to use a passed in named argument is via named arguments in the template of the component you are calling (which is not true). For example, `Foo` component in this case could easily be using `this.args.Name` in its JavaScript portion.

lib/rules/no-capital-arguments.js
```diff
@@ -0,0 +1,33 @@
...
+const ERROR_MESSAGE = 'Capital argument names is not supported';
+const AllowedPrefix = /[a-z]/;
+
+module.exports = class NoCapitalArguments extends Rule {
+  visitor() {
+    return {
+      AttrNode(node) {
+        if (node.name.startsWith('@')) {
+          let firstChar = node.name.charAt(1);
+          if (!AllowedPrefix.test(firstChar)) {
```
rwjblue, line 14: This isn't quite good enough. The current list of disallowed names can be found here:

https://github.com/emberjs/ember.js/blob/3a7c034cdc81c3ed05508f64c630e2041995d2f8/packages/ember-template-compiler/lib/plugins/assert-reserved-named-arguments.ts#L39-L43

```js
const RESERVED = ['@arguments', '@args', '@block', '@else'];

function isReserved(name: string): boolean {
  return RESERVED.indexOf(name) !== -1 || Boolean(name.match(/^@[^a-z]/));
}
```

rwjblue (changes requested):

test/unit/rules/no-capital-arguments-test.js
lifeart: @rwjblue I have added linting for `paths` as well, could we use fixer for paths?

lib/rules/no-capital-arguments.js
```diff
@@ -0,0 +1,60 @@
...
+      PathExpression(node) {
+        if (node.data) {
+          let part = node.parts[0] || '';
+          let firstChar = part.charAt(0);
+          let isReserved = this.isReserved(`@${part}`);
+          if (!AllowedPrefix.test(firstChar) || isReserved) {
+            if (this.mode === 'fix' && !isReserved) {
+              node.original = `@${firstChar.toLowerCase()}${part.slice(1)}${node.parts
+                .slice(1)
+                .join('.')}`;
```
rwjblue, line 28: I'd rather not have a fixer at all. Changing this to a lowercase value doesn't really fix anything (unless your callsites are all passing the argument with a lowercase as well).

lifeart: fixers removed

rwjblue (approved):

PR #1760, opened 2021-02-17
chore: Deprecating moduleId and moduleName in Rule base class

scalvert:
Implements #1346

### DIFF
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -6,7 +6,6 @@ const Scope = require('./internal/scope');
 // Constant to return from AST handlers so we know to remove/unregister them
 const REMOVE_HANDLER = {};
 
-const MODULE_ID = Symbol('_moduleId');
 const MODULE_NAME = Symbol('_moduleName');
 
 const loggedModules = new Set([]);
@@ -63,7 +62,6 @@ module.exports = class Base {
     this._configStack = [this.config];
 
     this._filePath = options.filePath;
-    this[MODULE_ID] = options.moduleId;
     this[MODULE_NAME] = options.moduleName;
     this._rawSource = options.rawSource;
 
@@ -75,27 +73,13 @@ module.exports = class Base {
     return this._configResolver.editorConfig();
   }
 
-  get _moduleId() {
-    const moduleIdKey = `${this.constructor.name}:moduleId`;
-
-    if (!loggedModules.has(moduleIdKey)) {
-      this._console.log('The `_moduleId` property is deprecated. Please use `_filePath` instead.');
-
-      this.loggedModules.push(moduleIdKey);
-    }
-
-    return this[MODULE_ID];
-  }
-
   get _moduleName() {
-    const moduleNameKey = `${this.constructor.name}:moduleName`;
-
-    if (!loggedModules.has(moduleNameKey)) {
+    if (!loggedModules.has(this.ruleName)) {
       this._console.log(
         'The `_moduleName` property is deprecated. Please use `_filePath` instead.'
       );
 
-      this.loggedModules.push(moduleNameKey);
+      this.loggedModules.push(this.ruleName);
     }
 
     return this[MODULE_NAME];
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -6,6 +6,11 @@ const Scope = require('./internal/scope');
 // Constant to return from AST handlers so we know to remove/unregister them
 const REMOVE_HANDLER = {};
 
+const MODULE_ID = Symbol('_moduleId');
+const MODULE_NAME = Symbol('_moduleName');
+
+const loggedModules = new Set([]);
+
 // If we need to log an error or warning about an instruction node (e.g. syntax
 // error), we want to only log it once, rather than having every rule log it,
 // spamming the user. So the rules push any nodes with syntax errors into this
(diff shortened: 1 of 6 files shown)


lib/rules/base.js
```diff
@@ -58,7 +63,8 @@ module.exports = class Base {
     this._configStack = [this.config];
 
     this._filePath = options.filePath;
-    this._moduleName = options.moduleName;
+    this[MODULE_ID] = options.moduleId;
```
rwjblue, line 66: I think we can kill this one, since we (apparently) didn't expose it before.

lib/rules/base.js
```diff
@@ -69,6 +75,32 @@ module.exports = class Base {
...
+    const moduleIdKey = `${this.constructor.name}:moduleId`;
+
+    if (!loggedModules.has(moduleIdKey)) {
+      this._console.log('The `_moduleId` property is deprecated. Please use `_filePath` instead.');
+
+      this.loggedModules.push(moduleIdKey);
+    }
+
+    return this[MODULE_ID];
+  }
```
rwjblue, line 88: Same as above, we don't have a preexisting `_moduleId` prop that is available on rules so I guess we don't need these getters.

lib/rules/base.js
```diff
@@ -69,6 +75,32 @@ module.exports = class Base {
...
+
+  get _moduleName() {
+    const moduleNameKey = `${this.constructor.name}:moduleName`;
+
+    if (!loggedModules.has(moduleNameKey)) {
```
rwjblue, line 93: We want to change this around to make it only log once _per rule_ (not once per module being linted, which is what is implemented here I think).

Off the top of my head, that would be something like:

```js

if (!loggedRules.has(this.ruleName)) {
  this._console.log(
    `The \`_moduleName\` property is deprecated (accessed from ${this.ruleName}). Please use \`_filePath\` instead.`
  );

  loggedRules.add(this.ruleName);
}
```

rwjblue (changes requested):
Left some inline comments to tweak / resolve, and I also think we need to account for `this._console` properly nooping when using `--quiet` and/or `--json` bin arguments.

lib/rules/base.js
lifeart: filePath will be public after https://github.com/ember-template-lint/ember-template-lint/pull/1729/files

lib/rules/base.js
scalvert: This is one log per Rule class, not once per module linted.

`ruleName` would work too, but it's effectively the same thing.

lib/rules/base.js
scalvert: Yep, removed.

lib/rules/base.js
scalvert: Yep, removed.

lib/rules/base.js
scalvert: Changed to `ruleName` 🤷‍♂️

rwjblue: thanks sorry, I originally misread the diff

lib/rules/base.js
```diff
@@ -69,6 +73,18 @@ module.exports = class Base {
     return this._configResolver.editorConfig();
   }
 
+  get _moduleName() {
+    if (!loggedModules.has(this.ruleName)) {
+      this._console.log(
+        'The `_moduleName` property is deprecated. Please use `_filePath` instead.'
+      );
+
+      this.loggedModules.push(this.ruleName);
```
rwjblue, line 82: This doesn't seem right, it is a module scoped variable (not an instance property)

lib/rules/base.js
```diff
@@ -69,6 +73,18 @@ module.exports = class Base {
     return this._configResolver.editorConfig();
   }
 
+  get _moduleName() {
+    if (!loggedModules.has(this.ruleName)) {
+      this._console.log(
+        'The `_moduleName` property is deprecated. Please use `_filePath` instead.'
```
rwjblue, line 79: Please include the rule name in the message, without it this is very hard to track down for folks.

lib/rules/base.js
scalvert: Derp. Mistake.

rwjblue (approved):

lib/rules/no-attrs-in-components.js
```diff
@@ -8,7 +8,7 @@ const componentTemplateRegex = new RegExp(
 
 module.exports = class NoAttrsInComponents extends Rule {
   isComponentTemplate() {
-    return componentTemplateRegex.test(this._moduleName);
+    return componentTemplateRegex.test(this._filePath);
```
rwjblue, line 11: I think this will need to fallback (triggering the deprecation) to avoid being breaking

```suggestion
    return componentTemplateRegex.test(this._filePath || this._moduleName);
```

lib/rules/no-outlet-outside-routes.js
```diff
@@ -22,7 +22,7 @@ module.exports = class NoOutletOutsideRoutes extends Rule {
   }
 
   visitor() {
-    this.__isRouteTemplate = isRouteTemplate(this._moduleName);
+    this.__isRouteTemplate = isRouteTemplate(this._filePath);
```
rwjblue, line 25:
```suggestion
    this.__isRouteTemplate = isRouteTemplate(this._filePath || this._moduleName);
```

rwjblue (changes requested):

rwjblue (approved):

Issue #1761, opened 2021-02-18
`require-input-label` fails when label is a sibling element


mckennagalvin:
`require-input-label` seems to throw errors for elements whose labels are defined outside of the input element.
e.g. the following code doesn't pass and gives an error for `form elements require a valid associated label`, even though it is valid markup for a11y:

```
<label for="foo">this is a label</label>
<textarea id="foo"/>
```

MelSumner:
I wasn't able to replicate this issue, but I think that this rule should support the textarea element anyway, so I submitted a PR that will cover it. 👍

amk221:
I'm also experiencing this with:

```handlebars
{{#let (unique-id) as |id|}}
  <label id={{id}}>
    {{@label}}
  </label>
  <select for={{id}}>
    {{#each @options as |option|}}
      ...
    {{/each}}
  </select>
{{/let}}
```

MelSumner:
This was resolved.

PR #1763, opened 2021-02-18
Added textarea element support to require-input-label rule

MelSumner:
If merged, this PR should resolve issue #1761 by adding support for the `<textarea />` element to the `require-input-label` rule. 

- added good and bad tests
- added check for `<textarea />` element


MelSumner:
After some more review, please hold on reviewing this for now- it might already be resolved in `next` and I need to double check a few more things.

MelSumner:
Closing this PR, since 1. it's addressed more thoroughly already in `next`, and 2. it wasn't really the issue after all.

Issue #1762, opened 2021-02-18
Add new rule to deprecate passing `classBinding` and `classNameBindings` to component invocations


rwjblue:
Deprecate passing `classBinding` and `classNameBindings` as arguments.

References: 

* https://github.com/emberjs/rfcs/blob/master/text/0691-deprecate-class-binding-and-class-name-bindings.md

rwjblue:
Fixed by #1767 and #1769

PR #1769, opened 2021-02-18
Add new rule: `no-class-bindings`

scalvert:
Implements #1762

### DIFF
--- a/docs/rule/no-class-bindings.md
+++ b/docs/rule/no-class-bindings.md
@@ -1,0 +1,49 @@
+# no-class-bindings
+
+It is possible to pass `classBinding` and `classNameBindings` as arguments to a component when invoked with curly syntax.
+
+```hbs
+{{some-component classNameBindings="foo:truthy:falsy"}}
+```
+
+These arguments are merged into the class attribute on the class, regardless of whether or not the component is a classic component which contains the classNameBindings logic.
+
+This rule disallows components accepting either `classBinding` or `classNameBindings`.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+{{some-thing classBinding="lol"}}no
+```
+
+```hbs
+<SomeThing @classBinding="lol" />
+```
+
+```hbs
+{{some-thing classNameBindings="lol:foo:bar"}}
+```
+
+```hbs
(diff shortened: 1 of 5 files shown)


docs/rule/no-class-bindings.md
```diff
@@ -0,0 +1,49 @@
+# no-class-bindings
+
+It is possible to pass `classBinding` and `classNameBindings` as arguments to a component when invoked with curly syntax.
+
+```hbs
+{{some-component classNameBindings="foo:truthy:falsy"}}
+```
+
+These arguments are merged into the class attribute on the class, regardless of whether or not the component is a classic component which contains the classNameBindings logic.
```
scalvert, line 9: I'm open to a better description. This one sounds janky.

docs/rule/no-class-bindings.md
```diff
@@ -0,0 +1,49 @@
...
+
+This rule disallows components accepting either `classBinding` or `classNameBindings`.
```
MelSumner, line 11:
```suggestion
This rule disallows use of `classBinding` or `classNameBindings` on components.
```

docs/rule/no-class-bindings.md
MelSumner, line 9:
```suggestion
These arguments are merged into the `class` attribute, whether or not the component is a classic component (which contains the `classNameBindings` logic).
```

MelSumner (approved):
A few prose suggestions but not blocking. 
Do you think we should include a `--fix` option?

Thank you for working on this issue! 👍

rwjblue:
> Do you think we should include a --fix option?

I don't know how to fix all cases (specifically `classNameBindings`), but we could fix `classNames` probably (refactoring in curlies to `class=` is all you do, in elements you'd `@classNames` -> `class=` as an attribute).

scalvert:
@rwjblue it's `classBinding` and `classNameBindings`, there is no `classNames` in this PR.

scalvert:
@MelSumner we could, but I'm unsure we can cover all cases. I'd prefer to avoid unless we're sure we can fix reasonably.

rwjblue (approved):

Issue #1766, opened 2021-02-18
Improve `this.log` for rule base class


rwjblue:
Currently calling `this.log` requires a few properties to be provided manually:

```js
        this.log({
          message: `some message`,
          line: node.loc && node.loc.start.line,
          column: node.loc && node.loc.start.column,
          source: this.sourceForNode(node),
        });
```

I'd like to add the ability to pass `node` and have `line`,`column`, and `source` inferred. Something like:

```js
        this.log({
          message: `Passing the \`${specifiedKey}\` property as an argument within templates is not allowed.`,
          node,
        });
```

lifeart:
Could we support node + key? (attr.name, attr.value, elementNode.tag) etc (may be an additional param or argument), node is too big (imho)

rwjblue:
@lifeart - Can you explain more what you mean? What would the key be used for? How is `node` to big?

rwjblue:
Also, FWIW, this was implemented over in #1770.

lifeart:
@rwjblue I mean, if we need to highlight component tag name, we don't need to highlight whole component block, highlighting `Foo` from `<Foo @name={{a}} />` instead of `<Foo @name={{a}} />`.

and idea is add property path, and linter will figure out where this property located and highlight it.

```
this.log({node, part: 'tag' });
this.log({node, part: 'attributes.0.value' });
this.log({node, part: 'hash.name.key' });

**Comment by rwjblue** [comment]
I'm not sure how that is better. In those cases, just pass those nodes in directly.

```js
this.log({
  message: 'whatever here',
  node: elementNode.attributes[0].value
})
```

PR #1770, opened 2021-02-18
Add simplified mechanism for rules to log warnings and errors

scalvert:
Implements #1766. 

TODO:

- [ ] Ensure to add attribution for @josephdsumner as some of this code was borrowed from #1771

### DIFF
--- a/dev/new-rule/templates/rule.js
+++ b/dev/new-rule/templates/rule.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const Rule = require('./base');
 const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
 
 // TODO Change template to the real error message that you want to report
 const ERROR_MESSAGE = 'Error Message to Report';
@@ -17,9 +17,7 @@ module.exports = class PlaceholderForRuleClass extends Rule {
         if (failingCondition) {
           this.log({
             message: ERROR_MESSAGE,
-            line: node.loc && node.loc.start.line,
-            column: node.loc && node.loc.start.column,
-            source: this.sourceForNode(node),
+            node,
           });
         }
       },
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -382,7 +382,6 @@ module.exports = class Base {
     this.config = last(this._configStack);
   }
 
-  // eslint-disable-next-line complexity
   _processInstructionNode(node) {
     if (!this._allowInlineConfig) {
       // inline configuration is disabled, do nothing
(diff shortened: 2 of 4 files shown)


dev/new-rule/templates/rule.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
-const Rule = require('./base');
 const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./base');
```
scalvert, line 4: I swapped these as they result in a lint error when the file is generated.

lib/rules/base.js
```diff
@@ -524,12 +515,25 @@ module.exports = class Base {
...
     }
+
+    if (result.node) {
+      let node = result.node;
+
+      result.line = node.loc && node.loc.start.line;
+      result.column = node.loc && node.loc.start.column;
+      result.source = this.sourceForNode(node);
+
+      delete result.node;
```
scalvert, line 534: Realized reading #1771 that we should be deleting this so the result is stable (thanks @josephdsumner!).

lib/rules/base.js
```diff
@@ -524,12 +515,34 @@ module.exports = class Base {
...
     if (this._filePath) {
       defaults.filePath = this._filePath;
     }
+
     if (this[MODULE_NAME]) {
       defaults.moduleId = this[MODULE_NAME];
     }
+
+    // perform the node prop expansion only if those properties don't exist in result
+    if (result.node) {
```
rwjblue, line 528: Let's leverage the `Object.assign` just below to simplify this a bit.

```js
let nodeExpansion = node && {
};

let reportedResult = Object.assign({}, defaults, nodeExpansion, result);
```

scalvert: Updated with a few tweaks.

lib/rules/base.js
```diff
@@ -520,17 +511,35 @@ module.exports = class Base {
...
+        column: node.loc && node.loc.start.column,
+        source: this.sourceForNode(node),
+      };
+
+      delete result.node;
+
+      locProps = Object.assign({}, nodeProps, result);
+    }
+
+    let reportedResult = Object.assign({}, defaults, locProps, result);
```
rwjblue, line 542: I don't _think_ we need two assign's here 🤔

rwjblue (commented):
I was thinking RE: this, and it might be easiest to just land the feature now (in 2.x) then after we release 3.0.0 (and are no longer maintaining a long lived `next` branch) we can update all existing rules to use the new API. 

The main reasoning here is that changing this in all the rules is likely to make future rebase/merging efforts into `next` branch difficult...

What do you think @scalvert?

lib/rules/base.js
scalvert: That's true, we can just do the single assign on 542.

scalvert:
@rwjblue agree.

lib/rules/base.js
scalvert: Done.

rwjblue (approved):

PR #1771, opened 2021-02-18
Add ability to pass `node` to `this.log`

josephdsumner:
@MelSumner @rwjblue 

Motivation: #1766 

If merged, this PR would allow a base-extending rule to pass the current `node` as part of the `results` object sent to the logger via `this.log(results)`.  The current behavior is for the logger to append the received `results`, as is, to the reported results. The new behavior is that if the `node` is passed directly, then it will be used to infer and set the values for `source`, `line`, and `column` of the result. In the event that `node` is passed **in addition to** any of these values, the value(s) inferred from `node` will be used. Other (collision-free) items passed directly in `results` -- e.g., `message`, `isFixable`, `fixedTemplate` -- would pass through to the reported results unaffected.  The rationale here is that we wouldn't have to immediately break or migrate existing rulesets.

### DIFF
--- a/lib/rules/base.js
+++ b/lib/rules/base.js
@@ -529,6 +529,16 @@ module.exports = class Base {
     if (this[MODULE_NAME]) {
       defaults.moduleId = this[MODULE_NAME];
     }
+    if (result.node) {
+      let rawResult = { ...result };
+      let inferredResults = {
+        line: rawResult.node.loc && rawResult.node.loc.start.line,
+        column: rawResult.node.loc && rawResult.node.loc.start.column,
+        source: this.sourceForNode(rawResult.node),
+      };
+      delete result.node;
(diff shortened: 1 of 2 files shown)


scalvert:
Hey @josephdsumner, thanks for the PR. I slightly beat you to it with #1770 :)

I'm going to close this one in favor of that.

lib/rules/base.js
```diff
@@ -529,6 +529,16 @@ module.exports = class Base {
     if (this[MODULE_NAME]) {
       defaults.moduleId = this[MODULE_NAME];
     }
+    if (result.node) {
+      let rawResult = { ...result };
```
rwjblue, line 533: FWIW, I think we can't use `...foo` syntax just yet

PR #1773, opened 2021-02-19
Add max-warnings option

mukilane:
Adds `--max-warning` option, similar to [eslint](https://eslint.org/docs/2.0.0/user-guide/command-line-interface#-max-warnings) and [sass-lint](https://github.com/sasstools/sass-lint/tree/master/docs/cli).

TODO

- [x] Update README

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -169,6 +169,10 @@ function parseArgv(_argv) {
         describe: 'Prevent inline configuration comments from changing config or rules',
         boolean: true,
       },
+      'max-warnings': {
+        describe: 'Number of warnings to trigger nonzero exit code',
+        type: 'number',
+      },
     })
     .help()
     .version();
@@ -273,7 +277,11 @@ async function run() {
   }
 
   let results = processResults(resultsAccumulator);
-  if (results.errorCount > 0) {
+
+  if (
+    results.errorCount > 0 ||
+    (!options.quiet && options.maxWarnings && results.warningCount > options.maxWarnings)
+  ) {
     process.exitCode = 1;
   }
 
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -64,6 +64,8 @@ describe('ember-template-lint executable', function () {
                         [array] [default: [\\"**/dist/**\\",\\"**/tmp/**\\",\\"**/node_modules/**\\"]]
             --no-inline-config          Prevent inline configuration comments from
                                         changing config or rules                 [boolean]
+            --max-warnings              Number of warnings to trigger nonzero exit code
+                                                                                  [number]
             --help                      Show help                                [boolean]
             --version                   Show version number                      [boolean]"
         `);
@@ -107,6 +109,8 @@ describe('ember-template-lint executable', function () {
                         [array] [default: [\\"**/dist/**\\",\\"**/tmp/**\\",\\"**/node_modules/**\\"]]
             --no-inline-config          Prevent inline configuration comments from
                                         changing config or rules                 [boolean]
+            --max-warnings              Number of warnings to trigger nonzero exit code
+                                                                                  [number]
             --help                      Show help                                [boolean]
             --version                   Show version number                      [boolean]"
         `);
@@ -381,6 +385,8 @@ describe('ember-template-lint executable', function () {
                         [array] [default: [\\"**/dist/**\\",\\"**/tmp/**\\",\\"**/node_modules/**\\"]]
             --no-inline-config          Prevent inline configuration comments from
                                         changing config or rules                 [boolean]
+            --max-warnings              Number of warnings to trigger nonzero exit code
+                                                                                  [number]
             --help                      Show help                                [boolean]
             --version                   Show version number                      [boolean]"
         `);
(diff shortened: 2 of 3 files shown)


bin/ember-template-lint.js
```diff
@@ -273,7 +278,10 @@ async function run() {
   }
 
   let results = processResults(resultsAccumulator);
-  if (results.errorCount > 0) {
+
+  if (options['max-warnings'] !== -1) {
+    process.exitCode = results.errorCount + results.warningCount >= options['max-warnings'] ? 1 : 0;
```
rwjblue, line 283: This behavior is a tad odd to me, and doesn't seem to match what eslint does.

For example (unless I'm misreading the diff here), if you invoke like `ember-template-lint --max-warnings=10 .` and you have 2 **errors** the command will still exit with an exit code of zero.

I think we need to check for `results.errorCount > 0 || results.maxWarnings > options.maxWarnings`, or something like that.

bin/ember-template-lint.js
```diff
@@ -169,6 +169,11 @@ function parseArgv(_argv) {
         describe: 'Prevent inline configuration comments from changing config or rules',
         boolean: true,
       },
+      'max-warnings': {
+        describe: 'Number of warnings to trigger nonzero exit code',
+        type: 'number',
+        default: -1,
```
rwjblue, line 175: Let's remove the default value. If we do, then we can do a simpler comparison (`options.maxWarnings && results.warningCount > options.maxWarnings`) without having to check for `-1`.

Also, to a user/human a value of `-1` is somewhat "odd" (like when they read the help output). If I read it, I'd assume that we actually wanted to error for any warnings! (which is definitely not your intent here).

test/acceptance/cli-test.js
```diff
@@ -1317,6 +1323,52 @@ describe('ember-template-lint executable', function () {
...
+                '<h2>Here too!!</h2><div>Bare strings are bad...</div><!-- bad html comment! -->',
+            },
+          },
+        });
+
+        let result = await run(['.', '--max-warnings=4']);
+
+        expect(result.exitCode).toEqual(0);
+        expect(result.stderr).toBeFalsy();
+      });
```
rwjblue, line 1370: This test doesn't seem to be testing the thing it says it is.

I think you need to tweak it like:

```suggestion
      it('should return exit code 0 if warnings is less or equal to max-warnings', async function () {
        project.setConfig({
          rules: {
            'no-bare-strings': 'warn',
            'no-html-comments': 'warn',
          },
        });
        project.write({
          app: {
            templates: {
              'application.hbs':
                '<h2>Here too!!</h2><div>Bare strings are bad...</div><!-- bad html comment! -->',
            },
          },
        });

        let result = await run(['.', '--max-warnings=3']);

        expect(result.exitCode).toEqual(0);
        expect(result.stderr).toBeFalsy();
      });

      it('should return exit code 1 if warnings is greater than max-warnings', async function () {
        project.setConfig({
          rules: {
            'no-bare-strings': 'warn',
            'no-html-comments': 'warn',
          },
        });
        project.write({
          app: {
            templates: {
              'application.hbs':
                '<h2>Here too!!</h2><div>Bare strings are bad...</div><!-- bad html comment! -->',
            },
          },
        });

        let result = await run(['.', '--max-warnings=2']);

        expect(result.exitCode).toEqual(1);
        expect(result.stderr).toMatchInlineSnapshot();
      });
```

test/acceptance/cli-test.js
```diff
@@ -1317,6 +1323,52 @@ describe('ember-template-lint executable', function () {
...
+                '<h2>Here too!!</h2><div>Bare strings are bad...</div><!-- bad html comment! -->',
+            },
+          },
+        });
+
+        let result = await run(['.', '--max-warnings=3']); // One less than total
+
+        expect(result.exitCode).toEqual(1);
+        expect(result.stderr).toBeFalsy();
+      });
```
rwjblue, line 1348: Same here, this test suggests at the issue I pointed out in the bin script logic but doesn't actually test it. This test should be about ensuring that _any_ errors cause exitCode === 1 (regardless of `--max-warnings` value).

Can you update to:

```suggestion
      it('should should return exit code 1 if total error is less than max-warnings', async function () {
        project.setConfig({
          rules: {
            'no-bare-strings': 'error',
            'no-html-comments': 'error',
          },
        });
        project.write({
          app: {
            templates: {
              'application.hbs':
                '<h2>Here too!!</h2><div>Bare strings are bad...</div><!-- bad html comment! -->',
            },
          },
        });

        let result = await run(['.', '--max-warnings=10000']);

        expect(result.exitCode).toEqual(1);
        expect(result.stderr).toMatchInlineSnapshot();
      });
```

rwjblue (changes requested):
I've left some inline comments on things to fix/tweak/address to ensure that it works like we want, but overall the feature sounds great. Thanks for working on it!

bin/ember-template-lint.js
mukilane: I had the intent of handling both error and warnings under the same flag, but that seems odd now. Have handled only the warnings and removed the default value.

rwjblue (approved):
Awesome, thank you @mukilane!

PR #1774, opened 2021-02-21
Add fixer for `no-class-bindings` rule.

rwjblue:


rwjblue (commented):
One thing to note here, is that `fixElement` and `fixMustacheLike` are really similar. I intentionally chose to duplicate here though, because there are lots of very subtle differences (need to use `b.sexpr` instead of `b.mustache`, need to use `node.hash.pairs` instead of `node.attributes`, etc). I originally had all of those behind a conditional in each location but it was very noisy and hard to grok.

lib/rules/no-class-bindings.js
```diff
@@ -1,27 +1,183 @@
...
+function dasherize(str) {
+  let decamelizedString = str.replace(STRING_DECAMELIZE_REGEXP, '$1_$2').toLowerCase();
+
+  return decamelizedString.replace(STRING_DASHERIZE_REGEXP, '-');
+}
+
 module.exports = class NoClassBindings extends Rule {
   visitor() {
-    function check(node) {
+    let check = (node) => {
```
scalvert, line 22: Why was this necessary?

scalvert (approved):

lib/rules/no-class-bindings.js
rwjblue: The function uses `this` internally. Previously, it worked because the visitor hooks themselves are called bound to the plugin instance but this refactor changed to _sometimes_ calling this function directly (e.g. when it wouldn't be bound to the plugin instance naturally). Changing to an arrow function meant that I didn't have to update to `check.call(this, node)` in the fixers below.

test/unit/rules/no-class-bindings-test.js
```diff
@@ -90,5 +94,340 @@ generateRuleTests({
...
+              "source": "@classNameBindings={{@something}}",
+            },
+          ]
+        `);
+      },
+    },
+
+    {
+      template: '<SomeThing @classNameBindings="isTruthy" />',
+      fixedTemplate: '<SomeThing class="{{if this.isTruthy "is-truthy"}}" />',
```
Turbo87, line 121: I have two questions about this:

- wouldn't `isTruthy` refer to the `isTruthy` of `SomeThing` instead of the calling context (`this`)?
- I roughly remember that classNameBindings were allowed to be strings too, results in something like `{{if this.isTruthy (kebab-case this.isTruthy)}}`. Am I misremembering that?

rwjblue: > wouldn't isTruthy refer to the isTruthy of SomeThing instead of the calling context (this)?

You are correct. I have introduced a bug in the fixer for this. 🤦 

> I roughly remember that classNameBindings were allowed to be strings too, results in something like {{if this.isTruthy (kebab-case this.isTruthy)}}. Am I misremembering that?

Yes, there are tests for this behavior already.

PR #1783, opened 2021-02-23
Updating new rule generator to use yeoman

scalvert:
Updates the new rule generator with the following changes:

- changes the script name to `yarn new:rule` (more descriptive)
- uses file templates to output content (these can be expanded and improved later if we want to add more generator options)
- correctly inserts and sorts `rules/index.js` with the new rule ID

Screen Shot 2021-02-23 at 9 24 17 AM

Additionally:

- fixes the `scripts/update-rules.js` script to not include Octane (it no longer has rules of its own)
- renames the `yarn update` script to `yarn update:readme`


.eslintrc.js
```diff
@@ -149,5 +149,16 @@ module.exports = {
         'import/no-dynamic-require': 'off',
       },
     },
+    {
+      files: ['dev/**/*.js'],
+      rules: {
+        'node/no-unpublished-require': [
```
rwjblue, line 155: We should remove this override and instead add `dev/` to `.npmignore`

dev/new-rule/index.js
```diff
@@ -1,119 +1,10 @@
-// eslint-disable-next-line eslint-comments/disable-enable-pair
-/* eslint-disable no-console */
+const { createEnv } = require('yeoman-environment');
 
-const fs = require('fs');
-const path = require('path');
+// eslint-disable-next-line wrap-iife
```
rwjblue, line 3: What is `wrap-iife`?

Can you write it as:

```js
async function main() {
  let env = createEnv();

  env.register(require.resolve('./new-rule-generator'), 'ember-template-lint:new-rule');

  await env.run('ember-template-lint:new-rule');
}

main();
```

This seems fine (old habbits with a "main function" die hard I guess), and avoids the linting comment

dev/new-rule/templates/rules-index.ejs
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = {
```
rwjblue, line 3: Let's put a comment in here saying this file is generated, and also add a `yarn update:rule-index` script to update it without having to add a new rule.

rwjblue: Then you can also add a test to ensure that `yarn update:rule-index` results in "no changes".

rwjblue (changes requested):

rwjblue:
Can you also update `CONTRIBUTING.md` (for both `yarn new` change and `yarn update` change)?

dev/new-rule/templates/rules-index.ejs
scalvert: So I _could_ do this, but if you look at what `updateRulesIndex` is doing, it's using yeoman's generators to generate the file. If I change this to move away from yeoman's `fs`, it also opts us out of the yeoman runLoop, which governs when the file is updated/written. I don't think we want to do that, as it results in strange interleaving with the yeoman write process.

I'm happy to do a similar thing in a separate util. Up to you.

dev/new-rule/index.js
scalvert: Yep, makes sense.

.eslintrc.js
scalvert: Done.

dev/new-rule/templates/rules-index.ejs
rwjblue: Chatted offline about this with @scalvert. Let's make an issue to track this follow on work, it isn't blocking this from landing (it's just another nice improvement IMHO).

rwjblue (approved):
Generally LGTM (once CI is happy).

Only remaining TODO (other than CI) is to add an issue so we remember the follow on infra work RE: regenerating `rules/index.js`.

package.json
```diff
@@ -20,11 +20,11 @@
     "lint:docs": "markdownlint **/*.md",
     "lint:js": "eslint . --cache",
     "lint:package-json": "sort-package-json --check",
-    "new": "node dev/new-rule/index.js",
+    "new:rule": "node dev/new-rule/index.js",
     "release": "release-it",
     "test": "npm-run-all lint:* test:*",
     "test:jest": "jest",
-    "update": "node ./scripts/update-rules.js"
+    "update:readme:": "node ./scripts/update-rules.js"
```
bmish, line 27: extra colon?

rwjblue: 😹

scalvert: Yep Fixed.

Issue #1782, opened 2021-02-23
Node engine support changes for 3.0.0


rwjblue:
* Node 13 is no longer supported by Node itself, we should remove it from our engines.
* We should update the Node 10 supported version to 10.25.0 (this lets us start working towards fs.promises usage)

PR #1785, opened 2021-02-23
Require Node >= 10.24

scalvert:
Implements #1782

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "yeoman-generator": "^5.0.1"
   },
   "engines": {
-    "node": ">= 10.25.* < 13.* || 12.* < 13.* || >= 14.*"
+    "node": ">= 10.25 < 11 || 12.* || >= 14.*"
   },
   "publishConfig": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "yeoman-generator": "^5.0.1"
   },
   "engines": {
-    "node": "10.* || 12.* || >= 14.*"
+    "node": ">= 10.25.* < 13.* || 12.* < 13.* || >= 14.*"
```
bmish, line 79:
```suggestion
    "node": ">= 10.25 < 11 || 12.* || >= 14.*"
```

11 and 13 both need to be excluded.

scalvert: You're right, @bmish. For some reason the semver validator I was using was giving me odd results. This makes sense.

bmish (approved):

scalvert:
@bmish @rwjblue it looks like 10.24 is the latest on the 10.* range. Updating to that.

Issue #1764, opened 2021-02-18
Add rule to deprecate passing positional arguments to <LinkTo>


rwjblue:
```hbs
Error:

{{link-to "About Us" "about"}}
          ~~~~~~~~~~~~~~~~~~
Replacement:

<LinkTo @route="about">About Us</LinkTo>
```

```hbs
Error:

{{#link-to "about"}}About Us{{/link-to}}
           ~~~~~~~
Replacement:

<LinkTo @route="about">About Us</LinkTo>
```

```hbs
Error:

{{#link-to "post" @post}}Read {{@post.title}}...{{/link-to}}
           ~~~~~~~~~~~~

Replacement:

<LinkTo @route="post" @model={{@post}}>Read {{@post.title}}...</LinkTo>
```

```hbs
Error:

{{#link-to "post.comment" @comment.post @comment}}
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  Comment by {{@comment.author.name}} on {{@comment.date}}
{{/link-to}}

Replacement:

<LinkTo @route="post.comment" @models={{array post comment}}>
  Comment by {{comment.author.name}} on {{comment.date}}
</LinkTo>
```

```hbs
Error:

{{#link-to "posts" (query-params direction="desc" showArchived=false)}}
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  Recent Posts
{{/link-to}}

Replacement:

<LinkTo @route="posts" @query={{hash direction="desc" showArchived=false}}>
  Recent Posts
</LinkTo>
```

References:

https://github.com/emberjs/rfcs/blob/master/text/0698-deprecate-link-to-positional-arguments.md

PR #1787, opened 2021-02-24
New rule: no-link-to-positional-params

scalvert:
Implements #1764. Messages match the spec in https://github.com/emberjs/rfcs/blob/master/text/0698-deprecate-link-to-positional-arguments.md.

This rule will be added to `recommended` in `next` for v3.0


bmish:
Typically we use `no-` or `require-` prefixes in rule names, i.e. `no-link-to-positional-params`.

scalvert:
@bmish sure, but there's other rules that are named `deprecated-*`. I was following that convention.

bmish:
The `deprecated-` naming convention is something I've only ever seen used by a few old ember-template-lint rules. I generally try to follow naming conventions found in eslint and common eslint plugins instead.

scalvert:
We should probably document that then. That would save folks time when they're contributing.

test/unit/rules/no-link-to-positional-params-test.js
```diff
@@ -0,0 +1,91 @@
...
+  good: [
+    '<LinkTo @route="about">About Us</LinkTo>',
+    '<LinkTo @route="post" @model={{@post}}>Read {{@post.title}}...</LinkTo>',
+    `<LinkTo @route="post.comment" @models={{array post comment}}>
+        Comment by {{comment.author.name}} on {{comment.date}}
+      </LinkTo>`,
+    `<LinkTo @route="posts" @query={{hash direction="desc" showArchived=false}}>
+        Recent Posts
+      </LinkTo>`,
+  ],
```
rwjblue, line 19: Need some good examples using `{{link-to`

rwjblue:
LGTM once we have more curly example tests.

test/unit/rules/no-link-to-positional-params-test.js
scalvert: Done

rwjblue (approved):

PR #1790, opened 2021-02-26
Remove non-strict variant of link-rel-noopener

rwjblue:


docs/rule/link-rel-noopener.md
```diff
@@ -33,9 +33,7 @@ This rule **allows** the following:
 The following values are valid configuration:
 
 * string -- `strict` for enabled and validating both noopener `and` noreferrer
```
bmish, line 35: Shouldn't we remove mention of `strict` since there's no option anymore?

bmish (approved):

bmish:
Also need to update the `recommended` config to remove the reference to `strict`.

docs/rule/link-rel-noopener.md
rwjblue: There is still an option for `strict` actually, it just has the same behavior as `true`.

rwjblue: I was vaguely worried about folks "internalizing" that they had to use `"strict"` for this rule's configuration (since that has been true for a very long time in the 2.x series), and being confused when they swapped to `true`.

bmish: I see. I'd personally rather do one of the following:
* remove the the `strict` option since it's redundant now
* or replace the `strict` option with a `lenient` option (default false) for people who want to opt back in to the old default behavior to ease migration

rwjblue: updated

scalvert (approved):

bmish (approved):

PR #1801, opened 2021-03-02
docs: avoid mentioning reverted change about `no-model-argument-in-route-templates` in v3 CHANGELOG

bmish:
This change (#1647) was reverted (no longer in recommended config) so it's misleading to include this item.


rwjblue (approved):
Thank you!

Issue #1802, opened 2021-03-02
no-unknown-arguments-for-builtin-components: Argument "@route" is required for <LinkTo /> component


Turbo87:
I've just seen this lint error with the v3 release, but from what I can tell this error is wrong. Using `LinkTo` with just `@query` works very well and should not be forbidden by the rule 🤔

rwjblue:
Yep, good catch thank you!

PR #1805, opened 2021-03-03
Ensure `no-unknown-arguments-for-builtin-components` allows `@query` only `<LinkTo>`'s

lifeart:
resolves https://github.com/ember-template-lint/ember-template-lint/issues/1802


rwjblue (approved):
Thanks for fixing so quickly @lifeart!

Issue #1800, opened 2021-03-01
no-bare-strings should check placeholder attribute on built-in Input components


jaydgruber:
The default config of `no-bare-strings` checks `placeholder` on an `<input>` element.
It should also check `placeholder` for `<Input>` and `<Textarea>` built-in components.
ref: https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input

rwjblue:
@jaydgruber - Totally agreed!

PR #1806, opened 2021-03-03
Ensure `no-bare-strings` check's `placeholder` attribute for `<Input />` and `<Textarea />`

lifeart:
resolves https://github.com/ember-template-lint/ember-template-lint/issues/1800

### DIFF
--- a/lib/rules/no-bare-strings.js
+++ b/lib/rules/no-bare-strings.js
@@ -35,6 +35,8 @@ const GLOBAL_ATTRIBUTES = [
 
 const TAG_ATTRIBUTES = {
   input: ['placeholder'],
+  Input: ['placeholder'],
+  Textarea: ['placeholder'],
   img: ['alt'],
 };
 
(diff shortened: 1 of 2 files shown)


lib/rules/no-bare-strings.js
```diff
@@ -35,6 +35,8 @@ const GLOBAL_ATTRIBUTES = [
 
 const TAG_ATTRIBUTES = {
   input: ['placeholder'],
+  Input: ['placeholder'],
+  Textarea: ['placeholder'],
```
rwjblue, line 39: Should we also check `@placeholder` for these two?

rwjblue: Looks like we did support those named arguments, though now with https://github.com/emberjs/rfcs/blob/master/text/0707-modernize-built-in-components-2.md they are deprecated.

I'd like to check them also though..

rwjblue (approved):

PR #1807, opened 2021-03-03
Remove `deprecated-each-syntax` from `recommended` config

### LINKED ISSUE #185: {{#each}} without params?
This is probably an edge case that may not be worth tackling, but I thought I'd throw it in just in case. I'm running into a situation where `deprecated-each-syntax` and `unused-block-params` seem to be at odds. I have an each block where I'm rendering something once for each item in a collection, but not referencing the items in the collection at all:

```handlebars
{{#each things}}
  <li>foo</li>
{{/each}}
```

This trips the `deprecated-each-syntax` rule, despite not using the `each x in things` syntax. Adding a throwaway block param (`{{#each things as |thing|}}`) trips the `unused-block-params` rules.

I see in [the rule](https://github.com/rwjblue/ember-template-lint/blob/master/lib/rules/deprecations/lint-deprecated-each-syntax.js#L57) that there's a case for this, presumably targeting an even older deprecated `{{#each}}` usage.

rwjblue:
This rule was written to help folks migrate away from these older `{{#each` syntaxes in the 1.x era. The assumption here is that if you have `{{#each something}}` that you are relying on "context shifting", but that is simply not the case. For example, you may use a `repeat` helper like:

```hbs
{{#each (repeat 3)}}
 do this three times
{{/each}}
```

In that case, you may not _need_ the block param (which would make this not trigger the rule, but would also trigger the "no-unused-block-params" rule).

This rule should be deprecated itself, and slated for removal in ember-template-lint@4.

Closes #185 (again)

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ Each rule has emojis denoting:
 |                            | [attribute-indentation](./docs/rule/attribute-indentation.md)                                             |
 | :nail_care:                | [block-indentation](./docs/rule/block-indentation.md)                                                     |
 | :white_check_mark:         | [builtin-component-arguments](./docs/rule/builtin-component-arguments.md)                                 |
-| :white_check_mark:         | [deprecated-each-syntax](./docs/rule/deprecated-each-syntax.md)                                           |
+|                            | [deprecated-each-syntax](./docs/rule/deprecated-each-syntax.md)                                           |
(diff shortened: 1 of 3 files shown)


bmish:
Can you also update the rule documentation to say/explain why it's deprecated?

rwjblue:
I created https://github.com/ember-template-lint/ember-template-lint/issues/1809 to track actually deprecating it and updating the documentation.

PR #1810, opened 2021-03-03
Add rules index generator

### LINKED ISSUE #1784: New Generator: update:rules-index
Building on #1783, we want to extra the rules index update templates to a separate generator, and compose that together with the new rule generator. This will allow us to have the following:

```bash
yarn new:rule # composes the new rule and rules index generators together

yarn update:rules-index # includes only the behavior to update the rules index file
```

scalvert:
Adds a new generator to regenerate the rules `index.js` file.

Closes #1784


rwjblue (commented):
Changes look good to me, can you:

* add a test that fails when the rules index is not in sync (e.g. if someone hand edits the index)
* update the contributing.md

scalvert:
Sounds good.

dev/generate.js
```diff
@@ -0,0 +1,16 @@
...
+
+const AVAILABLE_GENERATORS = ['new-rule', 'rules-index'];
+
+async function main() {
+  let generator = process.argv.slice(2);
+  let env = createEnv();
+
+  for (const generator of AVAILABLE_GENERATORS) {
+    env.register(require.resolve(`./${generator}-generator`), `ember-template-lint:${generator}`);
+  }
```
rwjblue, line 11: It's kinda odd to use `const generator` here to shadow the `let generator` just above, maybe change one of them?

package.json
```diff
@@ -35,6 +36,9 @@
       "<rootDir>/test/**/*-test.js"
     ]
   },
+  "resolutions": {
+    "tmp": "^0.2.1"
```
rwjblue, line 40: Hmm, why?

dev/generate.js
scalvert: Ya good call.

scalvert: Fixed.

package.json
scalvert: Removed.

rwjblue (approved):

PR #1812, opened 2021-03-04
Adding SARIF printer

scalvert:
Adds a new printer to output results in [SARIF](https://sarifweb.azurewebsites.net/) format. This allows ember-template-lint to integrate more tightly with github annotations when using applications. As Github natively utilizes SARIF as an interchange standard, this avoids the need to adapt the JSON results into SARIF externally.

The implementation is heavily based on [Microsoft's ESLint SARIF formatter](https://github.com/microsoft/sarif-sdk/blob/main/src/ESLint.Formatter/sarif.js).

We should discuss how printers are currently working. For now, this is implemented with a new command line option `--sarif`, but moving to a reporter-based system similar to what eslint provides would allow for greater flexibility.


lib/printers/sarif.js
```diff
@@ -0,0 +1,194 @@
...
+            const encoding = chardet.detect(contents);
+
+            // Encoding will be null if it could not be determined.
+            if (encoding) {
+              // Convert the content bytes to a UTF-8 string.
+              contentsUtf8 = utf8.encode(contents.toString(encoding.encoding));
+
+              sarifFiles[result.filePath].contents = {
+                text: contentsUtf8,
+              };
```
rwjblue, line 79: I don't fully grok this. We basically only operate on UTF-8 files...

lib/printers/sarif.js
```diff
@@ -0,0 +1,194 @@
...
+          },
+        };
+
+        let contentsUtf8;
+
+        if (embedFileContents) {
+          try {
+            // Try to get the file contents and encoding.
+            const contents = fs.readFileSync(result.filePath);
+            const encoding = chardet.detect(contents);
```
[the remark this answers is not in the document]
scalvert: We can probably just assume that it's `utf-8`?

lib/printers/sarif.js
scalvert: Yep, we can absolutely just trim this stuff out. It was largely cribbed from the eslint formatter, and I knew this would be a point of discussion. I wanted to get your 👀 on it to discuss this aspect before I made the assumption it should be `utf-8` always.

lib/printers/sarif.js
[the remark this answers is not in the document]
scalvert: Removed this.

lib/printers/sarif.js
scalvert: Removed this.

yarn.lock
```diff
@@ -1776,6 +1776,11 @@ chardet@^0.7.0:
   resolved "https://registry.yarnpkg.com/chardet/-/chardet-0.7.0.tgz#90094849f0937f2eedc2425d0d28a9e5f0cbad9e"
   integrity sha512-mT8iDcrh03qDGRRmoA2hmBJnxpllMR+0/0qlzjqZES6NdiWDcZkCNAk4rPFZ9Q85r27unkiNNg8ZOiwZXBHwcA==
 
+chardet@^1.3.0:
```
rwjblue, line 1779: Probably needs a fresh `yarn` (since these deps are removed)

bin/ember-template-lint.js
```diff
@@ -143,6 +143,10 @@ function parseArgv(_argv) {
         describe: 'Format output as json',
         boolean: true,
       },
+      sarif: {
+        describe: 'Format output as SARIF',
+        boolean: true,
+      },
```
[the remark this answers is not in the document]
scalvert: Agree. I knew this would come up, but just wanted to have something working before we did the formatters stuff.

yarn.lock
scalvert: Yep.

bin/ember-template-lint.js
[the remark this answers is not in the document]
scalvert: Since #1340 has been merged, I've updated to use the format option.

yarn.lock
scalvert: Done.

.github/workflows/ci.yml
```diff
@@ -50,9 +50,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: print node node-version
```
scalvert, line 53: I'd incorrectly merged this in with #1340. This reverts that change (it was only for debugging purposes).

rwjblue (commented):
Seems like the file needs to incorporate todos still (no clue how / what that should look like) based on `--include-todo` and/or `--update-todo`.

lib/formatters/sarif.js
```diff
@@ -0,0 +1,182 @@
...
+      ],
+    };
+    const sarifFiles = {};
+    const sarifArtifactIndices = {};
+    let nextArtifactIndex = 0;
+    const sarifRules = {};
+    const sarifRuleIndices = {};
+    let nextRuleIndex = 0;
+    const sarifResults = [];
+    const embedFileContents = process.env.SARIF_EMBER_TEMPLATE_LINT_EMBED === 'true';
```
rwjblue, line 45: Hmm, this is somewhat odd. Should we have a better way to specify this?

lib/formatters/sarif.js
```diff
@@ -0,0 +1,182 @@
...
+            if (message.rule && message.rule !== 'global') {
+              sarifRepresentation.ruleId = message.rule;
+
+              if (typeof sarifRules[message.rule] === 'undefined') {
+                sarifRuleIndices[message.rule] = nextRuleIndex++;
+
+                // Create a new entry in the rules dictionary.
+                sarifRules[message.rule] = {
+                  id: message.rule,
+                  helpUri: `https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/${message.rule}.md`,
```
rwjblue, line 103: Should probably not link to `master`, but instead link to `blog/v${currentVersion}/docs/rule/...`

lib/formatters/sarif.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    if (Object.keys(sarifRules).length > 0) {
+      for (const ruleId of Object.keys(sarifRules)) {
+        let rule = sarifRules[ruleId];
+        sarifLog.runs[0].tool.driver.rules.push(rule);
+      }
+    }
+
+    this.console.log(JSON.stringify(sarifLog, null, 2));
```
rwjblue, line 178: I'd prefer to have this save to a file, instead of `console.log` (otherwise you interleave any arbitrary stdout along with the actual sarif output and end up with something that is unparsable).

test/acceptance/cli-test.js
```diff
@@ -1099,6 +1099,182 @@ describe('ember-template-lint executable', function () {
       });
     });
 
+    describe('with sarif formatter', function () {
```
rwjblue, line 1102: Is there a SARIF validator tool? Can we use it here to ensure that our output is valid format?

lib/formatters/sarif.js
scalvert: What do you suggest? Ultimately, having a way to not bloat the SARIF log with file contents seems good.

lib/formatters/sarif.js
scalvert: We can do that. I can do something similar to what we do with Checkup.

test/acceptance/cli-test.js
scalvert: We can validate with ajv, if we want to go that route. The SARIF log has a schema.

lib/formatters/sarif.js
scalvert: If we do this, it seems like we should provide an `--output-file` option too, unless we just want to autogenerate SARIF log names. Thoughts?

lib/formatters/sarif.js
scalvert: Good call. Done.

test/acceptance/cli-test.js
scalvert: So for some reason ajv doesn't want to validate using the SARIF schema, as the `$schema` property _within_ the SARIF schema itself is erroring in ajv. Unsure why, but we should for sure look at doing this. Perhaps we can do a follow up?

lib/formatters/sarif.js
rwjblue: Maybe have it leverage `--verbose` flag?

lib/formatters/sarif.js
rwjblue: Ya, agree RE: adding `--output-file`. In addition we could choose a reasonable filename by default (to make `ember-template-lint . --format sarif` work without an `--output-path`).

test/acceptance/cli-test.js
rwjblue: Ya, seems fine to follow up. My hope was mostly just that since this isn't an arbitrary format that we are making up (like existing `--json` output) I'd really like **some** way to make sure we are compliant.

Maybe open a new issue and cross reference?

lib/formatters/sarif.js
scalvert: I like.

scalvert: Done.

lib/formatters/sarif.js
scalvert: Agreed.

rwjblue: Oh, another factor here is that we should emit to stdout when piping.

So I think the decision tree here is:

* if `--output-file` is provided, use it and save a file
* if piping, log the content to stdout
* choose a filename and save to working dir

scalvert: OK, I've added the functionality to pipe to stdout.

test/acceptance/cli-test.js
scalvert: Makes sense.

lib/formatters/default.js
```diff
@@ -14,12 +14,16 @@ class DefaultPrinter {
     let printOptions = {
       console: this.console,
       includeTodo: options.includeTodo,
+      isInteractive: process.stdout.isTTY,
+      outputFile: options.outputFile,
       quiet: options.quiet,
       updateTodo: options.updateTodo,
       verbose: options.verbose,
+      workingDirectory: options.workingDirectory,
```
rwjblue, line 22: Since `Rule` has `workingDir` already as a public API, lets do:

```suggestion
      workingDir: options.workingDirectory,
```

scalvert: Done.

test/formatters/sarif-test.js
```diff
@@ -0,0 +1,301 @@
...
+    process.chdir(ROOT);
+    await project.dispose();
+  });
+
+  it('should output sarif log to default path (in project working directory)', function () {
+    let sarifWindowsPattern = /Report\swritten\sto\s(.*\\ember-template-lint-report-\d{4}(?:-\d{2}){3}(?:_\d{2}){2}\.sarif)/;
+    let sarifUnixPattern = /Report\swrit{2}en\sto\s(.*\/ember-template-lint-report-\d{4}(?:-\d{2}){3}(?:_\d{2}){2}\.sarif)/;
+    let pattern = IS_WINDOWS ? sarifWindowsPattern : sarifUnixPattern;
+    let formatter = new Sarif(
+      Object.assign(DEFAULT_OPTIONS, {
```
rwjblue, line 205: Can you throw another `{}` in here? As it stands, you currently mutate `DEFAULT_OPTIONS` but don't "clean it up" in teardown.

```suggestion
      Object.assign({}, DEFAULT_OPTIONS, {
```

something like that (don't actually accept, it won't pass prettier)

test/formatters/sarif-test.js
```diff
@@ -0,0 +1,301 @@
...
+
+    formatter.print(RESULTS);
+  });
+
+  it('should output sarif log to custom relative path', function () {
+    let sarifWindowsPattern = /Report\swrit{2}en\sto\s(.*\\my-custom-file.sarif)/;
+    let sarifUnixPattern = /Report\swrit{2}en\sto\s(.*\/my-custom-file.sarif)/;
+    let pattern = IS_WINDOWS ? sarifWindowsPattern : sarifUnixPattern;
+    let formatter = new Sarif(
+      Object.assign(DEFAULT_OPTIONS, {
```
rwjblue, line 232: same here

test/formatters/sarif-test.js
```diff
@@ -0,0 +1,301 @@
...
+
+    formatter.print(RESULTS);
+  });
+
+  it('should output sarif log to custom absolute path', function () {
+    let sarifWindowsPattern = /Report\swrit{2}en\sto\s(.*\\subdir\\my-custom-file.sarif)/;
+    let sarifUnixPattern = /Report\swrit{2}en\sto\s(.*\/subdir\/my-custom-file.sarif)/;
+    let pattern = IS_WINDOWS ? sarifWindowsPattern : sarifUnixPattern;
+    let formatter = new Sarif(
+      Object.assign(DEFAULT_OPTIONS, {
```
rwjblue, line 261: and here 😸

test/formatters/sarif-test.js
```diff
@@ -0,0 +1,301 @@
...
+        workingDir: project.baseDir,
+      })
+    );
+
+    formatter.print(RESULTS);
+  });
+
+  it('should output sarif log JSON not using TTY', function () {
+    let formatter = new Sarif(
+      Object.assign(DEFAULT_OPTIONS, {
```
rwjblue, line 287: 🕺

test/formatters/sarif-test.js
scalvert: Yep.

scalvert: Done.

test/formatters/sarif-test.js
scalvert: Done.

test/formatters/sarif-test.js
scalvert: Done.

rwjblue (approved):

Issue #1821, opened 2021-03-06
Suggestions in no-unknown-built-in-component-arguments are often a bit misleading


rwjblue:
Some errors from https://github.com/ember-template-lint/ember-template-lint/issues/1820, specifically the "did you mean" text for these is pretty wild to me:

**Input**
```
error  "@autocomplete" is unknown argument for <Input /> component. Did you mean "@value"?  no-unknown-arguments-for-builtin-components
error  "@class" is unknown argument for <Input /> component. Did you mean "@escape-press"?  no-unknown-arguments-for-builtin-components
error  "@required" is unknown argument for <Input /> component. Did you mean "@escape-press"?  no-unknown-arguments-for-builtin-components
error  "@spellcheck" is unknown argument for <Input /> component. Did you mean "@checked"?  no-unknown-arguments-for-builtin-components
```

**LinkTo**
```
error  "@class" is unknown argument for <LinkTo /> component. Did you mean "@activeClass"?  no-unknown-arguments-for-builtin-components
error  "@invokeAction" is unknown argument for <LinkTo /> component. Did you mean "@activeClass"?  no-unknown-arguments-for-builtin-components
```

---

I'd like to:

* Teach the rule about all known arguments to the built-ins (even the deprecated ones), and for the deprecated ones either don't provide a suggestion per-se or suggest moving to the attribute
* Determine how we can tell the likelihood that our match is reasonable (`@invokeAction` -> `@activeClass` for example doesn't seem reasonable), and optionally not emit suggestion text if the match isn't confident enough.

rwjblue:
@lifeart - What do you think?

lifeart:
Lol, in my idea `Did you mean...` should handle typos, agree we should add knowledge about deprecated properties and extend messaging | add autofix for this case.

PR #1823, opened 2021-03-07
Fix suggestions for deprecated events and argument names in `no-unknown-built-in-component-arguments`

lifeart:
resolves: https://github.com/ember-template-lint/ember-template-lint/issues/1821


rwjblue (approved):
Great work here @lifeart, thank you!

PR #1824, opened 2021-03-07
Add missing ARIA roles to `no-invalid-role`

### LINKED ISSUE #1818: role="switch" cause no-invalid-role error
https://github.com/ember-template-lint/ember-template-lint/blob/8085aa967a60f7365627dcd074b60517e803ff31/lib/rules/no-invalid-role.js#L162

Per https://www.w3.org/WAI/PF/aria/roles#widget_roles switch role is valid one however it cause an error with v3 ember-lint-template as it's not in the list of valid roles. Looks like a defect.

Windvis:
I restructured the code into separate Sets since that makes it easier to compare with the documentation. Let me know if I should revert that.

The changes are split into separate commits to make it easier to review, but I can squash if needed.

Closes #1818

### DIFF
--- a/lib/rules/no-invalid-role.js
+++ b/lib/rules/no-invalid-role.js
@@ -106,70 +106,84 @@ const ELEMENTS_DISALLOWING_PRESENTATION_OR_NONE_ROLE = new Set([
 ]);
 
 // List of roles taken from https://www.w3.org/WAI/PF/aria/roles#widget_roles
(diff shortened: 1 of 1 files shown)


Windvis:
I didn't add any extra tests yet since I'm not sure what's expected exactly. I could add tests for the missing roles I added here, but there are other roles that don't have tests yet either.

rwjblue (approved):
This looks great, thank you @Windvis!

PR #1825, opened 2021-03-08
Emit full path when emitting output on CI.

rwjblue:
When running in CI environments, it is helpful to emit the full path to the file on disk (instead of the path relative to the specified working directory). For example, in GitHub Actions CI there are a number of matchers (RegExp's on standard out) that they use to add annotations to files.

Prior to this change, those annotations worked properly for applications that were located in the repo root path but failed to properly identify a file when the the application was not in the repo root. Specifically, when linting in a monorepo format (basically when the project's `package.json` is not also the location of `.git`) the annotations would not work properly.

This changes the output to always use the full path in CI conditions.  This should not regress users (seeing the relative path _looks_ better) since they don't often run in CI environments.

### DIFF
--- a/lib/formatters/default.js
+++ b/lib/formatters/default.js
@@ -17,6 +17,7 @@ class DefaultPrinter {
       quiet: options.quiet,
       updateTodo: options.updateTodo,
       verbose: options.verbose,
+      workingDir: options.workingDir,
     };
 
     // TODO: make format json work
--- a/lib/formatters/pretty.js
+++ b/lib/formatters/pretty.js
@@ -1,4 +1,7 @@
+const path = require('path');
+
 const chalk = require('chalk');
+const ciInfo = require('ci-info');
 
 const Linter = require('../linter');
 
@@ -99,7 +102,10 @@ class PrettyPrinter {
       .map((error) => PrettyPrinter._formatError(error, options))
       .join('\n');
 
-    return `${chalk.underline(filePath)}\n${errorsMessages}\n`;
+    let displayFilePath =
+      ciInfo.isCI && options.workingDir ? path.join(options.workingDir, filePath) : filePath;
+
+    return `${chalk.underline(displayFilePath)}\n${errorsMessages}\n`;
   }
 
   static _formatError(error, options = {}) {
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   "dependencies": {
     "@ember-template-lint/todo-utils": "^8.0.0-beta.3",
     "chalk": "^4.0.0",
+    "ci-info": "^3.1.1",
     "ember-template-recast": "^5.0.1",
     "find-up": "^5.0.0",
     "fuse.js": "^6.4.6",


lib/formatters/default.js
```diff
@@ -17,6 +17,7 @@ class DefaultPrinter {
       quiet: options.quiet,
       updateTodo: options.updateTodo,
       verbose: options.verbose,
+      workingDir: options.workingDir,
```
scalvert, line 20: We have both `workingDir` and `workingDirectory`. Which should we prefer?

lib/formatters/pretty.js
```diff
@@ -99,7 +102,10 @@ class PrettyPrinter {
       .map((error) => PrettyPrinter._formatError(error, options))
       .join('\n');
 
-    return `${chalk.underline(filePath)}\n${errorsMessages}\n`;
+    let displayFilePath =
```
scalvert, line 105: I know this is largely for CI only, but it'd be nice to have a test for this so we know what it's expected to look like in either case.

scalvert (approved):

lib/formatters/pretty.js
```diff
@@ -99,7 +102,10 @@ class PrettyPrinter {
...
+      ciInfo.isCI && options.workingDir ? path.join(options.workingDir, filePath) : filePath;
```
Turbo87, line 106: Is workingDir a resolved/absolute path?

lib/formatters/default.js
rwjblue: Where do we have `workingDirectory`?

rwjblue: I think rules have `this.workingDir`, thats why I used it here. But you are right, it is actually `options.workingDirectory` from the bin script parser.

lib/formatters/pretty.js
rwjblue: Ya, totally right. I need to think on how to do that, might have to use a "DI" style injection to the constructor in a unit test to force both modes (since we actually do run in CI, the `ci-info` package will be `true` all the time so tests for the non-CI case will fail).

lib/formatters/pretty.js
rwjblue: When we default it, it is but not when it is not defaulted.

https://github.com/ember-template-lint/ember-template-lint/blob/6ec1d058c31e1c323cf7937ace4beb83b63780c2/bin/ember-template-lint.js#L220-L222

Will fix to ensure it is always absolute.

rtablada: Not sure that it is blocking, but I think it might be nice to allow this to be configured as an option to show absolute paths. Personally as a dev used to eslint, dotnet build warnings, PHP warnings, and TS errors I'm more used to having absolute paths in local dev too so I disagree with the assumption of "seeing the relative path looks better"

rtablada:
👍 to this other than a request to allow user config to show absolute path even outside of CI (maybe a `options.showAbsolutePath ?? isCi`)

lib/formatters/default.js
scalvert: Do we need both? Can we just have one?

lib/formatters/pretty.js
scalvert: I think it's less about it "looking better" but rather it's a bit "easier on the eyes". I do think you have a point, and consistency across tools can be a strong driver for these types of features.

lifeart:
https://github.com/ember-template-lint/ember-template-lint/pull/2491 seems related

lifeart:
closing this PR because functionality already available using flag landed in https://github.com/ember-template-lint/ember-template-lint/pull/2491

PR #1841, opened 2021-03-13
Add new rule `no-with`

### LINKED ISSUE #1808: Add `no-deprecated-with` rule
Based on https://github.com/emberjs/rfcs/blob/master/text/0445-deprecate-with.md, we should add a rule to prevent usage of `{{#with`.

bertdeblock:
Closes #1808.

To do:
- [x] Add "fix" mode support (will land this separately)
- [x] Update Ember version in deprecation URL
- [x] Update Ember version in docs (updated the docs to match the deprecation guides)

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,7 @@ Each rule has emojis denoting:
 | [no-unused-block-params](./docs/rule/no-unused-block-params.md)                                           | ✅  |     |     |     |
 | [no-whitespace-for-layout](./docs/rule/no-whitespace-for-layout.md)                                       |     | 💅  | ⌨️  |     |
 | [no-whitespace-within-word](./docs/rule/no-whitespace-within-word.md)                                     |     | 💅  | ⌨️  |     |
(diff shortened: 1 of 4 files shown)


bmish:
I think we could probably just name this `no-with`. A lot of our rules deal with deprecations but we don't necessarily need to say that in the rule name.

bertdeblock:
@bmish Sure! Will update when I add support for fix mode. Though I'm still not completely sure what the logic behind the fix would be.

rwjblue:
@bertdeblock - I think we can land the fix separately from the initial rule.

bertdeblock:
Renamed the rule to `no-with`. Let me know if anything else.

bertdeblock:
Gentle ping 😃 Anything else I should do before this can be merged?

rwjblue:
@bertdeblock - Mind rebasing?

bertdeblock:
@rwjblue Done!

bmish:
@bertdeblock can you rebase one more time? I think we can merge this today.

bmish (approved):
Thanks!

PR #1857, opened 2021-03-17
Adding ignore option to no-passed-in-event-handlers

scalvert:
Adds the ability to ignore specific event handlers from triggering this rule. This is largely intended to assist in migration purposes, where the cost of requiring all callers to update existing APIs is high. This option provides an intermediate migration path for consumers that want to enable this rule to prevent most violations, while allowing some violations to be temporarily ignored.

### DIFF
--- a/docs/rule/no-passed-in-event-handlers.md
+++ b/docs/rule/no-passed-in-event-handlers.md
@@ -50,14 +50,21 @@ This rule **allows** the following:
 
   - object -- An object with the following keys:
 
-    - `ignore` -- An array of event handler names to ignore. Event handler names should exclude the @ when specifying those intended for named arguments. This rule will ensure both non-named arguments and named arguments are both ignored appropriately.
+    - `ignore` -- An object with the following keys/values:
+
+      - key: string -- The name of the element or mustache statement to ignore event handlers for
+      - value: array -- Event handler names. Event handler names should exclude the @ when specifying those intended for named arguments. This rule will ensure both non-named arguments and named arguments are both ignored appropriately.
 
       eg.
 
       Given the following configuration:
 
       ```json
-      { "ignore": [click] }
+      {
+        "ignore": {
+          "MyButton": ["click"]
+        }
+      }
       ```
 
       Both `click` and `@click` will be ignore as violations of this rule.
--- a/docs/rule/no-passed-in-event-handlers.md
+++ b/docs/rule/no-passed-in-event-handlers.md
@@ -44,6 +44,24 @@ This rule **allows** the following:
 {{foo onClick=this.handleClick}}
 ```
 
+## Configuration
+
+- boolean - `true` to enable / `false` to disable
+
+  - object -- An object with the following keys:
+
+    - `ignore` -- An array of event handler names to ignore. Event handler names should exclude the @ when specifying those intended for named arguments. This rule will ensure both non-named arguments and named arguments are both ignored appropriately.
+
+      eg.
+
+      Given the following configuration:
+
+      ```json
+      { "ignore": [click] }
+      ```
+
+      Both `click` and `@click` will be ignore as violations of this rule.
+
 ## Migration
 
 - create explicit component APIs for these events (e.g. `@click` -> `@onClick`)
--- a/docs/rule/no-passed-in-event-handlers.md
+++ b/docs/rule/no-passed-in-event-handlers.md
@@ -67,8 +67,6 @@ This rule **allows** the following:
       }
       ```
 
-      Both `click` and `@click` will be ignore as violations of this rule.
-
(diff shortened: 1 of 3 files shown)


MelSumner (approved):
great tests! Thanks for doing this so quickly! 👍

rwjblue:
Hmm, I'd rather this be based on the component being invoked (not the handler name).

scalvert:
So essentially:

```
{
  ignore: {
    'ArtDecoButton': ['click'],
    //...
  }
}
```

rwjblue:
Yes, exactly!

scalvert:
Okey dokey

docs/rule/no-passed-in-event-handlers.md
```diff
@@ -44,6 +44,24 @@ This rule **allows** the following:
...
+
+    - `ignore` -- An array of event handler names to ignore. Event handler names should exclude the @ when specifying those intended for named arguments. This rule will ensure both non-named arguments and named arguments are both ignored appropriately.
+
+      eg.
+
+      Given the following configuration:
+
+      ```json
+      { "ignore": [click] }
+      ```
```
rwjblue, line 61: This needs to be updated with the new config structure

docs/rule/no-passed-in-event-handlers.md
```diff
@@ -44,6 +44,24 @@ This rule **allows** the following:
...
+
+      Both `click` and `@click` will be ignore as violations of this rule.
```
rwjblue, line 63: We should have them provide the named argument name assuming angle bracket invocation, and then we'd internally handle `@click` when inside angle invocation (because `@click` in angle invocation is the same thing as `click` in mustaches). I think this is what the rule already does, I just want to make sure that the docs make that clear.

lib/rules/no-passed-in-event-handlers.js
```diff
@@ -40,7 +40,53 @@ const EVENT_HANDLER_METHODS = new Set([
   'drop',
 ]);
 
+const DEFAULT_CONFIG = {
+  ignore: [],
+};
```
rwjblue, line 45: This seems wrong now that we expect an object. I think the only reason it isn't throwing an error in the tests is that `let arr = []; arr.click` is perfectly "valid" (it'll just always be undefined).

lib/rules/no-passed-in-event-handlers.js
```diff
@@ -40,7 +40,53 @@ const EVENT_HANDLER_METHODS = new Set([
...
+}
+
 module.exports = class NoPassedInEventHandlers extends Rule {
+  parseConfig(config) {
+    switch (typeof config) {
+      case 'boolean': {
+        return config ? DEFAULT_CONFIG : false;
+      }
+      case 'object': {
+        return isValidConfigObjectFormat(config) ? config : DEFAULT_CONFIG;
```
rwjblue, line 73: When we aren't valid config wise, we should throw an error (not fallback to default).

test/unit/rules/no-passed-in-event-handlers-test.js
```diff
@@ -6,7 +6,7 @@ const generateRuleTests = require('../../helpers/rule-test-harness');
 generateRuleTests({
   name: 'no-passed-in-event-handlers',
 
-  config: 'true',
```
rwjblue, line 9: O_o

rwjblue (changes requested):

docs/rule/no-passed-in-event-handlers.md
scalvert: On it.

lib/rules/no-passed-in-event-handlers.js
scalvert: It doesn't seem wrong, it _is_ wrong. Updated.

lib/rules/no-passed-in-event-handlers.js
scalvert: I think returned false makes sense here, and is inline with what other rules do.

test/unit/rules/no-passed-in-event-handlers-test.js
scalvert: Ya, I have no idea why this was stringified...

docs/rule/no-passed-in-event-handlers.md
scalvert: Are you suggesting that we have them provide the handler in named arg syntax, eg. `@click`?

scalvert: OK fixed, based on our offline discussion.

rwjblue (approved):

PR #1861, opened 2021-03-19
Fix false positive with local `log` in `no-log` rule

bertdeblock:
Prevents local `{{log}}`s from being flagged.

### DIFF
--- a/lib/rules/no-log.js
+++ b/lib/rules/no-log.js
@@ -2,13 +2,13 @@
 
 const Rule = require('./_base');
 
-const message = 'Unexpected {{log}} usage.';
+const ERROR_MESSAGE = 'Unexpected {{log}} usage.';
 
 module.exports = class NoLog extends Rule {
   _checkForLog(node) {
-    if (node.path.original === 'log') {
+    if (node.path.original === 'log' && !this.isLocal(node)) {
       this.log({
-        message,
+        message: ERROR_MESSAGE,
         node,
       });
     }
@@ -27,4 +27,4 @@ module.exports = class NoLog extends Rule {
   }
 };
 
-module.exports.message = message;
+module.exports.ERROR_MESSAGE = ERROR_MESSAGE;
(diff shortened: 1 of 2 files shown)


lib/rules/no-log.js
```diff
@@ -2,31 +2,37 @@
 
 const Rule = require('./base');
 
-const message = 'Unexpected {{log}} usage.';
+const ERROR_MESSAGE = 'Unexpected {{log}} usage.';
```
bertdeblock, line 5: Renamed so it's more consistent with the newer rules.

lib/rules/no-log.js
```diff
@@ -2,31 +2,37 @@
...
+      BlockStatement(node) {
+        if (this.isLogNode(node)) {
+          this.logErrorMessage(node);
+        }
       },
 
-      BlockStatement(node) {
-        this._checkForLog(node);
+      MustacheStatement(node) {
+        if (this.isLogNode(node) && !this.isLocal(node)) {
```
bertdeblock, line 17: Noticed that `isLocal` sometimes returns `undefined` instead of a boolean. Is this expected?

rwjblue: No, it shouldn't return `undefined`. We can fix that over in scope.

lib/rules/no-log.js
```diff
@@ -2,31 +2,37 @@
...
-      });
-    }
-  }
-
   visitor() {
     return {
-      MustacheStatement(node) {
-        this._checkForLog(node);
+      BlockStatement(node) {
+        if (this.isLogNode(node)) {
```
rwjblue, line 11: I think both `BlockStatement` and `MustacheStatement` need to do the `!this.isLocal(node)` guard. 

Additionally (though this isn't this rules job) `{{#log` is **never** going to be the built-in log helper AFAIK.

rwjblue (changes requested):

lib/rules/no-log.js
bertdeblock: Updated. I was changing the existing code a bit too much to my personal preference. I reverted this. The diff should be cleaner now.

bmish:
Is this ready to merge?

bertdeblock:
It should be yes!

bmish (approved):
Thanks!

PR #1863, opened 2021-03-20
Fix `undefined` error message for `no-link-to-positional-params` rule

### LINKED ISSUE #1854: `no-link-to-positional-params` rule reports `undefined` violation message with position parameter and query-params
```hbs
{{link-to
  "Launchpad"
  "index"
  (query-params fetchScenario="default")
}}
```

```
tests/dummy/app/templates/index.hbs
  5:4  error  undefined  no-link-to-positional-params
```

CC: @scalvert

bertdeblock:
Closes #1854.

Sorry for the large overhaul. Hope that's okay.

- Removes the type checks as a link's route / label / model(s) can be anything (e.g. `StringLiteral`, `PathExpression` or `SubExpression`)
- Takes into account more than 3 params (e.g. label, route, model(s) and query params)
- Removes the `{{link-to "about"}}` bad case as that's not valid usage of the `{{link-to}}` component (it triggers a build error)

### DIFF
--- a/lib/rules/no-link-to-positional-params.js
+++ b/lib/rules/no-link-to-positional-params.js
@@ -6,62 +6,58 @@ module.exports = class NoLinkToPositionalParams extends Rule {
   visitor() {
     return {
       MustacheStatement(node) {
-        this.process(node);
+        this.process(node, { hasRequiredTextParam: true });
       },
 
       BlockStatement(node) {
-        this.process(node);
+        this.process(node, { hasRequiredTextParam: false });
       },
     };
   }
 
-  process(node) {
+  process(node, options) {
     if (node.path.original === 'link-to' && node.params.length > 0) {
-      let message = this.getMessage(node);
+      let message = this.getMessage(node, options);
 
       this.log({ message, node });
     }
   }
 
-  getMessage(node) {
-    function getError(message) {
-      return `Invoking the \`<LinkTo>\` component with positional arguments is deprecated. ${message}`;
+  getMessage(node, { hasRequiredTextParam }) {
+    let requiredParamsCount = hasRequiredTextParam ? 2 : 1; // Route name is always required, text only for inline `{{link-to}}` components.
+    let hasQueryParams = includesQueryParams(node.params);
+    let modelParamsCount = countModelParams(node.params, requiredParamsCount, hasQueryParams);
+    let namedArguments = ['`route`'];
+    let additionalInfo = '';
+
+    if (modelParamsCount === 1) {
+      namedArguments.push('`model`');
+    } else if (modelParamsCount > 1) {
+      namedArguments.push('`models`');
     }
 
-    if (node.params.length === 1 && node.params[0].type === 'StringLiteral') {
-      return getError('Instead, please use the equivalent named arguments (`@route`).');
+    if (hasQueryParams) {
+      namedArguments.push('`query` using the `hash` helper');
     }
 
-    if (node.params.length === 2) {
-      if (node.params.every((param) => param.type === 'StringLiteral')) {
-        return getError(`Instead, please use the equivalent named arguments (\`@route\`) and pass a
-  block for the link's content.`);
-      }
-
-      if (node.params[0].type === 'StringLiteral' && node.params[1].type === 'PathExpression') {
-        return getError('Instead, please use the equivalent named arguments (`@route`, `@model`).');
-      }
-
-      if (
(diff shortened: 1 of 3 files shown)


lib/rules/no-link-to-positional-params.js
```diff
@@ -6,62 +6,58 @@ module.exports = class NoLinkToPositionalParams extends Rule {
...
+    return param.path && param.path.original === 'query-params';
+  });
+}
+
+function countModelParams(params, requiredCount, hasQueryParams) {
+  return Math.max(params.length - (hasQueryParams ? requiredCount + 1 : requiredCount), 0);
+}
+
+function getErrorMessage(namedArguments, additionalInfo) {
+  return `Invoking the \`<LinkTo>\` component with positional arguments is deprecated. Instead, please use the equivalent named arguments (${namedArguments}).${additionalInfo}`;
```
bertdeblock, line 62: Was wondering if it would be better if we use `{{link-to}}` here as that would better match the user's code?

test/unit/rules/no-link-to-positional-params-test.js
```diff
@@ -28,23 +28,84 @@ generateRuleTests({
 
   bad: [
     {
-      template: '{{link-to "about"}}',
+      template: '{{link-to "About Us" "about"}}',
```
scalvert, line 31: I'd prefer that if we're changing the logic within this rule that we only add new test cases, not modify any existing ones. This helps us prevent regressions, since if we change the rule's logic and all the existing tests pass, we know we've preserved at minimum the existing functionality. 

Can we bring back the existing tests and just add new ones?

scalvert (changes requested):

test/unit/rules/no-link-to-positional-params-test.js
bertdeblock: Hi Steve, thanks for reviewing!

I agree with keeping existing tests to prevent regressions, I kept all the others, but in this case I find it strange to do so because:
- `{{link-to "about"}}` isn't valid use of the component, inline `{{link-to}}` components require _both a label and a route_
  - It doesn't compile in newer projects
  - It triggers a runtime error in older projects
  - We don't lint against `{{link-to}}` as well (no params at all, which also isn't valid use of the component)
- [The old logic assumed that the first string argument was always the route](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-link-to-positional-params.js#L31-L33) which isn't true for inline `{{link-to}}` components (the label comes first), meaning the provided error message wasn't completely correct
- The [other](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/rules/inline-link-to-test.js) [link-to](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/rules/no-link-to-tagname-test.js) tests don't have this case either, I'm assuming also for the reasons mentioned above

bertdeblock: I added the test again, though I think it's strange to lint something that doesn't compile/run.

scalvert: Ah, so you're saying the inline version of `link-to` doesn't allow a single route param, correct? I think you're right - the inline version requires two params. Thanks for finding this. Let's change it to the correct params. Appreciate you fixing this!

lib/rules/no-link-to-positional-params.js
scalvert: Ya, we should probably detect which variant they're using and output that. What do you think?

bertdeblock: We could probably just use `{{link-to}}`, but I guess we should then also omit the `@` char for the named arguments?

test/unit/rules/no-link-to-positional-params-test.js
bertdeblock: Correct, it needs 2 params to compile/run. So I'll delete this test case?

lib/rules/no-link-to-positional-params.js
scalvert: Ya it would be nice to try to replicate what was matched on in the error message, so it's as accurate as possible. Can we do that?

test/unit/rules/no-link-to-positional-params-test.js
scalvert: Yep, sounds good!

scalvert (approved):

PR #1865, opened 2021-03-20
Make sure `scope.isLocal()` returns a boolean

bertdeblock:
Where can/should I make a test for this?

### DIFF
--- a/lib/helpers/scope.js
+++ b/lib/helpers/scope.js
@@ -130,6 +130,8 @@ class Scope {
         return true;
       }
     }
+
+    return false;
   }
 
   get currentNode() {


bmish:
I think we should first move the utils/helpers in `lib/rules/internal/` to instead live in `lib/helpers` where all the rest of the utils/helpers live. Then add test files for each of them under `test/unit/helpers/`

bertdeblock:
Alright thanks! Will see if I can manage.

bmish:
Moving the files should be done in a separate PR by the way, and you don't necessarily have to do it now, but that's how I imagine it.

bmish:
Status on this?

bertdeblock:
I haven't looked into setting up tests for this yet.

bmish:
If all we're doing is adding `return false;` to make sure the function returns a boolean, I think we can merge it without additional tests.

bertdeblock:
Okay, will rebase.

bmish (approved):

PR #1874, opened 2021-03-27
Document the A11Y preset

### LINKED ISSUE #1677: Why is a11y preset is not listed in README
Why is `a11y` preset not listed in README?
Is there a reason for this, or can I safely add this to my config?

bertdeblock:
Description taken from [this comment](https://github.com/ember-template-lint/ember-template-lint/issues/1677#issuecomment-754905979).

Closes #1677.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,10 +188,11 @@ For more detailed information see [configuration](docs/configuration.md).
 
 ### Presets
 
-|                    | Name                                     | Description                                                                                                                                                                                                                                                                          |
-| :----------------- | :--------------------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| :white_check_mark: | [recommended](lib/config/recommended.js) | enables the recommended rules                                                                                                                                                                                                                                                        |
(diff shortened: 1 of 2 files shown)


MelSumner (commented):
Should this also include the 👍  in the rule table, to show which rules are in the a11y preset?

bertdeblock:
Totally forgot about that, but that makes sense yes. Btw I only choose the 👍 emoji because I didn't know what else would fit.

bertdeblock:
@MelSumner Updated accordingly.

scalvert:
Does 👍 make sense as the emoji here? It doesn't really say Accessibility to me (I don't know what the alternative would be).

bmish:
I don't think 👍 is a good emoji for this preset. How about ♿?

bertdeblock:
Agreed that thumbs up isn't very representative. Let me know what it should be and I'll update!

bertdeblock:
Rebased and updated the emoji to be ♿.

MelSumner (approved):
[issue (non-blocking)](https://conventionalcomments.org/) I don't think that there is a good icon for this right now. We should ship it as-is, and change it as soon as we figure out what to use instead. 

Thank you for doing this work! 👍

rwjblue:
@bertdeblock - Sorry for the run around here, let's use `:keyboard:` ⌨️ instead.

rwjblue (approved):
Thank you!

PR #1881, opened 2021-03-29
Check for link title being substring of link text in `no-invalid-link-title` rule

### LINKED ISSUE #1162: no-invalid-link-title: additional test(s)
Add an invalid test case that ensures that a title which is a substring of the link text is disallowed.

bertdeblock:
Makes the rule more strict.

Closes #1162.

### DIFF
--- a/docs/rule/no-invalid-link-title.md
+++ b/docs/rule/no-invalid-link-title.md
@@ -2,9 +2,9 @@
 
 ✅ The `extends: 'recommended'` property in a configuration file enables this rule.
 
-The `title` attribute is a useful way to give users extra context for a link. However, this content should be complementary content and should not be the exact same value as the link text.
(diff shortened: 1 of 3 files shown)


bertdeblock:
@MelSumner Would you mind checking if this resolves #1162 as intended?

rwjblue:
I think that this is a breaking change (we will now catch more scenarios than before), but I think we can still consider it a bug fix.

That seem right to y'all?

bertdeblock:
@rwjblue Yup, sounds right 👍.

bmish:
Can you rebase and target to `next` so we can include this in v4 as part of https://github.com/ember-template-lint/ember-template-lint/issues/1908?

bmish (approved):

Issue #1852, opened 2021-03-17
Request: separate out fixable functionality and removing stale todos from `--fix` flag


MelSumner:
## Summary 
It's confusing that the `--fix` flag will remove todos that have been resolved AND ALSO fix issues that are auto-fixable. A different command should be used to remove resolved todos.

## Details 
Scenario: A developer is working on their part of a large project. They are only responsible for folder `app/foo/` which had an a11y-related todo. 

Developer fixes the issue and then runs `yarn lint:hbs --fix` to remove the todo. 

Developer sees that the command resulted in these changes:
- the correct todo file was removed (related to file changes in the `app/foo/` folder
- auto-fixable issues in `app/bar/` folder were also completed (oh no!)

This adds an extra step to the developer workflow. The command that was intended to quickly and easily remove the todo has now made more work for them- they have to undo the changes they unintentionally made. 

It's also more tedious to try to run the `--fix` command specifically on the file where the issue was resolved. 

Then intention of the `--fix` to remove the todo was ease of use, and right now it's not meeting that criteria.

scalvert:
I can see this being a problem based on the use case you described. We wanted the `--fix` flag to remove todos to simplify the workflow, but if it's have the opposite effect I'm totally open to changing this.

Techn1x:
I'd really like this feature too. I have outlined some justifications in one of the comments for https://github.com/ember-template-lint/ember-template-lint/issues/1898

MelSumner:
We discussed this today and we're going to add a `--prune-todo` flag that will remove any invalid/fixed/expired todos. We'll update the message to instruct users to run this flag instead. 

Related #1873 #1905

MelSumner:
The flag ended up being `--clean-todo` and it's in version 3.4.0, now available. Closing issue.

PR #1895, opened 2021-04-14
Updating stale todo message with more targeted --fix instructions

scalvert:
The original error message when a todo was stale was instructing users to run `--fix`, but without any qualifying file path. This could cause users to run `ember-template-lint . --fix`, which would potentially fix not only the stale todo, but anything else fixable in the app.

While the `--fix` flag should always be considered 'safe', it could be considered jarring to be instructed to run a particular command, and have unrelated files change. This PR updates the message so that it instructs you to run `--fix` only on the file with the stale todo.

### DIFF
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -53,7 +53,7 @@ module.exports = class TodoHandler {
         for (const [, todo] of itemsToRemoveFromTodos) {
           results.push({
             rule: 'invalid-todo-violation-rule',
-            message: `Todo violation passes \`${todo.ruleId}\` rule. Please run \`--fix\` to remove this todo from the todo list.`,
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
I'm excited about this change, I think it will make it nicer for developers! Thank you!

rwjblue (approved):

Techn1x:
Will this still autofix all the other problems in that same file?

For example, if my file `app/zzz.hbs` has 50 lint violations, all marked as todo, then I fix 10 of them and get this "invalid-todo-violation-rule" showing up, so I run
`ember-template-lint app/zzz.hbs --fix`

It'll still run all the autofixing codemods on that file right?

Would be great if we could just run the autofixer for only the `invalid-todo-violation-rule` or something like that..

MelSumner:
@Techn1x yes, it will autofix that file. We believe that the `--fix` flag improves the template code accurately (and if not, a bug should be filed), and should be regularly used as part of maintaining quality code (which really is the entire point of this library). 💪 

However, you could weigh in on #1852 if you wanted to. 👍

PR #1897, opened 2021-04-15
Allow words containing redundant words for `require-valid-alt-text` rule

### LINKED ISSUE #1896: require-valid-alt-text incorrectly parsing alt
`logout` fails the `require-valid-alt-text` rule, likely because the word logout contains the word "logo" (which is a banned term).

We may want to do a more comprehensive test here than just an 'includes'
https://github.com/ember-template-lint/ember-template-lint/blob/62589045e5ed5c6ca14eed13e80f563f30fb8520/lib/rules/require-valid-alt-text.js#L146-L148

I could see this causing issues for other _valid_ alt text entries. For example `alt="my photography class"` would fail due to the word `photo` being present in the string

examples of other words in alt text that would cause this lint issue
photography
telephoto
analogous
logon
logout
pilgrimage
photon

bertdeblock:
Fixes #1896.

### DIFF
--- a/lib/rules/require-valid-alt-text.js
+++ b/lib/rules/require-valid-alt-text.js
@@ -143,8 +143,9 @@ module.exports = class RequireValidAltText extends Rule {
                     node,
                   });
                 } else {
+                  const normalizedAltValueWords = normalizedAltValue.split(' ');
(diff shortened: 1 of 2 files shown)


bmish (approved):

Techn1x:
Thanks!

PR #1899, opened 2021-04-15
updated the invalid violation message for visual distinction

MelSumner:
I've tried to update the message for the `invalid-todo-violation-rule` to appear more visually distinct. Since we can't sub-chalk this, I've added some 🎉 . 

What do you think?

### DIFF
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -53,7 +53,7 @@ module.exports = class TodoHandler {
         for (const [, todo] of itemsToRemoveFromTodos) {
           results.push({
             rule: 'invalid-todo-violation-rule',
-            message: `Todo violation passes \`${todo.ruleId}\` rule. Please run \`ember-template-lint ${todo.filePath} --fix\` to remove this todo from the todo list.`,
+            message: `Todo violation passes \`${todo.ruleId}\` rule! 🎉 🎉  Please run \`ember-template-lint ${todo.filePath} --fix\` to remove this todo from the todo list.`,
             filePath: todo.filePath,
             moduleId: todo.moduleId,
             severity: 2,
--- a/test/acceptance/todo-cli-test.js
+++ b/test/acceptance/todo-cli-test.js
@@ -321,7 +321,7 @@ describe('todo usage', () => {
       expect(result.exitCode).toEqual(1);
       expect(result.stdout).toMatchInlineSnapshot(`
         "app/templates/require-button-type.hbs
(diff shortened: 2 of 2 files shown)


lib/-private/todo-handler.js
```diff
@@ -53,7 +53,7 @@ module.exports = class TodoHandler {
         for (const [, todo] of itemsToRemoveFromTodos) {
           results.push({
             rule: 'invalid-todo-violation-rule',
-            message: `Todo violation passes \`${todo.ruleId}\` rule. Please run \`ember-template-lint ${todo.filePath} --fix\` to remove this todo from the todo list.`,
+            message: `Todo violation passes \`${todo.ruleId}\` rule! 🎉 🎉  Please run \`ember-template-lint ${todo.filePath} --fix\` to remove this todo from the todo list.`,
```
MelSumner, line 56:
```suggestion
            message: `Todo violation now passes the \`${todo.ruleId}\` rule! 🎉 🎉  Please run \`ember-template-lint ${todo.filePath} --fix\` to remove this todo from the todo list.`,
```

test/acceptance/todo-cli-test.js
```diff
@@ -321,7 +321,7 @@ describe('todo usage', () => {
       expect(result.exitCode).toEqual(1);
       expect(result.stdout).toMatchInlineSnapshot(`
         "app/templates/require-button-type.hbs
-          -:-  error  Todo violation passes \`require-button-type\` rule. Please run \`ember-template-lint app/templates/require-button-type.hbs --fix\` to remove this todo from the todo list.  invalid-todo-violation-rule
+          -:-  error  Todo violation passes \`require-button-type\` rule! 🎉 🎉  Please run \`ember-template-lint app/templates/require-button-type.hbs --fix\` to remove this todo from the todo list.  invalid-todo-violation-rule
```
MelSumner, line 324:
```suggestion
          -:-  error  Todo violation now passes the \`require-button-type\` rule! 🎉 🎉  Please run \`ember-template-lint app/templates/require-button-type.hbs --fix\` to remove this todo from the todo list.  invalid-todo-violation-rule
```

scalvert:
Would you mind adding a screen cap of what this looks like? I'm just wondering if the emojis at the beginning would help with more visual distinction...

Also it seems odd to have an error, but show 🎉 . WDYT?

MelSumner:
@scalvert 

> Also it seems odd to have an error, but show 🎉 . WDYT?

Yes! It's odd, which is what I'm going for. The goal is to grab the developer's attention.

scalvert:
What if we instead change it to a ✅ , and put it at the start of the message. That way when people look at the output, it'll be easy to scan and see that visual difference.

scalvert:
Ping ^

PR #1904, opened 2021-04-22
Ensures we always emit output file even with empty results

scalvert:
Currently we only emit the output file if we have results, but this causes issues for CI systems that always expect the presence of an output file, regardless of lint violations or not. This PR ensures we always emit a file if one is specified in options.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -420,11 +420,12 @@ async function run() {
     let hasTodos = options.includeTodo && results.todoCount;
     let hasUpdatedTodos = options.updateTodo;
 
-    if (hasErrors || hasWarnings || hasTodos || hasUpdatedTodos) {
-      let Printer = require('../lib/formatters/default');
-      let printer = new Printer(options);
-      printer.print(results, todoInfo);
-    }
+    let Printer = require('../lib/formatters/default');
+    let printer = new Printer({
+      ...options,
+      hasResultData: hasErrors || hasWarnings || hasTodos || hasUpdatedTodos,
+    });
+    printer.print(results, todoInfo);
   }
 }
 
--- a/lib/formatters/default.js
+++ b/lib/formatters/default.js
@@ -43,11 +43,6 @@ class DefaultPrinter {
         this.delegates.push(new SarifPrinter(printOptions));
         break;
       }
-      case 'json': {
-        let JsonPrinter = require('./json');
-        this.delegates.push(new JsonPrinter(printOptions));
-        break;
-      }
       default: {
         try {
           const CustomPrinter = createRequire(
--- a/lib/formatters/default.js
+++ b/lib/formatters/default.js
@@ -14,12 +14,14 @@ class DefaultPrinter {
     let printOptions = {
       console: this.console,
       includeTodo: options.includeTodo,
-      isInteractive: process.stdout.isTTY,
+      // env var is used to test output options via tests
+      isInteractive: process.stdout.isTTY || process.env['IS_TTY'] === '1',
       outputFile: options.outputFile,
       quiet: options.quiet,
       updateTodo: options.updateTodo,
       verbose: options.verbose,
       workingDir: options.workingDirectory,
+      hasResultData: options.hasResultData,
     };
 
     // TODO: add 'json' format to the list of those below. The plan is to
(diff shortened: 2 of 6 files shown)


bin/ember-template-lint.js
```diff
@@ -419,8 +419,9 @@ async function run() {
     let hasWarnings = results.warningCount > 0;
     let hasTodos = options.includeTodo && results.todoCount;
     let hasUpdatedTodos = options.updateTodo;
+    let alwaysEmitOutputFile = options.format === 'sarif' && options.outputFile;
```
rwjblue, line 422: Why do we do the SARIF check here? Is that because we haven't migrated the JSON formatter to use the new infrastructure yet?

rwjblue: I kinda think it should always be up to the formatter to decide if we should emit. The default one (pretty) could choose to print nothing if there were no results (the current behavior), but then it would be trivially possible to author a formatter that behaved like our built in "file emitting" ones. 

What do you think?

lib/formatters/default.js
```diff
@@ -14,7 +14,7 @@ class DefaultPrinter {
     let printOptions = {
       console: this.console,
       includeTodo: options.includeTodo,
-      isInteractive: process.stdout.isTTY,
+      isInteractive: process.stdout.isTTY || process.env['IS_TTY'] === '1',
```
rwjblue, line 17: What is the second check for? Forcing interactive mode in tests? Would you mind making an inline comment explaining why the fallback...

bin/ember-template-lint.js
scalvert: Ya I like this. So we're essentially saying removing the conditional check and letting formatters make the call. We can also populate these checks above for the options passed to formatters, for convenience.

lib/formatters/default.js
scalvert: Correct. Happy to add a comment.

bin/ember-template-lint.js
scalvert: OK I've moved this check into the formatter itself, in addition to the checks in all the respective formatters.

lib/formatters/default.js
scalvert: Done

lib/formatters/default.js
```diff
@@ -41,6 +43,11 @@ class DefaultPrinter {
         this.delegates.push(new SarifPrinter(printOptions));
         break;
       }
+      case 'json': {
```
rwjblue, line 46: We originally avoided adding json here because we didn't want `--formatter json` to mean the same thing as `--json`. Have we changed our mind about that?

scalvert: Well as per our comment above

```
    // TODO: add 'json' format to the list of those below. The plan is to
    // modify the 'json' formatter to write to a file, similar to the 'sarif' formatter.
```

I was doing that. Should we not now?

rwjblue: I agree that this satisfies that comment now, but I think I also wanted to rethink the specific format. 

@scalvert - Let's decouple this for now (e.g. remove this case), and tackle it in a dedicate PR. Specifically, we should either a) we should introduce a `--formatter json` that uses a different format or b) deprecate the `--json` flag and suggest folks use `--formatter json`. I think "a" is fine, but I'd like that in a separate PR so it has a dedicated changelog entry (since it will certainly affect most editor integrations).

scalvert: Agree. Removing.

scalvert: Done

lib/formatters/json.js
```diff
@@ -7,6 +8,10 @@ class JsonPrinter {
   }
 
   print(results) {
+    if (!this.options.hasResultData && !this.options.outputFile) {
```
rwjblue, line 11:
```suggestion
    if (!this.options.hasResultData) {
```

lib/formatters/json.js
```diff
@@ -23,7 +28,14 @@ class JsonPrinter {
...
 
-    this.console.log(JSON.stringify(filteredErrors, null, 2));
+    let json = JSON.stringify(filteredErrors, null, 2);
+
+    if (this.options.isInteractive && this.options.outputFile) {
+      let outputPath = writeOutputFile(json, 'json', this.options);
+      this.console.log(`Report written to ${outputPath}`);
+    } else {
+      this.console.log(json);
+    }
```
rwjblue, line 38:
```suggestion
    this.console.log(json);
```

rwjblue (commented):
I want to be careful not to support `--outputFile foo.json --json`, I've made some inline suggestions to prevent that.

PR #1907, opened 2021-04-23
Deprecate `--json` option in favor of `--format=json`.

rwjblue:
Deprecate (documentation wise) using `--json` option, in favor of using `--format=json`. Everything else is the same (same format, same console outputting behavior).

Note: A future PR will refactor the formatters to use a unified system for emitting when `--outputFile` is specified.


scalvert (approved):

bmish:
I added this to a list of breaking changes for the next major release: #1908

PR #1918, opened 2021-04-27
Update some wording in README.md

MelSumner:
If merged, this PR:

- updates some prose for readability
- changes "addon" to "library" for accuracy

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 [![npm version](https://badge.fury.io/js/ember-template-lint.svg)](https://badge.fury.io/js/ember-template-lint)
 [![Build Status](https://github.com/ember-template-lint/ember-template-lint/workflows/CI/badge.svg)](https://github.com/ember-template-lint/ember-template-lint/actions?query=workflow%3ACI)
 
-ember-template-lint will lint your handlebars template and return error results.
+Ember Template Lint is a library that will lint your handlebars template and return error results.
 
-For example, given the rule [`no-bare-strings`](docs/rule/no-bare-strings.md) is enabled, this template would be
+For example, if the rule [`no-bare-strings`](docs/rule/no-bare-strings.md) is enabled, this template would be
 in violation:
 
 ```hbs
@@ -18,7 +18,7 @@ the `no-bare-strings` rule found an error.
 
 ## Installation
 
-This addon is installed by default with new Ember apps, so check your package.json before installing to see if you need to install it.
+This library is installed by default with new Ember apps, so check your `package.json` file before installing to see if you need to install it.
 
 To install ember-template-lint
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,9 +3,9 @@
 [![npm version](https://badge.fury.io/js/ember-template-lint.svg)](https://badge.fury.io/js/ember-template-lint)
 [![Build Status](https://github.com/ember-template-lint/ember-template-lint/workflows/CI/badge.svg)](https://github.com/ember-template-lint/ember-template-lint/actions?query=workflow%3ACI)
 
-ember-template-lint will lint your handlebars template and return error results.
+Ember Template Lint is a library that will lint your handlebars template and return error results.
```
scalvert, line 6: Let's keep the name of the package here rather than the other. Other than that, looks great!

```suggestion
`ember-template-lint` is a library that will lint your handlebars template and return error results.
```

scalvert (changes requested):

rwjblue (approved):

MrChocolatine:
May I ask why this change?

> changes "addon" to "library" for accuracy

Everywhere in the Ember ecosystem we speak about "add-ons". I don't get the "accuracy" here if the wording is different than what is used everywhere else, in the official documentations etc.

bertdeblock:
@MrChocolatine

Addons are NPM packages specifically written for Ember which can augment Ember apps or even other Ember addons.
E.g. by automatically providing components, helpers, services, templates, and so on.
They can also augment Ember apps and addons at build time, e.g. preprocessing SCSS/LESS files and so much more.
The [Using addons](https://cli.emberjs.com/release/basic-use/using-addons/) section might give you more insights into what is possible with Ember addons.

Addons are easy recognizable because of their file and folder structure (e.g. the top-level `app` and `addon` folders) and they also have the `ember-addon` keyword specified in their package file, which is what `ember-cli` uses to recognize addons (I think) and include them in your app.

`ember-template-lint` isn’t an addon though, it’s a normal NPM package (also written for Ember of course), but it mainly provides a CLI for linting hbs template files. It doesn’t augment an Ember app the way an addon does.
`ember-cli-template-lint` was actually the addon version of `ember-template-lint`.
It automatically augmented your test suite by including a lint test for each template in your app.
It’s not part of the blueprint anymore as using `ember-template-lint` directly is now the preferred approach.

I’m sure others can explain it better than I can, but hopefully this makes it a bit clearer why the word `library` is more correct here.

MrChocolatine:
> I’m sure others can explain it better than I can, but hopefully this makes it a bit clearer why the word `library` is more correct here.

It does yes, it makes more sense now, thank you very much @bertdeblock.

Issue #1923, opened 2021-05-01
Extend no-forbidden-elements APIs


ventuno:
Currently [`no-forbidden-elements`](https://github.com/ventuno/ember-template-lint/blob/master/docs/rule/no-forbidden-elements.md) only provides the option to override the list of default elements that are forbidden.

I would suggest that we add a new option (e.g.: `extendForbidden`) that allows for an _extension_ of such default list instead of just replacing it. This is useful for people who want to provide an extended list of forbidden elements without having to keep it up to date with the list of default/recommended ones.

Additionally, I would like to be able to provide a custom error message that can be used in place of the default one to recommend replacement elements or point to a piece of documentation (perhaps internal to an organization) where to reference locally defined best practices.

rajasegar:
@ventuno Sounds like a good idea to me. But I am just the implementer of the rule. The original idea belongs to @MelSumner

ventuno:
Thanks @rajasegar, I looked at the original commit but not the original ticket. @MelSumner can you please take a look and advise?

scalvert:
This seems like a good addition to me. PRs are welcome!

ventuno:
Thanks @scalvert I actually already have a PR out for this: https://github.com/ember-template-lint/ember-template-lint/pull/1924.

rwjblue:
We had a similar ask for customization of `no-bare-strings` a little while back, and went a slightly different direction:

https://github.com/ember-template-lint/ember-template-lint/pull/1481

The API you proposed in #1924 definitely seems better than what we did there to me (though I would call it `additional*` not `extend*`), but I think this is a little bit of the wrong fit for the no-forbidden-elements rule.

---

@ventuno - Can you help me understand what your use case is here? Are you trying to prevent additonal HTML elements? Or maybe you are trying to prevent specific **component** invocations?

ventuno:
@rwjblue in our use case we are trying to prevent developers from using HTML `iframe`s in favor of our `TestAwareIFrame`. So it would be nice for us to:
1) add `iframe` to the list of default elements without having to redefine it (the approach used in https://github.com/ember-template-lint/ember-template-lint/pull/1481 is perfect as well);
2) customize the error message to reference some internal documentation explaining how to fix the error in case of `iframe`s.

Does that make sense?

ventuno:
[...]

PR #1924, opened 2021-05-01
Extend no-forbidden-elements APIs.

ventuno:
WIP implementation of https://github.com/ember-template-lint/ember-template-lint/issues/1923.


rwjblue (changes requested):
I'd like to ensure that you cannot use **both** `forbidden` and `additionalForbidden` in the same configuration. Can you implement that and add a test case?

ventuno:
@rwjblue sure, what do you think should be the expected behavior?
1) an exception is thrown;
2) either `forbidden` or `additionalForbidden` is ignored.

My vote goes for 1 as it's difficult to decide which behavior should be the default in this case.

Issue #1925, opened 2021-05-01
Support message placeholders for error messages


ventuno:
[ESLinst](https://eslint.org/)'s [`context.report`](https://eslint.org/docs/developer-guide/working-with-rules#contextreport) is the API used to report linting validations (similar to `ember-template-eslint`'s [`log`](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/base.js#L513) method).
`context.report`'s argument accepts an optional `data` object to be used to provide values for [placeholders](https://eslint.org/docs/developer-guide/working-with-rules#using-message-placeholders) in messages.

This is useful when a rule provides a default error message and the opportunity to customize such error message. Using placeholders users are still able to add information that are only available at runtime to their custom error message making it more useful and understandable.

My suggestion would be to update `log`'s input object to support a similar format of message and placeholders.

Any thoughts?

### References
* [logic that applies interpolation in ESLint](https://github.com/eslint/eslint/blob/c981fb1994cd04914042ced1980aa86b68ba7be9/lib/linter/report-translator.js#L222);
* [`interpolate` function](https://github.com/eslint/eslint/blob/c981fb1994cd04914042ced1980aa86b68ba7be9/lib/linter/interpolate.js).

scalvert:
I like this idea. Happy to accept PRs (I think you have a WIP started).

rwjblue:
I don't really understand why we would do this when we could just interpolate directly within the strings when we call `.log`? 🤔

rwjblue:
Can you help me better understand the actual use case?

ventuno:
Thanks @rwjblue the idea is that if you're defining a custom error message then you can use those placeholders there as well, not only in the original error message.

For example in a case like https://github.com/ember-template-lint/ember-template-lint/issues/1923 when using a custom message I could use a placeholder such as `{{nodeTag}}` to add more details to my custom error message.

scalvert:
Correct me if I'm wrong, but I think the intent is to add optional, additional information to existing messages. Is that correct, @ventuno?

ventuno:
Thanks @scalvert. Let me try to explain with an example.

Let's start with https://github.com/ember-template-lint/ember-template-lint/issues/1923. Currently error messages are generated by the rule itself and as @rwjblue points out that can be simply achieved with template strings.

[...]

PR #1926, opened 2021-05-01
Support message placeholders for error messages

ventuno:
WIP implementation of https://github.com/ember-template-lint/ember-template-lint/issues/1925:
* I simply copy-pasted the implementation of [ESLint's `interpolate`](https://github.com/eslint/eslint/blob/c981fb1994cd04914042ced1980aa86b68ba7be9/lib/linter/interpolate.js). Alternatively we could add a dependency to `eslint` and simply import it, but I'm not sure if that's a better idea.
* I'm currently interpolating parameters inside the `log` function, which seems the best place to do it. More than happy to move the logic around if there's a better spot.


rwjblue:
> we could add a dependency to eslint and simply import it, but I'm not sure if that's a better idea.

Ya, I definitely don't want to do that (having multiple versions of eslint in a node_modules graph is a bit of a nightmare).

ventuno:
@rwjblue that was just a thought. The function itself is simple enough to be duplicated :-).

rwjblue (changes requested):
Now that I understand the goal (see issue thread), this makes **waaaaaayyyyy** more sense to me.

Some high level issues to resolve here:

* I realize that `{{` is what ESLint uses (and that is why you used it here), but since we are operating on files that have legitimate `{{` contents in them, I'd rather use something else as the delimeters. I was thinking that we might be able to use lodash template syntax instead.
* We need to beef up the rule authoring documentation to indicate this as a best practice
* We need to review existing rules, and make reasonable suggestions about how they should use placeholders (I'd suggest doing an audit, then making an issue with a big checklist)
* We should add documentation to our end-user configuration documentation with how an application might do this
* We should update our rule testing harness to make it possible to customize and assert the correct messages are emitted (likely no changes needed to the harness, but we'll need to add some tests)
* We should update to ensure that each rule's documentation has an established way to indicate the placeholders and how they work

ventuno:
@rwjblue just to confirm, my assumption for:
> We need to review existing rules, and make reasonable suggestions about how they should use placeholders (I'd suggest doing an audit, then making an issue with a big checklist)

is that only existing rules which provide a custom configuration to override the error message should be part of the audit.
Is that correct?

ventuno:
@rwjblue thanks for looking into this:


> * I realize that {{ is what ESLint uses (and that is why you used it here), but since we are operating on files that have legitimate {{ contents in them, I'd rather use something else as the delimeters. I was thinking that we might be able to use lodash template syntax instead.

Replaced to use `<%`/`%>` instead.

> * We need to beef up the rule authoring documentation to indicate this as a best practice

Updated the [Building a rule object section of the Plugin Support guide](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/plugins.md#building-a-rule-object) accordingly. Let me know if there's a better document.

> * We need to review existing rules, and make reasonable suggestions about how they should use placeholders (I'd suggest doing an audit, then making an issue with a big checklist)

See https://github.com/ember-template-lint/ember-template-lint/issues/2005, based on [my previous comment](https://github.com/ember-template-lint/ember-template-lint/pull/1926#pullrequestreview-687575602) I only audited rules which provide the configuration to customize the error message. Perhaps we can consider a default way for error messages to be overwritten, then it makes sense to audit _all_ rules for placeholders.

> * We should add documentation to our end-user configuration documentation with how an application might do this

Can you clarify this point a bit more? I'm not sure what's the action item here.

> * We should update our rule testing harness to make it possible to customize and assert the correct messages are emitted (likely no changes needed to the harness, but we'll need to add some tests)

Updating `no-restricted-invocation` was a way to test the rule testing harness, tests started failing because the `data` object was not being asserted. Error messages are properly interpolated. Let me know if anything else is needed.

> * We should update to ensure that each rule's documentation has an established way to indicate the placeholders and how they work

I updated [dev/templates/doc.md](https://github.com/ember-template-lint/ember-template-lint/blob/master/dev/templates/doc.md) to include a new section to describe placeholders offered by a new rule and updated the existing docs for `no-restricted-invocation` accordingly.

rwjblue:
Mind rebasing?

ventuno:
@rwjblue closing this in favor of https://github.com/ember-template-lint/ember-template-lint/pull/2086. It had become too hard to rebase.

PR #1903, opened 2021-04-21
Start testing under Node 16

bmish:
Just released yesterday.

PR #1927, opened 2021-05-02
Compatibility with node 16

scalvert:
Fixes #1903


rwjblue:
Since yeoman-environment@3 dropped support for Node 10, we should guard those tests. In this case, we don't need to support local development with Node 10 and the tests here are for internal functionality (which is why guarding the tests in question is fine).

scalvert:
Once #1962 is merged this should be mergable.

rwjblue:
Actually, isn't this one closeable?

scalvert:
I think we were waiting on my fix for the generator tests, which I believe you merged.

scalvert:
Hmm it looks like the only changes were to the yeoman-environment dep. Should we add tests for node 16 in this PR instead?

scalvert:
I'm going to close this and open a new PR that adds Node 16 to our test matrix.

Issue #44, opened 2016-05-12
Rule proposal: no-action-on-submit-button


oriSomething:
``` hbs
{{! bad }}
<button {{action "cancel"}}>cancel</button>

{{! good }}
<button type="button" {{action "cancel"}}>cancel</button>
```

because there should be a submit `action` on `<form>` instead, to catch any submit event.
and if a `<button>` is not a submit button, there should be `type="button"`

rwjblue:
I think there is a tad more nuance here.  As I don't think that `<button {{action 'foo'}}></button>` itself is bad, it is only bad when contained within a `<form>`, right?

oriSomething:
it's bad for fastboot and for e-reader that might tell the use it's a submit button instead of just a button

oriSomething:
for fastboot, you'll get button the will refresh the page instead of doing nothing until the JS catch

viniciussbs:
I was here searching for a rule like "button-type-attributes", similar to the existing one "img-alt-attributes", that would require the presence of the attribute `type` on `<button>` to avoid wrong form submission. I didn't know that this is also a problem even outside form tags. For me it's OK to always enforce the presence of `type`.

Turbo87:
@viniciussbs https://github.com/ember-template-lint/ember-template-lint/pull/950 implemented exactly that :)

JoaoDsv:
I'm starting to work on this one 🙋

PR #1931, opened 2021-05-04
Add new rule `no-action-on-submit-button`

JoaoDsv:
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/44.

It implements the rule proposal of `no-action-on-submit-button`.

# no-action-on-submit-button

This rule requires all `<button>` elements with a `type="submit"` attribute to not have any action.

When the `type` attribute of `<button>` elements is `submit`, the action should be on the `<form>` element instead of directly on the button.

By default, the `type` attribute of `<button>` elements is `submit`.

## Examples

```hbs
<form>
  <button type='submit' {{on 'click' this.handleClick}} />
  <button type='submit' {{action 'handleClick'}} />
  <button {{on 'click' this.handleClick}} />
  <button {{action 'handleClick'}} />
</form>
```

```hbs
// In a <form>
<form>
  <button type='button' {{on 'click' this.handleClick}} />
  <button type='button' {{action 'handleClick'}} />
  <button type='submit' />
  <button />
</form>

// Outside a <form>
<button type='submit' {{on 'click' this.handleClick}} />
<button type='submit' {{action 'handleClick'}} />
<button {{on 'click' this.handleClick}} />
<button {{action 'handleClick'}} />
```

### DIFF
--- a/docs/rule/no-action-on-submit-button.md
+++ b/docs/rule/no-action-on-submit-button.md
@@ -1,0 +1,35 @@
+# no-action-on-submit-button
+
+This rule requires all `<button>` elements with a `type="submit"` attribute to not have any click action.
+
+When the `type` attribute of `<button>` elements is `submit`, the action should be on the `<form>` element instead of directly on the button.
+
+By default, the `type` attribute of `<button>` elements is `submit`.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<button type="submit" {{on "click" this.handleClick}} />
+<button type="submit" {{action "handleClick"}} />
+<button {{on "click" this.handleClick}} />
+<button {{action "handleClick"}} />
+```
+
+This rule **allows** the following:
+
+```hbs
+<button type="button" {{on "click" this.handleClick}} />
+<button type="button" {{action "handleClick"}} />
+<button type="submit" />
+<button />
+```
+
+## Related Rules
+
+- [require-button-type](require-button-type.md)
+
+## References
+
+- [HTML spec - the button element](https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-type)
--- a/docs/rule/no-action-on-submit-button.md
+++ b/docs/rule/no-action-on-submit-button.md
@@ -1,6 +1,6 @@
 # no-action-on-submit-button
 
-This rule requires all `<button>` elements with a `type="submit"` attribute to not have any click action.
+In a `<form>`, this rule requires all `<button>` elements with a `type="submit"` attribute to not have any click action.
 
 When the `type` attribute of `<button>` elements is `submit`, the action should be on the `<form>` element instead of directly on the button.
 
@@ -11,19 +11,30 @@ By default, the `type` attribute of `<button>` elements is `submit`.
 This rule **forbids** the following:
 
 ```hbs
-<button type="submit" {{on "click" this.handleClick}} />
-<button type="submit" {{action "handleClick"}} />
-<button {{on "click" this.handleClick}} />
-<button {{action "handleClick"}} />
+<form>
+  <button type='submit' {{on 'click' this.handleClick}} />
+  <button type='submit' {{action 'handleClick'}} />
+  <button {{on 'click' this.handleClick}} />
+  <button {{action 'handleClick'}} />
+</form>
 ```
 
 This rule **allows** the following:
 
 ```hbs
-<button type="button" {{on "click" this.handleClick}} />
-<button type="button" {{action "handleClick"}} />
-<button type="submit" />
-<button />
+// In a <form>
+<form>
+  <button type='button' {{on 'click' this.handleClick}} />
+  <button type='button' {{action 'handleClick'}} />
+  <button type='submit' />
+  <button />
+</form>
+
+// Outside a <form>
+<button type='submit' {{on 'click' this.handleClick}} />
+<button type='submit' {{action 'handleClick'}} />
+<button {{on 'click' this.handleClick}} />
+<button {{action 'handleClick'}} />
 ```
 
 ## Related Rules
--- a/lib/rules/no-action-on-submit-button.js
+++ b/lib/rules/no-action-on-submit-button.js
@@ -1,6 +1,8 @@
 import Rule from './_base.js';
+import hasParentTag from '../helpers/has-parent-tag.js';
 
-const ERROR_MESSAGE = 'A `<button>` with `type="submit"` should have no click action';
+const ERROR_MESSAGE =
+  'In a `<form>`, a `<button>` with `type="submit"` should have no click action';
 
 export default class NoActionOnSubmitButton extends Rule {
   logNode({ node, message }) {
@@ -43,10 +45,15 @@ export default class NoActionOnSubmitButton extends Rule {
     }
 
     return {
-      ElementNode(node) {
+      ElementNode(node, path) {
         let { tag, attributes, modifiers } = node;
 
         if (tag !== 'button') {
+          return;
+        }
+
+        // is this button in a <form>?
+        if (!hasParentTag(path, 'form')) {
           return;
         }
 
--- a/lib/rules/no-action-on-submit-button.js
+++ b/lib/rules/no-action-on-submit-button.js
@@ -1,0 +1,75 @@
+import Rule from './_base.js';
+
+const ERROR_MESSAGE = 'A `<button>` with `type="submit"` should have no click action';
+
+export default class NoActionOnSubmitButton extends Rule {
+  logNode({ node, message }) {
+    return this.log({
+      node,
+      message,
+      line: node.loc && node.loc.start.line,
+      column: node.loc && node.loc.start.column,
+      source: this.sourceForNode(node),
+    });
+  }
+
+  visitor() {
+    function isTypeAttribute(attribute) {
+      return attribute.name === 'type';
+    }
+
+    function isOnClickModifier(modifier) {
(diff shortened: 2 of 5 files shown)


test/unit/rules/no-action-on-submit-button-test.js
```diff
@@ -0,0 +1,75 @@
...
+  ],
+
+  bad: [
+    {
+      template: '<button {{action "handleClick"}} />',
+
+      result: {
+        message: ERROR_MESSAGE,
+        source: '<button {{action "handleClick"}} />',
+        moduleId: 'layout.hbs',
```
bertdeblock, line 36: Don't think that `moduleId` is needed for these cases.

bertdeblock (commented):
Should we also check if the parent `form` already has an action defined and if so, assume that the action on the submit button is an arbitrary event? E.g. for tracking purposes. Or would this be to far-fetched?

lib/rules/no-action-on-submit-button.js
```diff
@@ -0,0 +1,73 @@
...
+
+          return (
+            path.original === 'on' &&
+            params[0].value === 'click' &&
+            params[1].type === 'PathExpression'
+          );
+        }
+
+        let typeAttribute = attributes.find((it) => isTypeAttribe(it));
+        let onClickModifier = modifiers.find((it) => isOnClickModifier(it));
```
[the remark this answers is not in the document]
bertdeblock:
```js
let typeAttribute = attributes.find(isTypeAttribe);
```
Would do the same btw.

lib/rules/no-action-on-submit-button.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const Rule = require('./base');
```
bmish, line 3:
```suggestion
const Rule = require('./_base');
```

lib/rules/no-action-on-submit-button.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const Rule = require('./_base');
+
+const ERROR_MESSAGE = 'A `<button>` with `type="submit"` should have no action';
+
+module.exports = class NoActionOnSubmitButton extends Rule {
+  logNode({ node, message }) {
+    return this.log({
```
bmish, line 9:
```suggestion
    return this.log({
      node,
```

test/unit/rules/no-action-on-submit-button-test.js
bmish: Removed.

test/unit/rules/no-action-on-submit-button-test.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
```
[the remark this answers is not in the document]
bmish: I added these tests.

lib/rules/no-action-on-submit-button.js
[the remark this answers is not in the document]
bmish: The performance shouldn't be an issue, but yes this could be optimized if desired.

lib/rules/no-action-on-submit-button.js
```diff
@@ -0,0 +1,73 @@
...
+          return path.original === 'action' && params[0].type === 'PathExpression';
+        }
+
+        function isOnClickModifier(modifier) {
+          let { path, params } = modifier;
+
+          return (
+            path.original === 'on' &&
+            params[0].value === 'click' &&
+            params[1].type === 'PathExpression'
```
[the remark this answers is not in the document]
bmish: Fixed.

bmish:
I fixed the conflicts, tests, and many (but not all) of the comments.

There's another comment here that may need to be addressed still: https://github.com/ember-template-lint/ember-template-lint/issues/44#issuecomment-218965427

bertdeblock:
If we need to check if the button has a `form` parent, we could make [this function](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/require-button-type.js#L10-L20) a shared helper and use that here as well? Though, if this rule specifically checks for `type="submit"` and buttons without a type (we also have the `require-button-type` rule to prevent this) that already seems like a good start?

bmish:
@JoaoDsv if you'd like to finish this PR up, let us know. Otherwise, I think @bertdeblock and I will try to finish it up. @bertdeblock, I'll give you access to push updates to this PR, so we can finish addressing comments including reducing false positives and improving consistency.

JoaoDsv:
Thanks for your feedbacks @bmish @bertdeblock !
Sorry for the late answer, I'm happy to try my best to address your comments

Feel free to commit at any time to help me to get this PR merged, it will be very welcome 🙂

JoaoDsv:
I addressed the remaining feedbacks ✅ 

Except reducing the false positives by checking the presence a parent `form` element. I'll handle it ASAP and let you know!

docs/rule/no-action-on-submit-button.md
```diff
@@ -0,0 +1,35 @@
...
+<button {{action "handleClick"}} />
+```
+
+This rule **allows** the following:
+
+```hbs
+<button type="button" {{on "click" this.handleClick}} />
+<button type="button" {{action "handleClick"}} />
+<button type="submit" />
+<button />
```
MelSumner, line 26: This would be disallowed by the `require-button-type` rule.

JoaoDsv: Thanks for your comment @MelSumner. Sure, but aren't rules independent from each other?

My point is: should we take an other rule into account in the unit test of this current rule?

If users decide to disable the `require-button-type` rule, then this syntax will be allowed by the current rule, because it's not its role to prevent the missing type.

What is your opinion about it?

MelSumner: So I try to write new rules with existing rules in mind; I don't want a rule I create to fail another existing rule. 

But @rwjblue might have some more clarity here. 

(PS overall I am excited about this rule!)

test/unit/rules/no-action-on-submit-button-test.js
```diff
@@ -0,0 +1,115 @@
...
+    '<div type="submit"></div>',
+    '<div type="submit" {{action this.handleClick"}}></div>',
+    '<div type="submit" {{on "click" this.handleClick}}></div>',
+  ],
+
+  bad: [
+    {
+      template: '<button {{action this.handleClick}} />',
+
+      result: {
```
bmish, line 37: We have updated rules to use snapshot testing with `verifyResults()`. Mind switching to that?

JoaoDsv: Sure, let me do it right now

bmish:
Looks good except we still need the parent `<form>` check.

bmish:
@JoaoDsv thanks for your continued work on this. Let us know when it's ready (with my last comment addressed).

JoaoDsv:
Hi @bmish @bertdeblock 

I have belatedly addressed your request about the parent `<form>` in this last commit: https://github.com/ember-template-lint/ember-template-lint/pull/1931/commits/24567e2f33a4b50a44c2f8178c0899593df9bb24. Is it what you had in mind?

ℹ️ Nicely spotted, as this also seems to solve @rwjblue's related concern: https://github.com/ember-template-lint/ember-template-lint/issues/44#issuecomment-218965427.

Would you mind having a last review and if everything looks ready, approve it?

docs/rule/no-action-on-submit-button.md
JoaoDsv: Hi @MelSumner 

I'm back at work on this rule, which I'd forgotten for a long time 🙇 

This rule is not encouraging `<button />` without (explicit) type.  It's essentially a syntax that won't make this rule fail, so I listed it in the "valid" scenarios.

Would you prefer me to remove `<button />` from these scenarios?

- The downside is that this scenario won't be clearly mentioned while likely to happen.
- The advantage is that it won't encourage people using this syntax (not allowed by `require-button-type` dedicated rule)

test/unit/rules/no-action-on-submit-button-test.js
```diff
@@ -0,0 +1,203 @@
...
+    '<button {{action this.handleClick on="click"}}/>',
+    '<button {{on "click" this.handleClick}} />',
+    '<button type="submit" {{action this.handleClick}} />',
+    '<button type="submit" {{action this.handleClick on="click"}} />',
+    '<button type="submit" {{action (fn this.someAction "foo")}} />',
+    '<button type="submit" {{on "click" this.handleClick}} />',
+    '<button type="submit" {{on "click" (fn this.addNumber 123)}} />',
+  ],
+
+  bad: [
```
bmish, line 41: Shouldn't we add a test for a deeply-nested button like this to make sure we aren't only checking the direct parent?

* `<form><div><button {{action this.handleClick}} /></div></form>`

JoaoDsv: Good idea! I added it as last scenario ✅

bmish:
@JoaoDsv just to confirm, we should enable this as `recommended` in the next major release, right?

JoaoDsv:
> @JoaoDsv just to confirm, we should enable this as `recommended` in the next major release, right?

I'd indeed suggest to:
- add the rule in the next minor release
- flag the rule as `recommended` in the next major release

By doing so, we additionally have some time to implement a `--fix` option to make the upgrade to the next major smoother. Wdyt @bmish?

bmish (approved):
Sounds good, thanks for your work to get this finished!

PR #1932, opened 2021-05-04
bumped version of todo-utils to 8.1.0 (minor)

MelSumner:
(I know dependabot but I'd like to have it sooner). Also if we could get a release after this is merged I would be grateful. Thank you!


scalvert:
Why do we need to bump this explicitly?

PR #1933, opened 2021-05-05
Adds `--clean-todo` option to provide separate functionality to cleanup stale and expired todos

MelSumner:
If merged, this PR:

- bumps todo-utils dependency to 8.1.0 (minor)
- adds `--clean-todo` flag, to provide a way to remove expired (errorDate is before today) and invalid todo files. 
- updates instructions printed to terminal
- updates documentation to include new flag

### DIFF
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -26,7 +26,13 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding todo file. You can remove this file by running `--clean-todo`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an expired todo file. You'll see this error:
+
+```bash
+error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
+```
+
+You can fix this error/remove the expired todo file by running `--clean-todo`
 
 ```bash
 ember-template-lint . --clean-todo
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -2,7 +2,7 @@
 
 Linting is a fundamental tool to help ensure the quality of a codebase. Ensuring there are as few linting errors as possible (ideally 0), is a useful measure of a baseline of code hygiene.
 
-It's common to leverage linting not just for syntax adherence, but also to direct developers to employ standardized patterns. As such, it's a fairly routine activity to introduce new lint rules into a codebase. This introduction, while necessary, can cause unintended friction, such as:
+It's common to leverage linting not only for syntax adherence, but also to direct developers to employ standardized patterns. As such, it's a fairly routine activity to introduce new lint rules into a codebase. This introduction, while necessary, can cause unintended friction, such as:
 
 - new lint errors being introduced where they previously didn't exist
 - causing unintended delays to shipping new fixes and features
@@ -26,10 +26,10 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding todo file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding todo file. You can remove this file by running `--clean-todo`
 
 ```bash
-ember-template-lint . --fix
+ember-template-lint . --clean-todo
 ```
 
 ### Configuring Due Dates
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -26,13 +26,13 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an expired todo file. You'll see this error:
+If an error is fixed manually, `ember-template-lint` will let you know that there's a stale todo file. You'll see this error:
 
 ```bash
-error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
+error  Todo violation passes no-vague-rules rule. Please run `ember-template-lint /path/to/file.hbs --clean-todo` to remove this todo from the todo list.
 ```
 
-You can fix this error/remove the expired todo file by running `--clean-todo`
+You can fix this error/remove the stale todo file by running `--clean-todo`
 
 ```bash
 ember-template-lint . --clean-todo
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -42,15 +42,11 @@ module.exports = class TodoHandler {
       writeTodoOptions
     );
 
-    if (itemsToRemoveFromTodos.size > 0) {
+    if (remove.size > 0) {
       if (shouldFix) {
-        applyTodoChanges(
-          getTodoStorageDirPath(this._workingDir),
-          new Map(),
-          itemsToRemoveFromTodos
-        );
+        applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), remove);
       } else {
-        for (const [, todo] of itemsToRemoveFromTodos) {
+        for (const [, todo] of remove) {
           results.push({
             rule: 'invalid-todo-violation-rule',
             message: `Todo violation passes \`${todo.ruleId}\` rule. Please run \`ember-template-lint ${todo.filePath} --clean-todo\` to remove this todo from the todo list.`,
@@ -63,11 +59,11 @@ module.exports = class TodoHandler {
       }
     }
 
-    if (expiredTodos.size > 0) {
-      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expiredTodos);
+    if (expired.size > 0) {
+      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expired);
     }
 
-    for (const todo of [...existingTodos.values()]) {
+    for (const todo of [...stable.values()]) {
       let severity = getSeverity(todo);
 
       if (severity === ERROR_SEVERITY) {
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -42,6 +42,10 @@ module.exports = class TodoHandler {
       writeTodoOptions
     );
 
+    if (expired.size > 0) {
+      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expired);
+    }
+
     if (remove.size > 0) {
       if (shouldFix) {
         applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), remove);
@@ -57,10 +61,6 @@ module.exports = class TodoHandler {
           });
         }
       }
-    }
-
-    if (expired.size > 0) {
-      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expired);
     }
 
     for (const todo of [...stable.values()]) {
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -36,8 +36,7 @@ module.exports = class TodoHandler {
 
     let existingTodoFiles = await readTodosForFilePath(this._workingDir, writeTodoOptions.filePath);
 
-    // this should represent the batches - add, remove, stable, expired
-    let [, itemsToRemoveFromTodos, existingTodos, expiredTodos] = await getTodoBatches(
+    let [, remove, stable, expired] = await getTodoBatches(
       buildTodoData(this._workingDir, this._buildResult(results)),
       existingTodoFiles,
       writeTodoOptions
@@ -65,19 +64,7 @@ module.exports = class TodoHandler {
     }
 
     if (expiredTodos.size > 0) {
-      if (shouldFix) {
-        applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expiredTodos);
-      } else {
-        for (const [, todo] of expiredTodos) {
-          results.push({
-            rule: 'expired-todo-rule',
-            message: `Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.`,
-            filePath: todo.filePath,
-            severity: 2,
-            isFixable: true,
-          });
-        }
-      }
+      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expiredTodos);
     }
 
     for (const todo of [...existingTodos.values()]) {
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -36,7 +36,8 @@ module.exports = class TodoHandler {
 
     let existingTodoFiles = await readTodosForFilePath(this._workingDir, writeTodoOptions.filePath);
 
-    let [, itemsToRemoveFromTodos, existingTodos] = await getTodoBatches(
+    // this should represent the batches - add, remove, stable, expired
+    let [, itemsToRemoveFromTodos, existingTodos, expiredTodos] = await getTodoBatches(
       buildTodoData(this._workingDir, this._buildResult(results)),
       existingTodoFiles,
       writeTodoOptions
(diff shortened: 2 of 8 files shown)


docs/todos.md
```diff
@@ -26,10 +26,16 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding todo file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an expired todo file. You'll see this error:
```
scalvert, line 29: It's not `expired`, it's `stale`. 

```suggestion
If an error is fixed manually, `ember-template-lint` will let you know that there's a stale todo file. You'll see this error:
```

docs/todos.md
```diff
@@ -26,10 +26,16 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
...
 
 ```bash
-ember-template-lint . --fix
+error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
+```
+
+You can fix this error/remove the expired todo file by running `--clean-todo`
```
scalvert, line 35:
```suggestion
You can fix this error/remove the stale todo file by running `--clean-todo`
```

docs/todos.md
```diff
@@ -26,10 +26,16 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
 ember-template-lint . --include-todo
 ```
 
-If an error is fixed manually, `ember-template-lint` will let you know that there's an outstanding todo file. You can remove this file by running `--fix`
+If an error is fixed manually, `ember-template-lint` will let you know that there's an expired todo file. You'll see this error:
 
 ```bash
-ember-template-lint . --fix
+error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
```
scalvert, line 32:
```suggestion
error  Todo violation passes no-vague-rules rule. Please run `ember-template-lint /path/to/file.hbs --clean-todo` to remove this todo from the todo list.
```

lib/-private/todo-handler.js
```diff
@@ -36,7 +36,8 @@ module.exports = class TodoHandler {
 
     let existingTodoFiles = await readTodosForFilePath(this._workingDir, writeTodoOptions.filePath);
 
-    let [, itemsToRemoveFromTodos, existingTodos] = await getTodoBatches(
+    // this should represent the batches - add, remove, stable, expired
+    let [, itemsToRemoveFromTodos, existingTodos, expiredTodos] = await getTodoBatches(
```
scalvert, line 40: Let's just call this what it is rather than have a comment here (which doesn't serve much purpose).

```suggestion
    let [, remove, stable, expired] = await getTodoBatches(
```

lib/-private/todo-handler.js
```diff
@@ -63,6 +64,22 @@ module.exports = class TodoHandler {
...
+            rule: 'expired-todo-rule',
+            message: `Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.`,
+            filePath: todo.filePath,
+            severity: 2,
+            isFixable: true,
+          });
+        }
+      }
+    }
+
```
scalvert, line 82: I think we should evaluate this before the above conditional. Additionally, we do not want to push a new error for expired, we want to "opaquely" remove the expireds.

```suggestion
    if (expiredTodos.size > 0) {
      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expiredTodos);
    }

```

MelSumner:
A little bit stuck on `shouldRemove` being related to config options instead of the `remove` batch, and can't quite figure out where the expire batches should go. I'll need to pair or have some strong feedback (or PRs if more convenient) to ship this fix.

lib/linter.js
```diff
@@ -260,6 +260,7 @@ class Linter {
     return currentSource;
   }
 
+  // checks to see if config is being overridden so it doesn't remove something that should stay
```
scalvert, line 263: I'm not sure what this comment means per se, and also don't think it's necessary. Can we remove it?

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+
+      expect(result.exitCode).toEqual(1);
+      expect(result.stdout).toMatchInlineSnapshot(`
+        "app/templates/require-button-type.hbs
+          1:0  error  All \`<button>\` elements should have a valid \`type\` attribute  require-button-type
+          -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
+
+        ✖ 2 problems (2 errors, 0 warnings)
+          2 errors and 0 warnings potentially fixable with the \`--fix\` option."
+      `);
```
scalvert, line 381: This snapshot will need to be updated once you remove the additional error you added for expired.

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
       expect(todoDirs).toHaveLength(0);
     });
 
+    it('errors if a todo item has expired when running without params', async function () {
```
scalvert, line 343:
```suggestion
    it('removes expired todo file if a todo item has expired when running without params', async function () {
```

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+        "app/templates/require-button-type.hbs
+          1:0  error  All \`<button>\` elements should have a valid \`type\` attribute  require-button-type
+          -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
+
+        ✖ 2 problems (2 errors, 0 warnings)
+          2 errors and 0 warnings potentially fixable with the \`--fix\` option."
+      `);
+
+      // run clean-todo, and expect that this will delete the expired todo item
+      await run(['app/templates/require-button-type.hbs', '--clean-todo']);
```
scalvert, line 384: This can be removed, since we're _always_ removing expired.

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+
+      // run normally again and expect only the error
+      result = await run(['.']);
```
scalvert, line 387: This can be removed, since we're _always_ removing

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+      let todoDirs = fs.readdirSync(getTodoStorageDirPath(project.baseDir));
+
+      expect(result.exitCode).toEqual(1);
+      expect(result.stdout).toMatchInlineSnapshot(`
+      "app/templates/require-button-type.hbs
+        1:0  error  All \`<button>\` elements should have a valid \`type\` attribute  require-button-type
+
+      ✖ 1 problems (1 errors, 0 warnings)
+        1 errors and 0 warnings potentially fixable with the \`--fix\` option."
+    `);
```
scalvert, line 398: This snapshot will need to be updated once you remove the additional error you added for expired.

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+      expect(todoDirs).toHaveLength(0);
+
+      // run fix, and expect that this will fix the auto-fixable error
+      await run(['app/templates/require-button-type.hbs', '--fix']);
+
+      // run normally again and expect no errors and no todos
+      result = await run(['.']);
```
scalvert, line 405: 🔪

test/acceptance/todo-cli-test.js
```diff
@@ -340,6 +340,75 @@ describe('todo usage', () => {
...
+
+      expect(result.exitCode).toEqual(0);
+      expect(result.stdout).toEqual('');
+      expect(todoDirs).toHaveLength(0);
```
scalvert, line 409: 🔪

test/acceptance/todo-cli-test.js
```diff
@@ -1083,7 +1152,7 @@ describe('todo usage', () => {
         `);
     });
 
-    it('should set todo to error if errorDate has expired via config', async function () {
+    it('should set todo to error and provide expired message if errorDate has expired via config', async function () {
```
scalvert, line 1155: You'll need to revert this test name.

test/acceptance/todo-cli-test.js
```diff
@@ -1113,8 +1182,10 @@ describe('todo usage', () => {
       expect(result.stdout).toMatchInlineSnapshot(`
           "app/templates/application.hbs
             1:5  error  Non-translated string used  no-bare-strings
+            -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
 
-          ✖ 1 problems (1 errors, 0 warnings)"
+          ✖ 2 problems (2 errors, 0 warnings)
+            1 errors and 0 warnings potentially fixable with the \`--fix\` option."
```
scalvert, line 1188: You'll need to revert this snapshot.

test/acceptance/todo-cli-test.js
```diff
@@ -1145,12 +1216,14 @@ describe('todo usage', () => {
...
```
scalvert, line 1222: You'll need to revert this snapshot.

test/acceptance/todo-cli-test.js
```diff
@@ -1145,12 +1216,14 @@ describe('todo usage', () => {
...
         `);
     });
 
-    it('should set todo to error if errorDate has expired via option', async function () {
+    it('should set todo to error and display an expired todo message if errorDate has expired via option', async function () {
```
scalvert, line 1226: You'll need to revert this test name.

test/acceptance/todo-cli-test.js
```diff
@@ -1176,12 +1249,14 @@ describe('todo usage', () => {
       expect(result.stdout).toMatchInlineSnapshot(`
           "app/templates/application.hbs
             1:5  error  Non-translated string used  no-bare-strings
+            -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
 
-          ✖ 1 problems (1 errors, 0 warnings)"
+          ✖ 2 problems (2 errors, 0 warnings)
+            1 errors and 0 warnings potentially fixable with the \`--fix\` option."
```
scalvert, line 1255: You'll need to revert this snapshot.

test/acceptance/todo-cli-test.js
```diff
@@ -1176,12 +1249,14 @@ describe('todo usage', () => {
...
         `);
     });
 
-    it('should set todo to error if both warnDate and errorDate have expired via config', async function () {
+    it('should set todo to error and display an expired todo message if both warnDate and errorDate have expired via config', async function () {
```
scalvert, line 1259: You'll need to revert this test name.

test/acceptance/todo-cli-test.js
```diff
@@ -1212,12 +1287,14 @@ describe('todo usage', () => {
       expect(result.stdout).toMatchInlineSnapshot(`
           "app/templates/application.hbs
             1:5  error  Non-translated string used  no-bare-strings
+            -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
 
-          ✖ 1 problems (1 errors, 0 warnings)"
+          ✖ 2 problems (2 errors, 0 warnings)
+            1 errors and 0 warnings potentially fixable with the \`--fix\` option."
```
scalvert, line 1293: You'll need to revert this snapshot.

test/acceptance/todo-cli-test.js
```diff
@@ -1212,12 +1287,14 @@ describe('todo usage', () => {
...
         `);
     });
 
-    it('should set todo to error if both warnDate and errorDate have expired via options', async function () {
+    it('should set todo to error and display expired todo message if both warnDate and errorDate have expired via options', async function () {
```
scalvert, line 1297: You'll need to revert this test name.

test/acceptance/todo-cli-test.js
```diff
@@ -1243,8 +1320,10 @@ describe('todo usage', () => {
       expect(result.stdout).toMatchInlineSnapshot(`
           "app/templates/application.hbs
             1:5  error  Non-translated string used  no-bare-strings
+            -:-  error  Expired todos exist. Please run \`ember-template-lint --clean-todo\` to remove expired todos.  expired-todo-rule
```
scalvert, line 1323: You'll need to revert this snapshot.

scalvert (changes requested):

lib/linter.js
MelSumner: yes, it's just there for me to remember. I'll remove it before we ship it.

test/acceptance/todo-cli-test.js
MelSumner: This line is in other tests as well, should I remove it?

lib/-private/todo-handler.js
```diff
@@ -63,7 +59,11 @@ module.exports = class TodoHandler {
       }
     }
 
-    for (const todo of [...existingTodos.values()]) {
+    if (expired.size > 0) {
+      applyTodoChanges(getTodoStorageDirPath(this._workingDir), new Map(), expired);
+    }
```
scalvert, line 64: I think we should move this _above_ the condition above. Can we do that?

scalvert (changes requested):

test/acceptance/todo-cli-test.js
MelSumner: I think this removes the only place where the flag was tested. We okay with that?

MelSumner:
I think we should also discuss if we _really_ want to auto-delete expired todos, instead of the user controlling that themselves. 

The selling point of the todos was that teams could [plan when to do the work](https://noti.st/melsumner/i9uja6/continuous-accessibility#sjlGifr). By auto-deleting the expired todos, are we overriding their plans?

scalvert:
If they're expired, they're no longer useful, right? What are your specific concerns?

docs/todos.md
```diff
@@ -26,10 +26,16 @@ If you want to see todos as part of `ember-template-lint`'s output, you can incl
...
 
 ```bash
-ember-template-lint . --fix
+error  Todo violation passes no-vague-rules rule. Please run `ember-template-lint /path/to/file.hbs --clean-todo` to remove this todo from the todo list.
+```
+
+You can fix this error/remove the stale todo file by running `--clean-todo`
+
+```bash
+ember-template-lint . --clean-todo
```
scalvert, line 38: Based on the example above, I think we should we replace `.` with `/path/to/file.hbs` so that it's inline with the example. You can additionally add a blurb about how to clean all todos for all files.

test/acceptance/todo-cli-test.js
scalvert: But the invocation on line 371 removes expired todos, right? We're always removing expired todos, and you've explicitly created a todo that's expired, right? So, after line 371, this todo that we want to check for no longer exists, right? 

Why do we need to run `--clean-todo`?

test/acceptance/todo-cli-test.js
scalvert: Again, same as I said above. We've already removed the expired todo, which is what this test is for, right?

scalvert (changes requested):

MelSumner:
> If they're expired, they're no longer useful, right? What are your specific concerns?

They're expired in that they now are back in the error state, but it's an explicit forcing function that the user has to go _do something_ about it. Doing something would be the thing that they decide to do- remove the todo and re-create it, use it as a timer, etc. But the "something" seems like it would be good to keep in the user's control instead of us choosing it for them. 

When I think about it more, it seems inconsistent with the rest of how ember-template-works in general. We give options but we let them decide.

scalvert:
I'm still of the mind that you explicitly created a todo for a duration, that duration has expired, therefore the todo has served its purpose and should be deleted. Another way to think about it is, that file will not be used for anything else ever again - it's orphaned. We should be actively cleaning up these orphaned files.

rwjblue:
> They're expired in that they now are back in the error state, but it's an explicit forcing function that the user has to go do something about it.

Hmm, remember that when in that expired state (excluding the bug that this is fixing where expired todo items were in the "stable" bucket from `getTodoBatches`) the todo itself is completely unused. Whatever in-app error was being masked by the todo is now directly shown as an error in the normal linting flow.

Once you have the error back, you basically have two choices:

* Fix the actual problem 
* Create another todo with an updated decay timeline

As far as I can tell, those are your **only** options. The thing that is important here is that for **all cases** the old expired todo file is no longer useful, and will have to be deleted.

scalvert:
Yep, agreed. This is very much in bug fix + enhancement territory as far as I'm concerned. 

Thanks for doing this, @MelSumner.

scalvert (approved):

rwjblue (approved):

PR #1934, opened 2021-05-06
Fix false positives in `deprecated-inline-view-helper` rule

dwickern:
These usages of `view` should be allowed since they won't invoke the deprecated `{{view}}` helper.


bmish:
@dwickern can you fix the conflicts? Then I can merge this.

bmish (approved):
Thanks!

PR #1936, opened 2021-05-07
no-unknown-arguments-for-builtin-components: allow `@indeterminate`

dwickern:
`@indeterminate` needs to be passed as an argument:
https://ember-twiddle.com/843748d9d4e8bde8ae56824d28a21024?openFiles=templates.application%5C.hbs%2C


rwjblue:
Hmmm. Based on the implementation in Ember itself, folks should be passing `indeterminate` as an attribute.

https://github.com/emberjs/ember.js/blob/e53d6bb4bbb5230829eb589fdba2388de3cd28cf/packages/%40ember/-internals/glimmer/lib/components/input.ts#L273-L284

^ these are the remaining supported arguments.

Notice that `indeterminate` is listed here:

https://github.com/emberjs/ember.js/blob/e53d6bb4bbb5230829eb589fdba2388de3cd28cf/packages/%40ember/-internals/glimmer/lib/components/input.ts#L335-L336

Which then goes through:

https://github.com/emberjs/ember.js/blob/e53d6bb4bbb5230829eb589fdba2388de3cd28cf/packages/%40ember/-internals/glimmer/lib/components/abstract-input.ts#L221

---

Am I missing something?

dwickern:
This appears to be changed in Ember 3.27.0

In 3.26.1 and earlier, `@indeterminate` should be passed as an argument:
```hbs
{{! displays the indeterminate state }}
<Input @type="checkbox" @indeterminate={{true}}/>

{{! does NOT display the indeterminate state }}
<Input @type="checkbox" indeterminate={{true}}/>
```

In 3.27.0 and later, `indeterminate` is preferred but `@indeterminate` still works with a deprecation warning:
```hbs
{{! displays the indeterminate state, logs a deprecation warning }}
<Input @type="checkbox" @indeterminate={{true}}/>

{{! displays the indeterminate state }}
<Input @type="checkbox" indeterminate={{true}}/>
```

Given the new deprecation warnings, I'd say this lint rule is no longer relevant since 3.27

gorner:
Currently working on an app that's still on v3.28 (slowly working on upgrading) and noticed this being caught by TemplateLint – but `indeterminate={{...}}` without the `@` isn't working. Is there something I'm missing, a different dependency that could be affecting things etc.?

gorner:
Just following up on the comment above (in case there's anyone else that comes across this) – in the app in question (which is still running v3.28, at least for a bit longer), `indeterminate` now works as expected as a regular attribute without the `@`. I'm not sure I can pinpoint the exact fix, but at the time of my previous comment the app was still on the `classic` edition and had ember-cli-htmlbars v6 in the build tree from various addons (in addition to v7).

lifeart:
@gorner, thank you for PR! Since in latest ember versions all ok, I'm closing this PR.

PR #1945, opened 2021-05-10
Fixes `--clean-todo` flag to actually do the cleaning

scalvert:
The `--clean-todo` flag was missing from the todo-handler, resulting in this flag effectively being a noop. This fix ensures that both `--clean-todo` and `--fix` work for cleaning todos, though we prefer users run the narrowly scoped `--clean-todo`, as it changes less and results in more predictable effects.


MelSumner (approved):
Confirmed that this works to resolve the issue (pulled the branch down, used yarn link, and went through the workflow).

PR #1947, opened 2021-05-11
Moves expired todo removal under the `--clean-todo` flag

scalvert:
After implementing and testing out the `--clean-todo` flag in some of our applications, we determined that proactively removing expired todos _even when just running `ember-template-lint .`_ would result in confusion for users when files are deleted through the normal invocation of the CLI. As such, we've chosen to move the expired todo removal under the `--clean-todo` flag, which would clean both todos for removal and those expired when using that explicit command.

### DIFF
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -42,13 +42,13 @@ module.exports = class TodoHandler {
       writeTodoOptions
     );
 
-    if (expired.size > 0) {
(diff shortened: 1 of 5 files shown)


scalvert:
Unsure why CI is failing for linting since locally it's passing.

MelSumner (approved):
I've also tested this against an app to confirm that it behaves as desired.

Issue #1508, opened 2020-09-09
Update `no-invalid-interactive` for additional use cases


MelSumner:
Update the `no-invalid-interactive` rule to cover the following use case: 

Interactive elements marked with `role="presentation"` or `role="none"` are invalid.

This is to support WCAG Success Criteria 1.3.1- Information and Relationships. https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html

rajasegar:
I think this would make sense for updating the `no-invalid-role` rule to not allow `role="presentation"` or `role="none"` for interactive elements rather than `no-invalid-interactive` because the intention here is to catch the invalid roles. In addition, this would also keep the conceptual integrity of the rules to be more consistent with what they are intended to do.
I am thinking of adding more additional tests to `no-invalid-role` like:
```
<button type="button" role="presentation">Hello</button>
<button type="button" role="none">Hello</button>
```
@MelSumner Please let me know your thoughts and would appreciate some examples also.

MelSumner:
@rajasegar I'm also fine with that approach!

faith-or:
Adding extra tests in this PR: https://github.com/ember-template-lint/ember-template-lint/pull/1972

PR #1972, opened 2021-05-30
Adding extra tests to no-invalid-role rule

faith-or:
This PR adds additional tests for `no-invalid-role` rule and an additional example in the doc.

 Reference: https://github.com/ember-template-lint/ember-template-lint/issues/1508


scalvert:
Thanks for the addition!

bmish:
FYI we shouldn't have labeled this as an "enhancement" since it does not add user-facing improvements (it's a great change but should only have been labeled "documentation").

Issue #1578, opened 2020-10-22
Rule Request: `require-presentational-children`


MelSumner:
There are roles that require all children to be presentational. 

While browsers are required to automatically apply `role="presentation"` to all descendant elements of any element with a role that cannot support semantic children, it would be useful to have this in a linting rule, so that the developer can receive this feedback in their developer workflow instead of after an accessibility audit has occurred or a user has filed a bug. 

Supporting document: https://w3c.github.io/aria-practices/#children_presentational

Good: 
```
<li role="tab">Tab title</li>
```

Bad:
```
<li role="tab"><h3>Tab Title</h3></li>
```

faith-or:
Here's my attempt at this: https://github.com/ember-template-lint/ember-template-lint/pull/1982 

If @MelSumner  can give me some feedback. Not sure if I have the full list of non-semantic children.

PR #1982, opened 2021-06-01
Add new rule `require-presentational-children`

faith-or:
This PR adds the new rule `require-presentational-children` 

**Background**
Certain roles require all children to be presentational. Browsers are required to automatically apply role="presentation" to all descendant elements of any element with a role that cannot support semantic children. We add this rule so developers can get early feedback.

Supporting document: https://w3c.github.io/aria-practices/#children_presentational

Good:
```
<li role="tab">Tab title</li>
```
Bad:
```
<li role="tab"><h3>Tab Title</h3></li>
```
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/1578 

Testing:
```
Test Suites: 123 passed, 123 total
Tests:       6 skipped, 6192 passed, 6198 total
Snapshots:   65 passed, 65 total
Time:        20.38 s
Ran all test suites.
✨  Done in 24.03s.
```


lib/config/recommended.js
```diff
@@ -65,6 +65,7 @@ module.exports = {
     'require-iframe-title': 'error',
     'require-input-label': 'error',
     'require-lang-attribute': 'error',
+    'require-presentational-children': 'error',
```
MelSumner, line 68:
```suggestion
    'require-presentational-children': 'warn',
```
Let's change this to warn so it won't break people. 

Also please add this to the lib/config/a11y.js:

```
'require-presentational-children': 'error',
```

lib/rules/require-presentational-children.js
```diff
@@ -0,0 +1,101 @@
...
+  'progressbar',
+  'radio',
+  'scrollbar',
+  'separator',
+  'slider',
+  'switch',
+  'tab',
+]);
+
+const NON_SEMANTIC_TAGS = new Set([
```
MelSumner, line 54: We could do it this way, but I wonder if we should make a helper that has all of the semantic elements in it, and a helper that has all valid roles for author use it it. That way we can standardize across the app. 

cc'ing @Turbo87  and @rwjblue for input here.

test/unit/rules/require-presentational-children-test.js
```diff
@@ -0,0 +1,82 @@
...
+    '<div role="article"><h2>Hello</h2></div>',
+    `
+    <ul role="tablist">
+      <li role="presentation">
+        <a role="tab" href="#">Tab 1</a>
+      </li>
+    </ul>
+    `,
+    `
+    <ul role="presentation">
```
MelSumner, line 26: Let's take this one out, since we don't want our examples to be technically correct for the rule but otherwise promote something that is not best practice.

lib/rules/require-presentational-children.js
faith-or: Right like a central const file. 

I chose to check if the tag is non-semantic vs semantic because there are a lot of semantic tags and the list can grow.

test/unit/rules/require-presentational-children-test.js
faith-or: 👍

lib/config/recommended.js
faith-or: 👌

MelSumner (approved):
This looks good to me. @rwjblue what do you think?

lib/config/recommended.js
bertdeblock: Even though this won't break peoples lint build, this doesn't seem like a common thing to do for new rules. Personally I would prefer not to have new warnings pop up when updating ember-template-lint to the next minor.

bmish:
@faith-or can you rebase? I think we can merge this soon.

docs/rule/require-presentational-children.md
```diff
@@ -0,0 +1,54 @@
+# require-presentational-children
+
+:white_check_mark: The `extends: 'recommended'` property in a configuration file enables this rule.
```
bmish, line 3:
```suggestion
✅ The `extends: 'recommended'` property in a configuration file enables this rule.
```

lib/rules/require-presentational-children.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const { flatMap } = require('../helpers/javascript');
+const Rule = require('./base');
```
bmish, line 5:
```suggestion
const Rule = require('./_base');
```

lib/rules/require-presentational-children.js
```diff
@@ -0,0 +1,101 @@
...
+        if (!ROLES_THAT_CANNOT_SUPPORT_SEMANTIC_CHILDREN.has(roleAttrValue)) {
+          return;
+        }
+
+        const children = getAllElementNodeDescendants(node);
+
+        for (const child of children) {
+          // If it's not a non semantic tag, it's semantic
+          if (!NON_SEMANTIC_TAGS.has(child.tag)) {
+            this.log({
```
bmish, line 88:
```suggestion
            this.log({
              node,
```

bmish (approved):
Thanks!

faith-or:
> @faith-or can you rebase? I think we can merge this soon.

Sorry! Thanks for merging this in for me @bmish !!

PR #1964, opened 2021-05-25
Updated for todo-util configuration changes

MelSumner:
## Summary
If merged, this PR helps to resolve #1880 by consuming the changes made by https://github.com/ember-template-lint/ember-template-lint-todo-utils/pull/213. 

Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

## Tests
When run locally with the changes in the todo-utils PR, the tests for this branch all pass.

Issue #1880, opened 2021-03-29
[todos] Add days to warn/error configuration in lint todo config


scalvert:
Currently there's no serialized source of truth for configurations for days to warn/error. When creating todos, those days are configured in an ad hoc manner. We should extend our current configuration to include specific configurations for days to warn/error, per rule.

Example:

```
{
  "lintTodo": {
    "decayDays": {
      "warn": 5,
      "error": 10
    },
    "decayDaysByRule": {
      "link-href-attributes": { warn: 30, error: 90 },
      ...
    } 
  }
}
```

MelSumner:
Do you think that we should consider moving this to the lint config file instead of the package.json file? I can imagine this getting a bit unweildy to have rule settings in two places. What do you think?

rwjblue:
We originally chose to use package.json because the same settings would be read by multiple tools (not just ember-template-lint). I do agree that we might want to have a separate config file, but it can't (directly) be `.template-lintrc.js`.

MelSumner:
Discussion with folks who use ember-template-lint has revealed a preference for a separate config file so as to not bulk up package.json, so that's the route I'll pursue.

scalvert:
Thanks, @MelSumner. Any information/updates on this?

scalvert:
Proposed changes:

1. Support both configuration styles:
    - package.json configuration (supports simple config where you only supply `defaultDecayDays`
    - .lint-todorc.js configuration (supports larger, more complicated configuration)

2. Expand configuration as per original suggestion, with the exception of further qualifying the default decay days property key to disambiguate:

    ```
    {
      "lintTodo": {
        "defaultDecayDays": { "warn": 5, "error": 10 },
        "decayDaysByRule" : {
          "link-href-attributes": { warn: 30, error: 90 },
          ...
        }
      }
    }
    ```

scalvert:
On further discussions offline with @MelSumner, we realized we need to further subdivide the config into `engine`, like so:

```
{
  "eslint" {
    // ... TodoConfig
  },
  "ember-template-lint": {
    // ... TodoConfig
  }
}
```

MelSumner:
Also , supporting per-rule config via environment variables is considered out of scope for the time being.

PR #1993, opened 2021-06-08
feat(todos): Adds support for configuring todo decay days by rule

scalvert:
Implements #1880 and supersedes #1964.

This change adds support for using a config file, `.lint-todorc.js`. This config is a superset of the old config, which was simply:

```json
{
  "lintTodo": {
    "daysToDecay": {
      "warn": 5,
      "error": 10
    }
  }
}
```

The new structure adds a level of nesting, allowing you to specify which `engine` the config section applies to:

```json
 {
   "lintTodo": {
     "ember-template-lint": {
       "daysToDecay": {
         "warn": 5,
         "error": 10
       },
       "daysToDecayByRule": {
         "no-bare-strings": { "warn": 10, "error": 20 }
       }
     }
   }
 }
```
In addition, you can now configure `daysToDecay` by ruleID, allowing more fine-grained configuration of decay dates. It also provides a single source of truth for configuring those dates.

As of this change, we now support

1. Legacy configs, such as those in the first example above
2. Configuration in `package.json`
3. Configuration in `.lint-todorc.js`

### DIFF
--- a/test/acceptance/todo-cli-test.js
+++ b/test/acceptance/todo-cli-test.js
@@ -392,49 +392,6 @@ describe('todo usage', () => {
       expect(todoDirs).toHaveLength(0);
     });
 
-    it('removes expired todo file if a todo item has expired when running with --clean-todo', async function () {
-      project.setConfig({
-        rules: {
-          'require-button-type': true,
-        },
-      });
-
-      project.write({
-        app: {
-          templates: {
-            'require-button-type.hbs': '<button>Check Expiration</button>',
-          },
-        },
-      });
-
-      project.setLegacyTodoConfig({
-        error: 5,
-      });
-
-      // generate todo based on existing error
-      await run(['.', '--update-todo'], {
-        // change the date so errorDate is before today
-        env: {
-          TODO_CREATED_DATE: subDays(new Date(), 10).toJSON(),
-        },
-      });
-
-      // run normally and expect the issue to be back in the error state and there to be no todo
-      let result = await run(['.', '--clean-todo']);
-
-      let todoDirs = fs.readdirSync(getTodoStorageDirPath(project.baseDir));
-
-      expect(result.exitCode).toEqual(1);
-      expect(result.stdout).toMatchInlineSnapshot(`
-        "app/templates/require-button-type.hbs
-          1:0  error  All \`<button>\` elements should have a valid \`type\` attribute  require-button-type
-
-        ✖ 1 problems (1 errors, 0 warnings)
-          1 errors and 0 warnings potentially fixable with the \`--fix\` option."
-      `);
-      expect(todoDirs).toHaveLength(0);
-    });
-
     it('outputs empty summary for no todos or errors', async function () {
       project.setConfig({
         rules: {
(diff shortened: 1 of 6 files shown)


test/helpers/fake-project.js
```diff
@@ -87,16 +87,50 @@ module.exports = class FakeProject extends FixturifyProject {
...
       lintTodo: {
-        daysToDecay: todoConfig,
+        daysToDecay,
       },
     });
 
     this.writeSync();
   }
 
+  setPackageJsonTodoConfig(engine, daysToDecay, daysToDecayByRule) {
```
MelSumner, line 100: praise: I like these new naming conventions, it makes everything a little easier for me to reason about.

test/acceptance/todo-cli-test.js
```diff
@@ -764,583 +699,741 @@ describe('todo usage', () => {
...
 
-      const todos = [...(await readTodos(project.baseDir)).values()];
+              ✖ 1 problems (1 errors, 0 warnings)
+                1 errors and 0 warnings potentially fixable with the \`--fix\` option."
+            `);
+          expect(todoDirs).toHaveLength(0);
+        });
 
-      expect(result.exitCode).toEqual(0);
+        it('should error if daysToDecay.error is less than daysToDecay.warn in package.json', async function () {
```
MelSumner, line 783:
```suggestion
        it('should error if daysToDecay.error is less than daysToDecay.warn in the config', async function () {
```

**question:** since this group of tests loops through legacy todo configuration, package.json todo configuration, and .lint-todorc.js todo configuration, should these tests be updated to refer to "the config" rather than `package.json`?

MelSumner:
@scalvert confirmed that it throws the conflicting configuration file error if lint todo configurations exist in both `package.json` and `.lint-todorc.js`: 

image

test/helpers/fake-project.js
scalvert: Agree, consistency is good.

scalvert:
@MelSumner thanks for checking that. That's not a very nice user experience. We can do better by adding a util in todo-utils that checks for this case, and allows consumers to validate early, and fail with a nicer formatted message. WDYT?

test/acceptance/todo-cli-test.js
scalvert: Agreed!

rwjblue:
https://github.com/ember-template-lint/ember-template-lint-todo-utils/pull/242 lets us fix that issue (still needs tweaks here once its released to actually do it though).

test/acceptance/todo-cli-test.js
rwjblue: Small verbal nit-pick here: I don't want to consider the prior config as "legacy". I'd like to think of it as short hand (I'd expect most folks to use this syntax just to have the same values for all engine types), but if you _need_ the more detailed "engine + rule" specificity you can drop down and do the new syntax.

scalvert: That's fair. We can change this.

scalvert:
@MelSumner @rwjblue I updated the code to support validating the todo config beforehand, so it's cleaner output.

scalvert:
@rwjblue any feedback on this one? It'd be nice to get it merged sooner rather than later.

MelSumner (approved):
Looking forward to this feature! 👍

rwjblue (approved):

PR #1994, opened 2021-06-10
Fix `require-input-label` to allow inputs within a `<label>` to have an `id` attribute

MelSumner:
If merged, this PR resolves the issue where a form element nested inside of a `<label>` element was throwing a invalid "multiple labels" error if the form element also had an `id` attribute. 

Details: 
- Added condition to allow `<label>text<input id="foo" /></label>`.
- Updated tests to include additional valid scenarios.

Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

### DIFF
--- a/test/unit/rules/require-input-label-test.js
+++ b/test/unit/rules/require-input-label-test.js
@@ -94,7 +94,7 @@ generateRuleTests({
       },
     },
     {
-      template: '<label><input /></label>',
+      template: '<label><input></label>',
       result: {
         message: ERROR_MESSAGE,
         line: 1,
--- a/test/unit/rules/require-input-label-test.js
+++ b/test/unit/rules/require-input-label-test.js
@@ -99,7 +99,7 @@ generateRuleTests({
         message: ERROR_MESSAGE,
         line: 1,
         column: 7,
-        source: '<input />',
+        source: '<input>',
       },
     },
     {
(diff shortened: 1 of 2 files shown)


lifeart (approved):

test/unit/rules/require-input-label-test.js
```diff
@@ -91,12 +94,12 @@ generateRuleTests({
       },
     },
     {
-      template: '<label><input></label>',
+      template: '<label><input /></label>',
```
rwjblue, line 97: Please reset this to HTML syntax (not XHTML). I think having an XHTML example would be perfectly fine, but it should be a new test (not mutating an existing / unrelated one).

rwjblue (changes requested):

test/unit/rules/require-input-label-test.js
MelSumner, line 97:
```suggestion
      template: '<label><input></label>',
```

test/unit/rules/require-input-label-test.js
```diff
@@ -96,7 +99,7 @@ generateRuleTests({
         message: ERROR_MESSAGE,
         line: 1,
         column: 7,
-        source: '<input>',
+        source: '<input />',
```
MelSumner, line 102:
```suggestion
        source: '<input>',
```

test/unit/rules/require-input-label-test.js
MelSumner: Ok, changed it back.

rwjblue (approved):

Issue #1920, opened 2021-04-29
Make `builtin-component-arguments` work for `Textarea` too


boris-petrov:
The same [reasoning](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/builtin-component-arguments.md) can be applied to `Textarea` as for `Input`. Perhaps extending the rule for it would be nice.

PR #2013, opened 2021-06-25
Add `Textarea` support to `builtin-component-arguments` rule

lifeart:
resolves https://github.com/ember-template-lint/ember-template-lint/issues/1920

### DIFF
--- a/docs/rule/builtin-component-arguments.md
+++ b/docs/rule/builtin-component-arguments.md
@@ -13,8 +13,17 @@ This rule warns about `Input` component invocations that use the following attri
 * `type`
 * `value`
 
+The builtin `Textarea` component has several arguments that match attributes
+of the lower-case `textarea` HTML element. These arguments should be set via e.g.
+`@value`, instead of `value`, but it is easy to forget and can cause subtle
+issues.
+
+This rule warns about `Textarea` component invocations that use the following attributes instead of arguments:
+
+* `value`
+
 Please note that this rule currently only warns about these three attributes on
-the `Input` component, but might be extended in the future to also warn about
+the `Input`, `Textarea` components, but might be extended in the future to also warn about
 other attributes or builtin components.
 
 ## Examples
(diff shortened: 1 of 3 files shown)


docs/rule/builtin-component-arguments.md
```diff
@@ -13,8 +13,17 @@ This rule warns about `Input` component invocations that use the following attri
 * `type`
 * `value`
 
+The builtin `Textarea` component has several arguments that match attributes
+of the lower-case `textarea` HTML element. These arguments should be set via e.g.
```
rwjblue, line 17: I know this is pedantic, but `value` is **not** a valid attribute of `<textarea></textarea>`. It is however a valid **property** of the resulting DOM element. 😭 

```suggestion
The builtin `Textarea` component has several arguments that match properties
of the lower-case `textarea` HTML element. These arguments should be set via e.g.
```

rwjblue (changes requested):

docs/rule/builtin-component-arguments.md
```diff
@@ -13,8 +13,17 @@ This rule warns about `Input` component invocations that use the following attri
...
+`@value`, instead of `value`, but it is easy to forget and can cause subtle
+issues.
+
+This rule warns about `Textarea` component invocations that use the following attributes instead of arguments:
+
+* `value`
+
 Please note that this rule currently only warns about these three attributes on
-the `Input` component, but might be extended in the future to also warn about
+the `Input`, `Textarea` components, but might be extended in the future to also warn about
```
MelSumner, line 26:
```suggestion
the `Input`, and one property on the `Textarea` components, but might be extended in the future to also warn about
```

MelSumner (changes requested):
One small suggested edit to the docs change, otherwise looks good to me! 👍 @rwjblue looks like it's ready to re-review.

bmish (approved):

Issue #1859, opened 2021-03-18
`no-duplicate-id` rule false positive with block param


bmish:
The rule sees `id={{inputProperties.id}}` on multiple elements and flags a violation, but this is fine because the variable is scoped to the parent component.

```hbs
<MyComponent as |inputProperties|>
    <Input id={{inputProperties.id}} />
</MyComponent>

<MyComponent as |inputProperties|>
    <Input id={{inputProperties.id}} />
</MyComponent>
```

faith-or:
I have a potential fix: https://github.com/ember-template-lint/ember-template-lint/pull/2024

bertdeblock:
Can be closed right?

faith-or:
Yup @bmish / @scalvert  can you close this since the fix is merged already

PR #2024, opened 2021-07-10
Fix no-duplicate-id rule false positive with block param

faith-or:
This address issue #1859.

If we have multiple components using the same yield name convention, `no-duplicate-id` will flag it as an error:

```
<MyComponent as |inputProperties|>
    <Input id={{inputProperties.id}} />
</MyComponent>

<MyComponent as |inputProperties|>
    <Input id={{inputProperties.id}} />
</MyComponent>
```

Above is valid because the id is scoped to the parent component. 

In this PR:
-  Added a way to create unique id if a block param is used. Reason why we have false positives is because the ids are the same. Here we create unique ids so if user does above it would be valid. If user uses the same id in the same scope, we will flag the error.
- added tests

Ran `yarn test`:

```
Test Suites: 122 passed, 122 total
Tests:       6 skipped, 6175 passed, 6181 total
Snapshots:   64 passed, 64 total
Time:        26.663 s, estimated 29 s
Ran all test suites.
✨  Done in 30.99s.
```


lifeart (approved):

scalvert (approved):

scalvert:
Thanks, @faith-or!

PR #2032, opened 2021-07-23
Fixes the SARIF formatter to correctly filter output by severity

scalvert:
The SARIF formatter contained a bug where it would include all severities in results without respecting the `--quiet` or `--include-todo` options. this resulted in a discrepancy in the results when run locally using the pretty formatter and those that were gathered using the SARIF formatter.

### DIFF
--- a/lib/formatters/sarif.js
+++ b/lib/formatters/sarif.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const url = require('url');
 
 const writeOutputFile = require('../helpers/write-output-file');
+const Linter = require('../linter');
 
 function getResultLevel(message) {
   if (message.fatal || message.severity === 2) {
@@ -86,8 +87,24 @@ class SarifPrinter {
           }
         }
 
-        if (result.messages.length > 0) {
-          for (const message of result.messages) {
+        let fileErrors = result.messages;
+        let filteredMessages;
+
+        if (fileErrors.length > 0) {
+          let errorsFiltered = fileErrors.filter(
+            (error) => error.severity === Linter.ERROR_SEVERITY
+          );
+          let warnings = this.options.quiet
+            ? []
+            : fileErrors.filter((error) => error.severity === Linter.WARNING_SEVERITY);
+          let todos =
+            this.options.quiet || !this.options.includeTodo
+              ? []
+              : fileErrors.filter((error) => error.severity === Linter.TODO_SEVERITY);
+
+          filteredMessages = [...errorsFiltered, ...warnings, ...todos];
+
+          for (const message of filteredMessages) {
             const sarifRepresentation = {
               level: getResultLevel(message),
               message: {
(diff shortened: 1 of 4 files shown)


lib/formatters/sarif.js
```diff
@@ -86,8 +87,24 @@ class SarifPrinter {
...
+          );
+          let warnings = this.options.quiet
+            ? []
+            : fileErrors.filter((error) => error.severity === Linter.WARNING_SEVERITY);
+          let todos =
+            this.options.quiet || !this.options.includeTodo
+              ? []
+              : fileErrors.filter((error) => error.severity === Linter.TODO_SEVERITY);
+
+          filteredMessages = [...errorsFiltered, ...warnings, ...todos];
```
elwayman02, line 105: Would this cause the results to be grouped by severity rather than file, or are they already grouped by severity before entering this code block? Just wondering if this will result in a meaningful change in the ordering.

As an alternative, you could do something like this:

```
let filteredMessages = fileErrors.filter((error) => {
  let isError = error.severity === Linter.ERROR_SEVERITY;
  let isWarning = error.severity === Linter.WARNING_SEVERITY;
  let isTodo = error.severity === Linter.TODO_SEVERITY;

  return isError || (!this.options.quiet && isWarning) || (!this.options.quiet && this.options.includeTodo && isTodo);
)};
```

scalvert: This is precisely the logic we use in the [json formatter](https://github.com/ember-template-lint/ember-template-lint/blob/a9bb2700e57d7d1f232a6c20e30cdaa50ae839c9/lib/formatters/json.js#L14-L28). Order isn't stated or guaranteed, and isn't materially relevant.

lib/formatters/sarif.js
```diff
@@ -86,8 +87,24 @@ class SarifPrinter {
...
+
+        if (fileErrors.length > 0) {
+          let errorsFiltered = fileErrors.filter(
+            (error) => error.severity === Linter.ERROR_SEVERITY
+          );
+          let warnings = this.options.quiet
+            ? []
+            : fileErrors.filter((error) => error.severity === Linter.WARNING_SEVERITY);
+          let todos =
+            this.options.quiet || !this.options.includeTodo
```
rwjblue, line 101: Why is this an `||`? I would think that if you have specified `--include-todo --quiet` you should still get todos included...

scalvert: I think quiet implies only errors, no?

rwjblue (approved):

lib/formatters/sarif.js
scalvert: We're going to follow up with this in a later PR.

rwjblue: Ya, I agree that is what it does in `--json` but I don't particularly like it. I'll open a separate issue to track that though since we should keep them in sync together either way.

PR #2033, opened 2021-07-23
Upgrades to latest todo-utils to include new fuzzy matching functionality

scalvert:
Uses new fuzzy matching functionality defined in https://github.com/ember-template-lint/ember-template-lint-todo-utils/issues/246.

### DIFF
--- a/lib/-private/todo-handler.js
+++ b/lib/-private/todo-handler.js
@@ -1,5 +1,5 @@
 const {
-  buildTodoData,
+  buildTodoDatum,
   getTodoBatches,
   todoStorageDirExists,
   writeTodos,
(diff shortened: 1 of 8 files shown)


lib/-private/todo-handler.js
```diff
@@ -74,8 +70,8 @@ module.exports = class TodoHandler {
         (result) =>
           result.filePath === todo.filePath &&
           result.rule === todo.ruleId &&
-          result.line === todo.line &&
-          result.column === todo.column
+          result.line === todo.range.start.line &&
+          result.column === todo.range.start.column
```
rwjblue, line 74: I think this should also be updated to care RE: `source`, right?

scalvert: Hmm, it's almost as if we have to do a secondary match on these...

scalvert: We should discuss this more, but the crux of this issue is that once we identify matches as `stable`, we further need a way to associate the stable items with their associated lint results. This is critical for us to be able to mutate the severity. As it stands now, the new fuzzy matching means that the logic above won't work for fuzzy matches, as we can't associate the result anymore. We effectively need a way to brand the lint result onto the todo data, such that it can be compared when processing stable matches.

elwayman02:
Any progress on this bugfix? We're getting a lot of complaints from developers who are feeling pain points from this bug.

scalvert:
@elwayman02 to be clear, this isn't a bugfix. We designed the todo system with strict matching in mind, and are adding this extended functionality as an enhancement. 

I'm working through some of the more complex parts of the upgrade, as we need to major version bump the todo-utils package, and have a few items on our laundry list to get through before we do that. Once that's done, this PR will be in a shippable state.

lib/-private/todo-handler.js
scalvert: This has been resolved via https://github.com/ember-template-lint/ember-template-lint-todo-utils/pull/277

scalvert:
Going to merge this through, since @rwjblue and I have extensively paired on the full implementation.

Issue #1965, opened 2021-05-25
`require-input-label` Custom label component false positive


Windvis:
I noticed the `require-input-label` triggers an error in some cases when you use a custom label component, instead of the html element:

```hbs
<CustomLabel>
  <input />
</CustomLabel>
```

Maybe these cases can be covered by adding a list of allowed wrapper components as a configuration option for the `require-input-label` rule?

It's possible to work around it by simply adding the for / id combo anyway, so not a major issue.

MelSumner:
I don't think I can think of a way to filter for that (how could I differentiate between a custom label element and any other element?); I think that using the inline disable of the rule would be the best option in this case. 

If you have any ideas though, I'm all ears!

Windvis:
@MelSumner I think it should be possible with an allow list similar to how [`no-implicit-this` works](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-implicit-this.md#configuration) for example?

```js
module.exports = {
  extends: 'recommended',

  'require-input-label: {
    allow: ['CustomLabel']
  },
};
```

PR #2053, opened 2021-08-04
Add `labelTags` option for custom label components to `require-input-label` rule

lifeart:
resolves: https://github.com/ember-template-lint/ember-template-lint/issues/1965


rwjblue:
@lifeart - can you rebase?

MelSumner (approved):
This looks good to me, thank you for including an update to the documentation.

lib/rules/require-input-label.js
```diff
@@ -22,7 +31,57 @@ function hasValidLabelParent(path) {
...
+        }
+        break;
+    }
+
+    let errorMessage = createErrorMessage(
+      this.ruleName,
+      [
+        '  * boolean - `true` to enable / `false` to disable',
+        '  * object -- An object with the following keys:',
+        '    * `allow` -- An array of component / helper names for that may be called without arguments',
```
bmish, line 78: This description looks wrong.

docs/rule/require-input-label.md
```diff
@@ -81,6 +81,14 @@ This rule **allows** the following:
 * another option is to add an aria-label to the input element.
 * wrapping the input element in a label element is also allowed; however this is less flexible for styling purposes, so use with awareness.
 
+## Configuration
+
+ The following values are valid configuration:
+
+* boolean - `true` to enable / `false` to disable
+* object -- An object with the following keys:
+  * `allow` -- An array of component names for that may be used as label replacement
```
bmish, line 90: Can you rename `allow` to `labelComponentNames` for greater clarity?

bmish, line 90: Let's mention that this list is in addition to the HTML `label` tag.

```suggestion
  * `allow` -- An array of component names for that may be used as label replacements (in addition to the HTML `label` tag)
```

bmish (changes requested):

test/unit/rules/require-input-label-test.js
```diff
@@ -63,6 +63,15 @@ generateRuleTests({
...
+      config: { allow: ['CustomLabel'] },
+      template: '<CustomLabel><input /></CustomLabel>',
+    },
+    {
+      config: { allow: [/web-label/] },
+      template: '<web-label><input /></web-label>',
+    },
   ],
 
   bad: [
```
bmish, line 77: Let's add one invalid test case with a config but not the matching label.

docs/rule/require-input-label.md
bmish: Or maybe `customLabelComponents`?

lifeart: I think this is too long names, you have to read doc and copy it, quite impossible to remember

bmish: My concerns are:

* It's not clear what is being allowed by this `allow` option. Someone reading this option will have to look at the doc to understand what it refers to. I think option names should be able to stand on their own ("self-documenting").
* What if there are other types of things we want to allow in the future besides just component names? Using `allow` will make it harder for us to add other types of "allowed" options in the future.

I don't think length matters. People will not be typing this option name often.

lifeart: How about “allowTags”?

bmish: Okay. As long as you think that it will be clear that this refers to custom label component tags/names.

docs/rule/require-input-label.md
```diff
@@ -81,6 +81,14 @@ This rule **allows** the following:
 * another option is to add an aria-label to the input element.
 * wrapping the input element in a label element is also allowed; however this is less flexible for styling purposes, so use with awareness.
 
+## Configuration
+
+ The following values are valid configuration:
+
+* boolean - `true` to enable / `false` to disable
+* object -- An object with the following keys:
+  * `allow` -- An array of component names for that may be used as label replacements (in addition to the HTML `label` tag)
```
bmish, line 90:
```suggestion
  * `allow` -- An array of component names that may be used as label replacements (in addition to the HTML `label` tag)
```

docs/rule/require-input-label.md
lifeart: @bmish looks like `labelTags` fits here :)

bmish:
Just a few comments to address and then we can merge this hopefully today.

test/unit/rules/require-input-label-test.js
lifeart: done

lifeart:
@bmish pinging here

bmish:
@lifeart I thought of a potential issue. Components/helpers can be written either dasherized/mustache (`{{my-component}}`) or title case / angle bracket (`<MyComponent>`). With this rule option, the rule won't necessarily catch both forms.

The [no-restricted-invocations](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-restricted-invocations.md#configuration) rule has a similar `names` option that solves this problem by requiring the name to be specified in a certain format and then automatically checking both the dasherized/sentence case versions.

I think we should automatically handle both formats as well to ensure this rule reliably catches the desired custom label components.

lifeart:
@bmish you mean convert one component naming to another? or warn if it's not angle-bracket?

(I think supporting curly components here will be overhead, because it's too close to be "deprecated")

bmish:
I do think it's important to support curly components still. Many larger codebases have not fully migrated to angle brackets yet. And it's relatively easy to support it.

Here's what I recommend:
1. Require names passed to `allowTags` to be in angle-bracket format (`MyComponent` / `NestedScope::MyComponent`). If they aren't, assert/throw exception.
2. In the rule, check component names against both the passed-in angle-bracket names as well as the old-curly bracket names (`e.g. `my-component` / `nested-scope/my-component`).

lifeart:
@bmish not really, we could have overlap with `{{helper-name}}` (

bmish:
@lifeart I'm not sure we need to worry about collisions with helper names. Right now, I'm more worried that mustache components can't be used as custom label components.

I think this should be a valid test case:

```js
{
  config: { labelTags: ['CustomLabel'] },
  template: '{{#custom-label}}<input />{{/custom-label}}',
},
```

bmish (approved):

bmish:
I am merging this to get a release out now but I opened https://github.com/ember-template-lint/ember-template-lint/issues/2125 to track updating this option to handle mustache components (as a bug fix).

Issue #2028, opened 2021-07-16
Remove `keydown` linting in rule: no-down-event-binding


camario25:
no-down-event-binding rule includes `keydown` event:

https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-down-event-binding.md

I understand how binding to `mousedown` could cause a11y issues because it prevents a user from moving away from the target(preventing the action).  However, I do not see the benefit for avoiding `keydown`.  Once the key is pressed a user can’t really undo it.  But maybe I’m not seeing something.

It is known that keydown provides more uniform behavior across different browsers.
https://www.mutuallyhuman.com/blog/keydown-is-the-only-keyboard-event-we-need/

Hello @MelSumner , I see that you contributed to this rule.  What is the original reasoning behind including `keydown` along with `mousedown`? 

Talking with our a11y team, they agree that it is fine to use keydown without compromising a11y.

Please let me know, if the community agrees, I can work on submitting a PR.  
Thank you, Mario

jfdnc:
+1 on this issue. Our team is making the requisite template changes to update `ember-template-lint` and ran into this same issue. The w3.org documentation linked from the `no-down-event-binding` doc in this repo is specific to pointer events, and should not include key events. This rule should likely be reworked to something like `no-pointer-down-event-binding`.

andrewferk:
+1. In order to get the arrow keys to trigger, `keydown` is the recommended solution.

MelSumner:
@camario25 @jfdnc I'm supportive of having this rule refactored to be more correct 👍

MelSumner:
Update: an example where you would want to use keyup instead of keydown: On a [tabs](https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-2/tabs.html) component, you would want to configure the arrow keys that cycle between the tabs to use keyup instead of keydown- keyup responds once, whereas keydown would respond multiple times. This would be detrimental for folks with motor control issues. 

I wonder if we should keep this rule as is and instead encourage folks to disable it if they really feel like they need to use it.

drewlee:
[...]

Issue #1908, opened 2021-04-23
Plan V4 Release


bmish:
### Planned changes for v4

- [x] #2208
   - [x] #2209
   - [x] Should we expose `lib/helpers/create-error-message` (used [here](https://gitlab.com/michal-bryxi/open-source/ember-template-lint-plugin-tailwindcss/-/blob/429b1a41a7084968e6a5d55e792193c0749bfad7/src/rules/class-wrap.ts#L5))? For now, decided no.
- [x] #2262
- [x] #2176
- [x] Remove deprecated CLI options
   - [x] #2193
   - [x] #2207
- [ ] Remove deprecated properties / behavior
   - [ ] `// TODO: add a deprecation for accessing _filePath`
   - [ ] `The _moduleName property used in the '${this.ruleName}' rule has been removed. Please use _filePath instead.` 
- [x] Update core linting / rule reporting behavior
   - [x] #2228
   - [x] #2242
   - [x] #2245
   - [x] #2261
- [x] Remove deprecated rules
   - [x] #2191
   - [x] #2188
- [x] Update rule configs
   - [x] #2192
   - [x] #2211
   - [x] #2237 
   - [x] #2256
- [x] Update rule behavior 
   - [x] #1881 
   - [x] #2195
   - [x] #2089
   - [x] #2241
   - [x] #2278
   - [ ] #2054 (only will be included if PR gets finished)
- [x] Update test behavior
    - [x] #2216
    - [x] #2217
    - [x] #2230
    - [x] #2240
    - [x] #2279
      - [x] #2301 
- [x] Remove `pending` internals
   - [x] #2252 
- [x] Update core TODO functionality 
   - [x] #2255 
- [x] Update config file behavior
   - [x] #2236 
   - [x] #2238
   - [x] #2239
- [x] Update non-ESM dependencies
   - [x] #2172
   - [x] #2197
- [ ] Update ESM dependencies 
   - [ ] #2292 (dependency)
   - [x] #2295 (dev-dependency)
   - [x] #2293 (dependency)
   - [x] #2294 (dependency)
   - [x] #2296 (dependency)
- [x] New features unblocked in v4
    - [x] #2143
- [x] #2297

### Future changes for v5

* Require rules with options to provide a JSON Schema in `meta.options` like ESLint. This is a large feature and requires the `meta` property for rules to be implemented.

### Pre-release checklist:
* Release a minor/patch version ahead of the major release for any non-breaking changes
[...]

PR #2054, opened 2021-08-05
Rename and narrow rule `no-down-event-binding` to `no-pointer-down-event-binding`

jfdnc:
Recommended rule `no-down-event-binding` references [WCAG failure of success criteria](https://www.w3.org/WAI/WCAG21/Techniques/failures/F101) for `down` events that only pertain to pointer-down events. Current implementation of `no-down-event-binding` also applies to `keydown` events, which does not fit the intention of the WCAG docs, and therefore should not be included in this lint rule.

- updates related file names to reflect `no-pointer-down-event-binding`
- updates node check for inclusion in "mousedown"|"pointerdown"
- updates related doc and test files as appropriate


jfdnc:
Just a ping here as this PR and the [issue for it](https://github.com/ember-template-lint/ember-template-lint/issues/2028) aren't being acknowledged. This can be moved forward or not, I just want to be sure that the maintainers here are aware of the apparently incorrect application of this rule preference. Any thoughts or opinions are appreciated.

lib/config/a11y.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     'no-abstract-roles': 'error',
     'no-accesskey-attribute': 'error',
     'no-aria-hidden-body': 'error',
-    'no-down-event-binding': 'error',
+    'no-pointer-down-event-binding': 'error',
```
bertdeblock, line 9: Should probably move down to respect the order of this list. Same goes for the `recommended` config file.

lib/rules/no-pointer-down-event-binding.js
```diff
@@ -26,20 +27,20 @@ function isActionModifer(node) {
...
 /**
- * Check if an event name is a "down" event
+ * Check if an event name is a pointer "down" event
  *
  * @param {string} eventName
  * @return {boolean}
  */
-function isDownEvent(eventName) {
-  return eventName.toLowerCase().endsWith('down');
+function isPointerDownEvent(eventName) {
```
bertdeblock, line 35: I think it's best that this util still returns a boolean like it used to, now it returns either an array or `null`.

test/unit/rules/no-pointer-down-event-binding-test.js
```diff
@@ -1,61 +1,61 @@
...
+const { ERROR_MESSAGE } = require('../../../lib/rules/no-pointer-down-event-binding');
 const generateRuleTests = require('../../helpers/rule-test-harness');
 
 generateRuleTests({
-  name: 'no-down-event-binding',
+  name: 'no-pointer-down-event-binding',
 
   config: true,
 
   good: [
```
bertdeblock, line 11: Personally, I would keep the old good cases as they are still allowed and to make sure not to introduce any regressions.

bertdeblock (commented):
👋 Left a few suggestions.

Also wanted to point out that, even though the rename makes sense, this might break people's lint build if they use the old name in disable statements or for turning off the rule completely via the config file. So I think we should either keep the old rule name and rename later on _or_ mark this PR as breaking?

lib/config/a11y.js
jfdnc: Good catch, will alpha sort this

jfdnc:
> So I think we should either keep the old rule name and rename later on _or_ mark this PR as breaking?

@bertdeblock I think we should prefer to mark this as breaking. The original intent of introducing the `no-down-event-binding` rule was to address the issues with down-events pointed out [here](https://www.w3.org/WAI/WCAG21/Techniques/failures/F101). Maintaining both that rule and the one is unnecessarily over-specified, and `no-down-event-binding` would continue to block `key` events for no purpose.

Thanks for the feedback here, will make the suggested tweaks!

test/unit/rules/no-pointer-down-event-binding-test.js
jfdnc: I think if this pointer-down linting concern had been introduced first, we wouldn't have had a reason to include `key` events in the scope of the tests, so it made sense to me to remove those. I don't think anything is really gained by having those included. I can add those back in the "good" checks if you feel strongly on that point, though.

bertdeblock: Nope, no strong feelings on this one. I'm just a contributor btw, so not my decision to make. But I've seen this mentioned before in PRs for this project, that's why I mentioned it.

bmish:
Just FYI, the v4 release (https://github.com/ember-template-lint/ember-template-lint/issues/1908) is coming up in the next week or two. So if we want to make breaking changes, now is the time to do it. CC: @elwayman02

drewlee:
@jfdnc I'm inquiring to see if these changes can be picked back up again or maybe modified so they can be released as a non-breaking change. (Maybe make the key down event a warning instead of an error?) Would be happy to assist in any way. Thanks!
CC @bmish

bmish:
Usually it's best to use an opt-in rule option to avoid a breaking change. But note that we will have another major release coming up (#2319) if breaking changes are desired.

drewlee:
@bmish I'd like to have my team pick this up, as this PR seems to have gone stale and abandoned. The simplest path sounds like a breaking change that can be rolled into the v5 release. We will likely address via a new PR. Since I don't have insight into the exact date v5 is slated for, I just wanted to make sure we have enough time to land this. Thanks!

judithhinlung:
I think this refactor could be ready to merge with a few changes. I have updated this PR to pull in the changes from master since its last commit, and I added `onmousedown` and `onpointerdown` to the list of pointer-down events. Since this is a breaking change, I have also removed this rule from the list of recommended rules.

lib/config/3-x-recommended.js
```diff
@@ -43,6 +42,7 @@ export default {
     'no-outlet-outside-routes': 'error',
     'no-partial': 'error',
     'no-passed-in-event-handlers': 'error',
+    'no-pointer-down-event-binding': 'error',
```
drewlee, line 45: I believe the `3-x-recommended.js` config was specific to version 3.0 and is no longer in use/relevant, so this change isn't necessary.

judithhinlung:
Due to the name change, leaving the rule name in 3-x-recommended.js as `no-down-event-binding` will cause the following test failure:
 FAIL  test/unit/default-config-test.js
  ● default configurations › 3-x-recommended.js › should contain only valid rules

drewlee:
@bmish Can we get some eyes on this and make sure we're on the right track? Thanks!

lib/rules/no-pointer-down-event-binding.js
```diff
@@ -19,25 +20,27 @@ function isOnModifier(node) {
...
+  return ['mousedown', 'onmousedown', 'pointerdown', 'onpointerdown'].includes(
```
bmish, line 34: Can you declare this array in a `const` outside the function so it doesn't get recreated each time the function is called?

bmish (approved):
Looks ready to go as part of the next major release #2319.

lib/rules/no-pointer-down-event-binding.js
judithhinlung: Yes, I have updated the rule to declare the pointer-down event bindings in a set called during rule setup.

README.md
```diff
@@ -152,6 +151,8 @@ Each rule has emojis denoting:
 | [no-outlet-outside-routes](./docs/rule/no-outlet-outside-routes.md)                                       | ✅  |     |     |     |
 | [no-partial](./docs/rule/no-partial.md)                                                                   | ✅  |     |     |     |
 | [no-passed-in-event-handlers](./docs/rule/no-passed-in-event-handlers.md)                                 | ✅  |     |     |     |
+<<<<<<< HEAD
```
bertdeblock, line 154: 😬

docs/rule/no-action.md
```diff
@@ -4,7 +4,7 @@
 
 What's wrong with `{{action}}`?
 
-"Action" is an overloaded term in Ember parlance. Actions are:
```
drewlee, line 7: @judithhinlung It looks like you got a bad merge as there are a bunch of unrelated changes in this PR now.

judithhinlung:
I have reverted the previous rebase to merge with master again, and the previous issues should be resolved.

jfdnc:
Been out of the loop with OS things for a while, @judithhinlung thanks so much for picking this up! You rock 🎉

PR #2063, opened 2021-08-10
Update docs with .lint-todorc.js

tylerbecks:


docs/todos.md
```diff
@@ -44,28 +44,44 @@ todos can be created with optional due dates. These due dates allow for todos to
...
+       "daysToDecay": {
          "warn": 5,
          "error": 10
        }
      }
    }
    ```
 
-1. Via environment variables
+2. Via `.lint-todorc.js`
```
elwayman02, line 64: These should all stay as `1.`. That is the markdown notation for a numbered list, they will be numbered automatically. If you manually number them, then the numbers have to be updated by hand every time items are added/removed/reordered.

scalvert: Correct!

docs/todos.md
```diff
@@ -44,28 +44,44 @@ todos can be created with optional due dates. These due dates allow for todos to
...
 :bulb: Both `warn` and `error` are optional. The value for `warn` should be lower than the value of `error`.
 
-1. Via package.json configuration
+1. Via `package.json`
 
    ```json
    {
      "lintTodo": {
-       "decayDays": {
+       "daysToDecay": {
```
scalvert, line 56: Thanks for this catch!

docs/todos.md
```diff
@@ -44,28 +44,44 @@ todos can be created with optional due dates. These due dates allow for todos to
 
 Due dates can be configured in multiple ways, but all specify integers for `warn` and `error` to signify the number of days from the todo created date to decay the severity.
 
+Due dates can be configured on a per-rule basis as well with the `daysToDecayByRule` option.  See examples below.
```
scalvert, line 47: I'd actually love to move this down below in its own section, so we can call it out explicitly. I sort of view this as a layered approach; we do the simple thing first, followed by successively more complex.

Can we give this its own section, outside of enumerating the different ways you can set the configs. Basically, have a section with a 3 # heading, called something like `Configuring daysToDecay by rule`.

scalvert (changes requested):

scalvert:
Thanks for adding this, @tylerbecks!

Issue #2072, opened 2021-08-20
Add endLine/endColumn to rule result.


scalvert:
Currently our output does not include `endLine`/`endColumn`. Those values can be inferred from the `line`/`column` in combination with the source, though this is less convenient and differs from the schema that `eslint` provides. Though the latter isn't an issue currently, any future move to using `eslint` as the underlying linter engine will precipitate the need for this extra data.

To preempt this, the following changes are proposed:

- [x] refactor all log calls to include node, remove extraneous props that are no longer needed
- [x] add deprecation util, by rule, that emits a warning (console.log) if the log API is used without the node prop, in dev it should throw (inspect JEST env var). Ensure we respect `--quiet`
- [x] widen API to include `line`, `column`, `endLine`, `endColumn` (widen both JSON formatter AND internal data structure that's output via the Linter Node APIs)

These changes will ensure our linter output has a closer shape to that of `eslint`. 

cc/ @rwjblue as we discussed.

scalvert:
Resolved in #2073 and #2151

PR #2073, opened 2021-08-20
Ensures all log calls include `node`

scalvert:
Partial implementation of #2072.

While this PR looks huge (and it is!), it's mostly procedural changes related to each rules' log calls. I've subdivided each one into a separate commit in which I made the change, independently ran each test to verify correctness, and committed. The main change to take a look at is the logic in `log`, where we're now merging with potentially overridden values passed in to `log`.

Additionally, a new deprecation warning has been added to `log` to ensure consumers can migrate their rules in plugins without concern for them breaking. If using incorrect log usage in local rules, this deprecated usage will `throw` during tests.

### DIFF
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -544,8 +544,8 @@ module.exports = class {
       // to include line, column, and source. If the rule has passed custom values for these,
       // we want to respect those.
       nodeProps = {
-        line: 'line' in result ? result.line : (node.loc && node.loc.start.line),
-        column: 'column' in result ? result.column : (node.loc && node.loc.start.column),
+        line: 'line' in result ? result.line : node.loc && node.loc.start.line,
+        column: 'column' in result ? result.column : node.loc && node.loc.start.column,
         source: 'source' in result ? result.source : this.sourceForNode(node),
       };
     }
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -526,7 +526,7 @@ module.exports = class {
       if (!loggedRules.has(this.ruleName)) {
         let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result.`;
 
-        if (process.env.NODE_ENV === 'test') {
+        if (process.env.EMBER_TEMPLATE_LINT_DEV_MODE === '1') {
           throw new Error(message);
         } else {
           this._console.log(message);
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -544,9 +544,9 @@ module.exports = class {
       // to include line, column, and source. If the rule has passed custom values for these,
       // we want to respect those.
       nodeProps = {
-        line: result.line || (node.loc && node.loc.start.line),
-        column: result.column || (node.loc && node.loc.start.column),
-        source: result.source || this.sourceForNode(node),
+        line: 'line' in result || (node.loc && node.loc.start.line),
+        column: 'column' in result || (node.loc && node.loc.start.column),
+        source: 'source' in result || this.sourceForNode(node),
       };
     }
 
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -544,9 +544,9 @@ module.exports = class {
       // to include line, column, and source. If the rule has passed custom values for these,
       // we want to respect those.
       nodeProps = {
-        line: 'line' in result || (node.loc && node.loc.start.line),
-        column: 'column' in result || (node.loc && node.loc.start.column),
-        source: 'source' in result || this.sourceForNode(node),
+        line: 'line' in result ? result.line : (node.loc && node.loc.start.line),
+        column: 'column' in result ? result.column : (node.loc && node.loc.start.column),
+        source: 'source' in result ? result.source : this.sourceForNode(node),
       };
     }
 
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -524,11 +524,15 @@ module.exports = class {
 
     if (!result.node) {
       if (!loggedRules.has(this.ruleName)) {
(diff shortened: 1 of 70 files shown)


lib/rules/_base.js
```diff
@@ -521,16 +522,32 @@ module.exports = class {
...
+      // if we've passed node, we want to ensure that we're correctly expanding the properties
+      // to include line, column, and source. If the rule has passed custom values for these,
+      // we want to respect those.
       nodeProps = {
-        line: node.loc && node.loc.start.line,
-        column: node.loc && node.loc.start.column,
-        source: this.sourceForNode(node),
+        line: result.line || (node.loc && node.loc.start.line),
+        column: result.column || (node.loc && node.loc.start.column),
+        source: result.source || this.sourceForNode(node),
```
rwjblue, line 549: Hmm, I don't think this quite works. `result.line`/`result.column` could be `0` and therefore falsey. Can you add a test case for that?

lib/rules/attribute-indentation.js
```diff
@@ -243,6 +243,7 @@ module.exports = class AttributeIndentation extends Rule {
 
       this.log({
         message,
+        node,
         line: actual.line,
         column: actual.column,
```
rwjblue, line 248: Some rules add `node` and remove `line`/`column`/`source`, but others don't. Any particular reason?

lib/rules/_base.js
```diff
@@ -521,16 +522,32 @@ module.exports = class {
       defaults.filePath = this.filePath;
     }
 
-    // perform the node prop expansion only if those properties don't exist in result
+    if (!result.node) {
+      if (!loggedRules.has(this.ruleName)) {
+        let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result.`;
+
+        if (process.env.JEST_WORKER_ID !== undefined) {
```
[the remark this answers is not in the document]
scalvert: This was to track for local dev, but as we've discussed this isn't likely a good solution, since some plugins can _not_ use jest. Changing this to a custom env var, which we'll opt into in our own test harness.

lib/rules/_base.js
scalvert: Ya good call. Will fix.

lib/rules/block-indentation.js
```diff
@@ -254,9 +252,9 @@ module.exports = class BlockIndentation extends Rule {
 
       this.log({
         message: warning,
+        node,
         line: node.loc.end.line,
         column: node.loc.end.column,
```
scalvert, line 257: This is specifically using `end` values from `loc`, rather than the more conventional `start`, which is why it appears inconsistent but is actually not.

lib/rules/attribute-indentation.js
scalvert: Some rules override the values of `line`, `column` and `source`, which end up being distinctly different than those from `node`. In these cases, we don't remove those properties.

See https://github.com/ember-template-lint/ember-template-lint/pull/2073/files#r698866054, for example.

lib/rules/_base.js
```diff
@@ -521,16 +522,32 @@ module.exports = class {
       defaults.filePath = this.filePath;
     }
 
-    // perform the node prop expansion only if those properties don't exist in result
+    if (!result.node) {
+      if (!loggedRules.has(this.ruleName)) {
+        let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result.`;
```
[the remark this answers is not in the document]
scalvert: Sure, can do.

lib/rules/attribute-indentation.js
scalvert: I ran the following in the `log` call to check if any of the current rules are passing in the same line/column as their node:

```js
for (let prop of ['line', 'column']) {
  if (prop in result && result[prop] === result.node.loc['prop']) {
    throw new Error(`result.${prop} and result.node.loc.${prop} are equal`);
  }
}
```

No rules appear to pass the same values.

lib/rules/_base.js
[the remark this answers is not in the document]
scalvert: Updated.

lib/rules/_base.js
```diff
@@ -521,16 +522,32 @@ module.exports = class {
...
+      // if we've passed node, we want to ensure that we're correctly expanding the properties
+      // to include line, column, and source. If the rule has passed custom values for these,
+      // we want to respect those.
       nodeProps = {
-        line: node.loc && node.loc.start.line,
-        column: node.loc && node.loc.start.column,
-        source: this.sourceForNode(node),
+        line: 'line' in result || (node.loc && node.loc.start.line),
+        column: 'column' in result || (node.loc && node.loc.start.column),
+        source: 'source' in result || this.sourceForNode(node),
```
rwjblue, line 549:
```suggestion
        line: 'line' in result ? result.line : (node.loc && node.loc.start.line),
        column: 'column' in result ? result.column : (node.loc && node.loc.start.column),
        source: 'source' in result ? result.source : this.sourceForNode(node),
```

lib/rules/attribute-indentation.js
rwjblue: Thanks for checking that!

rwjblue (changes requested):

lib/rules/_base.js
scalvert: Oh derp. What a dork I am. Thanks for catching that obvious mistake.

scalvert:
@rwjblue every time I accept your suggestion I'm cursed with a Prettier error. I wish GH could just Pretterify the suggestions!

scalvert:
@rwjblue I think we're GTG.

PR #2080, opened 2021-08-27
Add `require-valid-named-block-naming-format` rule

### LINKED ISSUE #1772: Feature request: rule for consistent naming of named-blocks
I'm not sure what Ember's convention is on named-block naming (`<:foo-bar>` vs `<:fooBar>`) but a rule to force the "right" one would be nice.

bertdeblock:
Adds a new rule `require-valid-named-block-naming-format` to enforce consistent naming of named blocks.

Possible options are `camelCase` (default) and `kebab-case`.
`PascalCase` isn't an option because [named blocks must start with a lower case letter](https://github.com/glimmerjs/glimmer-vm/blob/3874a63ce863f94078d93d22cba66dbb23c496d3/packages/%40glimmer/integration-tests/test/syntax/named-blocks-test.ts#L74-L83).

Wasn't sure which format to pick as the default so just picked `camelCase`, but can easily change to `kebab-case`.

Closes #1772.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,7 @@ Each rule has emojis denoting:
 | [require-lang-attribute](./docs/rule/require-lang-attribute.md)                                           | ✅  |     | ⌨️  |     |
 | [require-splattributes](./docs/rule/require-splattributes.md)                                             |     |     |     |     |
 | [require-valid-alt-text](./docs/rule/require-valid-alt-text.md)                                           | ✅  |     | ⌨️  |     |
+| [require-valid-named-block-naming-format](./docs/rule/require-valid-named-block-naming-format.md)         |     |     |     |     |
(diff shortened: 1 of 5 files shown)


rwjblue (commented):
I think maybe we should only operate on the "own" block APIs. Specifically, a user doesn't control the APIs to **pass** named blocks (unless it is their own component that is being invoked), but they do control their own component template's and what they use inside them. The way I see it is that the `has-block`, `has-block-params`, and `yield` API's are "own" APIs and safe to enforce (because the author of those files logically controls the caller API). On the other hand, triggering a failure on the `<:foo-bar>` side for components that you don't control will largely just be annoying (and force you to add a bunch of ignores).

@bertdeblock - What do you think?

bertdeblock:
@rwjblue Good point! Hadn't thought about it that way. Updated accordingly.

rwjblue (approved):

Issue #2071, opened 2021-08-19
Passing a random argument to `LinkTo` yields a false positive on `no-unknown-arguments-for-builtin-components`


MikiDi:
Something like
```hbs
    <LinkTo
      class="auk-search-results-list__item"
      @route={{@route}}
      @models={{this.models}}
      @random="test"
      @query={{@query}}
      ...attributes
    >
```
yields
```
Passing the "@dragOver" argument to <LinkTo /> is deprecated.
Instead, please use the {{on}} modifier, i.e. "<LinkTo {{on "dragover" ...}} />" instead of "<LinkTo @dragOver={{...}} />".  no-unknown-arguments-for-builtin-components
```
Due to the fuzzy matching going on until it finds sómething https://github.com/ember-template-lint/ember-template-lint/blob/d50db604486b444065f8919e5484d25715f06d84/lib/rules/no-unknown-arguments-for-builtin-components.js#L297

MikiDi:
If I'm not mistaken it is this https://github.com/ember-template-lint/ember-template-lint/pull/1823/files#diff-0d35c8339def5f04d4dd503125f81691f1c1fc72a726331bb90e29cdab3f6654R303 addition that makes it "return to soon" to reach the "unknown-argument"-part

bertdeblock:
Maybe the fuzzy matching could be tweaked a bit using [some of these options](https://fusejs.io/api/options.html), though I'm guessing it will be difficult to make this bullet proof.

Maybe @lifeart has any ideas if this can be improved?

lifeart:
As I remember Fuse has scoring, and we could filter results by score here

lifeart:
But, likely we should emit "unknown argument" error first here, instead of saying that something is deprecated.

lifeart:
Error should be:

`"@random" is unknown argument for <LinkTo /> component. Did you mean "@dragOver"?`;

bertdeblock:
This can be closed since #2088 was merged and released in [v3.8.0](https://github.com/ember-template-lint/ember-template-lint/blob/master/CHANGELOG.md#v380-2021-10-01).

PR #2088, opened 2021-08-31
Bugfix: fixes fuzzy messages arguments suggestion

lifeart:
fixes incorrect messaging in https://github.com/ember-template-lint/ember-template-lint/issues/2071

### DIFF
--- a/lib/rules/no-unknown-arguments-for-builtin-components.js
+++ b/lib/rules/no-unknown-arguments-for-builtin-components.js
@@ -300,7 +300,7 @@ function ERROR_MESSAGE(tagName, argumentName) {
   }
 
   if (!hasArgumentsMatch) {
-    const msg = `"${argumentName}" is unknown argument for <${tagName} /> component.`;
+    const msg = `"${argumentName}" is not a known argument for the <${tagName} /> component.`;
     if (fuzzyResults.length) {
       return `${msg} Did you mean "@${fuzzyResults[0].item}"?`;
     } else {
--- a/lib/rules/no-unknown-arguments-for-builtin-components.js
+++ b/lib/rules/no-unknown-arguments-for-builtin-components.js
@@ -40,7 +40,7 @@ const KnownArguments = {
       'disabledClass',
     ],
     deprecatedArguments: {
-      // key -> replacement attr
+      // argument name -> replacement html attr
       '@active': '',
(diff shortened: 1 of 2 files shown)


lib/rules/no-unknown-arguments-for-builtin-components.js
```diff
@@ -292,28 +292,31 @@ function ERROR_MESSAGE(tagName, argumentName) {
...
     fuzzyResults = new Fuse(candidates).search(query);
     query = query.slice(0, -1);
   }
 
-  let closestKey = fuzzyResults.length ? `@${fuzzyResults[0].item}` : argumentName;
-
-  if (closestKey in deprecatedArgs) {
-    return deprecateArgument(tagName, closestKey, deprecatedArgs[closestKey]);
+  if (!hasArgumentsMatch) {
+    const msg = `"${argumentName}" is unknown argument for <${tagName} /> component.`;
```
scalvert, line 303: I would consider changing the message to this (and its associated test):

```suggestion
    const msg = `"${argumentName}" is not a known argument for the <${tagName} /> component.`;
```

test/unit/rules/no-unknown-arguments-for-builtin-components-test.js
```diff
@@ -210,5 +210,21 @@ generateRuleTests({
...
+
+    // LinkTo unknown argument
+
+    {
+      template:
+        ' <LinkTo class="auk-search-results-list__item" @route={{@route}} @models={{this.models}} @random="test" @query={{@query}} ...attributes >Hello</LinkTo>',
+      results: [
+        {
+          message:
+            '"@random" is unknown argument for <LinkTo /> component. Did you mean "@dragOver"?',
```
rwjblue, line 222:
```suggestion
            '"@random" is not a known argument for <LinkTo /> component. Did you mean "@dragOver"?',
```

lifeart:
I think we could introduce some refactoring helping rule here, like "require-valid-component-arguments", configurable by end-user (we could create els helper to discover it).

End-users could define / restrict arguments usage per component, having same API as in this rule.

```json
{
 'MyComponent': [ 'myArgument' ]
}
```

lifeart:
node 16 ci failing because of V8 error:

image

scalvert (approved):

PR #2089, opened 2021-09-02
Ensure `no-bare-strings` checks the arguments to `{{page-title ...}}` helper

### LINKED ISSUE #1886: [no-bare-strings] Detection of constant string in page-title helper
Currently no-bare-strings does not check helpers. It would be great, if it could detect use of constant strings in selected helper arguments. For example page-title helper is currently used in ember blueprints, but is not verified by linter if it was internationalized 

Sample code which should be detected
```
{{page-title "Foo"}}
````

bertdeblock:
Feel free to close if you guys feel this isn't a valid addition to the library.

Should be marked as `breaking` as this change might introduce additional errors.

Fixes #1886.

### DIFF
--- a/lib/rules/no-bare-strings.js
+++ b/lib/rules/no-bare-strings.js
@@ -23,6 +23,7 @@
  */
 
 const createErrorMessage = require('../helpers/create-error-message');
+const { match } = require('../helpers/node-matcher');
 const Rule = require('./_base');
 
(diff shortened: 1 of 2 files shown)


bmish:
Can you retarget to `next`? Then we'll merge this for the v4 release.

bmish (approved):
Thanks!

PR #2115, opened 2021-09-30
Converting --clean-todo to a default, providing --no-clean-todo option

scalvert:
Requiring users to explicitly call `--clean-todo` after pushing an artificial error in our results causes the following:

- a mismatched count of _actual_ errors vs. artificial errors from todos that need cleaning
- extra work for the user to clean the todos when they really just want them cleaned

In light of this, we've inversed the functionality, so that the process of cleaning todos is the default, and you can opt out by using `--no-clean-todo`. We're taking advantage of yargs' auto-inverting when using `--no` on boolean options.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -15,6 +15,7 @@ const {
   validateConfig,
 } = require('@ember-template-lint/todo-utils');
 const chalk = require('chalk');
+const ci = require('ci-info');
 const getStdin = require('get-stdin');
 const globby = require('globby');
 const isGlob = require('is-glob');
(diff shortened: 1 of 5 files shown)


elwayman02 (approved):

bin/ember-template-lint.js
```diff
@@ -194,7 +194,7 @@ function parseArgv(_argv) {
       },
       'clean-todo': {
         describe: 'Remove expired and invalid todo files',
-        default: false,
+        default: true,
```
rwjblue, line 197: @scalvert - What do you think about making the default here something like:

```js
const ci = require('ci-info')

const shouldCleanTodos = !ci.isCI;
```

Then you'd do this:

```suggestion
        default: shouldCleanTodos,
```

scalvert: Ya I think this is good. This handles the case we want WRT CI. I'll make that change.

rwjblue (approved):

PR #2123, opened 2021-10-04
Add `npm-package-json-lint`

bmish:
Follow-up to #1447.

This time, we're not just using the sorting functionality of [npm-package-json-lint](https://github.com/tclindner/npm-package-json-lint) (which is also provided by [sort-package-json](https://github.com/keithamus/sort-package-json)). Importantly, we're also ensuring that our dependencies use caret version ranges.


lifeart:
Ah, errors killing me:

image

lifeart:
It's very unclear how to "order" it in proper way

bmish:
@lifeart that's not related to this PR. But you can try `yarn eslint --fix .` for that. Import ordering should be autofixable.

Issue #2114, opened 2021-09-28
Proposal for new a11y check to disallow headings with no text content


glnster:
Currently, ember-template-lint does not prevent empty headings (h1, h2, etc.). Headings relay the structure of a webpage and provide a meaningful, hierarchical order of its content. If headings are empty or its text contents is inaccessible, this could confuse users or prevent them accessing sections of interest.

This proposes the authoring of a new rule that enforces that headings have accessible text content. Headings with no text content or content that is hidden (using `aria-hidden`) should be disallowed. This is essentially a port of JSXs A11y [heading-has-content](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/heading-has-content.md) rule.

Reference: [WCAG SC 2.4.6 Headings and Labels](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-descriptive.html)

MelSumner:
We should probably also check for an element with the role of heading and an aria-level attribute. 

Make sure we reference the WCAG guidelines in the doc and not the jsx rule, though. 👍

glnster:
> We should probably also check for an element with the role of heading and an aria-level attribute.
> 
> Make sure we reference the WCAG guidelines in the doc and not the jsx rule, though. 👍

Good call-outs! Thank you. Linked to the guideline to reference from.

PR #2126, opened 2021-10-05
WIP Adds new rule `no-empty-headings`

glnster:
### WIP / DRAFT
This PR adds the new rule `no-empty-headings` which addresses #2114 .

### Background
Headings relay the structure of a webpage and provide a meaningful, hierarchical order of its content. If headings are empty or its text contents is inaccessible, this could confuse users or prevent them from accessing sections of interest.

This new rule enforces headings have accessible text content. Headings with no text content or content that is hidden (e.g. using aria-hidden) should be disallowed.

Reference: [WCAG SC 2.4.6 Headings and Labels](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-descriptive.html)

**Allowed**:
```
# examples
<h1>Accessible Heading</h1>
<h1 aria-hidden="true">Valid Heading</h1>
<div role="heading"><span>Accessible text</span></div>
```

**Forbidden**:
```
# examples
<h1></h1>
<h1> </h1>
<h1><span aria-hidden=”true”>Inaccessible text</span></h1>
<h1><span hidden>text</span></h1>
```

### Testing

```
Test Suites: 128 passed, 128 total
Time:        62.737 s
Ran all test suites.
✨  Done in 80.92s.
```

### DIFF
--- a/docs/rule/no-empty-headings.md
+++ b/docs/rule/no-empty-headings.md
@@ -1,0 +1,51 @@
+# no-empty-headings
+
+Headings relay the structure of a webpage and provide a meaningful, hierarchical order of its content. If headings are empty or its text contents are inaccessible, this could confuse users or prevent them accessing sections of interest.
+
+Disallow headings (h1, h2, etc.) with no accessible text content.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<h*></h*>
+```
+
+```hbs
+<div role="heading"></div>
+```
+
+```hbs
+<h*><span aria-hidden="true">Inaccessible text<span></h*>
+```
+
+This rule **allows** the following:
+
+```hbs
+<h*>Heading Content</h*>
+```
+
+```hbs
+<h*><span>Text</span><h*>
+```
+
+```hbs
+<div role="heading">Heading Content</div>
+```
+
+```hbs
+<h* aria-hidden="true">Heading Content</h*>
+```
+
+```hbs
+<h* hidden>Heading Content</h*>
+```
+
+## Migration
+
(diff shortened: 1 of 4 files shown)


docs/rule/no-empty-headings.md
```diff
@@ -0,0 +1,51 @@
...
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<h*></h*>
+```
+
+```hbs
+<div role="heading"></div>
```
drewlee, line 16: Looks good, but let's tighten up these examples, in that `role="heading"` needs to be accompanied with `aria-level="{\d}"`. It would probably be worth checking w/screen readers to see whether `role="heading"` without `aria-level` is still identified as a heading (it may default to a specific level, IIRC). If it isn't being recognized, we are probably safe to ignore from being validated.

lib/rules/no-empty-headings.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const AstNodeInfo = require('../helpers/ast-node-info');
+const Rule = require('./_base');
+
+const ERROR_MESSAGE = 'Heading (h1, h2, etc.) must contain accessible text content.';
```
drewlee, line 6: This message should probably be more general to also encompass an ARIA heading, or output a different message for that use case.

drewlee:
I think we'd also need to update https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/a11y.js

docs/rule/no-empty-headings.md
glnster: VO+Safari defaults it to heading level 2. I haven't checked other screen readers yet, but I think for completeness I will add the aria-level attrib in anyway.

lib/rules/no-empty-headings.js
glnster: I updated this message to include ARIA:heading role elements.

PR #2127, opened 2021-10-07
Add `allowEmptyLinks` option (default `true`) to `no-invalid-link-text` rule

### LINKED ISSUE #2113: Proposal for new a11y rule to disallow empty anchor elements
Currently, ember-template-lint disallows unhelpful link text such as "click here" via the [no-invalid-link-text](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-link-text.md) rule. However, there is nothing in place to prevent empty link text, which is even more harmful. When a link contains no content, the link's `href` ends up as the accessible name. This results in a poor experience and requires the user to resort to trial and error to determine the purpose of the link.

This proposes the authoring of a new rule that disallows empty anchor elements and `LinkTo` helpers. Hidden links should be ignored, while alternatively labeled links (using `aria-label` or `aria-labelledby`) should pass. This is essentially a port of JSXs A11y [anchor-has-content](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-has-content.md) rule.

glnster:
This PR augments the existing `no-invalid-link-text` rule to also handle empty anchor element cases.
This fixes #2113.

### Background
`ember-template-lint` disallows unhelpful link text such as "click here" via the `no-invalid-link-text` rule. However, there is nothing in place to prevent empty link text, which is even more harmful. When a link contains no content, the link's `href` ends up as the accessible name. This results in a poor experience and requires the user to resort to trial and error to determine the purpose of the link.

When the `allowEmptyLinks` rule option is set to `false` (default is `true`), the additional checks implemented here disallow links that do not have accessible text content. Hidden links are ignored, and alternatively labeled links (using `aria-label` or `aria-labelledby`) are allowed/checked.

**Allowed**
```
<a href={{link}} aria-hidden="true">Read more about semantic html</a>
<a href={{link}} hidden>Read more about semantic html</a>
<a href={{link}} aria-label="Read more about semantic html">Read more...</a>
<a href={{link}} aria-labelledby="some-id">Read more...</a>
```

**Forbidden**
```
<LinkTo></LinkTo>
<a href={{link}}></a>
<a href={{link}}> </a>
<a href={{link}} aria-labelledby="">Click here</a>
<a href={{link}} aria-label="Click here">Some other text</a>
```

### Testing
```
Test Suites: 127 passed, 127 total
Time:        45.708 s, estimated 47 s
Ran all test suites.
✨  Done in 55.11s.
```

### DIFF
--- a/docs/rule/no-invalid-link-text.md
+++ b/docs/rule/no-invalid-link-text.md
@@ -20,7 +20,7 @@ This rule **forbids** the following:
 <a href={{link}}>more</a>
 ```
 
-*With `allowEmptyLinks = false`*
+*With `allowEmptyLinks: false`*
 
 ```hbs
 <LinkTo></LinkTo>
@@ -40,11 +40,17 @@ This rule **allows** the following:
 <a href={{link}}>Read more about semantic html</a>
 ```
 
-*With `allowEmptyLinks = false`*
+*With `allowEmptyLinks: false`*
 
 ```hbs
 <a href={{link}} aria-label="Read more about semantic html">...</a>
 ```
+
+## Configuration
+
+* boolean -- `true` for enabled / `false` for disabled
+* object -- Containing the following values:
+  * `allowEmptyLinks` - When `false`, empty links are not allowed. Defaults to `true`.
 
 ## References
 


bmish:
Protip: you'll want to say `fixes #123` in the PR description in order to properly link the PR to the issue.

bmish:
The change looks reasonable. However, as currently implemented, this would be a breaking change. My recommendation is that we move this new functionality behind a rule option `allowEmptyLinks` (default `true`). In the next major version, we can change the default to `false` or even remove the option.

glnster:
Thanks! Minor correction in the title I think? The default is `true`.

PR #2143, opened 2021-10-15
Add autofixer to `quotes` rule

courajs:
Waiting on:
- [x] [Merging quote-changing support to `ember-template-recast`](https://github.com/ember-template-lint/ember-template-recast/pull/653)
- [x] Release `ember-template-recast` including ^
- [x] Upgrade `ember-template-recast` version to `6.1.0`
- [x] Fixing the acceptance test which asserts on remaining unfixable rule violations. Of course I can just change this to have a different rule violation which doesn't currently have a fixer. But it seems better to have some always-failing rule for test purposes. Does anyone have ideas on the best way to add that? Preferably we wouldn't have to add an actual rule to `lib/rules`

Note: this won't autofix when escaping is required, blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698.

### DIFF
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -55,6 +55,7 @@ module.exports = class Quotes extends Rule {
       AttrNode(node) {
         if (!node.isValueless && node.quoteType === badChar) {
           if (attrValueHasChar(node.value, goodChar)) {
+            // TODO: Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
             return this.log({
               message,
               node,
@@ -77,6 +78,7 @@ module.exports = class Quotes extends Rule {
 
         if (node.quoteType === badChar) {
           if (node.value.includes(goodChar)) {
+            // TODO: Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
             return this.log({
               message,
               node,
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -54,6 +54,12 @@ module.exports = class Quotes extends Rule {
     return {
       AttrNode(node) {
         if (!node.isValueless && node.quoteType === badChar) {
+          if (attrValueHasChar(node.value, goodChar)) {
+            return this.log({
+              message,
+              node,
+            });
+          }
           if (this.mode === 'fix') {
             node.quoteType = goodChar;
           } else {
@@ -70,6 +76,13 @@ module.exports = class Quotes extends Rule {
         let errorSource = this.sourceForNode(path.parentNode);
 
         if (node.quoteType === badChar) {
+          if (node.value.includes(goodChar)) {
+            return this.log({
+              message,
+              node,
+              source: errorSource,
+            });
+          }
           if (this.mode === 'fix') {
             node.quoteType = goodChar;
           } else {
@@ -85,3 +98,14 @@ module.exports = class Quotes extends Rule {
     };
   }
 };
+
+function attrValueHasChar(node, ch) {
+  if (node.type === 'TextNode') {
+    return node.chars.includes(ch);
+  } else if (node.type === 'ConcatStatement') {
+    return node.parts.some((n) => {
+      return n.type === 'TextNode' && n.chars.includes(ch);
+    });
+  }
+  return false;
+}
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -36,14 +36,17 @@ module.exports = class Quotes extends Rule {
 
   visitor() {
     let badChar;
+    let goodChar;
     let message;
     switch (this.config) {
       case 'double':
         badChar = "'";
+        goodChar = '"';
         message = 'you must use double quotes in templates';
         break;
       case 'single':
         badChar = '"';
+        goodChar = "'";
         message = 'you must use single quotes in templates';
         break;
     }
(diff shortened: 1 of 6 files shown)


dcyriller:
@courajs it is now possible to target `next` to leverage ember-template-recast@6.1.1.

bmish:
@dcyriller I assume you mean ember-template-recast@6.1.0.

dcyriller:
6.1.0 also contains a bugfix, hence the patch version in my mind. But you're totally right. I have updated my comment.

courajs:
Ugh, fixing these acceptance tests is turning into a pain. But I'm determined! I will return to this later this week (•̀ᴗ•́)و

courajs:
I declare this done!

test/acceptance/public-api-test.js
```diff
@@ -286,31 +286,54 @@ describe('public api', function () {
...
         source: templateContents,
         filePath: templatePath,
         moduleId: templatePath.slice(0, -4),
       });
 
-      expect(result.messages).toEqual(expected);
-      expect(result.output).toEqual(templateContents);
-      expect(result.isFixed).toEqual(false);
+      expect(result).toMatchInlineSnapshot(
+        { messages: [{ filePath: expect.any(String) }] },
```
bmish, line 316: This line looks wrong. I believe `toMatchInlineSnapshot` should only have one argument.

test/acceptance/public-api-test.js
```diff
@@ -1394,31 +1417,54 @@ describe('public api', function () {
...
```
bmish, line 1447: This line looks wrong. I believe `toMatchInlineSnapshot` should only have one argument.

test/unit/rules/quotes-test.js
```diff
@@ -146,6 +169,7 @@ generateRuleTests({
     {
       config: 'single',
       template: '<input type="checkbox">',
```
bmish, line 171: Can we add test cases which combinations of different quote types?

```
<img alt="john's house">
```

I believe we would need to escape the quotes inside for this to fix properly:

```
<img alt='john\'s house'>
```

test/unit/rules/quotes-test.js
```diff
@@ -33,9 +33,23 @@ generateRuleTests({
   ],
 
   bad: [
+    {
+      config: 'double',
+      template: "{{component 'one {{thing}} two'}}",
+      fixedTemplate: '{{component "one {{thing}} two"}}',
+
+      result: {
```
bmish, line 41: Can we use `verifyResults` instead so we can have a convenient snapshot?

bmish (changes requested):

test/acceptance/public-api-test.js
courajs: This is a jest feature called "property matchers" - it compensates for things in the snapshot that can change without causing a failure. In this case, the temporary file location is different every run, so this tells jest to just make sure they are both strings.
https://jestjs.io/docs/snapshot-testing#property-matchers

bmish: Oh, neat!

test/acceptance/public-api-test.js
bmish: Nevermind, see https://github.com/ember-template-lint/ember-template-lint/pull/2143/files#r758654423

test/unit/rules/quotes-test.js
courajs: Oof, this is an excellent point. We'll actually need escape via html entities: https://html.spec.whatwg.org/dev/syntax.html#attributes-2

courajs: I think this [would be better fixed in `ember-template-recast`](https://github.com/ember-template-lint/ember-template-recast/issues/698)

test/unit/rules/quotes-test.js
courajs: done

courajs:
@bmish you raised an excellent issue about escapes, which I'm not really motivated to fix right now 😅 
Thanks for the review, but I think I'm gonna step away from this issue for now. Feel free to merge as-is, or close this PR.
If anyone else wants to get this over the finish line, you'd need to fix [this issue in `ember-template-recast`](https://github.com/ember-template-lint/ember-template-recast/issues/698)

bmish:
@courajs how about you just disable the autofix for this special case now? We can still fix in the simple case where there's no need to do escaping. Then that more advanced case can be handled as a lower-priority follow-up.

bmish (approved):
Nice work!

lib/rules/quotes.js
```diff
@@ -36,40 +36,76 @@ module.exports = class Quotes extends Rule {
...
 
     return {
       AttrNode(node) {
         if (!node.isValueless && node.quoteType === badChar) {
-          return this.log({
-            message,
-            node,
-          });
+          if (attrValueHasChar(node.value, goodChar)) {
+            return this.log({
```
bmish, line 58:
```suggestion
            // Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
            return this.log({
```

lib/rules/quotes.js
```diff
@@ -36,40 +36,76 @@ module.exports = class Quotes extends Rule {
...
 
-        if (source[0] === badChar) {
-          return this.log({
-            message,
-            node,
-            source: errorSource,
-          });
+        if (node.quoteType === badChar) {
+          if (node.value.includes(goodChar)) {
+            return this.log({
```
bmish, line 80:
```suggestion
            // Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
            return this.log({
```

test/unit/rules/quotes-test.js
```diff
@@ -165,6 +201,72 @@ generateRuleTests({
         `);
       },
     },
+    {
+      config: 'single',
```
bmish, line 205:
```suggestion
      // Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
      config: 'single',
```

test/unit/rules/quotes-test.js
```diff
@@ -165,6 +201,72 @@ generateRuleTests({
...
+              "rule": "quotes",
+              "severity": 2,
+              "source": "class='a \\"so-called\\" btn {{this.otherClass}}'",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      config: 'single',
```
bmish, line 249:
```suggestion
      // Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
      config: 'single',
```

test/unit/rules/quotes-test.js
```diff
@@ -165,6 +201,72 @@ generateRuleTests({
...
+              "rule": "quotes",
+              "severity": 2,
+              "source": "alt=\\"Abdul's house\\"",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      config: 'double',
```
bmish, line 227:
```suggestion
      // Autofix blocked on: https://github.com/ember-template-lint/ember-template-recast/issues/698
      config: 'double',
```

bmish (approved):

PR #2149, opened 2021-10-18
make sure "good" templates can actually fail in test harness

mansona:
We noticed an issue with our internal ember-template-lint plugin that is using the in-built test harness where there was no way to make a "good" template fail 🙈  

I linked this into our plugin locally and it fixes the issue, but I don't 100% know how to test this in ember-template-lint 🤔 if anyone has any ideas please let m know and I can give it a go.

## Extra information

I first tried this following code in the "good" templates: 

```
good: [
    {
      template: `<BadComponent @value="with config true" />`,
    },
    `{{bad-component value="non object"}}`,
  ],
  ```

**note:** the lint rule is to just essentially deprecating the `<BadComponent />` so it should fail whenever it sees that component. 

This code passes even though it shouldn't.

I then tried this: 

```
  good: [
    {
      config: true,
      template: `<BadComponent @value="with config true" />`,
    },
    {
      template: `<BadComponent @value="without config true" />`,
    },
    `{{bad-component value="non object"}}`,
  ],
```

and now **all three tests fail** as expected. So there is something odd where the `config` value is being saved between runs 🤔

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -147,5 +147,14 @@ module.exports = {
         'import/no-dynamic-require': 'off',
       },
     },
+    {
+      files: ['mocha-tests/**/*.js'],
+      env: {
+        mocha: true,
+      },
+      rules: {
+        'import/no-dynamic-require': 'off',
+      },
+    },
   ],
 };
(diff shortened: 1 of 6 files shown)


rwjblue:
@mansona - Have a chance to check that?

PR #2151, opened 2021-10-21
Add `endLine` / `endColumn` to generated rule test results

scalvert:
Implements the final part of #2072.


scalvert:
@rwjblue and I are discussing whether this is breaking or not - leaning towards yes.

PR #2164, opened 2021-11-02
Noops the console when using the SARIF formatter

scalvert:
We noop the console when either using `--quiet` or the `json` option, but neglect to do so when using formatters that output JSON. This can cause issues when redirecting stdout to a file, such as a JSON or SARIF file, because of deprecation warnings that are emitted, causing invalid JSON (console output is additionally redirected, and coupled with JSON output is written together to a file).

This fix ensures that we additionally noop the console when using either of the `json` or `sarif` formatters via the `--format` option. This ensures that if we redirect stdout that it's valid JSON.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -347,7 +347,10 @@ async function run() {
       config,
       rule: options.rule,
       allowInlineConfig: !options.noInlineConfig,
-      console: options.quiet || options.json || options.format === 'sarif' ? NOOP_CONSOLE : console,
+      console:
+        options.quiet || options.json || ['sarif', 'json'].includes(options.format)
+          ? NOOP_CONSOLE
+          : console,
     });
   } catch (error) {
     console.error(error.message);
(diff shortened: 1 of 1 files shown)


elwayman02 (approved):

bin/ember-template-lint.js
```diff
@@ -347,7 +347,10 @@ async function run() {
       config,
       rule: options.rule,
       allowInlineConfig: !options.noInlineConfig,
-      console: options.quiet || options.json ? NOOP_CONSOLE : console,
+      console:
+        options.quiet || options.json || ['sarif', 'json'].includes(options.format)
```
rwjblue, line 351: Can you hoist this to a variable or two above? This is getting somewhat hard to reason about...

rwjblue (approved):

bin/ember-template-lint.js
scalvert: Some might say it's unreasonable :)

PR #2165, opened 2021-11-03
Adding ruleId to deprecation message to track its origin

scalvert:
As part of #2072 a deprecation warning was added to the base rule's `log` method. This warning was emitted when a callee invoked the method, which indicated that usage without passing the AST node was deprecated. What it neglected to do is identify the source rule, which would help track the invalid usage back so it could be fixed. 

This PR adds the ruleId to the message so it can be identified and fixed.

### DIFF
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -524,7 +524,7 @@ module.exports = class {
 
     if (!result.node) {
       if (!loggedRules.has(this.ruleName)) {
-        let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result. Rule ID: ${this.ruleName}.`;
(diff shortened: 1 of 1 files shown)


lib/rules/_base.js
```diff
@@ -524,7 +524,7 @@ module.exports = class {
 
     if (!result.node) {
       if (!loggedRules.has(this.ruleName)) {
-        let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result.`;
+        let message = `Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result. Rule ID: ${this.ruleName}.`;
```
rwjblue, line 527:
```suggestion
        let message = `ember-template-lint: (${this.ruleName}) Calling the log method without passing the node is deprecated. Please ensure you pass node in the log method's result.`;
```

PR #2167, opened 2021-11-03
Slightly loosening requires for log method parameter requirements

scalvert:
This change refines the requirements for calls to the base rule's log method, which previously required the node itself, but now only requires the node if all location properties aren't provided.

This addresses situations where it may not be as clear what the node is that's being passed in, or when the node is the `body` itself.

### DIFF
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -518,7 +518,7 @@ module.exports = class {
       severity: this.severity,
     };
     let hasAllLocProps = ['line', 'column', 'endLine', 'endColumn'].every((prop) => prop in result);
-    debugger;
+
     if (this.filePath) {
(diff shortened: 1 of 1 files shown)


lib/rules/_base.js
```diff
@@ -517,14 +517,16 @@ module.exports = class {
       rule: this.ruleName,
       severity: this.severity,
     };
-
+    let hasAllLocProps = ['line', 'column', 'endLine', 'endColumn'].every((prop) => prop in result);
+    debugger;
```
rwjblue, line 521: 💀

scalvert: Already gone

rwjblue (approved):

PR #2168, opened 2021-11-03
Add new rule `no-autofocus-attribute`

### LINKED ISSUE #2166: Proposal for new a11y check to disallow use of the autofocus attribute
The autofocus attribute is a global attribute that indicates an element should be focused on page load. Autofocus reduces accessibility by moving users to an element without warning and context. Screen reader users may be disoriented by the unexpected navigation, and users of  small devices could miss important information preceding the focused element.

This proposes the authoring of a new rule that disallows use of the autofocus attribute. This is essentially a port of JSXs A11y [no-autofocus](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-autofocus.md) rule.

Reference: [WCAG SC 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html)

judithhinlung:
This PR adds the new rule `no-autofocus-attribute` which fixes #2166 .

### Background
The autofocus attribute is a global attribute that indicates an element should be focused on page load. Autofocus reduces accessibility by moving users to an element without warning and context. Screen reader users may be disoriented by the unexpected navigation, and users of small devices could miss important information preceding the focused element.

This new rule enforces disallowing the use of the autofocus attribute.

Reference: [WCAG SC 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html)

**Allowed**:

```
# examples
<input />
```

**Forbidden**:

```
# examples
<autofocus />
<input autofocus="true" />
```

### Testing
```
Test Suites: 129 passed, 129 total
Tests:       6 skipped, 6979 passed, 6985 total
Snapshots:   877 passed, 877 total
Time:        44.092 s
Ran all test suites.
✨ Done in 53.23s.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Each rule has emojis denoting:
 | [no-arguments-for-html-elements](./docs/rule/no-arguments-for-html-elements.md)                           | ✅  |     |     |     |
 | [no-aria-hidden-body](./docs/rule/no-aria-hidden-body.md)                                                 | ✅  |     | ⌨️  | 🔧  |
 | [no-attrs-in-components](./docs/rule/no-attrs-in-components.md)                                           | ✅  |     |     |     |
-| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                                     |     |     | ⌨️  |     |
+| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                           |     |     | ⌨️  |     |
 | [no-bare-strings](./docs/rule/no-bare-strings.md)                                                         |     |     |     |     |
 | [no-block-params-for-html-elements](./docs/rule/no-block-params-for-html-elements.md)                     | ✅  |     |     |     |
 | [no-capital-arguments](./docs/rule/no-capital-arguments.md)                                               |     |     |     |     |
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ Each rule has emojis denoting:
 | [no-arguments-for-html-elements](./docs/rule/no-arguments-for-html-elements.md)                           | ✅  |     |     |     |
 | [no-aria-hidden-body](./docs/rule/no-aria-hidden-body.md)                                                 | ✅  |     | ⌨️  | 🔧  |
 | [no-attrs-in-components](./docs/rule/no-attrs-in-components.md)                                           | ✅  |     |     |     |
+| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                           | ✅  |     |     |     |
 | [no-bare-strings](./docs/rule/no-bare-strings.md)                                                         |     |     |     |     |
 | [no-block-params-for-html-elements](./docs/rule/no-block-params-for-html-elements.md)                     | ✅  |     |     |     |
 | [no-capital-arguments](./docs/rule/no-capital-arguments.md)                                               |     |     |     |     |
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Each rule has emojis denoting:
 | [no-arguments-for-html-elements](./docs/rule/no-arguments-for-html-elements.md)                           | ✅  |     |     |     |
 | [no-aria-hidden-body](./docs/rule/no-aria-hidden-body.md)                                                 | ✅  |     | ⌨️  | 🔧  |
 | [no-attrs-in-components](./docs/rule/no-attrs-in-components.md)                                           | ✅  |     |     |     |
-| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                           | ✅  |     |     |     |
+| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                                     |     |     | ⌨️  |     |
 | [no-bare-strings](./docs/rule/no-bare-strings.md)                                                         |     |     |     |     |
 | [no-block-params-for-html-elements](./docs/rule/no-block-params-for-html-elements.md)                     | ✅  |     |     |     |
 | [no-capital-arguments](./docs/rule/no-capital-arguments.md)                                               |     |     |     |     |
--- a/docs/rule/no-autofocus-attribute.md
+++ b/docs/rule/no-autofocus-attribute.md
@@ -1,7 +1,5 @@
 # no-autofocus-attribute
 
-✅ The `extends: 'recommended'` property in a configuration file enables this rule.
-
 The autofocus attribute is a global attribute that indicates an element should be focused on page load. Autofocus reduces accessibility by moving users to an element without warning and context. Its use should be limited to form fields that serve as the main purpose of the page, such as the search input on the Google home page.
 
 This rule takes no arguments.
@@ -21,7 +19,7 @@ This rule **forbids** the following:
 ```
 
 ```hbs
-<input type="password" autofocus="true" />
+<input type="password" autofocus="autofocus" />
 ```
 
 ## References
--- a/docs/rule/no-autofocus-attribute.md
+++ b/docs/rule/no-autofocus-attribute.md
@@ -1,0 +1,30 @@
+# no-autofocus-attribute
+
+🔧 The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+
+Enforce no autofocus prop on element. The autofocus attribute is a global attribute that indicates an element should be focused on page load. Because autofocus reduces accessibility by moving users to an element without warning and context, it should not be used.
+
+This rule takes no arguments.
+
+## Examples
+
+This rule **allows** the following:
+
+```hbs
+<input />
+```
+
+This rule **forbids** the following:
+
+```hbs
+<input autofocus />
+```
+
+```hbs
+<input autofocus='true' />
+```
+
+## References
+
+- [MDN Web](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
+- [Focus Order: Understanding SC 2.4.3](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html)
--- a/docs/rule/no-autofocus-attribute.md
+++ b/docs/rule/no-autofocus-attribute.md
@@ -1,8 +1,8 @@
 # no-autofocus-attribute
 
-🔧 The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+✅ The `extends: 'recommended'` property in a configuration file enables this rule.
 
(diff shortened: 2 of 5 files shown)


test/unit/rules/no-autofocus-attribute-test.js
```diff
@@ -0,0 +1,98 @@
+'use strict';
+
+const generateRuleTests = require('../../helpers/rule-test-harness');
+
+generateRuleTests({
+  name: 'no-autofocus-attribute',
+
+  config: true,
+
+  good: ['<input />'],
```
bmish, line 10: Need more valid variations, e.g. `input notAutofocus=true`.

lib/rules/no-autofocus-attribute.js
```diff
@@ -0,0 +1,30 @@
...
+const errorMessage =
+  'No autofocus attribute allowed, as it reduces accessibility by moving users to an element without warning and context';
+
+module.exports = class NoAutofocusAttribute extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        const autofocusNode = AstNodeInfo.findAttribute(node, 'autofocus');
+        if (autofocusNode) {
+          if (this.mode === 'fix') {
```
bmish, line 15: Unfortunately, it's not safe to autofix this rule by deleting the attribute, as that would cause behavior changes. Autofixes are not supposed to cause behavior changes. This fix would be more appropriate for an [auto-suggestion](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) which supported by ESLint but unfortunately ember-template-lint doesn't have such a feature.

docs/rule/no-autofocus-attribute.md
```diff
@@ -0,0 +1,30 @@
+# no-autofocus-attribute
+
+🔧 The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+
+Enforce no autofocus prop on element. The autofocus attribute is a global attribute that indicates an element should be focused on page load. Because autofocus reduces accessibility by moving users to an element without warning and context, it should not be used.
```
bmish, line 5: I would avoid the `prop` abbreviation here so that's a bit unclear.

docs/rule/no-autofocus-attribute.md
```diff
@@ -0,0 +1,30 @@
...
+
+This rule takes no arguments.
+
+## Examples
+
+This rule **allows** the following:
```
bmish, line 11: I've used `autofocus` to get focus onto a central search textbox on a web page upon page load so the user can quickly begin typing, just like how google.com uses it on its big search box. I assume this is a very common pattern, so we should probably mention use this use case and offer suggestions for it?

bmish (changes requested):

README.md
```diff
@@ -107,6 +107,7 @@ Each rule has emojis denoting:
 | [no-arguments-for-html-elements](./docs/rule/no-arguments-for-html-elements.md)                           | ✅  |     |     |     |
 | [no-aria-hidden-body](./docs/rule/no-aria-hidden-body.md)                                                 | ✅  |     | ⌨️  | 🔧  |
 | [no-attrs-in-components](./docs/rule/no-attrs-in-components.md)                                           | ✅  |     |     |     |
+| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                           | ✅  |     |     |     |
```
ckundo, line 110: this should include the emoji for accessibility.

bertdeblock: And the ✅ icon should go for now I believe, since it's not a recommended rule.

test/unit/rules/no-autofocus-attribute-test.js
```diff
@@ -0,0 +1,98 @@
...
+              "rule": "no-autofocus-attribute",
+              "severity": 2,
+              "source": "autofocus=\\"true\\"",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      template: '<input autofocus={{false}} />',
```
bertdeblock, line 56: FYI, this won't render the `autofocus` attribute and thus not cause the a11y issue described. Not sure if this matters?

docs/rule/no-autofocus-attribute.md
```diff
@@ -0,0 +1,31 @@
+# no-autofocus-attribute
+
+✅ The `extends: 'recommended'` property in a configuration file enables this rule.
```
rwjblue: We should remove this line (we can't include a new rule in `recommended` until a major version bump release).

test/unit/rules/no-autofocus-attribute-test.js
```diff
@@ -0,0 +1,103 @@
...
+  config: true,
+
+  good: [
+    '<input />',
+    '<input type="text" disabled="true" />',
+    '<input type="password" disabled={{false}} />',
+    '{{input type="text" disabled=true}}',
+  ],
+
+  bad: [
```
rwjblue, line 17: Would you mind adding some `good` and `bad` examples with elements _other_ than `<input />`? This should also include some examples using component invocation (e.g. `<AwesomeButton autofocus />`

rwjblue (changes requested):

test/unit/rules/no-autofocus-attribute-test.js
rwjblue: Additionally, since DOM attributes are case insensitive it would be good to have a `bad` example for `<input autoFocus>`

rwjblue:
Overall 👍👍 on this, thanks for working on it @judithhinlung!

judithhinlung:
I have updated this PR with the following changes:
1. errorMessage has been changed to ERROR_MESSAGE.
2. Updated the README with the accessibility emoji and removed the ✅.
3. Added guidance for limiting usage of autofocus in docs.
4. Removed autofix.
5. Removed the invalid test and added more valid variations.

Coming soon: add test cases for other elements, curly-brace component invocations, and invocations using a component helper.

docs/rule/no-autofocus-attribute.md
```diff
@@ -0,0 +1,29 @@
+# no-autofocus-attribute
+
+The autofocus attribute is a global attribute that indicates an element should be focused on page load. Autofocus reduces accessibility by moving users to an element without warning and context. Its use should be limited to form fields that serve as the main purpose of the page, such as the search input on the Google home page.
+
+This rule takes no arguments.
```
bmish, line 5: No need to state this.

```suggestion
```

bmish, line 5:
```suggestion

```

docs/rule/no-autofocus-attribute.md
bmish: Follow-up question, due to this use case, would this rule ever be suitable to be `recommended`? CC: @rwjblue

README.md
```diff
@@ -107,6 +107,7 @@ Each rule has emojis denoting:
 | [no-arguments-for-html-elements](./docs/rule/no-arguments-for-html-elements.md)                           | ✅  |     |     |     |
 | [no-aria-hidden-body](./docs/rule/no-aria-hidden-body.md)                                                 | ✅  |     | ⌨️  | 🔧  |
 | [no-attrs-in-components](./docs/rule/no-attrs-in-components.md)                                           | ✅  |     |     |     |
+| [no-autofocus-attribute](./docs/rule/no-autofocus-attribute.md)                                                     |     |     | ⌨️  |     |
```
bmish, line 110: `yarn update:readme` should automatically fix this.

test/unit/rules/no-autofocus-attribute-test.js
```diff
@@ -0,0 +1,191 @@
...
+
+  good: [
+    '<input />',
+    '<input type="text" disabled="true" />',
+    '<input type="password" disabled={{false}} />',
+    '{{input type="text" disabled=true}}',
+    '{{component "input" type="text" disabled=true}}',
+    '<div></div>',
+    '<h1><span>Valid Heading</span></h1>',
+    '<CustomComponent />',
```
bmish, line 18:
```suggestion
    '<CustomComponent />',
    '<CustomComponent disabled />',
    '<CustomComponent disabled=true />',
```

test/unit/rules/no-autofocus-attribute-test.js
```diff
@@ -0,0 +1,191 @@
...
+
+generateRuleTests({
+  name: 'no-autofocus-attribute',
+
+  config: true,
+
+  good: [
+    '<input />',
+    '<input type="text" disabled="true" />',
+    '<input type="password" disabled={{false}} />',
```
bmish, line 13: another test case 

```suggestion
    '<input type="password" disabled={{false}} />',
    '<input type="password" disabled />',
```

bmish (approved):

bmish:
LGTM, is this still a draft?

bmish:
Will merge and release this in the next day if no other comments.

docs/rule/no-autofocus-attribute.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+This rule **forbids** the following:
+
+```hbs
+<input autofocus />
+```
+
+```hbs
+<input autofocus='true' />
```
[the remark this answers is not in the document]
MrChocolatine: Why not simply `<input autofocus>`? (I know the PR was merged)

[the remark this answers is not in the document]
bertdeblock: Because `autofocus="autofocus"` is also a valid way of defining the attribute and we also show `autofocus` in the docs as well:
https://github.com/ember-template-lint/ember-template-lint/pull/2168/files#diff-8ce4ad24f0b3c9100092b7fa447f7f763c1867cdd986320c798b0742ff28d275R16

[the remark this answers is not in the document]
MrChocolatine: Oh dang... I am sorry, I did not see it was also shown on the document a few lines above 😑 .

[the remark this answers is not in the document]
bertdeblock: No worries 😃

docs/rule/no-autofocus-attribute.md
MelSumner: disallowing autofocus should be a recommended rule.

bmish: @MelSumner I see. For the above use case I mentioned, do we just recommend people inline disable the rule for that?

MelSumner: Yes, that's the recommendation for all A11y rules that might have an exception that the author feels is valid.

PR #2020, opened 2021-07-01
Add `--fix` support for `eol-last` rule

rwjblue:

PR #2174, opened 2021-11-09
Fix false negative with `always` setting and template ending with non-text in `eol-last` rule

### LINKED ISSUE #2173: 'eol-last': 'always' is not taking into account with version 3.8.0 while it works 3.7.0
We have a very simple config:
```
  extends: 'recommended',

  rules: {
    'eol-last': 'always'
  }
```

using 3.7.0 we get an error (as expected)
Screen Shot 2021-11-09 at 9 01 23 AM

3.8.0 and up version passes without error (unexpected should fails)

The other rules seem to work as expected.

bmish:
Fixes #2173.

This was broken when an autofixer was added to this rule in https://github.com/ember-template-lint/ember-template-lint/pull/2020

### DIFF
--- a/lib/rules/eol-last.js
+++ b/lib/rules/eol-last.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { builders: b } = require('ember-template-recast');
+
 const createErrorMessage = require('../helpers/create-error-message');
 const Rule = require('./_base');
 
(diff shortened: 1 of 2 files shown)


lib/rules/eol-last.js
```diff
@@ -71,15 +71,21 @@ module.exports = class EolLast extends Rule {
...
               lastNode.chars = lastNode.chars.slice(0, -1);
-            } else if (!lastNode.chars.endsWith('\n') && this.config === 'always') {
+            } else if (
+              lastNode.type === 'TextNode' &&
+              !lastNode.chars.endsWith('\n') &&
+              this.config === 'always'
+            ) {
               lastNode.chars = `${lastNode.chars}\n`;
+            } else if (lastNode.type !== 'TextNode' && this.config === 'always') {
+              // TODO: add new TextNode with empty line to end of body
```
bmish, line 88: Will update PR once I figure out how to create a new `TextNode` with `\n` to add to the end of the file body.

dcyriller (approved):

PR #2176, opened 2021-11-15
Require Node `^12.22.0 || ^14.17.0 || >=16.0.0`

dcyriller:
This PR targets the `next` branch.

Node 10 and 15 are not supported anymore.

Usually we would have set the new Nodejs support to `12* || 14* || >=16`. But this time, publising ESM modules is possible provided a node version in the range `^12.22.0 || ^14.17.0 || >=16.0.0` . Hence the range proposed in this PR.


package.json
```diff
@@ -123,7 +123,7 @@
     }
   },
   "volta": {
-    "node": "12.18.3",
+    "node": "12.22.7",
```
bmish, line 126: If we're gonna have volta in here, let's have it use Node 16 for development?

bmish (approved):

PR #2184, opened 2021-11-15
Remove `common-tags` devDependency

dcyriller:
It was really used only once in test/unit/recommended-config-test.js. Maybe it be removed?


bmish:
I do think it can be valuable to have some ability to strip indentation from template literal strings. For example, in eslint-plugin-unicorn, we use [outdent](https://www.npmjs.com/package/outdent) very heavily for the same purpose.

dcyriller:
Thanks for your feedback, I'm closing it then.

PR #2193, opened 2021-11-17
Remove `--json` CLI option

bmish:
Follow-up to the deprecation of this option in #1907.

Use `--format=json` instead.

Part of v4 release (#1908).

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -149,11 +149,6 @@ function parseArgv(_argv) {
         type: 'string',
         default: 'pretty',
       },
-      json: {
-        describe: 'Format output as json',
-        deprecated: 'Use --format=json instead',
-        boolean: true,
-      },
       'output-file': {
         describe: 'Specify file to write report to',
         type: 'string',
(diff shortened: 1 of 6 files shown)


bin/ember-template-lint.js
```diff
@@ -265,7 +260,7 @@ function printPending(results, options) {
   }
   let pendingListString = JSON.stringify(pendingList, null, 2);
 
-  if (options.json) {
+  if (options.format === 'json') {
```
bmish, line 263: This is a bug fix. Previously, `--print-pending` only worked with `--json` but not `--format=json`.

bmish: CC: @scalvert

scalvert: Thank you.

scalvert (approved):

bin/ember-template-lint.js
bmish: Extracting the fix to master: https://github.com/ember-template-lint/ember-template-lint/pull/2210

Although it turns out both `--json` and `--print-pending` will be removed in v4.

PR #2196, opened 2021-11-17
Resolve `no-unused-disable` lint errors

bertdeblock:
Resolves:
```
'import/extensions' rule is disabled but never reported
```


bertdeblock:
Seems I'm derping? Locally, ESLint reported that these disable comments were not needed anymore.

PR #2207, opened 2021-11-19
Remove `--print-pending` CLI option

bmish:
```
WARNING: Print pending is deprecated. Use --update-todo instead.
```

Part of v4 release (#1908).


bmish:
Feedback welcome even after this is merged.

scalvert:
Thanks for doing this! I'm going to rip out all the pending functionality as part of 4.0.0 too. Will add it to the list in #1908

PR #2208, opened 2021-11-19
Add `exports` to package.json

bmish:
https://nodejs.org/api/packages.html#exports

This prevents users from importing files directly from this package.

[ESLint v8](https://eslint.org/docs/user-guide/migrating-to-8.0.0#the-lib-entrypoint-has-been-removed) also added this restriction.

Ideally, we'll hear from users after making this change if there are additional internals that we need to expose as part of our public NodeJS API in [lib/index.js](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/index.js).

Part of v4 release (#1908).

TODO:

- [ ] #2209
- [ ] What to do about users who may need to retrieve the default config from [no-bare-strings](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md)? `const { DEFAULT_CONFIG } = require('ember-template-lint/lib/rules/no-bare-strings');` #2213

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   },
   "license": "MIT",
   "author": "Robert Jackson <me@rwjblue.com>",
-  "main": "lib/index.js",
+  "exports": "./lib/index.js",
+  "main": "./lib/index.js",
   "bin": "./bin/ember-template-lint.js",
   "scripts": {
     "lint": "npm-run-all lint:* --continue-on-error",


package.json
```diff
@@ -19,7 +19,8 @@
   },
   "license": "MIT",
   "author": "Robert Jackson <me@rwjblue.com>",
-  "main": "lib/index.js",
+  "exports": "./lib/index.js",
```
dcyriller, line 22: I think we will need to expose `lib/helpers/rule-test-harness.js` as well. It is used by the plugins to test their rules ([example](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/blob/e5074af45ddbc7ee42e76d244a4e3d556a755f77/test/unit/rules/lint-prettier-test.js#L1)).

But maybe it is the good time to expose it through `lib/index.js`? Wdyt?

bmish: Good point, my preference would be that we expose that through `lib/index.js`, as I've implemented here: https://github.com/ember-template-lint/ember-template-lint/pull/2209

bmish:
I added to the description of this PR that there's another item we need to solve:

What to do about users who may need to retrieve the default config from [no-bare-strings](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md)? `const { DEFAULT_CONFIG } = require('ember-template-lint/lib/rules/no-bare-strings');`

dcyriller:
> What to do about users who may need to retrieve the default config from no-bare-strings? const { DEFAULT_CONFIG } = require('ember-template-lint/lib/rules/no-bare-strings');

It is difficult to tell if the default config from no-bare-strings is really consummed by some ember-template-lint plugins. TBH I have no idea.

I see two options:
- we don't expose it and wait for users to report an issue if they need it. This option has the advantage to simplify the maintenance, the drawback is that we could have to release a bug fix (in the best case which is the case where "a user reports an issue", but it could possible that noone reports an issue);
- we expose it through the `lib/index.js` file. This option has the advantage of being more gentle (no bug, no issue to report) but the drawback of increasing the cost of maintenance of ember-template-lint (a lot of files exposed, etc.)

I see your PR as the opportunity to keep the API surface under control. Then, I would not expose the default config from no-bare-strings.

But maybe there are other opinions.

bmish:
@dcyriller thanks for the feedback. I agree we should avoid exposing things unnecessarily.

I'm going to redesign the `no-bare-strings` rule configuration a bit so that there's no need for consumers to import the default config from the rule. The new behavior of the rule options will be to augment instead of replace the default config, as augmenting is typically what consumers want to do. If anyone really needs to replace the default config, then we can add boolean options to the rule `ignoreDefaultAllowList`, `ignoreDefaultGlobalAttributes`, `ignoreDefaultElementAttributes` (default false) to get the old behavior back.

dcyriller:
I see, thanks for the explanation. I gave it a try in the linked PR.

dcyriller:
FYI I noticed [ember-template-lint-plugin-tailwind](https://gitlab.com/michal-bryxi/open-source/ember-template-lint-plugin-tailwindcss/-/blob/429b1a41a7084968e6a5d55e792193c0749bfad7/src/rules/class-wrap.ts#L5) uses `lib/helpers/create-error-message`.

bmish:
@dcyriller good catch. I don't really like this clunky `createErrorMessage` function. ESLint definitely doesn't have an equivalent. I'm hoping we can replace this with a future schema API (#1371) or other built-in functionality.

I'm leaning toward not publishing `createErrorMessage` unless it's heavily used by the plugin ecosystem. If we discover that lots of plugins need it, we can add it to the public API until we have a better alternative.

PR #2213, opened 2021-11-20
Change `no-bare-strings` rule options to augment instead of replace the default config 

dcyriller:
## Description

This PR would remove the need to import the default config from the `no-bare-strings` rule.

v3 recommended configuration:

```javascript
const { DEFAULT_CONFIG } = require('ember-template-lint/lib/rules/no-bare-strings');
const additionalCharsToIgnore = ['a', 'b', 'c'];
module.exports = {
  rules: {
    'no-bare-strings': [...DEFAULT_CONFIG.allowlist, ...additionalCharsToIgnore],
  },
};
```

In v4 the configuration would have to be updated to:

```javascript
const additionalCharsToIgnore = ['a', 'b', 'c'];
module.exports = {
  rules: {
    'no-bare-strings': additionalCharsToIgnore,
  },
};
```

To exclude the default allowlist, the configuration would be:

```javascript
module.exports = {
  rules: {
    'no-bare-strings': {
      allowlist: additionalCharsToIgnore,
      ignoreDefaultAllowList: true,
    },
  },
};
```

## Related Discussions

If the PR is accepted, I will also add `ignoreDefaultGlobalAttributes` and `ignoreDefaultElementAttributes`. It implements [the first part of your comment]( https://github.com/ember-template-lint/ember-template-lint/pull/2208#issuecomment-974532630) @bmish

### DIFF
--- a/docs/migration/v4.md
+++ b/docs/migration/v4.md
@@ -1,1 +1,40 @@
 # Migrating to v4.0.0
+
+## Changes to rule options
+
+### [no-bare-strings](../rule/no-bare-strings.md)
+
+The rules options now augment instead of replace the default config.
+
+In v3, the recommended configuration was:
+
+```javascript
+const { DEFAULT_CONFIG } = require('ember-template-lint/lib/rules/no-bare-strings');
+const additionalCharsToIgnore = ['a', 'b', 'c'];
+const additionalAttributesToCheck = ['data-foo'];
+
+module.exports = {
+  rules: {
+    'no-bare-strings': {
+      allowlist: [...DEFAULT_CONFIG.allowlist, ...additionalCharsToIgnore],
+      globalAttributes: [...DEFAULT_CONFIG.globalAttributes, ...additionalAttributesToCheck],
+    },
+  },
+};
+```
+
+In v4 the configuration should be updated to:
+
+```javascript
+const additionalCharsToIgnore = ['a', 'b', 'c'];
+const additionalAttributesToCheck = ['data-foo'];
+
+module.exports = {
+  rules: {
+    'no-bare-strings': {
+      allowlist: additionalCharsToIgnore,
+      globalAttributes: additionalAttributesToCheck,
+    },
+  },
+};
+```
--- a/docs/rule/no-bare-strings.md
+++ b/docs/rule/no-bare-strings.md
@@ -22,32 +22,17 @@ This rule **allows** the following:
  The following values are valid configuration:
 
 * boolean -- `true` for enabled / `false` for disabled
-* array -- an array of allowlisted strings
+* array -- An array of allowlisted strings (extends the default config)
 * object -- An object with the following keys:
-  * `allowlist` -- An array of allowlisted strings
-  * `globalAttributes` -- An array of attributes to check on every element.
-  * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
+  * `allowlist` -- An array of allowlisted strings (extends the default config)
+  * `globalAttributes` -- An array of attributes to check on every element (extends the default config).
+  * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name (extends the default config).
 
 When the config value of `true` is used the following configuration is used:
 
 * `allowlist` - refer to the `DEFAULT_CONFIG.allowlist` property in the [rule](../lib/rules/no-bare-strings.js)
 * `globalAttributes` - `title`, `aria-label`, `aria-placeholder`, `aria-roledescription`, `aria-valuetext`
 * `elementAttributes` - `{ img: ['alt'], input: ['placeholder'] }`
-
-The `DEFAULT_CONFIG` is available as an export. Example use in configuration:
-
-```javascript
-const {
-  DEFAULT_CONFIG
-} = require('ember-template-lint/lib/rules/no-bare-strings');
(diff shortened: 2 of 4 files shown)


docs/rule/no-bare-strings.md
```diff
@@ -22,32 +22,20 @@ This rule **allows** the following:
  The following values are valid configuration:
 
 * boolean -- `true` for enabled / `false` for disabled
-* array -- an array of allowlisted strings
+* array -- An array of allowlisted strings
 * object -- An object with the following keys:
   * `allowlist` -- An array of allowlisted strings
   * `globalAttributes` -- An array of attributes to check on every element.
   * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
+  * `ignoreDefaultAllowList` -- A boolean to optionaly ignore default allowlist
```
bmish, line 30: How about we actually call this `includeDefaultAllowList` (same change for other new options too)? That removes the negative--I usually try to avoid negatives in option names. Also we will want to make sure we mention the default value here.

docs/rule/no-bare-strings.md
```diff
@@ -22,32 +22,20 @@ This rule **allows** the following:
  The following values are valid configuration:
 
 * boolean -- `true` for enabled / `false` for disabled
-* array -- an array of allowlisted strings
+* array -- An array of allowlisted strings
 * object -- An object with the following keys:
   * `allowlist` -- An array of allowlisted strings
```
bmish, line 27:
```suggestion
  * `allowlist` -- An array of allowlisted strings (extends the default config)
```

test/unit/rules/no-bare-strings-test.js
```diff
@@ -100,6 +87,18 @@ generateRuleTests({
...
+      // extend the allowlist list
+      config: ['/', '"'],
+      template: '{{t "foo"}} / &lpar;"{{name}}"&rpar;',
+    },
+
+    {
+      // extend the allowlist list
+      config: { allowlist: ['/', '"'] },
+      template: '{{t "foo"}} / &lpar;"{{name}}"&rpar;',
+    },
```
bmish, line 100: Should add a valid test case that ignores the default allow list and uses something from it.

docs/migration/v4.md
```diff
@@ -1 +1,43 @@
 # Migrating to v4.0.0
+
+## Changes to rule options
+
+- [no-bare-strings](../rule/no-bare-strings.md) now has a `ignoreDefaultAllowList` option. It can be used to exclude the default allowlist. The `allowlist` you provide (as an array or through the `allowlist` key) is now merged with the default allowlist defined in the [rule](../lib/rules/no-bare-strings.js)'s `DEFAULT_CONFIG.allowlist` property'.
```
bmish, line 5: Since this section is so large, I think we should use a header for it `### [no-bare-strings](../rule/no-bare-strings.md)`.

bmish (changes requested):
Thank you! Great start on this.

bmish:
@dcyriller do you have a chance to finish this up? If not, I'll give it a try, since I'd like to release an alpha in the next day or so (note: this alpha doesn't need to have everything included).

dcyriller:
@bmish I added two commits:  https://github.com/ember-template-lint/ember-template-lint/pull/2213/commits/86549995c7591df3f8ee0ebcbedf0ba4e052effe and https://github.com/ember-template-lint/ember-template-lint/pull/2213/commits/76f8b78aa2be6b61b2cf2ad08426e000050707aa and rebased. Please let me know how it sounds.

docs/rule/no-bare-strings.md
```diff
@@ -22,32 +22,20 @@ This rule **allows** the following:
  The following values are valid configuration:
 
 * boolean -- `true` for enabled / `false` for disabled
-* array -- an array of allowlisted strings
+* array -- An array of allowlisted strings (extends the default config)
 * object -- An object with the following keys:
-  * `allowlist` -- An array of allowlisted strings
+  * `allowlist` -- An array of allowlisted strings (extends the default config)
   * `globalAttributes` -- An array of attributes to check on every element.
   * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
```
bmish, line 29: @dcyriller I want to make sure that all of these options extend the default config now instead of the previous replace behavior. Can you make sure this behavior is documented and tested for `globalAttributes` and `elementAttributes`?

```suggestion
  * `globalAttributes` -- An array of attributes to check on every element (extends the default config)
  * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name (extends the default config)
```

My top priority for this PR is changing the default behavior of these existing options to all be consistent like this.

I would actually be fine if we don't add any new options in this PR (`includeDefaultAllowList`, `ignoreDefaultGlobalAttributes`, `ignoreDefaultElementAttributes`). These options can be added at a later time if someone requests them (I'm not sure anyone actually wants to ignore the default config). Maybe we should just leave them out to keep the rule simpler.

bmish (changes requested):

bmish:
Hey @dcyriller so unfortunately I was in a hurry and I ended up implementing this same change in #2241 before seeing that you were working on it too. It looks like we came up with virtually the same solutions. I will make sure to adopt all improvements found in each PR and then will merge one of them.

dcyriller:
Superseded by the other PR :)

PR #2215, opened 2021-11-20
Fix Node version on CI

ddzz:
`volta-cli/action` does not install Node when a `node-version` is not specified, which means this job has been using the Node version included with `ubuntu-latest` (`14.x`) instead of `12.x`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: volta-cli/action@v1
+        with:
+          node-version: 12.x
 
       - name: install dependencies
         run: yarn install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: volta-cli/action@v1
+        with:
+          node-version: 12.x
```
bmish, line 23: It looks [volta-cli/action](https://github.com/volta-cli/action) updates the `volta` field in package.json when Node 12 is specified here, and that causes the later `git diff --exit-code` step to fail since there's a diff now.

I think what we should do here instead is don't specify the Node version in this step (let it default to Node 16 which is specified in the volta field in package.json) and then just ensure that Node 16 is excluded from the `nodeX` job below.

Let's also update the `floating-dependencies` job to run on Node 16 too. We should use Node 16 everywhere unless we're explicitly trying to test old Node versions.

bmish (changes requested):

bmish (approved):
Thanks!

PR #2216, opened 2021-11-22
Disallow unexpected properties in rule test cases

bmish:
There are many possible properties that can be used when writing test cases for rules. It's easy to mistakenly make a typo or incorrectly pass these properties without noticing.

This PR will assert/throw when an unexpected property is passed to the rule test harness / in a test case.

ESLint v7 implemented this same feature:
* https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class
   > It fails test cases if any unknown properties are found in the objects in the errors property. 
* https://github.com/eslint/eslint/pull/12096
* https://github.com/eslint/rfcs/blob/main/designs/2019-rule-tester-improvements/README.md#4-changing-the-errors-property-of-objects-to-fail-on-unknown-properties

TODO:

- [x] Nevermind for now: Consider using a parameter / schema validation library like ajv / json schema. Or maybe not since ESLint didn't use any library to do this.
- [x] Add a test.

Part of v4 release (#1908).

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -27,6 +27,153 @@ function cloneSimpleObject(obj) {
   return JSON.parse(JSON.stringify(obj || {}));
 }
 
+const ALLOWED_GENERATE_RULE_TEST_PROPERTIES = new Set([
+  'bad',
+  'config',
+  'error',
+  'focusMethod',
+  'good',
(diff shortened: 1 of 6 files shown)


bmish:
Merging to get this release moving, but please comment with any feedback even after it's merged.

PR #2217, opened 2021-11-22
Require auto-fixable test cases to assert the fixed template

bmish:
It's best practice to test the autofix output of a test case. Any test case that produces an autofix should assert what the fixed template will look like.

This PR requires that fixable test cases assert the `fixedTemplate`.

ESLint v7 implemented this same feature:
* https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class
   > It fails test cases if the rule under test provides an autofix but a test case doesn't have an output property. Add an output property to test cases to test the rule's autofix functionality.
* https://github.com/eslint/rfcs/blob/main/designs/2019-rule-tester-improvements/README.md#2-ensuring-to-test-autofix
* There is also a lint rule [eslint-plugin/consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/consistent-output.md) designed to help ESLint rule authors remember to assert the autofix output of their test cases.

TODO:

- [x] Add a test.
- [x] Ensure this asserts the correct number of times and the assert is helpful.
- [x] Handle when this uses `verifyResults` too?

Part of v4 release (#1908).

```
  ● no-negated-condition › {{#if (not condition)}}<img>{{/if}}: logs errors

    assert(received)

    Expected value to be equal to:
      true
    Received:
      false
    
    Message:
      fixable test cases must assert the `fixedTemplate`

      382 |         for (const act of actual) {
      383 |           if (act.isFixable) {
    > 384 |             assert(
          |             ^
      385 |               typeof item.fixedTemplate === 'string',
      386 |               'fixable test cases must assert the `fixedTemplate`'
      387 |             );

      at Object.<anonymous> (lib/helpers/rule-test-harness.js:384:13)

Test Suites: 1 failed, 1 total
```

### DIFF
--- a/lib/helpers/rule-test-harness.js
+++ b/lib/helpers/rule-test-harness.js
@@ -379,6 +379,15 @@ module.exports = function generateRuleTests(...args) {
 
         assert(actual.length > 0, '`bad` test cases should always emit at least one failure');
 
+        for (const act of actual) {
+          if (act.isFixable) {
+            assert(
+              typeof item.fixedTemplate === 'string',
+              'fixable test cases must assert the `fixedTemplate`'
+            );
+          }
+        }
+
         if (item.fatal) {
           assert.strictEqual(actual.length, 1); // can't have more than one fatal error
           delete actual[0].source; // remove the source (stack trace is not easy to assert)
--- a/test/acceptance/rule-test.js
+++ b/test/acceptance/rule-test.js
@@ -676,4 +676,73 @@ describe('regression tests', function () {
       ]
     `);
   });
+
+  test('with fixable test missing `fixedTemplate` assertion', async function () {
+    let group;
+    defaultTestHarness({
+      groupingMethod(name, callback) {
+        group = new Group(name, callback);
+      },
+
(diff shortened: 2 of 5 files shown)


lifeart (approved):

lib/helpers/rule-test-harness.js
```diff
@@ -243,6 +243,15 @@ module.exports = function generateRuleTests({
             );
           }
 
+          for (const act of actual) {
+            if (act.isFixable) {
+              assert(
+                item.fixedTemplate !== undefined,
+                'fixable test cases must assert the `fixedTemplate`'
```
lifeart, line 250: could we show "template" itself in error or somehow point to exact case to developer?

bmish: @lifeart I added an example of the output to the PR description. The `assert` shows up under the specific test case, so I don't think we need to repeat it.

bmish:
Merging to get this release moving, but please comment with any feedback even after it's merged.

PR #2228, opened 2021-11-23
When logging a rule violation, require passing the node or all loc properties

bmish:
This behavior previously was only enforced when the environment variable `EMBER_TEMPLATE_LINT_DEV_MODE` was set to `1`. Now, that environmental variable is removed, and this behavior is always enforced.

The original functionality here was implemented in:
* https://github.com/ember-template-lint/ember-template-lint/pull/2073
* https://github.com/ember-template-lint/ember-template-lint/issues/2072
* https://github.com/ember-template-lint/ember-template-lint/pull/2167

Part of v4 release (#1908).


bmish:
Merging now for upcoming alpha release, but feedback welcome even after merging.

PR #2229, opened 2021-11-23
Add `only` to allowed test case properties

bmish:
Follow-up to #2216.

This should not go in the changelog as it's a bug fix to an unreleased change.


bmish:
Nevermind, looks like we haven't implemented this ability to focus/run a specific test yet, like eslint/mocha/jest have.

PR #2241, opened 2021-11-24
Change `no-bare-strings` rule options to augment instead of replace the default config 

bmish:
This makes it easier to use the rule and eliminates the need to manually import the default config from the rule (which is no longer allowed as of https://github.com/ember-template-lint/ember-template-lint/pull/2208).

Most of this was inspired by the work @dcyriller did in https://github.com/ember-template-lint/ember-template-lint/pull/2213. I made a few additional improvements.

### DIFF
--- a/lib/rules/no-bare-strings.js
+++ b/lib/rules/no-bare-strings.js
@@ -1,32 +1,10 @@
 'use strict';
 
-/*
- Disallows the use of bare strings in a template
-
- ```
- {{!-- good  --}}
- <div>{{evaluatesToAString}}</div>
- <div>{{'A string'}}</div>
-
- {{!-- bad --}}
- <div>A bare string</div>
- ```
-
- The following values are valid configuration:
-
-   * boolean -- `true` for enabled / `false` for disabled
-   * array -- an array of allowlisted strings
-   * object -- An object with the following keys:
-     * `allowlist` -- An array of allowlisted strings
-     * `globalAttributes` -- An array of attributes to check on every element.
-     * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
- */
-
 const createErrorMessage = require('../helpers/create-error-message');
 const { match } = require('../helpers/node-matcher');
 const Rule = require('./_base');
 
-const GLOBAL_ATTRIBUTES = [
+const DEFAULT_GLOBAL_ATTRIBUTES = [
   'title',
   'aria-label',
   'aria-placeholder',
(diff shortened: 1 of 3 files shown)


lib/rules/no-bare-strings.js
```diff
@@ -1,40 +1,18 @@
 'use strict';
 
-/*
```
bmish, line 3: @dcyriller removed this redundant comment.

lib/rules/no-bare-strings.js
```diff
@@ -1,40 +1,18 @@
...
-     * `globalAttributes` -- An array of attributes to check on every element.
-     * `elementAttributes` -- An object whose keys are tag names and value is an array of attributes to check for that tag name.
- */
-
 const createErrorMessage = require('../helpers/create-error-message');
 const { match } = require('../helpers/node-matcher');
 const Rule = require('./_base');
 
-const GLOBAL_ATTRIBUTES = [
+const DEFAULT_GLOBAL_ATTRIBUTES = [
```
bmish, line 7: @dcyriller renamed these for consistency.

lib/rules/no-bare-strings.js
```diff
@@ -147,6 +125,23 @@ function sanitizeConfigArray(allowlist = []) {
   return allowlist.filter((option) => option !== '').sort((a, b) => b.length - a.length);
 }
 
+/**
+ * Example:
+ * obj1 = { img: ['data-alt'] } }
+ * obj2 = { img: ['alt'] } }
+ * result = { img: ['data-alt', 'alt'] } }
+ */
+function mergeObjects(obj1 = {}, obj2 = {}) {
```
bmish, line 134: @dcyriller this is a very similar solution to what you implemented with `augmentedWith`.

dcyriller: nice

lib/rules/no-bare-strings.js
```diff
@@ -113,8 +91,8 @@ const IGNORED_ELEMENTS = new Set(['pre', 'script', 'style', 'template', 'textare
 // Character entity reference chart: https://dev.w3.org/html5/html-author/charref
 const DEFAULT_CONFIG = {
   allowlist: DEFAULT_ALLOWLIST,
-  globalAttributes: GLOBAL_ATTRIBUTES,
-  elementAttributes: TAG_ATTRIBUTES,
+  globalAttributes: DEFAULT_GLOBAL_ATTRIBUTES,
+  elementAttributes: DEFAULT_ELEMENT_ATTRIBUTES,
```
dcyriller, line 95: maybe something like:
```suggestion
  elementAttributes: DEFAULT_ATTRIBUTES_BY_ELEMENT,
```

dcyriller (approved):

lib/rules/no-bare-strings.js
bmish: I agree that's a better name, but I want to be consistent with the public option name for now.

PR #2252, opened 2021-11-29
Remove deprecated pending functionality

scalvert:
As part of v4.0.0, the deprecated pending functionality is being removed in favor of the new todo system. This PR removes the internals of the pending feature.

Part of v4 release (#1908).


lib/get-config.js
```diff
@@ -133,6 +125,12 @@ function validateRootProperties(source) {
     if (!KNOWN_ROOT_PROPERTIES.has(key)) {
       throw new Error(`Unknown top-level configuration property detected: ${key}`);
     }
+
+    if ('pending' in source) {
+      throw new Error(
```
bmish, line 130: This new exception is dead code because the above exception will fire first when it encounters the unknown `pending` property. Personally, I think the existing exception is sufficient, and we don't necessarily need to add a new one.

test/acceptance/cli-test.js
```diff
@@ -614,9 +608,9 @@ describe('ember-template-lint executable', function () {
           'app/templates/application.hbs',
           '  1:4  error  Non-translated string used  no-bare-strings',
           '  1:24  error  Non-translated string used  no-bare-strings',
-          '  1:53  warning  HTML comment detected  no-html-comments',
+          '  1:53  error  HTML comment detected  no-html-comments',
```
bmish, line 611: The title of this test `should print properly formatted error and warning messages` is outdated now?

test/acceptance/todo-cli-test.js
```diff
@@ -89,7 +83,7 @@ describe('todo usage', () => {
       expect(result.stdout).toBeTruthy();
     });
 
-    it('errors when config.pending and `.lint-todo` dir coexist', async function () {
+    it('errors when config.pending is present', async function () {
```
bmish, line 86: I'm not sure this test is useful anymore. Perhaps it should be deleted.

There's already a separate test for the `Unknown top-level configuration property` exception.

lib/get-config.js
scalvert: Hmmm good point. I can remove it.

test/acceptance/todo-cli-test.js
scalvert: Yep, will remove with the extra validation code.

test/acceptance/cli-test.js
scalvert: Ya I'll change the name.

bmish (approved):

rwjblue (approved):

PR #2254, opened 2021-11-29
Remove misleading awaits in tests

courajs:
[`FakeProject#dispose`](https://github.com/ember-template-lint/ember-template-lint/blob/893bea44d79127af66c4837f7334e3e372ab0360/test/helpers/fake-project.js#L143) is not async and does not return a promise. (Neither does [`FixturifyProject#dispose`](https://github.com/stefanpenner/node-fixturify-project/blob/v2.1.1/index.ts#L310))

So, these awaits are misleading. I bumped into this because of some
confusing `require-atomic-updates` eslint errors

See test/helpers/fake-project.js or
https://github.com/stefanpenner/node-fixturify-project/blob/v2.1.1/index.ts#L310

### DIFF
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -20,9 +20,9 @@ describe('ember-template-lint executable', function () {
     project.chdir();
   });
 
-  afterEach(async function () {
+  afterEach(function () {
     process.chdir(ROOT);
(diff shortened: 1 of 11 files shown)


bmish (approved):
Thanks!

PR #2262, opened 2021-12-04
Convert to ESM

bmish:
This PR migrates this plugin to [ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). I did everything manually.

This will be a breaking change in that it will require ember-template-lint plugin authors and consumers of our NodeJS API to convert to ESM. It should not be a breaking change for CLI users. As of now, I'm not planning to add a [build step for commonjs fallback](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77) which has downsides.

See my PR comments for more details about why things changed.

### Disabled test

I had to skip `test/unit/get-config-test.js` during CI in Node versions before 16 (see ci.yml). While this test passes in Node 16, it fails in other Node versions. Tests in this file which trigger `resolveProjectConfig` and thus `await import()` for config/plugin loading cause this error:

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.
```

This seems like a bug in jest's handling of ESM or even a V8 bug. Possibly related: https://github.com/facebook/jest/issues/11438

### TODO in this PR (note: not a complete list)

- [x] Fix import cycles detected by `import/no-cycle` lint rule
- [x] Add file extensions in imports
- [x] Check consistency of export statements
- [x] Convert `update:readme` script to ESM so it can import ESM rule files
- [x] Add comment to top of auto-generated rules/index.js and config/index.js files saying that they're auto-generated
- [x] Update docs
- [x] Add handling/testing for loading `.cjs`  file extension in config/plugin files
- [x] Add handling/testing for loading `.mjs`  file extension in config/plugin files
- [x] Convert `dev/` folder to ESM

### TODO after this PR

- Update ESM-only dependencies
- Change `Linter` class to use underscores for private members and improve `node-api.md`
- Sanity-check default vs. named exports for our NodeJS API

### Demonstration PRs:

- https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/pull/191

Part of v4 release (#1908).


bmish:
I've gotten this down to only 2 failing test suites. There appears to be one major issue remaining to solve. Open to advice on whether I'm going down the right path.

It seems we need to support both CJS and ESM formats for .template-lintrc.js config files and index.js plugin files. CJS is the format we use today, but if these files need to import rules directly, which are now written in ESM, then the files need to be changed to ESM (only ESM files can import ESM).

Example of importing rule directly in config/plugin file:

```js
import inlineComponent from './rules/inline-component.js';

export default {
  plugins: [
    {
      name: 'plugin1',
      rules: {
        'inline-component': inlineComponent,
      },
    },
  ],
};
```

So inside of `resolveProjectConfig()` and `requirePlugin()`, I think we'll need to:
* Determine if the config/plugin file is CJS or ESM (how?)
* Call `require()` if CJS or `await import()` for ESM

The main issue being that a lot of functions need to be changed to async if the loading has to be async.

scalvert:
Should we just require that consumers migrate to ESM configs/plugins? The thing with ESM is it's like TS; once part of your ecosystem moves to using it, it's a bit of a cascading effect. 

Should we instead provide a migration tool to convert cjs to esm for v4 for configs, and require plugins to convert themselves (there can't be a ton of these)?

bmish:
> Should we just require that consumers migrate to ESM configs/plugins? The thing with ESM is it's like TS; once part of your ecosystem moves to using it, it's a bit of a cascading effect.
> 
> Should we instead provide a migration tool to convert cjs to esm for v4 for configs, and require plugins to convert themselves (there can't be a ton of these)?

Yeah so plugins will need to convert to ESM and I think that's fine as it's clearly the future.

In terms of supporting CJS/ESM config files, it may be easier than I was expecting since `await import()` will support both: https://discord.com/channels/480462759797063690/666416704418611233

So I'm currently working on switching over to `await import()` for config loading...just a bit tricky since a bunch of stuff internally needs to change to async as a result. Will report back soon.

scalvert:
Ya I knew `await import()` would work when importing ESM from CJS, but thought you had to use `createRequire` to do the inverse. Glad to hear it works.

bin/ember-template-lint.js
```diff
@@ -481,7 +482,6 @@ async function run() {
   let hasTodos = options.includeTodo && results.todoCount;
   let hasUpdatedTodos = options.updateTodo;
 
-  let Printer = require('../lib/formatters/default');
```
bmish, line 484: This is now a static import at top of file. Could possibly be moved back to dynamic `await import()` but static import seems fine.

lib/get-config.js
```diff
@@ -28,16 +24,21 @@ const ALLOWED_ERROR_CODES = new Set([
...
   // throws exception if not found
-  let pluginPath = resolve.sync(pluginName, { basedir });
+  let pluginPath = resolve.sync(pluginName, { basedir, extensions: ['.js', '.mjs', '.cjs'] });
 
-  return require(pluginPath); // eslint-disable-line import/no-dynamic-require
+  try {
+    const { default: plugin } = await import(pluginPath);
+    return plugin;
+  } catch {
+    return require(pluginPath); // eslint-disable-line import/no-dynamic-require
```
bmish, line 37: Fallback for loading CJS still seems to be necessary...

lib/helpers/determine-rule-config.js
```diff
@@ -0,0 +1,36 @@
...
+    case 'off':
+      return { config: false, severity: IGNORE_SEVERITY };
+    case 'warn':
+      return { config: true, severity: WARNING_SEVERITY };
+    case 'error':
+      return { config: true, severity: ERROR_SEVERITY };
+  }
+}
+
+export default function determineRuleConfig(ruleData) {
```
bmish, line 14: Extracted to separate helper file to avoid import cycle.

lib/linter.js
```diff
@@ -31,23 +31,18 @@ function buildErrorMessage(filePath, error) {
...
-class Linter {
+export default class Linter {
   constructor(_options) {
     let options = _options || {};
 
     this.options = options;
     this.console = options.console || console;
     this.workingDir = options.workingDir || process.cwd();
-
-    this.loadConfig();
```
bmish, line 42: Config loading can't happen in `Linter` class constructor anymore since it's async. It now happens when calling `Litner` class public functions like `verify`.

lib/linter.js
```diff
@@ -281,6 +279,13 @@ class Linter {
     return fileResults;
   }
 
+  async getConfig() {
```
bmish, line 282: Config should now be accessed through `getConfig` to ensure it's loaded instead of directly accessing the `config` member variable.

lib/rules/index.js
```diff
@@ -1,5 +1,214 @@
-'use strict';
+// STOP: This file is autogenerated. Do not modify by hand.
```
bmish, line 1: This file now has to be auto-generated by a new script `yarn update:indices` since `requireIndex` doesn't work with ESM imports.

scripts/update-rules.js
```diff
@@ -1,18 +1,24 @@
...
+import path, { dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import prettier from 'prettier';
 
-const fs = require('fs');
-const path = require('path');
+import a11y from '../lib/config/a11y.js';
+import recommended from '../lib/config/recommended.js';
+import stylistic from '../lib/config/stylistic.js';
+import rules from '../lib/rules/index.js';
```
bmish, line 10: Had to convert this script to ESM because it imports the rules index which is ESM.

test/acceptance/public-api-test.js
```diff
@@ -32,22 +33,25 @@ describe('public api', function () {
         '.template-lintrc.js': "throw Error('error happening during config loading');\n",
       });
 
-      expect(() => {
-        new Linter({
```
bmish, line 36: Config loading no longer happens in constructor so changed to call `loadConfig` in this test.

test/acceptance/public-api-test.js
```diff
@@ -1118,15 +1131,16 @@ describe('public api', function () {
       project.dispose();
     });
 
-    it('[.html] does not identify errors for ember-cli default app/index.html (3.20)', async function () {
+    it('[.html] does not identify errors (except for no-forbidden-elements) for ember-cli default app/index.html (3.20)', async function () {
       // reset config to default value
       project.setConfig();
```
bmish, line 1136: Calling `project.setConfig();` here previously had no effect because the config was already loaded when the `Linter` was constructed in `beforeEach`, so this test wasn't actually using the full recommended config. Now this does have an effect since the config is loaded when `verify` is called later in this test. There are actually violations of `no-forbidden-elements` and `require-lang-attribute` so I had to disable or fix those.

test/fixtures/with-inline-plugins/.template-lintrc.js
```diff
@@ -1,9 +1,12 @@
-module.exports = {
+
+import inlineComponent from './rules/inline-component.js';
+
+export default {
   plugins: [
     {
       name: 'plugin1',
       rules: {
-        'inline-component': require('./rules/lint-inline-component')
```
bmish, line 6: Any fixture directly importing a rule had to be converted to ESM since rules are now ESM.

bmish, line 6: The `lint-` prefix on rule names is an artifact that was mostly removed years ago. I finished removing it on these fixture rules.

.eslintrc.cjs
```diff
@@ -1,7 +1,9 @@
 module.exports = {
+  parser: '@babel/eslint-parser',
```
bmish, line 2: `@babel/eslint-parser` added for parsing top-level await syntax now allowed in ESM.

bin/ember-template-lint.js
```diff
@@ -481,22 +250,11 @@ async function run() {
...
 
-// exports are for easier unit testing
-module.exports = {
-  _parseArgv: parseArgv,
-  _expandFileGlobs: expandFileGlobs,
-  _getFilesToLint: getFilesToLint,
-  _getPossibleOptionNames: getPossibleOptionNames,
-};
-
-if (require.main === module) {
```
bmish, line 500: Moved the exported helper functions from this file to a new helper file `lib/helpers/cli.js` since it's tricky in ESM to have a binary script that can also be imported.

scalvert:
Wow, @bmish. Epic PR, and thank you! Will take a look and see if there's any issues.

bin/ember-template-lint.js
scalvert: I think this is fine since we're always loading a printer.

bmish: Okay, I did this in a few places since static imports don't require switching stuff to await, but probably wouldn't be difficult to move it back to dynamic (using `await import()`) later if needed.

bmish:
I'll plan to release this as a beta tomorrow. Feedback always welcome even after merging. If anyone plans to review, let me know.

lib/get-config.js
```diff
@@ -411,9 +393,9 @@ function getProjectConfig(workingDir, options) {
     ensureRootProperties(config, source);
     validateRootProperties(source);
 
-    config.plugins = processPlugins(workingDir, source.plugins, options);
-    config.loadedRules = processLoadedRules(workingDir, config, options);
-    config.loadedConfigurations = processLoadedConfigurations(workingDir, config, options);
+    config.plugins = await processPlugins(workingDir, source.plugins, options);
+    config.loadedRules = await processLoadedRules(workingDir, config, options);
```
lifeart, line 397: Looks like we could do it in parallel (promise all)

lifeart (approved):

lifeart:
@bmish great work!

.eslintrc.cjs
```diff
@@ -82,7 +84,7 @@ module.exports = {
     'filenames/match-regex': ['error', '^.?[a-z0-9-]+$'], // Kebab-case.
 
     // Optional import rules:
-    'import/extensions': 'error',
+    'import/extensions': ['error', 'always'],
```
bmish, line 87: This enforces that we include file extensions in all import statements.

package.json
```diff
@@ -34,8 +35,10 @@
     "new:rule": "node dev/generate.js new-rule",
     "release": "release-it",
     "test": "npm-run-all lint:* test:*",
-    "test:jest": "jest --coverage",
-    "update:readme": "node ./scripts/update-rules.js"
+    "test:jest": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --coverage",
```
bmish, line 38: `--experimental-vm-modules` needed for running ESM tests with jest.

https://jestjs.io/docs/ecmascript-modules

package.json
```diff
@@ -69,16 +72,17 @@
     "globby": "^11.0.4",
     "is-glob": "^4.0.3",
     "micromatch": "^4.0.4",
-    "requireindex": "^1.2.0",
```
bmish, line 72: `requireIndex` doesn't work with ESM imports. Now using new script `yarn update:indices`.

lib/get-config.js
bmish: I just tried this and the tests all pass:

```js
await Promise.all([
      (config.plugins = await processPlugins(workingDir, source.plugins, options)),
      (config.loadedRules = await processLoadedRules(workingDir, config, options)),
      (config.loadedConfigurations = await processLoadedConfigurations(
        workingDir,
        config,
        options
      )),
]);
```

However, I'm still a bit concerned that this might not be safe? Don't we depend on plugins to be loaded first before processing loaded rules? Either way, let's consider this optimization in a separate PR due to the risk.

PR #2270, opened 2021-12-06
Add helpful assertion message when fatal test case error message doesn't match

jamescdavis:
This is very useful when setting up error test cases. When including the actual message, you get:

Screen Shot 2021-12-06 at 1 35 05 PM

instead of just:

Screen Shot 2021-12-06 at 1 39 39 PM

which helps to quickly see the message you need to match.


lib/helpers/rule-test-harness.js
```diff
@@ -365,7 +365,7 @@ module.exports = function generateRuleTests({
             delete expectedResults[i].rule;
             delete element.source;
 
-            assert.ok(element.message.includes(expectedResults[i].message));
+            assert.ok(element.message.includes(expectedResults[i].message), element.message);
```
bmish, line 368: I don't think the message alone is enough, instead let's actually tell the user what we are asserting: `expected "foo" to contain "bar"`.

bmish:
@jamescdavis should we merge this as-is or should we make the improvement I suggested?

lifeart:
Here is updated version: 
image

NullVoxPopuli (approved):

PR #2276, opened 2021-12-14
Add new rule `no-invalid-aria-attributes`

### LINKED ISSUE #2275: Proposal for new a11y check to disallow invalid ARIA attributes
Currently, Ember-template-lint does not prevent invalid ARIA attributes, or any ARIA-* properties not found in the [WAI-ARIA States and Properties spec](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties), on an element. Using invalid ARIA attributes reduces usability by failing to provide the element it belongs to with the intended accessibility function.

This proposes the authoring of a new rule that enforces that ARIA attributes should be valid, and invalid attributes will generate an error. This rule is analogous to JSXs A11y [aria-props]((https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-props.md) rule.

Reference:
[Supported States and Properties | Accessible Rich Internet Applications (WAI-ARIA) 1.0](https://www.w3.org/TR/wai-aria-1.0/states_and_properties)

judithhinlung:
This PR adds the new rule no-invalid-aria-attributes which fixes #2275 .

### Background
Currently, Ember-template-lint does not prevent invalid ARIA attributes, or any ARIA-* properties not found in the [WAI-ARIA States and Properties spec](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties), on an element. Using invalid ARIA attributes reduces usability by failing to provide the element it belongs to with the intended accessibility function.

This proposes the authoring of a new rule that enforces that ARIA attributes should be valid, and invalid attributes will generate an error. 
Reference:
[Supported States and Properties | Accessible Rich Internet Applications (WAI-ARIA) 1.0](https://www.w3.org/TR/wai-aria-1.0/states_and_properties) 

**Allowed**:

```
# examples
<input type="email" aria-required="true" />
<div role="region" aria-live="polite">Valid live region</div>
```

**Forbidden**:

```
# examples
<input type="text" aria-text="inaccessible text" />
<div role="alert" aria--live="polite"></div>
```

### Testing
Test Suites: 130 passed, 130 total
Tests:       6 skipped, 7051 passed, 7057 total
Snapshots:   887 passed, 887 total
Time:        50.697 s, estimated 55 s
Ran all test suites.
✨  Done in 62.73s.


lib/rules/no-invalid-aria-attributes.js
```diff
@@ -0,0 +1,89 @@
...
+  for (let attribute of attributes) {
+    if (attribute.name.startsWith('aria-')) {
+      ariaAttributes.push(attribute);
+    }
+  }
+  return ariaAttributes;
+}
+
+module.exports = class NoInvalidAriaAttributes extends Rule {
+  logNode({ node, message }) {
```
bmish, line 66: This helper doesn't seem to be necessary, can we remove it?

lib/rules/no-invalid-aria-attributes.js
```diff
@@ -0,0 +1,89 @@
...
+  'aria-selected',
+  'aria-setsize',
+  'aria-sort',
+  'aria-valuemax',
+  'aria-valuemin',
+  'aria-valuenow',
+  'aria-valuetext',
+]);
+
+function getAriaAttributes(node) {
```
bmish, line 54: It seems slightly inefficient to create a list of all these attributes. Might I suggest we just iterate over the list of attributes like this:

```
for (const attribute of attributes) {
  if (attribute.name.startsWith('aria-') && !VALID_ARIA_ATTRIBUTES.has(attribute.name)) {
    // ...
  }
}
``` 

This way, we avoid creating a list. Note: if there's a good reason to create a separate list in the future, we can do it, but it just doesn't seem necessary right now.

lib/rules/no-invalid-aria-attributes.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+const Rule = require('./_base');
+
+const VALID_ARIA_ATTRIBUTES = new Set([
```
bmish, line 5: Can you add a comment above this explaining where to get this list from?

docs/rule/no-invalid-aria-attributes.md
```diff
@@ -0,0 +1,22 @@
+# no-invalid-aria-attributes
+
+This rule checks for the use of invalid ARIA attributes, or any aria-\* property on an element that is not included in the [WAI-ARIA States and Properties spec](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties).
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<input type="text" aria-require="true" />
```
bmish, line 10: In this example, we want to make it really clear this is not a real attribute, so let's name it something like:
```suggestion
<input type="text" aria-not-real="true" />
```

docs/rule/no-invalid-aria-attributes.md
```diff
@@ -0,0 +1,22 @@
+# no-invalid-aria-attributes
+
+This rule checks for the use of invalid ARIA attributes, or any aria-\* property on an element that is not included in the [WAI-ARIA States and Properties spec](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties).
```
bmish, line 3: Can you explain what happens when someone uses a non-existent aria attribute? E.g. `A non-existent ARIA attribute has no effect.`

docs/rule/no-invalid-aria-attributes.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Examples
+
+This rule **forbids** the following:
```
bmish, line 7: Is there anything else we can validate about ARIA attributes? Anything we validate about their values or what attributes they are used on? It's easier to make this rule comprehensive now and cover multiple things, since adding violations later will be a breaking change.

lib/config/a11y.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     'no-duplicate-landmark-elements': 'error',
     'no-empty-headings': 'error',
     'no-heading-inside-button': 'error',
+    'no-invalid-aria-attributes': 'error',
```
bmish, line 16: I imagine we should add this to `recommended` in the upcoming v4 release right?

docs/rule/no-invalid-aria-attributes.md
bmish, line 3: Is there any legitimate use case for ARIA attributes not mentioned in that list? If so, we might want to mention them.  Or would this always be caused by typos?

bmish (changes requested):

README.md
```diff
@@ -130,6 +130,7 @@ Each rule has emojis denoting:
 | [no-inline-styles](./docs/rule/no-inline-styles.md)                                                       | ✅  |     |     |     |
 | [no-input-block](./docs/rule/no-input-block.md)                                                           | ✅  |     |     |     |
 | [no-input-tagname](./docs/rule/no-input-tagname.md)                                                       | ✅  |     |     |     |
+| [no-invalid-aria-attributes](./docs/rule/no-invalid-aria-attributes.md)                                                     |     |     | ⌨️  |     |
```
drewlee, line 133: The whitespace formatting is off for this new line, which is causing rendering issues for all the subsequent table rows.

bmish: `yarn update:readme` to fix this.

judithhinlung:
[aria-query](https://www.npmjs.com/package/aria-query) is a package that contains all the existing ARIA attributes and their valid types and values, which would be very useful in implementing the attribute-value checks. However, when looking at the other lint rules, I don't see any of them using an external import. Is it preferred for ETL rules to be self-contained, or could I leverage this package in my implementation of this rule?

bmish:
> [aria-query](https://www.npmjs.com/package/aria-query) is a package that contains all the existing ARIA attributes and their valid types and values, which would be very useful in implementing the attribute-value checks. However, when looking at the other lint rules, I don't see any of them using an external import. Is it preferred for ETL rules to be self-contained, or could I leverage this package in my implementation of this rule?

@judithhinlung I'm opening to adding this dependency, since it sounds like it will make this lint rule much more powerful. Let's get it a try.

judithhinlung:
I have updated this PR to use the NPM package [Aria-query](https://www.npmjs.com/package/aria-query) as the source of truth for ARIA validation. This PR now also includes validation for ARIA property types and the use of permitted values. The no-invalid-aria-attribute rule is now essentially a direct port of the JSX rules [aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-props.md) and [Aria-proptypes](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-proptypes.md).

bmish:
@judithhinlung it looks like there is missing test coverage so the tests are failing. Note that we need to ensure every line in the rule file is covered by tests.

```
  no-invalid-aria-attributes.js                  |   88.24 |     80.3 |     100 |   88.24 | 26-28,42,44,48,54,80,87                         
```

lib/rules/no-invalid-aria-attributes.js
bmish: Let's remove this `logNode` helper since it's no better than just calling `this.log` directly.

lib/rules/no-invalid-aria-attributes.js
```diff
@@ -0,0 +1,170 @@
+'use strict';
+
+const ariaQuery = require('aria-query');
+
+const Rule = require('./_base');
+
+const VALID_ARIA_ATTRIBUTES = ariaQuery.aria;
+
+function createInvalidAriaAttributeMessage(name) {
+  return `${name} is not a valid ARIA attribute.`;
```
bmish, line 10: It's not clear what is meant here by "not valid". I think this is trying to say that the ARIA attribute is unknown or unrecognized / non-existent. Can we instead say something like:
```suggestion
  return `${name} is an unrecognized ARIA attribute.`;
```

bmish (changes requested):
Nice work on this so far.

bmish:
@judithhinlung do you think we can finish this up this week? If so, we can include it as a `recommended` rule in the upcoming v4 release (#1908) which I'd like to release later this week.

judithhinlung:
Hey @bmish unfortunately I don't think I will be able to finish this until sometime next week, when I return from my vacation.

judithhinlung:
I have updated this PR with more test coverage, removed the extraneous logNode helper, and modified the invalid-ARIA-attribute message to say that the attribute is not recognized.

package.json
```diff
@@ -61,6 +61,7 @@
     ]
   },
   "dependencies": {
+    "aria-query": "^5.0.0",
```
drewlee, line 64: @judithhinlung Looks like the build is failing because the listed deps are out of alphabetical order. `aria-query` should come after `@lint-todo/utils`.

lib/rules/no-invalid-aria-attributes.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+}
+
+function isBoolean(value) {
+  return typeof value === 'boolean' || value === 'true' || value === 'false';
+}
+
+function isNumeric(value) {
+  if (typeof value === 'string') {
+    value = Number.parseInt(value, 10);
```
bmish, line 40: Normally it's good to avoid modifying the input parameters. Maybe just use `Number.isInteger()`? Not sure why this is checking for booleans.

bmish (approved):
Looks good. Is this still a draft or is it ready?

bmish:
Thanks for your great work on this!

PR #2271, opened 2021-12-09
Add autofixer to `attribute-indentation` rule

VincentMolinie:
The code is not clean at all but it's a beginning. Any help is welcome on the cleaning part 🙏

PR #2281, opened 2021-12-16
Add autofixer to `block-indentation` rule

VincentMolinie:


bmish:
Hope to see this and #2271 land!

VincentMolinie:
> Hope to see this and #2271 land!

Yeah this one seems good I think. `attribute-indetation` needs some reqorks as I'm actually fixing `block-identation` inside it 😅

VincentMolinie:
tests pass @bmish 🙌

bmish:
Looks like you need to run `yarn update:readme`.

lib/rules/block-indentation.js
```diff
@@ -21,7 +21,7 @@
 
  <div>
   <p>{{t "greeting"}}</p>
- </div>
+  </div>
```
bmish, line 24: This looks wrong?

lib/rules/block-indentation.js
```diff
@@ -161,17 +175,33 @@ module.exports = class BlockIndentation extends Rule {
...
   visitor() {
     this._elementStack = [];
+    this.seen = new Set();
+    this.nodeToSource = new WeakMap();
 
     return {
+      Template(node) {
+        if (this.mode === 'fix') {
+          if (
+            node.body[0] &&
```
bmish, line 185: Better way to write this is: `node.body.length > 0`

lib/rules/block-indentation.js
```diff
@@ -161,17 +175,33 @@ module.exports = class BlockIndentation extends Rule {
...
+            node.body[0].type === 'TextNode' &&
+            !node.body[0].chars.replace(/ +/, '')
```
bmish, line 187: Can you extract a helper function for this `isFoo(str)` to make it more clear what this is testing?

lib/rules/block-indentation.js
```diff
@@ -81,6 +83,18 @@ function isControlChar(char) {
   return char === '~' || char === '{' || char === '}';
 }
 
+function get(node, path) {
```
bmish, line 86: Can you add a jsdoc comment to this helper function explaining what exactly it's for, maybe providing a basic example of the usage.

bmish (approved):
This is looking pretty good, thanks for working on it. I think we need a bit more high-level commenting to help people follow what's going on since the logic is a bit complex.

bmish (changes requested):
This is looking pretty good, thanks for working on it. I think we need a bit more high-level commenting, maybe some clearer naming to help people follow what's going on since the logic is a bit complex.

VincentMolinie:
There is an issue on thé Indent. I did a fix at first but rollbacked It cause I didn't know if I should do It in this PR.
This code IS considered valid :
```html
<div>
</div>
  <span>
  </span>
```

The fix of thé rule and thé autofix are quitté easy, let me know if you want It in this PR

lib/rules/block-indentation.js
VincentMolinie: That's thé purpose. It says Bad example but It's actually a  good one 😅

lib/rules/block-indentation.js
VincentMolinie: Like I said thé indentation of block start IS validated only for thé first node while It should validated all. This code disappear if WE want to really fix this issue

lib/rules/block-indentation.js
```diff
@@ -180,56 +219,132 @@ module.exports = class BlockIndentation extends Rule {
...
+      }
+
+      // Process all the child with the path to access the child from the current node
+      for (const { path: childPath, child } of childrenWithPath) {
+        if (['BlockStatement', 'ElementNode'].includes(child.type)) {
+          this.process(child, childPath);
+        }
+      }
+
+      // Process all the child with the path to access the child from the current node
```
bmish, line 279: Duplicate comment from above?

test/unit/rules/block-indentation-test.js
```diff
@@ -227,21 +227,36 @@ generateRuleTests({
         '{{/if}}',
       ].join('\n'),
     },
-  ],
+  ].filter((value, index) => index >= 0),
```
bmish, line 230: Why is this here?

bmish:
> There is an issue on thé Indent. I did a fix at first but rollbacked It cause I didn't know if I should do It in this PR. This code IS considered valid :
> 
> ```
> <div>
> </div>
>   <span>
>   </span>
> ```
> 
> The fix of thé rule and thé autofix are quitté easy, let me know if you want It in this PR

@VincentMolinie it should be fine for us to fix the bug as a follow-up PR. Thanks for working on this!

test/unit/rules/block-indentation-test.js
VincentMolinie: Oops 😂

lib/rules/block-indentation.js
VincentMolinie: Good catch 👍 \

VincentMolinie:
> 0e20e0c

Okey so you want me to rollback: 0e20e0c ?

VincentMolinie:
> > There is an issue on thé Indent. I did a fix at first but rollbacked It cause I didn't know if I should do It in this PR. This code IS considered valid :
> > ```
> > <div>
> > </div>
> >   <span>
> >   </span>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The fix of thé rule and thé autofix are quitté easy, let me know if you want It in this PR
> 
> @VincentMolinie it should be fine for us to fix the bug as a follow-up PR. Thanks for working on this!

Okey so you want me to rollback: 0e20e0c ?

bmish:
@VincentMolinie would it be easy for you to extract that bug fix to a follow-up PR after I merge this soon? If so, let's do that quickly. If too inconvenient, we can keep them together.

VincentMolinie:
Done ✅  @bmish

bmish:
@VincentMolinie thanks, did you see the above comment about the bug extraction? Just wondering what you think.

VincentMolinie:
Yes I removed the commit 😉

bmish (approved):
Thank you!

bmish:
@VincentMolinie excellent, can you create that other PR now?

VincentMolinie:
It's right here: https://github.com/ember-template-lint/ember-template-lint/pull/2318

PR #2291, opened 2021-12-23
Fix broken markdown links

bmish:
Found one of these broken links manually then used used [markdown-link-check](https://github.com/tcort/markdown-link-check) to find the rest.

### DIFF
--- a/docs/rule/require-each-key.md
+++ b/docs/rule/require-each-key.md
@@ -79,4 +79,4 @@ This rule **allows** the following:
 ## References
 
 * [Specifying Keys](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each#specifying-keys)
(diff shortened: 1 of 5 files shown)


docs/rule/require-each-key.md
```diff
@@ -79,4 +79,4 @@ This rule **allows** the following:
 ## References
 
 * [Specifying Keys](https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each#specifying-keys)
-* [The Immutable Pattern in Tracked Properties](https://glimmerjs.com/guides/tracked-properties)
+* [The Immutable Pattern in Tracked Properties](https://glimmerjs.com/guides/tracked-properties) (TODO: fix this link)
```
bmish, line 82: This guide disappeared and I don't know what to replace it with currently.

Issue #2307, opened 2021-12-30
In 3.25, remove no-dynamic-subexpression-invocations from the recommended rule set


NullVoxPopuli:
Added in https://github.com/ember-template-lint/ember-template-lint/pull/1716/
In the PR, I didn't see any scoping of the rule to self-disable post 3.25.
Present in recommended list here: https://github.com/ember-template-lint/ember-template-lint/blob/next/lib/config/recommended.js#L25
Added to recommended here: https://github.com/ember-template-lint/ember-template-lint/commit/eaed378a9f89b33f3eae3db72a0180908669ddff#diff-b50ab033af46255b11243e8d0bfc4e5972d85c76006693ecd9e1a9b5acaf5b4a

In 3.25+, (it's _very common_ (for people who have been wanting locally scoped things for ages))  to do:
```hbs
{{#if (this.someFunction 1 2)}}
{{/if}}

{{ (this.someFunction a b) }}

{{yield (hash a=(this.someFunction x z))}}
```

PR #2308, opened 2021-12-30
Remove `no-dynamic-subexpression-invocations` rule from `recommended` config

NullVoxPopuli:
In order to keep this lint, it needs to be a no-op for Ember 3.25+
I don't know how much of the ecosystem is on 3.25+ yet, but I have 
_many_ projects on 3.25+
(and using local helpers)

Fixes https://github.com/ember-template-lint/ember-template-lint/issues/2307


bmish:
Can you remove it from the list in the v4 migration guide too?

bmish (approved):
Thanks1

PR #2315, opened 2022-01-03
Remove `no-model-argument-in-route-templates` rule from `recommended` config

Windvis:
Related discussion: https://github.com/ember-template-lint/ember-template-lint/issues/1908#issuecomment-1004010809


bmish (approved):
Thanks!

PR #2316, opened 2022-01-03
Remove `no-mut-helper` rule from `recommended` config

Windvis:
Related discussion: https://github.com/ember-template-lint/ember-template-lint/issues/1908#issuecomment-1004016193


bmish (approved):
Thanks!

PR #2332, opened 2022-01-11
Fix bug with newlines and cleanup `no-unbalanced-curlies` rule

### LINKED ISSUE #2330: no-unbalanced-curlies false positive
The no-unbalanced-curlies rule gives a false positive if you have an escaped start curlies and then a newline before the end curlies:

```hbs
\{{foo
}}
```

This diff shows the problem in the test suite:

```diff
diff --git a/test/unit/rules/no-unbalanced-curlies-test.js b/test/unit/rules/no-unbalanced-curlies-test.js
index 056beea1..7b965b4e 100644
--- a/test/unit/rules/no-unbalanced-curlies-test.js
+++ b/test/unit/rules/no-unbalanced-curlies-test.js
@@ -13,6 +13,7 @@ generateRuleTests({
     '\\{{foo}}',
     '\\{{foo}}\\{{foo}}',
     '\\{{foo}}{{foo}}',
+    '\\{{foo\n}}',
   ],
```

I don't really follow why the existing rule is trying to parse one line at a time, so I'm unsure what the best fix is.

pablobm:
Fixes #2330, which reports a false positive when there's a newline inside escaped curlies.

I have been away from Ember and JS for a bit, but I was able to figure things out a bit from my original PR at #1222. If I have this right, the line-by-line solution (mentioned in the issue) was intended to properly detect the location of error in order to report it. Without thinking much, it looks like I decided to parse each individual line, instead of the whole TextNode. This is corrected in the first two commits (one as a simple fix, another to avoid multiple parses), thus fixing the issue.

Additionally, I noticed the test coverage warning and had a look into this too. I removed unused code that I don't remember why I added in the first place. I suspect I copied it from some other rule, but now I can't find any other example. Now coverage for this test is at 100%.

Finally, I changed the name of the constant `DISALLOWED_CHARS` to `SUSPECT_CHARS` as it better describes its function: having these chars is necessary but not sufficient for the rule to be triggered.

### DIFF
--- a/lib/rules/no-unbalanced-curlies.js
+++ b/lib/rules/no-unbalanced-curlies.js
@@ -17,17 +17,6 @@ export default class NoUnbalancedCurlies extends Rule {
         }
         let source = this.sourceForNode(node);
 
-        let { loc } = node;
-
-        if (!loc) {
-          this.log({
-            message: ERROR_MESSAGE,
-            source,
-            node,
-          });
-          return;
-        }
-
         let isMustache = false;
         try {
           let result = parse(chars);
@@ -43,6 +32,7 @@ export default class NoUnbalancedCurlies extends Rule {
           return;
         }
 
+        let { loc } = node;
         let lineNum = loc.start.line;
         let colNum = loc.start.column;
         let lines = chars.match(reLines);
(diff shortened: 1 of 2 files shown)


bmish (approved):
Thanks!

Issue #2321, opened 2022-01-06
no-this-in-template-only-components triggers for route templates


runspired:
which obviously are not components, and in this case there are also controller files alongside.

Setup: components are in the `components/` directory using template co-location, routes, route templates, and controllers are in pods. Ostensibly this may be the issue, as the rule likely doesn't try to determine if a template has a matching route/controller vs component file.

elidupuis:
Hi all 👋 

I've recently hit this issue where our components are in pods (in the `components/` directory) and this rule is flagging all the `this.foo` occurrences. The biggest issue I see at this point, though, is the autofix behaviour: when running with `--fix` all `this.foo` cases are converted to `@foo` which breaks the functionality of the components.

I've simply disabled this rule for now but it would be great to have a fix for apps using component pods. I'm definitely willing to pitch in but I'm not exactly sure where to start.

Is this still something worth digging into?

dknutsen:
I'm also seeing this triggered for route templates in an in-repo engine (standard templates, non-pods)

PR #2335, opened 2022-01-12
partial-fix: default pods paths to false for `no-this-in-template-only-components`

runspired:
helps with #2321


lifeart (approved):

bmish:
Can we add a rule test for this?

elidupuis:
@bmish Interestingly, it appears that there _is_ a test for route templates already: https://github.com/ember-template-lint/ember-template-lint/blob/1ed2f4650e228279a5010ffc65594ec08aa9efe6/test/unit/rules/no-this-in-template-only-components-test.js#L13-L18

But something is missing because this rule definitely flags `this.foo` usage in our route templates.

Issue #2336, opened 2022-01-11
ESM error in v4 consuming plugins in a Windows env


charlesfries:
I'm seeing the following error after updating `ember-template-lint` and `ember-template-lint-plugin-prettier` to 4.0.0.

```
(node:1584) UnhandledPromiseRejectionWarning: Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\Users\cfries\Documents\Repositories\etlpp\node_modules\ember-template-lint-plugin-prettier\ember-template-lint-plugin-prettier.js
require() of ES modules is not supported.
require() of C:\Users\cfries\Documents\Repositories\etlpp\node_modules\ember-template-lint-plugin-prettier\ember-template-lint-plugin-prettier.js from C:\Users\cfries\Documents\Repositories\etlpp\node_modules\ember-template-lint\lib\get-config.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename ember-template-lint-plugin-prettier.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from C:\Users\cfries\Documents\Repositories\etlpp\node_modules\ember-template-lint-plugin-prettier\package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1102:13)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:14)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at requirePlugin (file:///C:/Users/cfries/Documents/Repositories/etlpp/node_modules/ember-template-lint/lib/get-config.js:38:12)
    at async processPlugins (file:///C:/Users/cfries/Documents/Repositories/etlpp/node_modules/ember-template-lint/lib/get-config.js:150:16)
    at async getProjectConfig (file:///C:/Users/cfries/Documents/Repositories/etlpp/node_modules/ember-template-lint/lib/get-config.js:396:22)
    at async Linter.loadConfig (file:///C:/Users/cfries/Documents/Repositories/etlpp/node_modules/ember-template-lint/lib/linter.js:48:19)
    at async Linter.verify (file:///C:/Users/cfries/Documents/Repositories/etlpp/node_modules/ember-template-lint/lib/linter.js:304:7)
[...]

PR #2339, opened 2022-01-13
Convert path to URL before import of config/plugin for Windows support

lifeart:
fixes https://github.com/ember-template-lint/ember-template-lint/issues/2336

I'm not sure how to test it, because chain for this function is quite long, should we create unit test for 'requirePlugin' function itself?

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -91,7 +91,7 @@ export async function resolveProjectConfig(workingDir, options) {
   options.resolvedConfigPath = configPath;
 
   try {
-    const { default: config } = await import(configPath);
+    const { default: config } = await import(pathToFileURL(configPath));
     return config;
   } catch (error) {
     // Fallback to ensure we can load CJS configs in Node versions before 16 (TODO: remove eventually).
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -502,12 +502,7 @@ describe('ember-template-lint executable', function () {
         expect(path.isAbsolute(configPath)).toBeTruthy();
         expect(fs.existsSync(configPath)).toBeTruthy();
 
-        let result = await runBin(
-          '--filename',
-          'app/templates/application.hbs',
-          '--config-path',
-          configPath
-        );
+        let result = await runBin('app/templates/application.hbs', '--config-path', configPath);
 
         expect(result.exitCode).toEqual(0);
         expect(result.stdout).toBeFalsy();
--- a/test/acceptance/cli-test.js
+++ b/test/acceptance/cli-test.js
@@ -483,12 +483,12 @@ describe('ember-template-lint executable', function () {
   describe('multiplatform config resolve', function () {
     describe('given absolute config path and path to single file without errors', function () {
       it('should exit without error and any console output', async function () {
-        project.setConfig({
+        await project.setConfig({
           rules: {
             'no-bare-strings': false,
           },
         });
-        project.write({
+        await project.write({
           app: {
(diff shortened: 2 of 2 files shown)


lib/get-config.js
```diff
@@ -31,7 +32,7 @@ async function requirePlugin(workingDir, pluginName, fromConfigPath) {
   let pluginPath = resolve.sync(pluginName, { basedir, extensions: ['.js', '.mjs', '.cjs'] });
 
   try {
-    const { default: plugin } = await import(pluginPath);
+    const { default: plugin } = await import(pathToFileURL(pluginPath));
```
bmish, line 35: Wouldn't we also need this for the config loading on line 84 too?

lifeart: good catch! pushed

test/acceptance/cli-test.js
```diff
@@ -442,6 +442,37 @@ describe('ember-template-lint executable', function () {
     });
   });
 
+  describe('multiplatform config resolve', function () {
+    describe('given absolute config path and path to single file without errors', function () {
```
bmish, line 446: Question, is this different from the existing tests for `--config-path`? Does it actually test a Windows-style path like `C:\Users\...`. And does it fail before the fix?

lifeart: it should test windows-style path in our windows CI env
no failures before

lifeart: also, it should fail here: https://github.com/ember-template-lint/ember-template-lint/blob/master/test/helpers/fake-project.js#L68
but not failing, we could experiment with some prefixes ('/' before 'C'), or node versions to catch this error

lifeart: in original bugreport #2336 it's failing for custom plugins (as I understand), likely we could also try to test custom plugin here..

lib/get-config.js
```diff
@@ -31,7 +32,7 @@ async function requirePlugin(workingDir, pluginName, fromConfigPath) {
   let pluginPath = resolve.sync(pluginName, { basedir, extensions: ['.js', '.mjs', '.cjs'] });
```
bmish, line 32: Could the reason we end up needing this be related to `resolve.sync` usage here? I wonder if we can solve this by removing the `resolve.sync` somehow instead? 

Although this link shows your similar workaround: https://github.com/nodejs/help/issues/3380

Also, I'm still hoping we can figure out actual tests for this that fail beforehand.

charlesfries:
Anything I can do to help get this merged

lib/get-config.js
lifeart: @bmish you mean change this logic in scope of this PR? (it's already in codebase)

locks:
@bmish @lifeart what's the latest status on this?

bmish:
From what I remember, we had trouble reproducing the issue in the test and so I wasn't confident that the test actually verified it was fixed.

lflo5727:
Is there any update on this PR? 
This issue is blocking us from adding a custom ember-template-lint plugin. We have the same issue described here: https://github.com/ember-template-lint/ember-template-lint/issues/2336 

If not is there any decent workaround?

bmish:
If someone can confirm that this PR fixes the issue, we might be able to move forward with it.

lflo5727:
@bmish I just pulled this branch down for ember-template-lint and tested. This absolutely fixes the issue. 
Would be great to get this pushed along

bmish (approved):
Thanks! Let's give this a try.

PR #2348, opened 2022-01-18
Support the `modifier` helper in the `modifier-name-case` rule

bertdeblock:
At the moment, the `modifier-name-case` rule throws the following error when using the `modifier` helper:

Screenshot 2022-01-18 at 11 22 53

This PR aims to fix that.

### DIFF
--- a/lib/rules/modifier-name-case.js
+++ b/lib/rules/modifier-name-case.js
@@ -1,4 +1,5 @@
 import dasherize from '../helpers/dasherize-component-name.js';
+import { match } from '../helpers/node-matcher.js';
 import Rule from './_base.js';
 
 function generateErrorMessage(modifierName) {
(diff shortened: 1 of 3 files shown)


lib/rules/modifier-name-case.js
```diff
@@ -12,15 +13,37 @@ export default class ModifierNameCase extends Rule {
...
         let modifierName = node.path.original;
 
-        if (modifierName === dasherize(modifierName)) {
+        if (typeof modifierName === 'string') {
+          this._validateModifierName(modifierName, node.path);
+        }
+      },
+
+      SubExpression(node) {
+        if (isModifierHelper(node) === false) {
```
bmish, line 22:
```suggestion
        if (!isModifierHelper(node)) {
```

lib/rules/modifier-name-case.js
```diff
@@ -12,15 +13,37 @@ export default class ModifierNameCase extends Rule {
...
+        let nameParam = node.params[0];
+
+        if (nameParam) {
+          this._validateModifierName(nameParam.original, nameParam);
+        }
       },
     };
   }
+
+  _validateModifierName(modifierName, node) {
```
bmish, line 35: Maybe just pass in a single `node` parameter and then create the `modifierName` inside the function?

bmish (approved):
CC original rule author @mansona just in case they had feedback.

bertdeblock:
Pushed a small update for the docs. Forgot to add the parenthesis. Should've been `{{(modifier "didInsert")}}`.

bmish:
@bertdeblock can you link me to some documentation for that `modifier` helper that shows some examples?

bertdeblock:
@bmish At the moment, the only documentation that I know of is the [Contextual Helpers and Modifiers RFC](https://github.com/emberjs/rfcs/blob/master/text/0432-contextual-helpers.md).

bmish:
@bertdeblock thanks for sharing. Based on that link, I don't see parenthesis used in: `{{modifier "foo"}}`

bertdeblock:
Parenthesis are needed otherwise you are invoking `modifier` _as_ a modifier, while in fact it's a helper. Maybe my example in the docs just isn't clear enough?

Would something like this be better?

```hbs
<input {{(if @shouldAutoFocus (modifier "auto-focus"))}} />
```

bmish:
@bertdeblock thanks for clarifying. Up to you on how to document it.

bertdeblock:
@bmish Updated the examples in the docs. Should be more realistic and clear (hopefully 😬).

PR #2359, opened 2022-01-28
Fix monorepo support by allowing import to do what it's good at (and …

NullVoxPopuli:
…ignoring paths 🙃)

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -27,16 +27,35 @@ const ALLOWED_ERROR_CODES = new Set([
 async function requirePlugin(workingDir, pluginName, fromConfigPath) {
   let basedir = fromConfigPath === undefined ? workingDir : path.dirname(fromConfigPath);
 
+  let plugin;
+
+  try {
+    let importedModule = await import(pluginName);
+    plugin = importedModule.default;
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
+  }
+
   // throws exception if not found
   let pluginPath = resolve.sync(pluginName, { basedir, extensions: ['.js', '.mjs', '.cjs'] });
 
-  try {
-    const { default: plugin } = await import(pluginPath);
-    return plugin;
-  } catch {
-    // Fallback to ensure we can load CJS plugins in Node versions before 16 (TODO: remove eventually).
-    return require(pluginPath); // eslint-disable-line import/no-dynamic-require
+  if (!plugin) {
+    try {
+      let importedModule = await import(pluginPath);
+      plugin = importedModule.default;
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
   }
+
+  if (!plugin) {
+    // Fallback to ensure we can load CJS plugins in Node versions before 16 (TODO: remove eventually).
+    plugin = require(pluginPath); // eslint-disable-line import/no-dynamic-require
+  }
+
+  return plugin;
 }
 
 export async function resolveProjectConfig(workingDir, options) {
--- a/test/acceptance/monorepo-test.js
+++ b/test/acceptance/monorepo-test.js
@@ -1,0 +1,116 @@
+import { execa } from 'execa';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import Project from '../helpers/fake-project.js';
+import setupEnvVar from '../helpers/setup-env-var.js';
+
+const cliPath = fileURLToPath(new URL('../..', import.meta.url));
+const binPath = path.join(cliPath, 'bin/ember-template-lint.js');
+
+function run(project, args, options = {}) {
+  options.reject = false;
+  options.cwd = options.cwd || project.path('.');
+
+  return execa(process.execPath, [binPath, ...args], options);
+}
+
+export async function newTmpDir() {
+  const tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), `ember-template-lint-${Date.now()}-`));
+
+  return tmpDir;
+}
+
+export async function newMonorepo(workspaces) {
+  let root = await newTmpDir();
+
+  await execa('git', ['init'], { cwd: root });
+
+  await fs.writeFile(
+    path.join(root, 'package.json'),
+    `{ "workspaces": [${workspaces.map((w) => `"${w}"`)}], "private": true }`
+  );
+
+  let projects = [];
+  for (let workspace of workspaces) {
+    let safeName = workspace.replaceAll('/', '-');
+    let project = new Project(`test-${safeName}`, '0.0.0');
+
+    project.baseDir = path.join(root, workspace);
+
+    await fs.mkdir(project.baseDir, { recursive: true });
+
+    project.setDefaultFiles();
+    project.linkDependency('ember-template-lint', cliPath);
+    project.writeSync();
+  }
+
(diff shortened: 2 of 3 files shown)


lib/get-config.js
```diff
@@ -27,16 +27,35 @@ const ALLOWED_ERROR_CODES = new Set([
 async function requirePlugin(workingDir, pluginName, fromConfigPath) {
   let basedir = fromConfigPath === undefined ? workingDir : path.dirname(fromConfigPath);
 
+  let plugin;
+
+  try {
+    let importedModule = await import(pluginName);
```
NullVoxPopuli, line 33: Without _trying_ this first, we can't actually use plugins from `node_modules` :see_no_evil: 

For example, prior to this change, we'd get this error:
```
<repo>/node_modules/resolve/lib/sync.js:102
    var err = new Error("Cannot find module '" + x + "' from '" + parent + "'");
              ^

Error: Cannot find module 'ember-template-lint-plugin-tailwindcss' from '<repo>/frontend'
    at Function.resolveSync [as sync] (<repo>/node_modules/resolve/lib/sync.js:102:15)
    at requirePlugin (file://<repo>/node_modules/ember-template-lint/lib/get-config.js:31:28)
    at processPlugins (file://<repo>/node_modules/ember-template-lint/lib/get-config.js:150:22)
    at getProjectConfig (file://<repo>/node_modules/ember-template-lint/lib/get-config.js:396:28)
    at async Linter.loadConfig (file://<repo>/node_modules/ember-template-lint/lib/linter.js:48:19)
    at async Linter.verify (file://<repo>/node_modules/ember-template-lint/lib/linter.js:304:7)
    at async run (file://<repo>/node_modules/ember-template-lint/bin/ember-template-lint.js:211:21) {
  code: 'MODULE_NOT_FOUND'
}
error Command failed with exit code 1.    }

```

bmish:
We'll need tests for this that fail before the change and pass after.

lib/get-config.js
bmish: Can you explain more about how the monorepo is setup?

NullVoxPopuli: The issue is present in _any_ monorepo where the ember project is not at the root, where the root's node_modules contains ember-template-lint (so almost _every_ monorepo)

NullVoxPopuli:
> We'll need tests for this that fail before the change and pass after.

How do you want that to happen?
Are there existing tests where you generate whole projects and additional plugins?

bmish:
@NullVoxPopuli yes we have tests / test fixtures that set up fake projects and plugins. We will need to do something similar for this monorepo situation.

NullVoxPopuli:
@bmish I've added the beginnings of a test, the main thing is that I need to set the baseDir of a `Project` and that's not supported in the current version of fixurify project -- and "just upgrading" breaks a bunch of stuff.

NullVoxPopuli:
@bmish anyway to move forward without the test? or is there some other way to simulate monorepo? I suppose I could _not_ use fixturify-project for the monorepo test :thinking: 

This is just blocking adoption within _all_ monorepos :upside_down_face: 

imo, letting node do the module resolution by default (instead of via file path) is the more correct thing to do anyway :partying_face:

bmish:
Having tests for all changes is very important to me, especially for a change like this. I made sure all the changes that went into the v4 release were comprehensively tested. So it will be great for us to get this monorepo scenario tested. I can try to help with the `fixturify-project` upgrade over the next few days if you haven't figured that out. Thanks for starting on the monorepo test in the meantime.

NullVoxPopuli:
> Having tests for all changes is very important to me

I'm with ya there -- especially with big projects.

I don't think I want to deal with fixturify changes :upside_down_face: 
Should I wait for you to upgrade that? or should I proceed with these tests without fixturify?

bmish:
My preferred path is that we get the `fixturify-project` upgrade in so we can utilize that. However, I spent a bit of time trying to figure it out but didn't make much progress and it was not fun...I'll see if I can try again in next few days.

bmish:
Tests for this should hopefully be unblocked now that this has been merged:
* https://github.com/ember-template-lint/ember-template-lint/pull/2407

lib/get-config.js
scalvert: Are you suggesting that your monorepo has `ember-template-lint` in the root (presumably as a devDep), and the plugins in subdirs?

NullVoxPopuli: Just that Monorepo hoisting can place the package anywhere, we can't assume the location

scalvert: Sure, but typically the idea in a monorepo is to keep your tool devDeps in the root package.json. If you did this would this not work for you?

scalvert: OK sure. I suppose from my perspective that's part of the benefit of monorepos - minimizing tool dependencies so you have less maintenance. But of course that's not super relevant, though what I was getting at was trying to first determine if this was an issue that could be handled via a recommended setup vs. a code change.

NullVoxPopuli: makes sense -- important question though -- why were we _not_ relying on the built-in node resolution algo and trying to find deps ourselves?

rwjblue:
Taking another crack at this over in https://github.com/ember-template-lint/ember-template-lint/pull/2471

lib/get-config.js
rwjblue: > why were we not relying on the built-in node resolution algo (which happens for us via import/require) and trying to find deps ourselves?

Because the node resolution algorithm is actually wrong in this case 😸. Specifically, the node modules algorithm would suggest that the plugin **must** be resolvable from whatever location `ember-template-lint` itself is installed at. **Which is not correct!** The plugin will be installed somewhere relative to the `.template-lintrc.js` file **not** where ever** `ember-template-lint` happens to be installed. This is the point of the current `resolve.sync` code.

rwjblue:
More tests being added in https://github.com/ember-template-lint/ember-template-lint/pull/2473

rwjblue:
At the moment, I'm not able to reproduce a failure (in either #2471 or #2473). I think we need a better understanding of the failure (preferably via a reproduction in the "clone this, npm install, npm test" form) that @NullVoxPopuli was trying to fix.

NullVoxPopuli:
full repro steps:

_Terminal 1_
```bash
git clone git@github.com:NullVoxPopuli/limber.git
cd limber
git checkout etl-repro
yarn
```

_Terminal 2_
```bash
git@gitlab.com:NullVoxPopuli/ember-template-lint-plugin-tailwindcss.git
cd ember-template-lint-plugin-tailwindcss
git checkout vitest
yarn
yarn build
yarn link
```

_Back in Terminal 1_
```bash
yarn link ember-template-lint-plugin-tailwindcss
cd frontend
yarn link ember-template-lint-plugin-tailwindcss
yarn lint:hbs
```

rwjblue:
Thank you @NullVoxPopuli! Finally reproduced this style of failure over in https://github.com/ember-template-lint/ember-template-lint/pull/2475.

The good news is that it is specifically limited to cases where `resolve.sync` does not find a main entry point, but `import()` would. That is mostly going to be when you have `exports` setup but no `main` (but not _only_ that scenario, that just seems like the most likely one).

NullVoxPopuli:
> The good news is that it is specifically limited to cases where resolve.sync does not find a main entry point, but import() would. That is mostly going to be when you have exports setup but no main (but not only that scenario, that just seems like the most likely one).

I missed this.
Hmm.

This doesn't seem as bad as i though.
Will need to take a look

Issue #2319, opened 2022-01-05
Plan v5 Release


bmish:
### Planned changes for v5

* #2669
* #2606
* Update configs:
   * #2671
   * #2679
   * DONE: Confirm that `a11y` and `stylistic` configs are up-to-date
* Rename rules
   * #2054
   * #2675
* Adjust rule option defaults
   * #2674
   * #2672
   * #2673
   * #2657
* Update dependencies
   * #2676
   * #2677 
   * #2641
   * #2680
* https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/issues/236
* TODO: let us know if you want to include any other breaking changes

Done already?
* Fix serious monorepo plugin / Windows issues: https://discord.com/channels/480462759797063690/666416704418611233/949340263485177866 

Future changes not yet ready:

* Match ESLint behavior
   * [Allow any file extension to be matched in globs to match ESLint's behavior](https://github.com/ember-template-lint/ember-template-lint/pull/2366#issuecomment-1035008823)
   * Print out full paths by default (variation of https://github.com/ember-template-lint/ember-template-lint/pull/2491#issuecomment-1115253179)
   * #2366
* Add more rules to `stylistic` config
   * #2575 (this rule might be too noisy/opinionated to include)
* New `recommended` rules
   * @runspired [requested](https://discord.com/channels/480462759797063690/666416704418611233/927660252420259860) to have [no-model-argument-in-route-templates](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-model-argument-in-route-templates.md) considered as a `recommended` rule but the rule and Ember documentation may need tweaks before this would make sense
* Re-evaluate whether we should be publicly-exporting so many [helpers](https://github.com/ember-template-lint/ember-template-lint/blob/c9730dd5b3e07b0b727e3f5a5b2bb4795ca53b97/lib/index.js#L10) in our Node API

For reference, [v4.0.0](https://github.com/ember-template-lint/ember-template-lint/releases/tag/v4.0.0) was released 2022-01-05, discussion in #1908.

v5 Preparation PR: https://github.com/ember-template-lint/ember-template-lint/pull/2670

Pre-release versions:
* https://github.com/ember-template-lint/ember-template-lint/releases/tag/v5.0.0-alpha.0

NullVoxPopuli:
when are you wanting to release v5?

bmish:
[...]

PR #2366, opened 2022-02-07
check for svg to extend the supportedExtensions

SkoebaSteve:
We have relied for over a year on ember template lint to lint our svg files for certain issues(embedded images, text, etc). This is useful, since we can rely on the integration we already use for our template file lint.

We used to have our files only nested one deep, so no need for doing a glob search pattern. We reorganized our folder structure and will now have svgs nested in multiple folders deep.

When wrapping it in \"public/**/*.svg\", it does seem to do a recursive glob search, but in this case it executes the executeGlobby function and that hardcodes the supported extensions to be `let supportedExtensions = new Set(['.hbs', '.handlebars']);`

This change proposes to allow for linting of SVG files as well. I'm open to ideas and suggestions, one idea I had is to extend it with`.html` files as well for example.

ps. I was looking into adding a test scenario but I'm not sure where to best place this so any suggestion would be great!

### DIFF
--- a/lib/helpers/cli.js
+++ b/lib/helpers/cli.js
@@ -221,6 +221,10 @@ export function getPossibleOptionNames(specifiedOptions) {
 function executeGlobby(workingDir, pattern, ignore) {
   let supportedExtensions = new Set(['.hbs', '.handlebars']);
 
+  if (path.extname(pattern) === '.svg') {
+    supportedExtensions.add('.svg');
+  }
+
   // `--no-ignore-pattern` results in `ignorePattern === [false]`
   let options =
     ignore[0] === false ? { cwd: workingDir } : { cwd: workingDir, gitignore: true, ignore };


lib/helpers/cli.js
```diff
@@ -221,6 +221,10 @@ export function getPossibleOptionNames(specifiedOptions) {
 function executeGlobby(workingDir, pattern, ignore) {
   let supportedExtensions = new Set(['.hbs', '.handlebars']);
 
+  if (path.extname(pattern) === '.svg') {
```
bmish, line 224: Hmm, I don't think that treating a regexp pattern as a path is robust enough. A regexp pattern doesn't necessarily look like a path, so while this might work sometime, it isn't really a reliable way to do this.

bmish (commented):
There's some useful context in these PRs / linked issues about our current behavior: 

* https://github.com/ember-template-lint/ember-template-lint/pull/1556
* https://github.com/ember-template-lint/ember-template-lint/pull/1232

I haven't had a chance to review all of this yet so I don't have a recommendation yet.

I do want to call out ESLint's behavior, as long-term, we've been moving toward ember-template-lint matching ESLint's behavior.

https://eslint.org/docs/user-guide/command-line-interface#--ext
> --ext
> 
> This option allows you to specify which file extensions ESLint will use when searching for target files in the directories you specify. By default, ESLint lints *.js files and the files that match the overrides entries of your configuration.
> 
> Note: --ext is only used when the arguments are directories. If you use glob patterns or file names, then --ext is ignored.
>
> For example, npx eslint lib/* --ext .js will match all files within the lib/ directory, regardless of extension.

SkoebaSteve:
> Note: --ext is only used when the arguments are directories. If you use glob patterns or file names, then --ext is ignored.
> 
> For example, npx eslint lib/* --ext .js will match all files within the lib/ directory, regardless of extension.

Following  ESLint's behaviour would be problematic in this case, since we would prefer to be able to use a glob pattern.

SkoebaSteve:
> I think ESLint's behavior would actually work for you: I tested with ESLint and any file matching a glob passed to ESLint will be linted, not limited to predefined file extensions. 

@bmish that sounds great, thanks for checking!

> Since we have been trying to align ember-template-lint with ESLint's behavior, we could move toward this, but it wouldn't be for a while given that we just had a major release.

Ok I understand, maybe the best thing for us now is to wait for this then. Pretty new to ember-template-lint but I'm willing to help out, is there anything you can think of?

bmish:
Just a heads up that we're going to release the v5 version in the next week or two (https://github.com/ember-template-lint/ember-template-lint/issues/2319). It would be a good time to make any breaking changes.

PR #2376, opened 2022-02-10
Add new rule `no-obscure-array-access`

jsturgis:
Adding a rule to discourage usage of the following syntax in the template `{{list.[0]}}`

See: https://discord.com/channels/480462759797063690/486548111221719040/542753450144956436
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/787.


test/unit/rules/no-obscure-array-access-test.js
```diff
@@ -0,0 +1,65 @@
...
+    "{{get this.list '0'}}",
+    '{{foo bar @list}}',
+    'Just a regular text in the template bar.[1] bar.1',
+    '<Foo foo="bar.[1]" />',
+  ],
+
+  bad: [
+    {
+      template: '{{foo bar=this.list.[0]}}',
+      result: {
```
bmish, line 21: Can you update this to use `verifyResults` snapshot testing like the other rule tests. We prefer that now.

docs/rule/no-obscure-array-access.md
```diff
@@ -0,0 +1,39 @@
...
+{{foo bar (get @list '1')}}
+```
+
+```hbs
+<Foo @bar={{get list '0'}} />
+```
+
+## References
+
+- [Ember discord discussion in ember-cli channel on 02/06/19](https://discord.com/channels/480462759797063690/486548111221719040)
```
bmish, line 39: This link doesn't seem to link me to the specific message, only the channel. Can we link to the specific message?

docs/rule/no-obscure-array-access.md
```diff
@@ -0,0 +1,39 @@
...
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+{{foo bar=list.[0]}}
+```
+
+```hbs
+{{foo bar @list.[1]}}
```
bmish, line 16: Is this missing the equals sign?

And I actually think one good and one bad example are probably enough here. We don't need to show every possible variation of the surrounding syntax. Only the specific array access is what matters. And having too many examples can be distracting.

bad

```hbs
<Foo @bar={{list.[0]}} />
```
good

```hbs
<Foo @bar={{get list '0'}} />
```

bmish (changes requested):

docs/rule/no-obscure-array-access.md
jsturgis: Yeah good catch.

lib/rules/no-obscure-array-access.js
```diff
@@ -0,0 +1,23 @@
+import Rule from './_base.js';
+
+export function ERROR_MESSAGE(expression) {
+  return `Obscure expressions are prohibited. ${expression}. Please use Ember's get helper instead. i.e {{get list '0'}} `;
```
bmish, line 4: We don't need to repeat the text in error message. Let's remove that.

bmish (changes requested):

docs/rule/no-obscure-array-access.md
bmish: Do you think we should remove unnecessary examples as I mentioned?

lib/rules/no-obscure-array-access.js
jsturgis: Fixed.

docs/rule/no-obscure-array-access.md
bmish: Thoughts on this?

lib/rules/no-obscure-array-access.js
```diff
@@ -0,0 +1,79 @@
+import { builders } from 'ember-template-recast';
+
+import Rule from './_base.js';
+
+export function ERROR_MESSAGE() {
+  return `Obscure expressions are prohibited. Please use Ember's get helper instead. e.g. {{get @list '0'}} `;
+}
+
+function buildParams(node) {
```
bmish, line 9: Can you add commenting all throughout this file explaining what is going on at each step? The logic is a bit complex.

bmish (changes requested):

docs/rule/no-obscure-array-access.md
```diff
@@ -0,0 +1,40 @@
+# no-obscure-array-access
+
+Using obscure expressions `{{list.[1].name}}` is discouraged and is likely to be deprecated soon.
```
chrisrng, line 3: We should mention that this break Prettier too iirc (https://github.com/prettier/prettier/issues/4908#issuecomment-682125029)

There's also a codemod to help https://github.com/mahirshah/ember-array-index-template-codemod

jsturgis: Wow nice thanks for this.

jsturgis: From [the discord](https://discord.com/channels/480462759797063690/486548111221719040/542754258689458197) it seems like the recommended way to access the array index is:

```
{{get (get users 0) 'name'}}
```
Nesting the array index getter inside the non array path getter, that is how I implemented the fixer codemod for this rule. If we don't need to nest it makes the fixer much simpler and I can do basically what that codemod is doing.

jsturgis: For prettier, it doesn't break when I try it locally, it just converts `[0]` to `0`.

chrisrng: Ah nice maybe that was fixed it used to break

docs/rule/no-obscure-array-access.md
jsturgis: I missed that originally, I will do it now.

lib/rules/no-obscure-array-access.js
bmish: Thanks for adding the comments. It's extremely valuable to have the autofixer. The logic can be further simplified later but it's important to have comments and solid tests now.

bmish (approved):
@jsturgis can you fix CI?

bmish:
Also highly recommend doing a quick test of this rule on some real-world codebases to ensure it works properly and doesn't crash.

lib/rules/no-obscure-array-access.js
jsturgis: Cool, I've simplified the fixer now so it should be a littler more straightforward.

bmish: Wow, nice!

bmish:
Thank you!

jsturgis:
> Also highly recommend doing a quick test of this rule on some real-world codebases to ensure it works properly and doesn't crash.

Good idea thanks, I ran this against the Linkedin codebase and it worked as expected.

boris-petrov:
@jsturgis, @bmish - I think this rule should also handle `{{foo.0}}` - this actually does work but is still bad practice so the rule should catch it.

bmish:
@jsturgis can you ensure that's tested?

PR #2388, opened 2022-02-16
[no-obscure-array-access] Handle paths that contain digits

jsturgis:
This change updates the regexp to handle pathExpressions that contains digits.


bmish (approved):
Thanks!

PR #2389, opened 2022-02-19
Fix string handling errors with `no-invalid-aria-attributes` rule

### LINKED ISSUE #2385: [no-invalid-aria-attributes] Doesn't work at all?
I don't understand - even using a simple `<button type="button" aria-haspopup="true">` leads to `The value for aria-haspopup must be a single token from the following: false, true, menu, listbox, tree, grid, dialog.`. Am I doing something wrong?

judithhinlung:
This rule addresses string handling bugs in the `no-invalid-aria-attributes` lint rule, which fixes #2385.

### Background
Currently this rule fails when handling "true" and "false" values for aria-attributes that include them in their list of permitted values. Also, concat statements, or strings containing interpolated variables, are not accepted. 

This PR proposes the following fixes:
- Convert all boolean values in the list of permitted values to strings prior to validating against a given attribute value.
- Skip validation of concat statements, since we cannot know the values of the interpolated variables.

**Allowed**:

```
# examples
<button aria-haspopup="true"></button>
<div aria-label="{{@name}} "foo" {{@bar}}"></div>
```


bmish:
@judithhinlung is this ready or still in progress?

bmish (approved):
Thanks!

rrenno:
Thank you so much, this cleared up most of our false-positive cases. 

I think there is still an additional use case that conflicts with the [no-unnecessary-concat](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-unnecessary-concat.md) rule.

`<div aria-label="{{@name}}"></div>` is now correctly allowed by this rule, however `no-unnecessary-concat` prefers `<div aria-label={{@name}}></div>`. Both however correctly evaluate to a string. 

Should something like `aria-label={{@name}}` be allowed -- at least by an optional flag? If yes, I can open a new issue.

bmish:
I think we should make sure the rules are compatible with each other by default. Can you open a new issue?

Issue #2395, opened 2022-02-23
Refactor printers to return built output as string vs. direct printing


scalvert:
The current architecture of the printers/formatters (we should unify the terminology here!) has the following issues:

1. Printers print output themselves, making them hard to test without passing in a mock console to output to
2. The `output-file` option only works on certain printers that have implemented it, and is completely ignored on others. This makes it very hard to know what the right magical combination of formatters is, and which will do what you want.
3. The fact that each printer needs to implement output file functionality itself means there's lots of duplicated code, and lots of potential for problems/bugs. Additionally, anyone implementing a custom printer will need to also do this themselves, placing an additional burden on our consumers.

I'd like to propose the following:

1. We unify on the name formatters, as that more closely matches our static analysis tool counterparts.
2. We refactor the formatters to return a string.
3. We centralize the output file functionality such that any formatter can output either to stdout or to a file. This avoids the need to know the secret incantation to pass to the CLI in order to get predictable results.
4. We optionally don't support multiple formatters as is partly implemented right now. If we think we need to support multiple formatters in a single run, we can revisit this.

Interested in your thoughts, @rwjblue @bmish

scalvert:
Unsure whether to consider this breaking, since we don't really document creating custom formatters, and there aren't many out there right now other than the ones I've created (which I'm happy to update!). Part of this change could/should be adding documentation for how to author formatters.

scalvert:
Resolved in #2401

PR #2401, opened 2022-02-25
Change formatters to return string output rather than outputting to stdout/output files

scalvert:
Implements #2395.

Solves:
- Inconsistencies between formatters in combination with `--output-file` (some formatters would write to a file, some would not. This appeared 'broken', so this could be considered a bugfix)
- The need for each formatter to handle output files themselves (including external formatters)

Changes:
- Renames `printers` to `formatters`
- Renames `print` method to `format`
- Changes how formatters handle the output to return strings vs. console logging internally
- Centralized the formatter output into the bin script, and provided a consistent way to output to an output file irrespective of formatter


bin/ember-template-lint.js
```diff
@@ -245,16 +249,30 @@ async function run() {
...
   let hasUpdatedTodos = options.updateTodo;
 
-  let printer = new Printer({
+  let formatter = loadFormatter({
     ...options,
     hasResultData: hasErrors || hasWarnings || hasTodos || hasUpdatedTodos,
   });
-  printer.print(results, todoInfo);
+
+  let output = formatter.format(results, todoInfo);
```
rwjblue, line 266: Let's check if `typeof formatter.format === 'function'` and if not fallback to existing behavior (makign this non-breaking)

rwjblue (commented):
Overall, this is a great change! Left one inline comment RE: how to make it not breaking, but let's do it!

bin/ember-template-lint.js
scalvert: 100%. This would be easy to switch back to using `print`. Will make that change.

bin/ember-template-lint.js
```diff
@@ -245,16 +249,34 @@ async function run() {
...
+        formatter.print(results, todoInfo);
+
+  if (output) {
+    if ('output-file' in options) {
+      let outputPath = writeOutputFile(output, formatter.outputFileType || 'txt', options);
+      console.log(`Report written to ${outputPath}`);
+    } else {
+      console.log(output);
+    }
+  }
```
rwjblue, line 279: Let's tweak this so that we don't accidentally "leak" the ability to return a value to `formatter.print`. I'm thinking we actually want something like (but probably don't take inline suggestion, due to prettier issues):

```suggestion
  if (typeof formatter.format === 'function') {
    if ('output-file' in options) {
      let outputPath = writeOutputFile(output, formatter.outputFileType || 'txt', options);
      console.log(`Report written to ${outputPath}`);
    } else {
      console.log(output);
    }
  } else {
    // support legacy formatters
    formatter.print(results, todoInfo);
  }
```

lib/formatters/json.js
```diff
@@ -1,17 +1,19 @@
 import Linter from '../linter.js';
 
-export default class JsonPrinter {
+export default class JsonFormatter {
+  outputFileType = 'json';
```
rwjblue, line 4: Is this used?

rwjblue: I see, it is used in the bin script to tell `writeOutputFile` what extension to use by default. 

Should we name it something different to make it clear that it's just a default extension? (if the user provides a file name, this value is not used at all)

lib/formatters/sarif.js
```diff
@@ -77,17 +70,13 @@ export default class SarifPrinter {
         };
 
         if (embedFileContents) {
-          try {
```
rwjblue, line 80: Why does this `try`/`catch` get deleted? Think it's not needed?

rwjblue (approved):

lib/formatters/sarif.js
scalvert: Ya. I mean I mainly deleted it because we were inline outputting the error and you don't really get any benefit there.

PR #2413, opened 2022-03-07
Add new rule `no-scope-outside-table-headings`

### LINKED ISSUE #2412: Proposal to add new a11y rule restricting the use of the scope attribute to `th` elements
Currently, ember-template-lint does not disallow HTML elements other than the <th> element from using the scope attribute. Scope is used on `<th>` elements to clarify the relationship between a table's header cells and data cells for screen readers. Because placing this attribute on other HTML elements has no effect, it is a best practice that only `<th>` elements declare a value for scope.

This proposes the authoring of a new rule that enforces that the scope attribute is only used on `<th>` elements. This is essentially a port of JSXs A11y [scope rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/scope.md).

References: 
- [H63: Using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/TR/WCAG20-TECHS/H63.html)
- [Tables Concepts](https://www.w3.org/WAI/tutorials/tables/)

judithhinlung:
This PR adds the new rule `scope` which fixes #2412 .
 
### Background
The scope attribute is used on `<th>` elements to clarify the relationship between a table's header cells and data cells for screen readers. Because placing this attribute on other HTML elements has no effect, only `<th>` elements should declare a value for scope.

This new rule enforces that the scope attribute is only used on `<th>` elements.  

References: 
- [H63: Using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/TR/WCAG20-TECHS/H63.html)
- [Tables Concepts](https://www.w3.org/WAI/tutorials/tables/)

 **Allowed**:
 
 ```
# examples
<th scope="row">some header cell</th>
<CustomHeader scope={{foo}} />
```

**Forbidden**:

```
# examples
<a scope="col"></a>
<table scope="rowgroup"></table>
```

### Testing
```
Test Suites: 130 passed, 130 total
Tests:       5 skipped, 7318 passed, 7323 total
Snapshots:   945 passed, 945 total
Time:        125.954 s
Ran all test suites.
✨  Done in 133.92s.

```

### DIFF
--- a/docs/rule/no-scope-outside-table-headings.md
+++ b/docs/rule/no-scope-outside-table-headings.md
@@ -22,4 +22,5 @@ This rule **allows** the following:
 
 ## References
 
-- [Tables Concepts](https://www.w3.org/WAI/tutorials/tables/)
+- [HTML \<th\> scope Attribute](https://www.w3schools.com/tags/att_th_scope.asp)
+- [scope - eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/scope.md)
--- a/docs/rule/no-scope-outside-table-headings.md
+++ b/docs/rule/no-scope-outside-table-headings.md
@@ -1,4 +1,4 @@
-# scope
+# no-scope-outside-table-headings
 
 The scope attribute is used on `<th>` elements to clarify the relationship between a table's header cells and data cells for screen readers. Scope is set to "row" or "col" for header cells that refer to a given row or column. For header cells that reference multiple rows or columns, set the scope attribute to "rowgroup" and "colgroup" and define the range for the rows or columns.
 
--- a/test/unit/rules/no-scope-outside-table-headings-test.js
+++ b/test/unit/rules/no-scope-outside-table-headings-test.js
@@ -8,13 +8,15 @@ generateRuleTests({
   good: [
     '<th scope="row">Some table heading></th>',
     `
-<table>
-<th scope="col">Table header</th>
-<td>Some data</td>
-</table>
-`,
+    <table>
+      <th scope="col">Table header</th>
+      <td>Some data</td>
+    </table>
+    `,
     '<CustomComponent scope />',
+    '<CustomComponent scope="row" />',
     '<CustomComponent scope={{foo}} />',
+    '{{foo-component scope="row"}}',
   ],
   bad: [
     {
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by scalvert** [comment]
Hey @judithhinlung, I know this is still in progress, but can we rename the rule to something a bit less general? Maybe something like `require-valid-scope` or something?

File: test/unit/rules/no-scope-outside-table-headings-test.js [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+    '<th scope="row">Some table heading></th>',
+    `
+<table>
+<th scope="col">Table header</th>
+<td>Some data</td>
+</table>
+`,
+    '<CustomComponent scope />',
+    '<CustomComponent scope={{foo}} />',
+  ],
```
bmish, line 18: Can we add these to `good`:
* `{{foo-component scope="row"}}`
* `<CustomComponent scope="row" />`

test/unit/rules/no-scope-outside-table-headings-test.js
```diff
@@ -0,0 +1,53 @@
...
+  bad: [
```
bmish, line 19: Can we add this to bad just to show that we don't care what the value is?
* `<td scope>`

bmish (changes requested):

docs/rule/no-scope-outside-table-headings.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+This rule **allows** the following:
+
+```hbs
+<th scope="row">A header cell</th>
+<CustomHeader scope={{foo}} />
+```
+
+## References
```
bmish, line 23: Can you also link to something specific about `scope`? https://www.w3schools.com/tags/att_th_scope.asp

bmish (approved):

docs/rule/no-scope-outside-table-headings.md
bmish, line 23: Can you also link to 

* https://www.w3.org/TR/WCAG20-TECHS/H63.html
* And the corresponding React lint rule

bmish: Can you also link to 

* https://www.w3.org/TR/WCAG20-TECHS/H63.html
* And the JSXs A11y [scope rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/scope.md).

judithhinlung:
Hi @bmish, is it okay to merge this pr?

Issue #2420, opened 2022-03-16
`no-whitespace-within-word` false positive with dashes & I


mydea:
I.e. the following text:

```
Hello - I like this.
```

Triggers an error by the `no-whitespace-within-word` rule. The reason is, that it treats the dash as a word character, I guess.

rrenno:
I think this PR caused a regression for the ability to use digits back-to-back, i.e. 

```javascript
/\d+-\d+/
```

Created an issue here: https://github.com/ember-template-lint/ember-template-lint/issues/2464

PR #2421, opened 2022-03-16
Handle dashes in `no-whitespace-within-word` rule

mydea:
This allows the following:

```
Hello - I like this
```

So basically, it considers dashes as "whitespace" for the purpose of this rule, instead of as characters/

Closes https://github.com/ember-template-lint/ember-template-lint/issues/2420

### DIFF
--- a/lib/rules/no-whitespace-within-word.js
+++ b/lib/rules/no-whitespace-within-word.js
@@ -57,8 +57,10 @@ const whitespaceCharacterList = new Set([
   '&ic;',
 ]);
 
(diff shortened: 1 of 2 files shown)


bmish (approved):
Seems reasonable, thanks!

rrenno:
I think this PR caused a regression for the ability to use digits back-to-back, i.e. 

```javascript
/\d+-\d+/
```

Created an issue here: https://github.com/ember-template-lint/ember-template-lint/issues/2464

PR #2422, opened 2022-03-16
Add `&times` to `no-bare-strings` default allow list

jbottigliero:
Not sure if this exclusion is intentional or not, but it seems to be in line with other HTML entities in the allow list...


jbottigliero:
Thanks for the heads up @bmish; I've opened #2423, which seems closer to that desired functionality... for my use case I'll likely revert to an `allowlist`, but it seems like others might find the `allowHTMLCharacters` configuration useful? Happy to close this out and continue the conversation in the new PR – or close both out! 😅

lifeart (approved):

NullVoxPopuli (approved):

Issue #1131, opened 2020-02-21
[no-bare-strings] Automatically allow all HTML entities


boris-petrov:
Things like `&times;` should be ignored by default by the `no-bare-strings` rule because they are built-in HTML constructs.

Forked from https://github.com/ember-template-lint/ember-template-lint/issues/662.

cc @rwjblue

rwjblue:
Thank you @boris-petrov!

rwjblue:
For context, I think that we can implement this by adding all known HTML entities to the default allow list. Thankfully, simple-html-tokenizer bundles the list of entities for us.

Turbo87:
**why** should we allow any of these entities?

lifeart:
@Turbo87 this entities seems multilingual (in most cases)

Turbo87:
maybe some of them, but definitely not all. and I'm even doubtful about the rest. also, even if they are the same across all locales, why not still put them in the translations to be consistent?

lifeart:
there is already list of `& * ;` items https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-bare-strings.js#L45

I think real-life use-cases for it it's pretty rare and we don't need to warn about it (because it can be meaningful inside app context - for example "calculator" app with special symbols)

bmish:
I agree with having an option `allowHTMLEnttities` which defaults to true to maintain the current behavior.

initram:
If we allow all HTML entities then we are not maintaining current behavior. Entities like &oslash; and &otilde; are not allowed by default today. I think it would weaken the rule to allow all HTML enties, especially if you look at the set of all HTML5 entities.

There is some merit to expanding the default white-list, but at some point we would have to tell people to write their own white-lists if they want more symbols allowed. If we allow &times; should we also allow the text "×"? Today we have a default whitelist that contains a bunch of commonly used symbols and their html entity equivalent.

fusion2004:
@initram I think my main problem is I don't want to have to maintain the entire list. I just wanted to add one, and get the rest of default whitelist.

rwjblue:
Ya, but we can work on that part regardless I think. A couple of ways:

* expose the default array for importing, so folks can do `[...defaults, 'thing-here']`
* add another option that is something like `additionalAllowedStrings` or something

PR #2423, opened 2022-03-17
feat:  adds `allowHTMLCharacters` configuration property for `no-bare-strings` rule

jbottigliero:
A more in-depth pass at #2422 implementing the functionality described in https://github.com/ember-template-lint/ember-template-lint/issues/1131.


jbottigliero:
Let me know if this looks closer to the direction you all were hoping to go with this functionality, if so, I'll get the tests cleaned up!

lifeart:
seems we may insert data https://github.com/tildeio/simple-html-tokenizer/blob/15b1d6e7a825e68fe7a68a7080e272c374adbc27/src/generated/html5-named-char-refs.ts, without dependency

Issue #2333, opened 2022-01-12
separate rule violations with the same line, column and rule name cannot both create a TODO


runspired:
Example failure (1 TODO created, 1 violation not converted to TODO)

```
  2:2  todo  <div> has a role of tab, it cannot have semantic descendants like <h5>  require-presentational-children
  2:2  error  <div> has a role of tab, it cannot have semantic descendants like <a>  require-presentational-children
  ```
  
  note the offending `h5` and `a` elements are not on line 2 in the associated file, that is merely where the div opens.
  
  Offending content (sorry for the 😱 code)
  
  ```hbs
  <div class="card analytics-runner-card no-radius">
    <div
    class="card-header analytics-runner-card-header no-radius"
    role="tab"
    id="columnsAria"
  >
    <h5 class="mb-0">
      <a
        data-toggle="collapse"
        class="nav-link"
        data-parent="#accordion"
        href="#columns"
        aria-expanded="false"
        aria-controls="columns"
      >
        {{t "analytics.column_order"}}
      </a>
    </h5>
  </div>
 </div>
  ```

**Comment by bryanhickerson** [comment]
I have the same issue, I had to turn off `require-presentational-children` as a result.

**Comment by rwjblue** [comment]
Hmmm. Seems to me like the issue is with which node we are reporting as incorrect (RE: line + col). We should either report only a single error for this case (which would make todo's work fine), or report individual errors but using the specific invalid child as the line+col instead of always using the parent's node.

**Comment by drewlee** [comment]
I've run into the same issue but using `ember-template-lint` v4.3.0 and single file todos. Given the following example markup will yield 2 lint violations.
```html
<div role="button">
  <p>Text content</p>
  <strong>Supplementary content</strong>
</div>
```
```shell
  7:0  error  <div> has a role of button, it cannot have semantic descendants like <p>  require-presentational-children
  7:0  error  <div> has a role of button, it cannot have semantic descendants like <strong>  require-presentational-children

✖ 2 problems (2 errors, 0 warnings)
error Command failed with exit code 1.
```
After generating todos, running `ember-template-lint` results in the following output.
```shell
$ ember-template-lint .
app/templates/application.hbs
  7:0  error  <div> has a role of button, it cannot have semantic descendants like <strong>  require-presentational-children

✖ 1 problems (1 errors, 0 warnings)
error Command failed with exit code 1.
```

[...]

PR #2433, opened 2022-03-22
Fixes require-presentational-children to return correct line/column of node

scalvert:
Fixes: #2333


scalvert:
cc/ @runspired @drewlee @bryanhickerson

drewlee:
Looks like we need to update the associated test snapshots since we're now outputting the child node details instead of the main node.

drewlee (approved):
Thanks, Steve!

scalvert:
@faith-or any chance you could review this to see if it makes sense?

faith-or (approved):
LGTM

PR #2434, opened 2022-03-23
Extract `hasParentTag` utility function

JoaoDsv:
The `hasFormParent` function only exists in the `lib/rules/require-button-type.js` rule file.

We want to be able to reuse it in other rule, so we are extracting it in a more generic version.

_FYI it will soon be used to complete [this rule](https://github.com/ember-template-lint/ember-template-lint/pull/1931)._

### DIFF
--- a/lib/helpers/has-parent-tag.js
+++ b/lib/helpers/has-parent-tag.js
@@ -1,0 +1,11 @@
+import { match } from './node-matcher.js';
+
+export default function hasParentTag(path, tag) {
+  let parents = [...path.parents()];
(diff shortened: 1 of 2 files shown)


bmish:
This looks fine, is this still a draft?

JoaoDsv:
> This looks fine, is this still a draft?

@bmish It's indeed not in draft anymore, I just wanted your opinion on it, as you was the requestor of this extraction. Feel free to approve it if it looks good to you 🙏

bmish (approved):
Thanks!

PR #2435, opened 2022-03-24
Add CLI option `--no-error-on-unmatched-pattern`

CvX:
#2245 added the behavior (inspired by eslint) to error out when no matching files are found.

This PR adds a related flag (also from eslint) `--no-error-on-unmatched-pattern` that can disable this behavior when needed.

### DIFF
--- a/lib/helpers/cli.js
+++ b/lib/helpers/cli.js
@@ -12,6 +12,13 @@ import camelize from './camelize.js';
 
 const STDIN = '/dev/stdin';
 
+class NoMatchingFilesError extends Error {
+  constructor(...params) {
+    super(...params);
+    this.name = 'NoMatchingFilesError';
+  }
+}
+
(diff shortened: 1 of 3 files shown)


lib/helpers/cli.js
```diff
@@ -113,6 +120,10 @@ export function parseArgv(_argv) {
       describe: 'Number of warnings to trigger nonzero exit code',
       type: 'number',
     },
+    'no-error-on-unmatched-pattern': {
+      describe: 'Prevent errors when pattern is unmatched',
+      boolean: true,
```
bmish, line 125: Can we add `default: true` here and in the help-text for clarity?

bmish (changes requested):

lib/helpers/cli.js
bmish: Maybe good as a follow-up?

Issue #2464, opened 2022-04-12
Regression from `Handle dashes in no-whitespace-within-word rule` PR #2421


rrenno:
Super awesome effort on #2421 to fix the dash support, however it appears this broke the ability to use digits back-to-back, i.e. 

```javascript
/\d+-\d+/
```

So for example the following in our code base now throws a `no-whitespace-within-word` error whereas before it did not. The `5-7` being the formerly allowed text.
```handlebars
<li>
    <b>Select a subset.</b> For Java/Python/C/C++, we recommend starting with requiring perhaps 5-7 per week.
</li>
```

It's worth notiing that digits with a space pass linting:
```handlebars
<li>
    <b>Select a subset.</b> For Java/Python/C/C++, we recommend starting with requiring perhaps 5 - 7 per week.
</li>
```
Also alphabet characters without a space pass linting: 
```handlebars
<li>
    <b>Select a subset.</b> For Java/Python/C/C++, we recommend starting with requiring perhaps five-seven per week.
</li>
```

mydea:
Hmm, I guess the problem here is that dashes are basically considered "spaces" now by the rule, which means it basically interprets it as `perhaps 5 7 per week`. Whereas before it considered `5-7` one "word". IMHO the "correct" solution is probably to allow/ignore numbers, I'd say?

mydea:
I made a PR https://github.com/ember-template-lint/ember-template-lint/pull/2465 rewriting the implementation to use a regex instead. IMHO this is more robust and also handles this case properly!

PR #2465, opened 2022-04-13
Better handling of dashes in `no-whitespace-within-word` rule

mydea:
This should be more robust than the previous implementation.

This fixes https://github.com/ember-template-lint/ember-template-lint/issues/2464, and also handles multi-whitespace properly (e.g. `A&nbsp;&nbsp;B C`, for example). You can configure the regex for the single-character lookup (e.g. if you also want to handle some special characters etc.), by default we use `[a-zA-Z]`.

I am also not 100% sure if we _really_ need to include the full list of possible white-space characters here, or if it is good enough to just have the main ones (e.g. `&nbsp;`) in there - or maybe also make this configurable. Not sure if there _is_ a noticeable performance impact having a rather large regex, but I would imagine for most users/use cases this would be a _really_ rare scenario to run into `X&InvisibleComma;X&InvisibleComma;X&InvisibleComma;` or something like that.

I can either leave it as it is, or reduce the character set somewhat, or reduce it to a smaller subset and make it configurable for users who really need the full list? My feeling is a subset like this maybe:

```js
[
  '&nbsp;',
  '&NonBreakingSpace;',
  '&thinsp;',
  '&ThinSpace;',
  '&hairsp;',
  '&VeryThinSpace;',
  '&ThickSpace;',
  '&ZeroWidthSpace;',
  '&NegativeVeryThinSpace;',
  '&NegativeThinSpace;',
  '&NegativeMediumSpace;',
  '&NegativeThickSpace;',
  '&MediumSpace;',
  '&ThickSpace;',
]
```

So basically the "named" whitespace entites, may be enough?

### DIFF
--- a/lib/rules/no-whitespace-within-word.js
+++ b/lib/rules/no-whitespace-within-word.js
@@ -2,9 +2,8 @@ import Rule from './_base.js';
 
 const ERROR_MESSAGE = 'Excess whitespace in layout detected.';
 
-const whitespaceCharacterList = new Set([
+const WHITESPACE_ENTITY_LIST = [
   '&#32;',
-  ' ',
   '&#160;',
   '&nbsp;',
   '&NonBreakingSpace;',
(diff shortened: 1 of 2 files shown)


docs/rule/no-whitespace-within-word.md
```diff
@@ -46,6 +46,10 @@ This rule **allows** the following:
 
 This rule uses the heuristic of letter, whitespace character, letter, whitespace character, letter which makes it a good candidate for most use cases, but not ideal for some languages (such as Japanese).
 
+## Configuration
+
+* object - { 'character-regex': '[a-zA-Z]' } -Specify a custom regex (as string) to identify single characters. The default matches uppercase & lowercase characters.
```
bmish, line 51: Could it be overkill to have an option for this? I feel like this rule should just work and shouldn't require customization to handle basic use cases.

mydea: My idea was to account to be able to opt-in for non-latin characters, e.g. Chinese or Arabic (although I guess for those more symbol-based locales it makes sense to _not_ check them anyhow, but not an expert on that by any means) or something like German Umlaute, if somebody really needs this. But I'm also happy to remove the config option!

bmish: I think it's a nice goal for this rule to be able to handle different languages, but I don't think this option is practical currently. What are the chances someone using a different language is going to formulate the regexp for their own language to provide to this option? Pretty low I suspect.

The rule doc currently mentions `but not ideal for some languages` so I'd rather just focus on English now and omit this obscure option which.

That said, if there are more user-friendly ways we can make this rule generic and able to handle different languages, I'm all for it.

mydea: :+1: OK, I removed the config option :) IMHO it should be good enough for most cases, and it should, if, only produce false negatives, not false positives, which is OK for a linting rule like this I'd say anyhow.

bmish (approved):
Thanks!

docs/rule/no-whitespace-within-word.md
bmish: We can consider an option or other improvements later if needed.

PR #2466, opened 2022-04-13
Remove validation of mustache statements in `no-invalid-aria-attributes` rule

### LINKED ISSUE #2405: Rule conflict between `no-invalid-aria-attributes` and `no-unnecessary-concat`
`no-invalid-aria-attributes` was recently patched to allow for handlebars as acceptable for aria attributes, #2389 e.g.: 
`<div aria-label="{{@name}}"></div>` 

However `no-unnecessary-concat` prefers `<div aria-label={{@name}}></div>` but `no-invalid-aria-attributes` marks this as invalid without the quotes. Both however correctly evaluate to a string.

It seems like `no-invalid-aria-attributes` should be aware of  `no-unnecessary-concat` requirements -- opening this issue per https://github.com/ember-template-lint/ember-template-lint/pull/2389#issuecomment-1053609716

### LINKED ISSUE #2451: `no-invalid-aria-attributes` fails when it is passed a valid if statement
We have a component in the code which has a different (translated) aria-label based on some check:

```
  <button
    aria-label={{if @isNew (t "actions.add") (t "actions.edit")}}
    ....
  />
  ```
  I think this should pass the linting check since there will always be a valid string assigned to the aria-label attribute?
  
  I tried this case in the existing unit test for this rule and the test indeed fails, I can try to come up with a proposal for a fix but I'm not familiar with this linting code base.

### PR DESCRIPTION
**Pull request by judithhinlung**
This PR fixes #2405 and fixes #2451.

Besides string, number, and boolean literals, mustache statements often also contains helpers, variables, and other nested statements. Since it is not possible to check what a given mustache statement evaluates to at the template level, this PR removes their validation to prevent valid ARIA attribute values in complex statements from failing the rule.

**Allowed**:
```
# examples
<div aria-label={{this.foo}} />
<div aria-label={{if @label @label @placeholder}} />
<button aria-label={{if @isNew (t "actions.add") (t "actions.edit")}}></button>
```

### Testing
Test Suites: 135 passed, 135 total
Tests:       5 skipped, 7459 passed, 7464 total
Snapshots:   969 passed, 969 total
Time:        96.051 s
Ran all test suites.
✨  Done in 107.73s.

### DIFF
--- a/docs/rule/no-invalid-aria-attributes.md
+++ b/docs/rule/no-invalid-aria-attributes.md
@@ -11,7 +11,7 @@ This rule **forbids** the following:
 ```hbs
 <input type="text" aria-not-real="true" />
 <div role="region" aria-live="bogus">Inaccessible live region</div>',
-<button type="submit" aria-invalid={{if this.foo "true" "woosh"}}>Submit</button>
+<button type="submit" disabled="true" aria-disabled="123">Submit</button>
 ```
 
 This rule **allows** the following:
--- a/lib/rules/no-invalid-aria-attributes.js
+++ b/lib/rules/no-invalid-aria-attributes.js
@@ -84,23 +84,6 @@ function validityCheck(expectedType, permittedValues, allowUndefined, value) {
   }
 }
 
-function getValuesFromMustache(mustacheNode) {
-  let valuesList = [];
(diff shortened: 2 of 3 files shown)


bmish (approved):
Thanks!

PR #2468, opened 2022-04-14
Ensure configuration can be resolved from a parent directory when no `--config-path` override is present (e.g. within a monorepo with `.template-lintrc.js` in the monorepo root).

### LINKED ISSUE #2467: Can we use find-up when looking for the config file?
Currently in `resolveProjectConfig` (https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/get-config.js#L50), we use `findUpSync` when `options.configPath` is not present, but not if otherwise. Can we update the `if` block to use find-up as well? `options.configPath` has a default value (`.template-lintrc.js`), so most of the time we won't even go into the `else` block. Using `find-up` would make managing config in yarn workspaces easier. Thanks.

```js
  if (options.configPath) {
    // update this to use `find-up`?
    configPath = path.resolve(workingDir, options.configPath);
    // ...
  } else if (options.configPath === false) {
    return {};
  } else {
    configPath = findUpSync(CONFIG_FILE_NAME, {
      cwd: workingDir,
    });
    // ....
  }
```

scalvert:
Enhance config discovery to allow configs to be in parent directories. This is particularly useful for monorepos, where subpackages invoke ember-template-lint, but it's desirable to have a single config file in the root.

Fixes #2467 

/cc @NullVoxPopuli

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -67,11 +67,13 @@ function getTodoConfigFromCommandLineOptions(options) {
 }
 
 function _isOverridingConfig(options) {
+  let defaultArgs = parseArgv(['fake-file-to-get-default-options.hbs']);
+
   return Boolean(
-    options.config ||
-      options.rule ||
-      options.inlineConfig === false ||
-      options.configPath !== '.template-lintrc.js'
+    options.config !== defaultArgs.config ||
+      options.rule !== defaultArgs.rule ||
+      options.inlineConfig !== defaultArgs.inlineConfig ||
+      options.configPath !== defaultArgs.configPath
   );
 }
 
--- a/test/acceptance/todo-cli-test.js
+++ b/test/acceptance/todo-cli-test.js
@@ -6,8 +6,6 @@ import Project from '../helpers/fake-project.js';
 import run from '../helpers/run.js';
 import setupEnvVar from '../helpers/setup-env-var.js';
 
-const ROOT = process.cwd();
-
 jest.setTimeout(10_000);
 
 function buildReadOptions() {
@@ -25,7 +23,6 @@ describe('todo usage', () => {
   });
 
   afterEach(function () {
-    process.chdir(ROOT);
     project.dispose();
   });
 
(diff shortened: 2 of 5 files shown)


lib/get-config.js
```diff
@@ -51,20 +44,11 @@ export async function resolveProjectConfig(workingDir, options) {
...
-      // needed in order to ensure that we only squelch module missing errors for
-      // the config path itself (and not other files that the config path may
-      // require itself)
-      require.resolve(configPath);
-    } catch (error) {
-      let isModuleMissingError = ALLOWED_ERROR_CODES.has(error.code);
-      if (!isModuleMissingError) {
-        throw error;
-      }
+    configPath = findUpSync(options.configPath, {
```
rwjblue, line 47: This isn't quite right, the intent of the code here is that anyone explicitly passing `--config-path` does **not** get "Find up" behavior (and this change here breaks that intent). The real bug is that `--config-path` is explicitly defaulted in our bin scripts arguments, so we can't fall into the fallback `else` clause down below that already uses `findUp`. 

Chatted with @scalvert about this, and he is going to update and add some additional tests to cover those cases.

scalvert: Updated as per our chat.

bin/ember-template-lint.js
```diff
@@ -67,11 +67,13 @@ function getTodoConfigFromCommandLineOptions(options) {
...
+
   return Boolean(
-    options.config ||
-      options.rule ||
-      options.inlineConfig === false ||
-      options.configPath !== '.template-lintrc.js'
+    options.config !== defaultArgs.config ||
+      options.rule !== defaultArgs.rule ||
+      options.inlineConfig !== defaultArgs.inlineConfig ||
+      options.configPath !== defaultArgs.configPath
```
rwjblue, line 76: The point of this is to ensure that the values we are comparing against is **always** guaranteed to be the default values.

test/acceptance/todo-cli-test.js
```diff
@@ -25,7 +23,6 @@ describe('todo usage', () => {
   });
 
   afterEach(function () {
-    process.chdir(ROOT);
```
rwjblue, line 28: Note: this isn't needed because `project.dispose()` handles it for us automatically

scalvert:
You've obliterated my PR title....

😄

rwjblue (approved):

scalvert:
cc/ @lifeart 

In case this is useful for UELS and config discovery from a single file.

PR #2471, opened 2022-04-14
Add tests for plugin resolution within a monorepo setup

rwjblue:
Adds tests inspired by those added in #2359, but without the need to `git clone` + `yarn install` + `yarn build` external packages in the test harness.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -3,6 +3,7 @@ import { findUpSync } from 'find-up';
 import micromatch from 'micromatch';
 import { createRequire } from 'node:module';
 import path from 'node:path';
+import { pathToFileURL } from 'node:url';
 import resolve from 'resolve';
 
 import defaultConfigurations from './config/index.js';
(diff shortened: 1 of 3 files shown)


rwjblue:
These tests pass without changing anything about the implementation.

@NullVoxPopuli - Can you copy one of these tests and tweak it in order to fail? I'm still not sure what is actually broken in your use case...

lib/get-config.js
```diff
@@ -39,7 +40,9 @@ async function requirePlugin(workingDir, pluginName, fromConfigPath) {
   let pluginPath = resolve.sync(pluginName, { basedir, extensions: ['.js', '.mjs', '.cjs'] });
 
   try {
-    const { default: plugin } = await import(pluginPath);
+    let pluginURL = pathToFileURL(pluginPath);
```
rwjblue, line 43: Note: this definitely fixes a bug in plugin resolution, but only for windows users (we previously attempted to `import('c:/foo/bar/baz.js')` which is not allowed, you have to do `import('file:///c:/foo/bar/baz.js')`).

test/acceptance/monorepo-test.js
```diff
@@ -0,0 +1,165 @@
...
+import run from '../helpers/run.js';
+import setupEnvVar from '../helpers/setup-env-var.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+describe('monorepo setups', function () {
+  setupEnvVar('FORCE_COLOR', '0');
+  setupEnvVar('LC_ALL', 'en_US');
+
+  // Fake project
```
scalvert, line 16:
```suggestion
```

scalvert (approved):

lib/get-config.js
bmish: This should be labeled as a `bug fix` then right?

PR #2474, opened 2022-04-15
Ensure `--compact-todo` functions properly 

scalvert:
The compacting code in `@lint-todo/utils` had a bug, which was fixed (https://github.com/lint-todo/utils/pull/376). This PR bumps the minimum version to this new release.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -131,9 +131,7 @@ async function run() {
   }
 
   if (options.compactTodo) {
-    let { compacted } = compactTodoStorageFile(options.workingDirectory, {
-      engine: 'ember-template-lint',
-    });
+    let { compacted } = compactTodoStorageFile(options.workingDirectory);
     _console.log(`Removed ${compacted} todos in .lint-todo storage file`);
     process.exitCode = 0;
     return;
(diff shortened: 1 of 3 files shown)


bin/ember-template-lint.js
```diff
@@ -131,7 +131,9 @@ async function run() {
   }
 
   if (options.compactTodo) {
-    let { compacted } = compactTodoStorageFile(options.workingDirectory);
+    let { compacted } = compactTodoStorageFile(options.workingDirectory, {
+      engine: 'ember-template-lint',
+    });
```
rwjblue, line 136: We were just chatting about this, I think we should consider making `compactTodoStorageFile` work across all engines. As it stands, I worry that we might drop unrelated todos from other engines during compaction...

Might need another test for this over in the todo utils...

rwjblue (changes requested):

bin/ember-template-lint.js
scalvert: Done and done.

elwayman02 (approved):

bin/ember-template-lint.js
rwjblue: Hmm, I would have thought that you don't need the second arg anymore?

rwjblue (approved):

PR #2475, opened 2022-04-15
Add test and implementation for plugin using `exports` in `package.json`.

rwjblue:
This currently fails, because of how we use `resolve.sync` to discover the full resolvable path in order to `import()` (in `lib/get-config.js`), and `resolve` doesn't itself support ESM resolution semantics yet (see https://github.com/browserify/resolve/issues/222).

Reproduces the failure @nullvoxpopuli ran into in https://github.com/ember-template-lint/ember-template-lint/pull/2359


NullVoxPopuli (approved):
I think ultimately we should be using `require.resolve` (or something similar that handles resolving `package.json#exports` -- it's a complex enough thing where I think an additional dependency would be worth it -- should one exist that allows us resolving from particular working directories -- I have a hunch one does exist, because embroider _does a lot_ of this task.

This is improvement for now -- and we can improve further later if needed.

PR #2478, opened 2022-04-19
Add `validateValues` option to `require-lang-attribute` rule

### LINKED ISSUE #2477: Proposal to update `require-lang-attribute` rule with value validation
Currently, the `require-lang-attribute` rule checks for the presence of the `lang` attribute but not whether its value is valid. This issue proposes updating the rule so that, if the `lang` attribute contains a text value, check that its value is a known [IETF's BCP 47 language tag](https://tools.ietf.org/search/bcp47). When the `lang` attribute is properly assigned, both assistive technologies and conventional user agents will be able to render text more accurately. 

This update is essentially a port of JSX's  [`lang` rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/lang.md).

Reference: [lang - HTML: HyperText Markup Language - MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

judithhinlung:
This PR fixes #2477.

### Background
Currently, the `require-lang-attribute` rule checks for the presence of the `lang` attribute but not whether its value is valid. This issue proposes updating the rule so that, if the `lang` attribute contains a text value, check that its value is a known [IETF's BCP 47 language tag](https://tools.ietf.org/search/bcp47). When this attribute is properly assigned, both assistive technologies and conventional user agents will be able to render text more accurately. 

Reference: [lang - HTML: HyperText Markup Language - MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

**Allowed**:
```
# examples
<html lang="en"></html>
<html lang="en-US"></html>
<html lang={{lang}}></html>
```

**Forbidden**:
```
# examples
<html></html>
<html lang=""></html>
<html lang="abracadabra"></Html>
```

### Testing
Test Suites: 136 passed, 136 total
Tests:       5 skipped, 7478 passed, 7483 total
Snapshots:   981 passed, 981 total
Time:        95.504 s
Ran all test suites.
✨  Done in 101.16s.

### DIFF
--- a/docs/rule/require-lang-attribute.md
+++ b/docs/rule/require-lang-attribute.md
@@ -14,11 +14,11 @@ When the language of the page cannot be identified, the integrity of the above i
 
 Consider any of the following use cases:
 
-* the application developer is unaware that Ember now includes the lang attribute
-* the application does not require internationalization
-* the application's content is in a language that is not English
-* an end-user with a screen reader turned on, whose operating system (OS) is set to a different language, navigates to that page with their screen reader turned on
-* the screen reader would attempt to read the page in the language that is defined by the lang attribute on the page, but the supporting element information ("button", "link", etc) is read out in the language that is set by the operating system.
+- the application developer is unaware that Ember now includes the lang attribute
+- the application does not require internationalization
+- the application's content is in a language that is not English
+- an end-user with a screen reader turned on, whose operating system (OS) is set to a different language, navigates to that page with their screen reader turned on
+- the screen reader would attempt to read the page in the language that is defined by the lang attribute on the page, but the supporting element information ("button", "link", etc) is read out in the language that is set by the operating system.
 
 ## Examples
 
@@ -29,7 +29,7 @@ This rule **forbids** the following:
 ```
 
 ```hbs
-<html lang=""></html>
+<html lang=''></html>
 ```
 
 ```hbs
@@ -39,11 +39,11 @@ This rule **forbids** the following:
 This rule **allows** the following:
 
 ```hbs
-<html lang="en"></html>
+<html lang='en'></html>
 ```
 
 ```hbs
-<html lang="en-US"></html>
+<html lang='en-US'></html>
 ```
 
 ```hbs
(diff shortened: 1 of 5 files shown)


bmish:
It looks like this is a breaking change right now? If so, we would want to hide this behavior behind an option `requireValidValue` (default `false`), and then wait until the next major release (#2319) to enable it by default.

test/unit/rules/require-lang-attribute-test.js
```diff
@@ -3,7 +3,62 @@ import generateRuleTests from '../../helpers/rule-test-harness.js';
...
+              "source": "<html lang=\\"\\"></html>",
+            },
+          ]
+        `);
+      },
+    },
+  ],
+});
+
+generateRuleTests({
```
bmish, line 56: Hmm, I think we probably want to avoid having an additional `generateRuleTests` call. Instead, we should pass `config` to specific test cases.

docs/rule/require-lang-attribute.md
```diff
@@ -54,6 +54,15 @@ This rule **allows** the following:
...
 Add the `lang` attribute to the `app/index.html` file in your Ember app. If you use an internationalization addon like `ember-intl`, note that this will not conflict with that addon.
 
+## Configuration
+
+- boolean -- if `true`, default configuration is applied
+  (`validateValues: false`), see below for details
+
+- object -- containing the following property:
+  - boolean -- `validateValues` -- if `true`, the rule checks whether the value in the `lang` attribute is a known IETF's BCP 47 language tag
+    (default: `false`)
```
bmish, line 64: We could indicate here that we should enable this by default in the next major release as a reminder:

```suggestion
    (default: `false`) (TODO: enable by default in next major release)
```

bmish (changes requested):

test/unit/rules/require-lang-attribute-test.js
```diff
@@ -4,8 +4,35 @@ generateRuleTests({
   name: 'require-lang-attribute',
 
   config: true,
-
-  good: ['<html lang="en"></html>', '<html lang="en-US"></html>', '<html lang={{lang}}></html>'],
+  good: [
+    '<html lang="en"></html>',
+    '<html lang="en-US"></html>',
+    '<html lang={{lang}}></html>',
```
bmish, line 10: We need to make sure we have a test for the default behavior of allowing invalid values when an option isn't specified.
```suggestion
    '<html lang={{lang}}></html>',
    '<html lang="hurrah"></html>'
```

bmish (approved):
Thanks!

PR #2324, opened 2022-01-07
Implement support embedded templates

NullVoxPopuli:
Continuation of: https://github.com/ember-template-lint/ember-template-lint/pull/2306

Goals:
 - support linting templates embedded in js/ts
   - including tests
  - support `<template>`
  - ensure parsing JS doesn't break linting
  
Non-Goals:
  - introducing this behavior as a default -- should be opt-in for now, as it'd be a breaking change (people's tests can be in all sorts of violations)
  
Stretch Goals:
 - linting arbitrary files with embedded templates, such as markdown / glimdown  
  
For `<template>`,
it may hinge on the acceptance of https://github.com/emberjs/rfcs/pull/779  

Todo:

- [ ] more tests
- [x] ensure surrounding JS doesn't get linted, this project only lints templates
- [ ] ensure lines / columns match for the overall file
- [x] fix ts parsing
- [ ] track imported / declared variables so that `no-implicit-this` doesn't incorrectly fire

Related work:
 - https://github.com/ember-cli/eslint-plugin-ember/pull/1395/files

PR #2483, opened 2022-04-22
Implement support embedded templates

ventuno:
Continue the work started in https://github.com/ember-template-lint/ember-template-lint/pull/2324


lib/linter.js
```diff
@@ -13,6 +14,12 @@ import {
 } from './helpers/severity.js';
 
 const MAX_AUTOFIX_PASSES = 10;
+// We need to disable some rules when parsing embedded templates
+// This is because some of the errors are expected to be caught by ESLint
+const MUTED_RULES_FOR_EMBEDDED_TEMPLATES = new Set([
+  // ESLint will determine whether or not a template is using a local variable which will not require the keyword this
```
NullVoxPopuli, line 20: thanks for adding the comment!

test/acceptance/editors-integration-test.js
```diff
@@ -66,4 +66,362 @@ describe('editors integration', function () {
...
+
+        let template = fs.readFileSync(path.resolve('some-module.gjs'), { encoding: 'utf8' });
+        expect(template).toBe(
+          'export const SomeComponent = <template>\n' +
+            '  <button type="button"></button>\n' +
+            '</template>'
+        );
+      });
+
+      it('no-implicit-this is muted for embedded templates', async function () {
```
NullVoxPopuli, line 412: :raised_hands:

test/acceptance/editors-integration-test.js
```diff
@@ -66,4 +66,362 @@ describe('editors integration', function () {
       expect(template).toBe('<button type="button"></button>');
     });
   });
+
+  describe('with embedded templates', function () {
+    describe('hbs (tests and manual low-level usage)', function () {
```
NullVoxPopuli, line 71: for other reviewers, even though this isn't related to `<template>`, this way of creating components (hbs + setComponentTemplate) works as early as 3.13 (iirc?), and has not been able to be linted _until now_. Additionally, folks could choose to have a different set of lints for their tests' templates, if they want.

lib/extract-templates.js
```diff
@@ -0,0 +1,133 @@
+import { parseTemplates } from 'ember-template-imports/lib/parse-templates.js';
+
+export const SUPPORTED_EXTENSIONS = ['js', 'ts', 'gjs', 'gts'];
```
NullVoxPopuli, line 3: I wonder if there is a way to have these be opt-in? :thinking: 

I've been curious if these supported extensions will have an impact on an existing codebase.

Any chance, you can look in to that, @ventuno?

Ideally, with nothing but default configs, js/ts (tests, for example) remain unlinted. -- we can flip the default in a major tho

NullVoxPopuli (approved):
Looking good! just a couple questions

lib/extract-templates.js
ventuno: Thanks @NullVoxPopuli, the only way I can think of doing this, is by using a command line argument which would enable parsing for script files. Thoughts?

chriskrycho: The way ESLint etc. approach this is allowing you to specify what to include in your config, so that's probably the forward-and-backward-compatible way to do it here: instead of hard-coding all the extensions, introduce a config option which specifies extensions to tackle, supplies a default of just `.hbs`, and then allows folks to opt in new extensions as well, with an explanation of how those will be handled. (Perhaps we need two things: template-only file extensions and embedded template extensions?)

ventuno: `ember-template-lint` has the [`ignore-pattern`](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/helpers/cli.js#L104) argument which could be used to this effect.
What about adding to the list of default values all patterns we want to exclude?
Something like:
```
'ignore-pattern': {
  describe: 'Specify custom ignore pattern (can be disabled with --no-ignore-pattern)',
  type: 'array',
  default: [..., '**/*.js', '**/*.gjs', ...],
},
```
provided that works as expected.

ventuno: @NullVoxPopuli, @chriskrycho so I just updated the code to ignore script files by default and added tests. Let me know what you think, but this should be the most straightforward way to do this.

lib/extract-templates.js
```diff
@@ -0,0 +1,133 @@
...
+    return [makeTemplateInfo(LOCATION_START, moduleSource)];
+  }
+
+  let parsed = parseTemplates(moduleSource, relativePath, 'template');
+
+  // If after parsing we found no templates and we had no relative path,
+  // then we assume we had a hbs file in input.
+  if (parsed.length === 0 && !relativePath) {
+    return [makeTemplateInfo(LOCATION_START, moduleSource)];
+  }
```
rwjblue, line 33: It seems vaguely odd to have this be a fallback, I think we expect (at least at the moment) for 99% of the things we run against are going to be stand alone `.hbs` files. Can we move this first, and do an extension check _before_ `let mightHaveTemplates`?

lib/extract-templates.js
```diff
@@ -0,0 +1,133 @@
...
+   * 6 }
+   *
+   * The extracted template will be:
+   * 1 \n
+   * 2    {{debugger}}\n
+   *
+   * The coordinates of the template in the source file are: { line: 3, column: 14 }.
+   * The coordinates of the error in the template are: { line: 2, column: 4 }.
+   *
+   * Thus, we need to always subtract one before adding the template location.
```
rwjblue, line 100: Is this actually true? I think it's only because you are thinking of the newline as not part of the template.

I would expect this to actually not need "line" adjustment, right? (read this as if it were the only contents in a `.gjs` file, which is a valid contents AFAICT)

```js
<template>{{debugger}}</template>
```

lib/extract-templates.js
```diff
@@ -0,0 +1,133 @@
...
+ * @param {object} result
+ * @param {number} result.line
+ * @param {number} result.endLine
+ * @param {number} result.column
+ * @param {number} result.endColumn
+ *
+ * @returns {object}
+ */
+export function coordinatesOfResult(templateInfo, result) {
+  /**
```
rwjblue, line 84: Can you update the docs here to have a high level explanation of what the point of the function is? I assume it's something like

```suggestion
/**
 * Processes results and corrects for template location offsets.
 *
 * @param {TemplateInfo} templateInfo
 * @param {object} result
 * @param {number} result.line
 * @param {number} result.endLine
 * @param {number} result.column
 * @param {number} result.endColumn
 *
 * @returns {object}
 */
export function coordinatesOfResult(templateInfo, result) {
  /**
```

lib/extract-templates.js
```diff
@@ -0,0 +1,133 @@
+import { parseTemplates } from 'ember-template-imports/lib/parse-templates.js';
+
+export const SUPPORTED_EXTENSIONS = ['.js', '.ts', '.gjs', '.gts'];
```
rwjblue, line 3: Does `ember-template-imports/lib/parse-template.js` actually parse `hbs` and `precompileTemplate` from `.js`/`.ts` files? If not, I don't think we should add `.js`/`.ts` here just yet.

lib/linter.js
```diff
@@ -309,34 +316,54 @@ export default class Linter {
...
-      rawSource: source,
-    });
-    results.push(...failures);
-
-    for (let rule of rules) {
       try {
-        let visitor = await rule.getVisitor();
-        transform(templateAST, () => visitor);
-        results.push(...rule.results);
+        templateAST = parse(source);
```
rwjblue, line 328: Personally, I'd prefer to correct the location information at this level instead of modifying result info.

Specifically, I'd like to:

1. in the short term, traverse this AST and update all the `.loc` properties to the actually correct values if we are starting "from an offset" (e.g. with `.gjs`/`.gts`)
2. update `@glimmer/syntax` (and `ember-template-recast`) to accept start info for the source string (so that `.loc` info is generated correctly to begin with)

lib/linter.js
```diff
@@ -309,34 +316,54 @@ export default class Linter {
...
+        fileConfig,
+        filePath: options.filePath,
+        configResolver: options.configResolver,
+        rawSource: source,
+      });
+      results.push(...failures);
+
+      for (let rule of rules) {
+        try {
+          if (templateInfo.isEmbedded && MUTED_RULES_FOR_EMBEDDED_TEMPLATES.has(rule.ruleName)) {
```
rwjblue, line 345: I'd like this to be removed, in favor of rules doing logic based on strict/loose modes

rwjblue (changes requested):

lib/extract-templates.js
ventuno: My understanding is that we want to output the line in the source file, not just the template section.
In the example below, we expect the location of the error to be `{ line: 4, column: 4 }`.
```
   * 1 import { hbs } from 'ember-cli-htmlbars'\n
   * 2 export class SomeComponent extends Component<Args> {\n
   * 3     <template>\n
   * 4         {{debugger}}\n
   * 5     </template>\n
   * 6 }
```
Now the location of the template content is `{ line: 3, column: 14 }` which is where the location after the last character in `<template>` (which is the `\n`).
This means the template is:
```
   * 1 \n
   * 2    {{debugger}}\n
```
The linter will report an error on location `{ line: 2, column: 4 }` of the template. If we simply compute the location in the overall source file without any adjustment the error location ends up being `{ line: 5, column: 4 }` which is incorrect. Thoughts?

lib/extract-templates.js
ventuno: If `relativePath` is `undefined` it means we're reading a file from `stdin` which means we don't have any information about the file format. I thought the best approach in this case would be to try and parse the file as if it were a script and then if an empty array is returned, we simply return the entire file.

If a `relativePath` is present, we parse the extensions and return on line 24 if the extensions of that file is not in `SUPPORTED_EXTENSIONS`.

lib/extract-templates.js
ventuno: @NullVoxPopuli and @chriskrycho are probably going to know better than me, but as far as I understand [`parseTemplates`](https://github.com/ember-template-imports/ember-template-imports/blob/master/src/parse-templates.ts) is currently able to parse content inside `<template></template>` and ``hbs` ` ``.

lib/linter.js
```diff
@@ -13,6 +14,12 @@ import {
 } from './helpers/severity.js';
 
 const MAX_AUTOFIX_PASSES = 10;
+// We need to disable some rules when parsing embedded templates
+// This is because some of the errors are expected to be caught by ESLint
+const MUTED_RULES_FOR_EMBEDDED_TEMPLATES = new Set([
+  // ESLint will determine whether or not a template is using a local variable which will not require the keyword this
+  'no-implicit-this',
+]);
```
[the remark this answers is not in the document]
ventuno: That sounds good to me. Assuming we're going to use the file path to determine this, we should still consider what should be the behavior when parsing input from stdin in which may not have a file path to use.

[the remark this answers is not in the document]
ventuno: So I updated this which required me to change the base rule implementation to make sure that `this.filePath` is defined _before_ `parseConfig` is called.
I removed `MUTED_RULES_FOR_EMBEDDED_TEMPLATES` and let `no-implicit-this` return `false` if the `filePath` has a "script file" extension. I also assumed the error should be _active_ if `filePath` is `undefined` (e.g.: the content is passed into the stdin without a `--filename` argument).

lib/linter.js
```diff
@@ -309,34 +316,54 @@ export default class Linter {
     }
 
     let hasBOM = options.source.codePointAt(0) === 0xfe_ff;
-    let source = hasBOM ? options.source.slice(1) : options.source;
+    let rawSource = hasBOM ? options.source.slice(1) : options.source;
```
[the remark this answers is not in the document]
ventuno: Renamed back to `source` and used `templateInfo.template` elsewhere.

lib/linter.js
ventuno: I'm fine with this approach, but is it really worth doing another AST traversal just to update the `loc`ation as a temporary solution? I'm not talking in terms of performance, just adding that complexity to the code if it's eventually going to be moved inside `parse`.

lib/extract-templates.js
NullVoxPopuli: Yup. Is Handy for folks manually using setComponentTemplate (pre `<template>`)

rwjblue: Hmm. Doesn't actually seem to have any direct tests for `parseTemplates` which makes me very uncomfortable with the assertion that it works for all cases of content you might have as templates in `.js`/`.ts` files.

lib/linter.js
rwjblue: IMHO, that's totally up to you. If you'd prefer to just do the second option, I'm 100% fine with that.

lib/extract-templates.js
rwjblue: Gotcha, makes sense now (with the stdin context).

lib/linter.js
ventuno: Just for me to clarify, are you fine with this being a TODO (I'm more than happy to tackle this next) or is this a blocker and I should update `parse` first?

lib/extract-templates.js
ventuno: I discovered this because ["returns a "fatal" result object if an error occurs during parsing"](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/acceptance/public-api-test.js#L497) was failing. I will add one more test here that explicitly calls out a missing `filePath`.

lib/extract-templates.js
rwjblue: I created a super simple test harness over in https://github.com/ember-template-imports/ember-template-imports/pull/36 so we can continue to flesh things out, but as you can see from those tests if we use `parseTemplates` directly like this we have a **TON** of problems. It will identify _**every**_ tagged template expression found in the file!!

lib/extract-templates.js
```diff
@@ -0,0 +1,135 @@
+import { parseTemplates } from 'ember-template-imports/lib/parse-templates.js';
```
rwjblue, line 1: I think we actually need to do some utility extraction over in ember-template-imports. `parseTemplates` is trying to find all possible matches, but it doesn't care at all about the semantic meaning of them. 

We probably need to be using some extraction of `process-embedded-templates`, which leverages `parse-static-imports` so that we can make sure that we only process *specific* imported tagged template expressions.

NullVoxPopuli: Sounds like a good thing to do in a followup, but I don't thing it should block this PR.

(getting a solid test suite is most important, so that such a cleanup can be done safely!)

rwjblue: No, if we land this with `parseTemplates` we would be attempting to lint _all_ tagged template expression contents as if they were handlebars templates. This is fundamentally wrong, and a very large blocker.

rwjblue: Code like:

```js

export default class Foo extends Component {
  indentedThing = stripIndent`
    Hahaha, this is just a plain string. Definitely not a template.
  `
}
```

That `stripIndent` will be linted by ember-template-lint as if it were a template.

lib/extract-templates.js
rwjblue: Gotcha, thank you for the detailed walkthrough (the examples comparing the overall contents of the file wiht line numbers really were helpful).

rwjblue: Can you confirm that we have tests for `<template>foo</template>` (errors that happen on the same line as the `<template>` tag)?

lib/rules/_base.js
```diff
@@ -50,6 +50,9 @@ export default class {
     this.results = [];
     this.mode = options.shouldFix ? 'fix' : 'report';
 
+    this.workingDir = options.workingDir;
+    // TODO: add a deprecation for accessing _filePath
+    this.filePath = this._filePath = options.filePath;
```
rwjblue, line 55: why move these?

lib/extract-templates.js
NullVoxPopuli: ah, I see. so parseTemplates is _just way wrong_ atm.

test/acceptance/editors-integration-test.js
```diff
@@ -66,4 +66,362 @@ describe('editors integration', function () {
       expect(template).toBe('<button type="button"></button>');
     });
   });
+
+  describe('with embedded templates', function () {
```
rwjblue, line 70: Can you add a test for the example I posted in another comment:

```js
export default class Foo extends Service {
  indentedThing = stripIndent`
    Hahaha, this is just a plain string. Definitely not a template.
  `
}
```

The test should assert that the contents of `stripIndent` are **not** linted (I expect this test to fail with the current implementation).

rwjblue (changes requested):

lib/extract-templates.js
rwjblue: Ya, it's very wrong for what _we_ want to do, but inside `ember-template-imports` it is used by `process-embedded-templates` properly as far as I can tell.

lib/rules/_base.js
rwjblue: I see, sorry missed the other comment https://github.com/ember-template-lint/ember-template-lint/pull/2483#discussion_r859214553

lib/extract-templates.js
ventuno: Just added a number of tests to [`extract-template-test.js`](https://github.com/ember-template-lint/ember-template-lint/pull/2483/files#diff-0435102285f4156c51a7c0b981f321d413251423fa542fd1af521078a8cae2b2). To cover all cases.

lib/extract-templates.js
ventuno: I had "for a template which does not start with a new line, the error location is correct" already, I also added "for a template defined on one line, the error location is correct" as well.

test/acceptance/editors-integration-test.js
ventuno: Added 'does not parse regular strings as a template' to prove this is actually failing as _not_ expected.

NullVoxPopuli:
where is this PR at? it's been a while since I looked into this :sweat_smile:

ventuno:
Thanks @NullVoxPopuli for checking in. This should be ready for a final round. Latest status update: https://discord.com/channels/480462759797063690/967432082630541354/983742401036320849.

nlfurniss:
@chriskrycho @rwjblue do you mind taking a look?

lib/extract-templates.js
```diff
@@ -0,0 +1,164 @@
...
+      {
+        importPath: 'htmlbars-inline-precompile',
+        importIdentifier: 'default',
+      },
+      {
+        importPath: '@ember/template-compilation',
+        importIdentifier: 'precompileTemplate',
+      },
+    ],
+  });
```
rwjblue, line 54: In order to go this route, I think we need to deal with https://github.com/ember-template-imports/ember-template-imports/issues/42. We need to know (from the return value of `parseTemplates`) _which_ of these imports were used (so that we know things like "is this strict mode or loose mode?").

We might be able to start off without that info, but I think it will eventually be required. The experience (in editor) when using `<template>` should tell you when you are accidentally relying on `<Input />` or `{{hash}}`/`{{get}}` without an import ([see the RFC](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#keywords) for details on keywords vs helpers).

lib/extract-templates.js
rwjblue: I think this is still outstanding (need a description added to the docs)

lib/rules/no-implicit-this.js
```diff
@@ -52,7 +53,7 @@ const ARGLESS_DEFAULT_BLUEPRINT = [
 
 export default class NoImplicitThis extends Rule {
   parseConfig(config) {
-    if (config === false || config === undefined) {
+    if (config === false || config === undefined || isSupportedScriptFileExtension(this.filePath)) {
```
rwjblue, line 56: Hmm, I don't think this is quite right. We don't care if it is in a `.js`/`.gjs` file we care if the usage was via a known strict mode system like `<template>`.  There is nothing that says you can't use `import { hbs }from 'ember-cli-htmlbars'` inside a `.gjs` file.

This is getting back to the same thing I was referring to above (and requires https://github.com/ember-template-imports/ember-template-imports/issues/42 to be completed in order to resolve).

I think we'll need to complete https://github.com/ember-template-imports/ember-template-imports/issues/42 and then add an additional property on the `Rule` base class. Something like `this.isStrictMode` or somesuch...

rwjblue (changes requested):
OK, did another pass of review and left a few inline comments. 

Some outstanding questions:

- Can you add tests that ensure that `.js`/`.gjs`/&c are not linted by default (e.g. if I run `ember-template-lint .`)? We absolutely should **make** this the default in time, but IMO we shouldn't do that all at one time, let's make it opt-in first (by folks explicitly globbing, or using `--no-ignore`, etc)
- We need to thread through `this.isStrictMode` or something to the `Rule` base class, and add associated tests.
- We need to beef up the testing harness to allow for individual rules testing inline templates (`hbs`, `<template>`, etc) with the normal test harness.

lib/linter.js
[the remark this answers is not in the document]
rwjblue: Resolving this for now, but we need to complete https://github.com/ember-template-imports/ember-template-imports/issues/42 and thread through `this.isStrictMode` kinda thing into `Rule`

lib/extract-templates.js
ventuno: Sorry about that. I must have missed this somehow.

lib/extract-templates.js
ventuno: What I have implemented so far is kind of a "best effort" approach, because we can't always know the file type based on the information passed to `ember-template-lint` (see the [discord conversation about this](https://discord.com/channels/480462759797063690/518154533143183377/967169732849598474)).

I'm down to either wait for https://github.com/ember-template-imports/ember-template-imports/issues/42 to be done (looks like @nlfurniss already made some progress) or help complete it, but what I've done here is the best we can do with the information we have so far.

Let's resolve the conversation about this here.

lib/rules/no-implicit-this.js
ventuno: Let's consolidate the conversation about https://github.com/ember-template-imports/ember-template-imports/issues/42 in [this thread](https://github.com/ember-template-lint/ember-template-lint/pull/2483#discussion_r938027129).

ventuno:
> OK, did another pass of review and left a few inline comments.
> 
> Some outstanding questions:
> 
>     * Can you add tests that ensure that `.js`/`.gjs`/&c are not linted by default (e.g. if I run `ember-template-lint .`)? We absolutely should **make** this the default in time, but IMO we shouldn't do that all at one time, let's make it opt-in first (by folks explicitly globbing, or using `--no-ignore`, etc)

I already edited the existing tests in [cli-test.js](https://github.com/ember-template-lint/ember-template-lint/pull/2483/files#diff-08030efb2275758201c5160d222db24848291b8e03c0a415f23e82ccc9ad357f) to this effect.
`should respect dirs ignored by default` tests that `bar.gjs` is ignored, and `should allow to disable dirs ignored by default` triggers an error for `other.gjs`. Let me know if something is missing.

> 
>     * We need to thread through `this.isStrictMode` or something to the `Rule` base class, and add associated tests.

Let's consolidate the conversation about https://github.com/ember-template-imports/ember-template-imports/issues/42 in [this thread](https://github.com/ember-template-lint/ember-template-lint/pull/2483#discussion_r938027129).

> 
>     * We need to beef up the testing harness to allow for individual rules testing inline templates (`hbs`, `<template>`, etc) with the normal test harness.

Can you clarify this a bit, I'm not sure I understand what you're looking for here.

lib/extract-templates.js
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @returns {TemplateInfo}
+ */
+function makeTemplateInfo({ line, column }, template, templateInfo, isEmbedded) {
+  return {
+    line,
+    column,
+    template,
+    isEmbedded,
+    isStrictMode: !isEmbedded || isStrictMode(templateInfo),
```
rwjblue, line 97: `isStrictMode` is basically determined by which import path you used (or if it came from `<template>`), from the list of imports that are in the PR at the moment the following are strict:

* `import { hbs } from 'ember-template-imports';`
* `<template></template>`

ventuno: @rwjblue I updated `isStrictMode` based on your comments. Right now I'm checking that `importIdentifier === 'hbs' && importPath === 'ember-template-imports'`.
This becomes an issue when the identifier is aliased as in

```import { hbs as theHbs } from 'ember-template-imports'; ```

because `parseTemplates` will return `{ importIdentifier: 'theHbs', importPath: 'ember-template-imports' }` which of course breaks the above check. Maybe we should return the un-aliased version from `parseTemplates` or remove the `importIdentifier === 'hbs'` check from `isStrictMode`.
I added tests for these cases but we need to resolve this conversation before they can pass.

rwjblue (approved):

PR #2487, opened 2022-04-26
Add `checkAllHTMLElements` option to `no-redundant-landmark-role` rule to lint against all HTML elements with default ARIA roles

### LINKED ISSUE #2485: Proposal to expand `no-redundant-landmark-role` to lint against redundancy for all elements with ARIA role support
This proposes expanding the coverage of `no-redundant-landmark-role` to lint against all HTML elements with default or implicit ARIA roles.  Since these semantic elements already have browser support for their respective ARIA roles, adding another role attribute will add  and should be avoided.
This change will make `no-redundant-landmark-role` analogous to [JSX's `no-redundant-roles](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-redundant-roles.md) rule. 

Reference: [ARIA Spec, ARIA Adds Nothing to Default Semantics of Most HTML Elements](https://www.w3.org/TR/using-aria/#aria-does-nothing)

judithhinlung:
This PR fixes #2485.

### Background
Most HTML elements have default/implicit ARIA roles, and setting a role that matches an element's default has no effect. 

This PR adds an option, `checkAllElementRoles`, in the config for the existing `no-redundant-landmark-role` rule to expand its coverage to lint against all HTML elements with default roles.
This PR also adds the `nav` element and the `navigation` role to the list of allowed element-role pairings, as [advised by the W3 for accessibility]((/https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element#Example:The_.3Cnav.3E_element)), and the `footer` element and `contentinfo` role are removed due to [redundancy in their functions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/contentinfo_role#description).

References:
- - [ARIA Spec, ARIA Adds Nothing to Default Semantics of Most HTML Elements](https://www.w3.org/TR/using-aria/#aria-does-nothing)
- [JSX `no-redundant-roles` rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-redundant-roles.md)

**Allowed**:

```
# examples
<nav role="navigation"></nav>
<form role="search"></form>
<input type="text" />
```

**Forbidden**:

```
# examples
<footer role="contentinfo"></footer>
<input type="checkbox" name="agree" value="checkbox1" role="checkbox" />
<select name="color" id="color" role="listbox" multiple><option value="default-color">black</option></select>
```

### Testing
Tests:       5 skipped, 7545 passed, 7550 total
Snapshots:   990 passed, 990 total
Time:        139.005 s, estimated 159 s
Ran all test suites.
✨  Done in 152.69s.

### DIFF
--- a/docs/rule/no-redundant-landmark-role.md
+++ b/docs/rule/no-redundant-landmark-role.md
@@ -6,7 +6,10 @@
 
 If a landmark element is used, any role provided will either be redundant or incorrect.
 This rule adds support for landmark elements, to ensure that no role attribute is placed on any of
-the landmark elements, with a few exceptions.
+the landmark elements, with the following exceptions:
+
+- a `nav` element with the `navigation` role to [make the structure of the page more accessible to user agents](/https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element#Example:The_.3Cnav.3E_element)s
+- a `form` element with the `search` role to [identify the form's search functionality](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role#examples)
 
 ## Examples
 
@@ -25,7 +28,7 @@ This rule **forbids** the following:
 ```
 
 ```hbs
-<nav role="navigation"></nav>
+<footer role="contentinfo"></footer>
 ```
 
 ```hbs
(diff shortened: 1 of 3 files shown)


docs/rule/no-redundant-landmark-role.md
```diff
@@ -39,11 +42,20 @@ This rule **allows** the following:
...
-<footer role="contentinfo"></footer>
+<nav role="navigation"></nav>
 ```
 
+## Configuration
+
+- boolean -- if `true`, default configuration is applied
+
+- object -- containing the following property:
+  - boolean -- `checkAllElementRoles` -- if `true`, the rule checks for redundancy between any semantic HTML element with a default/implicit ARIA role and the role provided
```
bmish, line 53: Perhaps "checking roles" is already implied by the name of the rule? What do you think about `checkAllHTMLElements` as the option name?

test/unit/rules/no-redundant-landmark-role-test.js
```diff
@@ -7,9 +7,46 @@ generateRuleTests({
 
   good: [
```
bmish, line 8: Can you add a commen to indicate which of these test cases is checking the behavior when `checkAllElementRoles` is not specified?

bmish (commented):
Does the rule need to be renamed eventually?

judithhinlung:
Yes I think renaming this rule eventually will make its purpose more clear.

bmish:
@judithhinlung can you suggest a future name for it? That way, we can add it to the list of breaking changes in #2319.

bmish (approved):
Thanks!

judithhinlung:
I think renaming this to `no-redundant-role` would capture the essence of the rule.

Issue #2488, opened 2022-04-27
Github Lint/Build Annotations Are Not Associated in Subdirectory Ember Projects


rtablada:
When an ember app is located in a repository sub-directory (as is common in monorepo setups), Github actions is not able to associate lint error output with the appropriate files in the Ember project.

It seems that Github Actions associates build errors based on either the full path or the path relative to the git root.

Currently (and going back to at least 2.9), Ember Template Lint outputs the file path relative to the ember project with no documented way of configuring this output. There may be undocumented or monkey patch ways to change this behavior, but these should be better surfaced and likely set as default (at least in CI) so that annotations work for both root projects and those in monorepo setups.

I've created an example monorepo project along with PRs to show the missing annotations.

* Missing Annotations in basic 3.20 app (template lint 2.9.1)
  * [Files (with no annotations)](https://github.com/rtablada/ember-template-lint-monorepo/pull/1/files)
  * [Lint Action Output](https://github.com/rtablada/ember-template-lint-monorepo/pull/1/checks)
* Missing Annotations in 4.3 app (template lint 4.3.0)
  * [Files (with no annotations)](https://github.com/rtablada/ember-template-lint-monorepo/pull/2/files)
  * [Lint Action Output](https://github.com/rtablada/ember-template-lint-monorepo/pull/2/checks)
* Working Annotations in 3.20 app (template lint with hotfix linked below)
  * [Files (with no annotations)](https://github.com/rtablada/ember-template-lint-monorepo/pull/3/files)
  * [Lint Action Output](https://github.com/rtablada/ember-template-lint-monorepo/pull/3/checks)

I created a small hotfix branch for this issue based on the old 2.9 branch, but I'm not sure that this is the best actual approach and will always print the full path (even in local dev where this is likely extra unwanted noise) https://github.com/rtablada/ember-template-lint/commit/d1c5b2d75cdc4f00b29f2d2696fe4f96cc5afc7f

PR #2491, opened 2022-04-29
Add Option to Print Full File Paths

rtablada:
This is a possible solution to #2488.

The new added `--print-full-path` option will set the pretty printer (I was unsure if this should also impact serif and JSON printers) to print the full directory path.

A more complicated option would be to have some way of defining a git root which would give a cleaner output though still allowing nested annotations to be discoverable. Since GH and most annotation/lint readers have support for full path, I went with the option to print full path. Doing a relative path or print git root would add more complexity to the CLI and would require a bit more to clean up output and things.

If requested, I can add the full path to also change the JSON and serif formatters along with associated tests.

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -102,6 +102,12 @@ Specify a configuration to override the normally loaded config file:
 ember-template-lint . --config='{"extends": "a11y"}'
 ```
 
+Specify to print full file paths (may be needed for CI build annotations in monorepos)
+
+```bash
+ember-template-lint --print-full-path
+```
+
 Provide a custom formatter from a relative path
 
 ```bash
--- a/lib/formatters/load-formatter.js
+++ b/lib/formatters/load-formatter.js
@@ -12,6 +12,7 @@ import SarifFormatter from './sarif.js';
 
 export function loadFormatter(options) {
   let formatOptions = {
+    printFullPath: options.printFullPath,
     includeTodo: options.includeTodo,
     // env var is used to test output options via tests
     isInteractive: process.stdout.isTTY || process.env['IS_TTY'] === '1',
--- a/lib/formatters/pretty.js
+++ b/lib/formatters/pretty.js
@@ -1,4 +1,5 @@
 import chalk from 'chalk';
+import path from 'node:path';
 
 import { WARNING_SEVERITY, TODO_SEVERITY } from '../helpers/severity.js';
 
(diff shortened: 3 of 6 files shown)


rwjblue:
I think adding an option for this is probably fine, but I wonder if there is something that we can do to make the default experience a bit better. Specifically, I'd like to use full path whenever the repo root is something different from the working directory (which is the `process.cwd()` when `ember-template-lint` is invoked).

Additionally, I was chatting with @scalvert about a very related issue just the other day. I think we ultimately need the internal results to always include full path (though doing this will require some changes to the todo utils IIRC; @scalvert c/d?), and then the pretty reporter can decide if it should print the full path or relative path.

scalvert:
Yep I also think this is a reasonable short term solution, though it's somewhat annoying to have to explicitly use the option to output the paths. Overall I think we need to fix this internally, as @rwjblue mentioned. I personally think we should mirror what eslint does in terms of what it prints out, which is the absolute path.

bmish:
I agree with improving our default behavior to match ESLint's behavior (which is a direction we are trying to move in) which sounds like it would solve this issue more elegantly. We could change this as a breaking change in the upcoming v5 release (#2319) and avoid having to add a short-lived option?

bmish:
@rtablada would you be interested to open a separate PR changing the default behavior? That way we can have that change lined up and ready-to-go the next major release.

rtablada:
@bmish so the desired behavior for v5 would be to always print full paths for all 3 writers (basically the behavior of this PR without the CLI flag)?

bmish:
Yes, always using full paths seems safest, simplest, and preferred given it matches ESLint's behavior. @rwjblue is that okay with you or do you still prefer the more sophisticated behavior you suggested earlier?

In terms of getting the full path, @rwjblue suggested earlier that internal refactoring to always handle the full path should be part of this (instead of just resolving at the end as shown in the current PR).

rtablada:
Sorry got pulled off to other 🔥s and health/personal stuff. I'm back in support/upgrade land for the next 2 weeks so have cycles for OSS again.

I'll make a spike to always print full path and see what that does to the todo system.

bmish:
Just a heads up that we're going to release the v5 version in the next week or two (#2319). It would be a good time to switch to use full paths by default if anyone can implement this.

rtablada:
Converting this to draft. I am on some other projects/priorities so don't have time to bring this over the finish line sadly.

Would really appreciate if someone in Ember OSS could update or implement full path outputs for inline lint on GH Actions and warnings for monorepos.

lifeart:
I think we could try to land it

NullVoxPopuli (approved):

PR #2497, opened 2022-05-02
Add new rule `require-media-caption`

### LINKED ISSUE #2496: Proposal for new a11y check to enforce that media elements have captions
Currently, Ember Template Lint does not enforce that media elements, such as `<audio>` and `<video>`, provide the `track` element with a `kind` attribute value of `captions` (`<track kind="captions" />`). Captions are essential for users who are deaf, as well as those for whom the media content is unavailable, to access the portion of the content conveyed via the audio track. 

This proposes a new rule to enforce that media elements provide a track with  captions.

Reference: - [Understanding Success Criterion 1.2.2: Captions (Prerecorded)](https://www.w3.org/WAI/WCAG21/Understanding/captions-prerecorded.html)

judithhinlung:
This PR adds the new rule require-media-caption which fixes #2496.

### background
Captions are essential for users who are deaf, as well as those for whom the media content is unavailable, to access synchronized media content. By providing a transcript of the spoken content as well as descriptions of other relevant non-speech information, captions allow users access to the information conveyed via the audio track.

Reference: - [Understanding Success Criterion 1.2.2: Captions (Prerecorded)](https://www.w3.org/WAI/WCAG21/Understanding/captions-prerecorded.html)

**Allowed**:
```
# examples
<audio><track kind="captions" /></audio>
<video muted="true"></video>
<video muted></video>
```
**Forbidden**:
```
# examples
<audio></audio>
<video><track /></video>
<video><track kind="descriptions" /></video>
<audio muted="false"></audio>
```

### Testing
Test Suites: 137 passed, 137 total
Tests:       5 skipped, 7534 passed, 7539 total
Snapshots:   985 passed, 985 total
Time:        109.104 s
Ran all test suites.
✨  Done in 115.11s.

### DIFF
--- a/docs/rule/require-media-caption.md
+++ b/docs/rule/require-media-caption.md
@@ -1,0 +1,41 @@
+# require-media-caption
+
+Captions provide a text version of the spoken and non-spoken audio information for media. They are essential for making audio and video content accessible for users who are deaf as well as those for whom the media is unavailable (similar to `alt` text on an image when it is unable to load).
+
+Captions should contain all relevant information needed to help users understand the media content, which may include a transcription of the dialogue and descriptions of meaningful sound effects. They are synchronized with the media to allow users access to the portion of the content conveyed via the audio track. Note that when audio or video components include the `muted` attribute, however, captions are *not* necessary.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<audio></audio>
+```
+
+```hbs
+<video><track /></video>
+```
+
+```hbs
+<video><track kind="descriptions" /></video>
+```
+
+This rule **allows** the following:
+
+```hbs
+<audio><track kind="captions"></audio>
+```
+
+```hbs
+<video muted="true"></video>
+```
+
+```hbs
+<video muted></video>
+```
+
+## References
+
+- [Captions_Subtitles _ Web Accessibility Initiative (WAI) _ W3C](https://www.w3.org/WAI/media/av/captions/)
+- [Understanding Success Criterion 1.2.2: Captions (Prerecorded)](https://www.w3.org/WAI/WCAG21/Understanding/captions-prerecorded.html)
+- [media-has-caption - eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/media-has-caption.md)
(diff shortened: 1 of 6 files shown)


test/unit/rules/require-media-caption-test.js
```diff
@@ -0,0 +1,63 @@
...
+      template: '<video><track kind="subtitles" /></video>',
+      result: {
+        message: 'Media elements such as <audio> and <video> must have a <track> for captions.',
+        line: 1,
+        column: 0,
+        source: '<video><track kind="subtitles" /></video>',
+      },
+    },
+    {
+      template: '<audio muted="false"></audio>',
```
bmish, line 45: Shouldn't we test with `muted=false` too?

test/unit/rules/require-media-caption-test.js
```diff
@@ -0,0 +1,63 @@
...
+    '<audio muted="true"></audio>',
+    '<video muted></video>',
+    '<audio muted={{this.muted}}></audio>',
+    '<video><track kind="captions" /><track kind="descriptions" /></video>',
+  ],
+
+  bad: [
+    {
+      template: '<video></video>',
+      result: {
```
bmish, line 19: Let's use `verifyResults` snapshot testing for all these test cases.

bmish (changes requested):
Do you think this rule is suitable to become a `recommended` rule in the next major release?

judithhinlung:
Thanks for the review @bmish , I have updated the tests to use verifyResults and added a test case validating muted=false fails as expected.

judithhinlung:
> Do you think this rule is suitable to become a `recommended` rule in the next major release?

Yes I agree that this would be a good `recommended` rule.

docs/rule/require-media-caption.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+```hbs
+<video muted="true"></video>
+```
+
+```hbs
+<video muted></video>
+```
+
```
bmish, line 36: I think we can remove this example. We don't really need to show different syntaxes for the same thing.
```suggestion
```

docs/rule/require-media-caption.md
```diff
@@ -0,0 +1,41 @@
...
+```hbs
+<video><track /></video>
+```
+
+```hbs
+<video><track kind="descriptions" /></video>
+```
+
+This rule **allows** the following:
```
bmish, line 23: Can we also show a `video` properly using captions?

bmish (approved):
Thanks!

PR #2501, opened 2022-05-09
Fix --version

### LINKED ISSUE #2431: Executable flag '--version' does not echo correct version
Installed latest version locally via npm.

Expecting '4.3.0', but output is '0.0.1'.

```
$ node -v
v16.13.0

$ npm -v
8.5.5

$ npm i --save-dev  ember-template-lint@latest

$ npm list ember-template-lint
└── ember-template-lint@4.3.0

$ ./node_modules/ember-template-lint/bin/ember-template-lint.js  --version
0.0.1
```

When installed globally, output is 'unknown'

```
$ npm i -g  ember-template-lint@latest
$ ember-template-lint --version
unknown
```

jamescdavis:
fixes #2431

### DIFF
--- a/lib/helpers/cli.js
+++ b/lib/helpers/cli.js
@@ -5,12 +5,14 @@ import { globbySync } from 'globby';
 import isGlob from 'is-glob';
 import micromatch from 'micromatch';
 import fs from 'node:fs';
+import { createRequire } from 'node:module';
 import path from 'node:path';
 import yargs from 'yargs';
 
 import camelize from './camelize.js';
 
 const STDIN = '/dev/stdin';
+const require = createRequire(import.meta.url);
 
 class NoMatchingFilesError extends Error {
   constructor(...params) {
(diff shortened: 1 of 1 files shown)


lib/helpers/cli.js
```diff
@@ -130,7 +132,7 @@ export function parseArgv(_argv) {
     .usage('$0 [options] [files..]')
     .options(specifiedOptions)
     .help()
-    .version();
+    .version(require('../../package.json').version);
```
bmish, line 135: This seems to indicate there is a yargs bug in retrieving the version, perhaps when ESM is present? Is there an active yargs issue for this? I'd prefer it was fixed there. If we are going to add a workaround, we would at least need to comment to explain why it's needed.

jamescdavis: You're suspicion is correct!: https://github.com/yargs/yargs/issues/2122

jamescdavis: I've verified https://github.com/yargs/yargs/pull/2123 fixes the issue, so hopefully that will go in soon!

jamescdavis:
This is bug in yargs: https://github.com/yargs/yargs/issues/2122

PR #2506, opened 2022-05-12
Converts to using npm over yarn 1

scalvert:
This is something @rwjblue and I have been discussing lately, and in fact we're moving as many of our OSS repos to using npm now over yarn 1. We're doing this for the following reasons:

- yarn 1 is no longer maintained, and bugs that exist in yarn 1 aren't prioritized to fix
- yarn used to be faster, but npm (as of v7) is blazing fast
- yarn was the only one to support workspaces, but now npm supports them to (as of v7). This is unrelated to this repo, but is a _thing_.
- yarn's registry will often timeout, causing frustrating re-runs of CI jobs. Example: https://github.com/ember-template-lint/ember-template-lint/runs/6412481173?check_suite_focus=true 

Once this change lands, contributors should follow these steps:

1. `rm -rf node_modules`
2. `npm i`
3. `npm test` to ensure things are working.


.github/workflows/ci.yml
```diff
@@ -100,6 +100,6 @@ jobs:
       - uses: volta-cli/action@v1
 
       - name: install dependencies
-        run: yarn install --no-lockfile
+        run: npm install --no-package-lock
```
bmish, line 103: `npm ci`?

.github/workflows/ci.yml
```diff
@@ -22,13 +22,13 @@ jobs:
       - uses: volta-cli/action@v1
 
       - name: install dependencies
-        run: yarn install --frozen-lockfile
+        run: npm install --frozen-lockfile
```
bmish, line 25:
```suggestion
        run: npm ci
```

bmish (commented):
We need to add `engine-strict=true` to `.npmrc` to ensure that installing dependencies still enforces the Node engine check since NPM doesn't do this by default (to ensure we only use dependencies that are compatible with our supported Node versions, more info in https://github.com/release-it/release-it/pull/890).

https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict

.github/workflows/ci.yml
scalvert: Oops, missed that one!

.github/workflows/ci.yml
scalvert: Missed this one too!

scalvert: Fixed.

.github/workflows/ci.yml
scalvert: Fixed.

bmish:
@scalvert we should update that job to use `npm ci --force` to ignore the engine check only in Node 12. That way we can have `engine-strict=true` to `.npmrc`.

scalvert:
> @scalvert we should update that job to use `npm ci --force` to ignore the engine check only in Node 12. That way we can add `engine-strict=true` to `.npmrc` applying everywhere else.

Nice, was just looking for this. Thanks!

.github/workflows/ci.yml
```diff
@@ -100,6 +100,6 @@ jobs:
       - uses: volta-cli/action@v1
 
       - name: install dependencies
-        run: yarn install --no-lockfile
+        run: npm ci
```
bmish, line 103: For the floating dependencies job, we specifically need to make sure it runs without a lockfile, try `npm ci --no-package-lock`.

scalvert: I thought I'd added that. WTF is going on...

package.json
```diff
@@ -26,9 +26,9 @@
   "scripts": {
     "lint": "npm-run-all --continue-on-error --parallel lint:*",
     "lint:docs": "markdownlint \"**/*.md\"",
-    "lint:docs:fix": "yarn lint:docs --fix",
+    "lint:docs:fix": "npm run lint:docs --fix",
```
bmish, line 29: To pass options to these NPM commands, I believe we need `-- --fix` for this to work...

scalvert: Another good catch.

scalvert: Fixed.

.github/workflows/ci.yml
scalvert: Fixed.

bmish (approved):

PR #2507, opened 2022-05-13
Add new rule `no-aria-unsupported-elements`

### LINKED ISSUE #2504: Proposal for new a11y check to disallow ARIA props on elements that do not support them
Currently, `ember-template-lint` does not enforce that certain reserved and non-visible DOM elements, such as `meta`, `html`, `script`, and `style`, do not contain the `role` and/or `aria-*` props.

This proposes a new rule to enforce that these elements do not contain ARIA props. Essentially, a port of JSX's A11y [aria-unsupported-elements](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-unsupported-elements.md) rule.

Reference: [WCAG SC 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)

thegilby:
This PR adds the new rule `no-aria-unsupported-elements` which fixes #2504.

# Background
Certain reserved DOM elements like `meta`, `html`, `script`, `style` do not support ARIA roles, states and properties. This is often because they are not visible. This rule enforces that these DOM elements do not contain the `role` and/or `aria-*` props.

Reference: [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)

## Allowed

```
# examples
<meta charset="UTF-8" />
<html lang="en"></html>
<script></script>
```

## Forbidden:

```
# examples
<meta charset="UTF-8" aria-hidden="false" />
<html lang="en" role="application"></html>
<script aria-hidden="false"></script>
```

## Testing

```
Test Suites: 138 passed, 138 total
Tests:       5 skipped, 7601 passed, 7606 total
Snapshots:   999 passed, 999 total
Time:        114.355 s, estimated 124 s
Ran all test suites.
```

### DIFF
--- a/docs/rule/no-aria-unsupported-elements.md
+++ b/docs/rule/no-aria-unsupported-elements.md
@@ -1,6 +1,15 @@
 # no-aria-unsupported-elements
 
-Certain reserved DOM elements like `meta`, `html`, `script`, `style` do not support ARIA roles, states and properties. This is often because they are not visible. This rule enforces that these DOM elements do not contain the `role` and/or `aria-*` props.
+Certain reserved DOM elements do not support ARIA roles, states and properties. This is often because they are not visible.
+
+These elements include:
+
+- `html`
+- `meta`
+- `script`
+- `style`
+
+This rule enforces that these DOM elements do not contain the `role` and/or `aria-*` props.
 
 ## Examples
 
--- a/docs/rule/no-aria-unsupported-elements.md
+++ b/docs/rule/no-aria-unsupported-elements.md
@@ -1,0 +1,38 @@
+# no-aria-unsupported-elements
+
+Certain reserved DOM elements like `meta`, `html`, `script`, `style` do not support ARIA roles, states and properties. This is often because they are not visible. This rule enforces that these DOM elements do not contain the `role` and/or `aria-*` props.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<meta charset="UTF-8" aria-hidden="false" />
+```
+
+```hbs
+<html lang="en" role="application"></html>
+```
+
+```hbs
+<script aria-hidden="false"></script>
+```
+
+This rule **allows** the following:
+
+```hbs
+<meta charset="UTF-8" />
+```
+
+```hbs
+<html lang="en"></html>
+```
+
+```hbs
+<script></script>
+```
+
+## References
+
+- [Understanding Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
+- [aria-unsupported-elements - eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-unsupported-elements.md)
--- a/lib/rules/no-aria-unsupported-elements.js
+++ b/lib/rules/no-aria-unsupported-elements.js
@@ -1,0 +1,40 @@
+import AstNodeInfo from '../helpers/ast-node-info.js';
+import Rule from './_base.js';
+
+export function ERROR_MESSAGE_ARIA_UNSUPPORTED_PROPERTY(tag, name) {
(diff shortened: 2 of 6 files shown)


lib/rules/no-aria-unsupported-elements.js
```diff
@@ -0,0 +1,40 @@
+import AstNodeInfo from '../helpers/ast-node-info.js';
+import Rule from './_base.js';
+
+export function ERROR_MESSAGE_ARIA_UNSUPPORTED_PROPERTY(tag, name) {
+  return `The <${tag}> element does not support the use of ARIA roles, states, and properties such as ${name}`;
+}
+const RESERVED_ELEMENTS = new Set(['html', 'meta', 'script', 'style']);
+
+export default class NoAriaUnsupportedElements extends Rule {
+  isReserved(tag) {
```
bmish, line 10: I would omit this function and just use `RESERVED_ELEMENTS.has(tag)` directly because:
* it's trivial
* if we really wanted to keep it, we should define it outside the class since it doesn't use any class state

lib/rules/no-aria-unsupported-elements.js
```diff
@@ -0,0 +1,40 @@
+import AstNodeInfo from '../helpers/ast-node-info.js';
+import Rule from './_base.js';
+
+export function ERROR_MESSAGE_ARIA_UNSUPPORTED_PROPERTY(tag, name) {
+  return `The <${tag}> element does not support the use of ARIA roles, states, and properties such as ${name}`;
+}
+const RESERVED_ELEMENTS = new Set(['html', 'meta', 'script', 'style']);
```
bmish, line 7: What about the name `NON_ARIA_ELEMENTS`? The meaning of "reserved" isn't clear to me.

test/unit/rules/no-aria-unsupported-elements-test.js
```diff
@@ -0,0 +1,72 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-aria-unsupported-elements',
+
+  config: true,
+
+  good: ['<meta charset="UTF-8" />', '<html lang="en"></html>', '<script></script>'],
```
bmish, line 8: Valid test cases we need to add to ensure other elements aren't restricted:
* `<div></div>`
* `<div aria-foo="true"></div>`
* `<div role="foo"></div>`

docs/rule/no-aria-unsupported-elements.md
```diff
@@ -0,0 +1,38 @@
+# no-aria-unsupported-elements
+
+Certain reserved DOM elements like `meta`, `html`, `script`, `style` do not support ARIA roles, states and properties. This is often because they are not visible. This rule enforces that these DOM elements do not contain the `role` and/or `aria-*` props.
```
bmish, line 3: Can we turn this list of elements into a bullet-point list here? That has nicer readability, especially in case the list becomes longer.

bmish (changes requested):

lib/rules/no-aria-unsupported-elements.js
thegilby: Agreed. That makes it much cleaner. Thanks!

test/unit/rules/no-aria-unsupported-elements-test.js
thegilby: Good shout out. Added these cases.

docs/rule/no-aria-unsupported-elements.md
thegilby: Cleaned up the wording a bit. Thanks!

bmish:
Will plan to make this a `recommended` rule in v5 (#2319) (let me know if you disagree with that).

thegilby:
> Will plan to make this a `recommended` rule in v5 (#2319) (let me know if you disagree with that).

No problem with that at all @bmish! Thank you! 🙇🏽‍♂️

PR #2511, opened 2022-05-16
Add new rule `no-unsupported-role-attributes`

### LINKED ISSUE #510: Executable does not respect .template-lintrc.js
To get tests passing we defined pending rules in `.template-lintrc.js`. This config is picked up correctly when running the ember tests but when running the `ember-template-lint` executable the config is ignored and it produces errors. 

As the linter executable run is added to `.travis.yml` by default in cli in `3.4` we first noticed this when running on travis and then confirmed by running locally.

judithhinlung:
This PR adds the new rule no-unsupported-role-attributes which fixes #510 .

### Background
Many ARIA states and properties are only available to elements with particular roles. This ensures that the appropriate information gets exposed to a browser's accessibility API for the given element.

This rule enforces that elements only contain ARIA attributes supported by its declared role. This is essentially a port of JSX's [role-supports-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-supports-aria-props.md) rule.

**Allowed**:
```
# examples
<div role="heading" aria-level="1" />
<input type="image" aria-atomic />
<CustomComponent role="textbox" aria-required="true" />
```
**Forbidden**:
```
# examples
<div role="link" href="#" aria-checked />
<input type="checkbox" aria-invalid="grammar" />
<CustomComponent role="listbox" aria-level="2" />
```

Reference: [Using ARIA, Roles, States, and Properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques)

### Testing
Test Suites: 128 passed, 128 total
Time:        62.737 s
Ran all test suites.
✨  Done in 80.92s.

### DIFF
--- a/lib/rules/no-unsupported-role-attributes.js
+++ b/lib/rules/no-unsupported-role-attributes.js
@@ -83,6 +83,49 @@ export default class NoUnsupportedRoleAttributes extends Rule {
           }
         }
       },
+      MustacheStatement(node) {
+        const roleAttribute = node.hash.pairs.find((pair) => pair.key === 'role');
+        if (!roleAttribute) {
+          return;
+        }
+        if (!roleAttribute.value.type === 'StringLiteral') {
+          return;
+        }
+        let role = roleAttribute.value.original;
+
+        // Skip validation for elements with unknown ARIA roles
+        if (!role) {
+          return;
+        }
+        const roleDefinition = roles.get(role);
+        if (!roleDefinition) {
+          return;
+        }
+
+        // Get a list of the ARIA attributes defined for this element
+        let foundAriaAttributes = [];
+        for (const pair of node.hash.pairs) {
+          if (pair.key.startsWith('aria-')) {
+            foundAriaAttributes.push(pair);
+          }
+        }
+
+        // Check that each ARIA attribute found is within the supported property set for the role
+        const supportedProps = Object.keys(roleDefinition.props);
+        for (let attribute of foundAriaAttributes) {
+          if (!supportedProps.includes(attribute.key)) {
+            if (this.mode === 'fix') {
+              node.hash.pairs = node.hash.pairs.filter((pair) => pair !== attribute);
+            } else {
+              this.log({
+                message: createUnsupportedAttributeErrorMessage(attribute.key, role, undefined),
+                node,
+                isFixable: true,
+              });
+            }
+          }
+        }
+      },
     };
   }
 }
--- a/lib/rules/no-unsupported-role-attributes.js
+++ b/lib/rules/no-unsupported-role-attributes.js
@@ -1,0 +1,84 @@
+import { roles, elementRoles } from 'aria-query';
+
+import AstNodeInfo from '../helpers/ast-node-info.js';
+import Rule from './_base.js';
+
+function createUnsupportedAttributeErrorMessage(attr, role, element) {
+  if (element) {
+    return `The attribute ${attr} is not supported by the element ${element} with the implicit role of ${role}`;
+  } else {
+    return `The attribute ${attr} is not supported by the role ${role}`;
+  }
+}
+function getImplicitRole(element, typeAttribute) {
+  if (element === 'input') {
+    for (let key of elementRoles.keys()) {
+      if (key.name === element) {
+        let attributes = key.attributes;
+        for (let attribute of attributes) {
+          if (attribute.name === 'type' && attribute.value === typeAttribute) {
+            return elementRoles.get(key)[0];
+          }
+        }
+      }
+    }
+  } else {
+    return elementRoles.get(elementRoles.keys().find((key) => key.name === element))[0];
+  }
+}
+
+export default class NoUnsupportedRoleAttributes extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        let element, typeAttribute, role;
+        if (AstNodeInfo.hasAttribute(node, 'role')) {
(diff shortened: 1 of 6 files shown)


test/unit/rules/no-unsupported-role-attributes-test.js
```diff
@@ -0,0 +1,233 @@
...
+    '<CustomComponent role="banner" />',
+    '<div role="textbox" aria-required={{this.required}} aria-errormessage={{this.error}}></div>',
+    '<dialog />',
+    '<a href="#" aria-describedby=""></a>',
+    '<menu type="toolbar" aria-hidden="true" />',
+    '<menuitem type="command" aria-labelledby={{this.label}} />',
+    '<input type="image" aria-atomic />',
+    '<input type="submit" aria-disabled="true" />',
+    '<select aria-expanded="false" aria-controls="ctrlID" />',
+  ],
```
bmish, line 21: Are these examples with custom/unknown elements all valid cases? In that we don't enforce anything on them?
* `<foo role="button" aria-fake="true">`
* `<foo-component role="button" aria-fake="true">`
* `<FooComponent role="button" aria-fake="true">`
* `<foo role="fake" aria-fake="true">`

lib/rules/no-unsupported-role-attributes.js
```diff
@@ -0,0 +1,84 @@
...
+      if (key.name === element) {
+        let attributes = key.attributes;
+        for (let attribute of attributes) {
+          if (attribute.name === 'type' && attribute.value === typeAttribute) {
+            return elementRoles.get(key)[0];
+          }
+        }
+      }
+    }
+  } else {
```
bmish, line 25: Can we use an early return instead here?

I found a lint rule that I think would enforce this: https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/prefer-early-return.md

lib/rules/no-unsupported-role-attributes.js
```diff
@@ -0,0 +1,84 @@
...
+
+export default class NoUnsupportedRoleAttributes extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
+        let element, typeAttribute, role;
+        if (AstNodeInfo.hasAttribute(node, 'role')) {
+          let roleAttrNode = AstNodeInfo.findAttribute(node, 'role');
+          if (roleAttrNode.value.type === 'TextNode') {
+            role = roleAttrNode.value.chars || '';
```
bmish, line 38: I think `null` or `undefined` is more clear for indicating that we couldn't find a role / type.

test/unit/rules/no-unsupported-role-attributes-test.js
bmish, line 21: Should have a valid test case to indicate that any random non-aria attribute is allowed:
* `<div role="heading" foo="true" />`
* `<div type="button" foo="true" />`

lib/rules/no-unsupported-role-attributes.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+  } else {
+    return elementRoles.get(elementRoles.keys().find((key) => key.name === element))[0];
+  }
+}
+
+export default class NoUnsupportedRoleAttributes extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
```
bmish, line 33: Maybe add a few more summary comments to explain what's going on at a high-level like: 

"First look to see if a role is explicitly defined"
"Otherwise check for an implicit role from the type"

test/unit/rules/no-unsupported-role-attributes-test.js
judithhinlung: The first three cases would fail the tests since aria-fake is not a supported ARIA property for the `button` role. They will fail in a similar way as <div role="option" aria-notreal="bogus" aria-selected="false" />.
The last case should already be caught by the [no-invalid-role](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-role.md) rule.

bmish: Got it, just want to make sure we include every variation.

Can you make sure we have test cases with the different styles of elements / custom components?

* `<foo>`
* `<FooComponent>`
* `{{foo-component}}`

judithhinlung: Currentlythere are good and bad test cases for elements <foo> and angle-bracket components <CustomComponent />.  However curly-component invocations {{foo}} are allowed to pass, because it could be difficult to differentiate between component arguments and attribute values in these invocations.

lib/rules/no-unsupported-role-attributes.js
judithhinlung: Yes thank you for pointing this out, the code has been updated with an early return.

test/unit/rules/no-unsupported-role-attributes-test.js
bmish: Okay, I think it could still be worth adding test cases for the curly components, even if the output is wrong, and then commenting to say why these test cases are wrong / have this limitation right now. This helps make it clear what the behavior should be in each situation and even what potential improvements could be made to the rule later or at least what improvements were considered.

bmish: Thanks! I think it's important to continue covering/testing mustache statements/components for completeness.

bmish (approved):
Thanks!

PR #2513, opened 2022-05-18
Update overrides example

jamescdavis:
Specifying overrides with a module name is not valid. This updates the docs to indicate a full file name with extension is required.


bmish:
Do you happen to know of the PR where the the change from module to file occurred? I see the documentation was originally added in: https://github.com/ember-template-lint/ember-template-lint/pull/1117

jamescdavis:
Actually, AFAICT, it _never_ worked with module and always expected a file path with extension since it was introduced in #1117 (I checked out out the merge commit and tested).

MelSumner:
@bmish this seems like a useful update to the docs; would it be possible to get this merged?

bmish (approved):
Thanks!

Issue #2128, opened 2021-10-08
Clean up remaining instances of moduleId


carakessler:
Currently, there is an expectation that a `moduleId` is passed in via `options` param when calling `verify` via node API (although the jsdoc comments list out the different options, and `moduleId` is NOT there). Without it, if ignore or pending items are configured, it will error out (they expect a moduleId). 

Here is an example of the stack trace: 
```
"message": "Expected input to be a string",
                                "stack": {
                                    "frames": [
                                        {
                                            "module": "Function.picomatch.test (/Users/ckessler/voyager-web_trunk/node_modules/picomatch/lib/picomatch.js:119:11)"
                                        },
                                        {
                                            "module": "matcher (/Users/ckessler/voyager-web_trunk/node_modules/picomatch/lib/picomatch.js:67:50)"
                                        },
                                        {
                                            "module": "/Users/ckessler/voyager-web_trunk/node_modules/ember-template-lint/lib/-private/module-status-cache.js:33:61"
                                        },
                                        {
                                            "module": "Array.find (<anonymous>)"
                                        },
                                        {
                                            "module": "ModuleStatusCache.lookupIgnore (/Users/ckessler/voyager-web_trunk/node_modules/ember-template-lint/lib/-private/module-status-cache.js:33:45)"
                                        },
                                        {
                                            "module": "ModuleStatusCache.getConfigForFile (/Users/ckessler/voyager-web_trunk/node_modules/ember-template-lint/lib/-private/module-status-cache.js:73:26)"
                                        },
                                        {
                                            "module": "Linter.verify (/Users/ckessler/voyager-web_trunk/node_modules/ember-template-lint/lib/linter.js:308:46)"
                                        },
```

To fix this, I would recommend just removing moduleId from needing to be passed in externally, and just do the file path stripping internally when necessary (the only diff between `moduleId` and `filePath`). Additionally, the required options should be verified in the `verify` function (pun intended).

PR #2515, opened 2022-05-19
Clean up `moduleId`, add ability to specify `ignore` via file path with extension

jamescdavis:
As described in #2128, when `ignore` is configured, `moduleId` is still required when calling the node API's `verify()`  despite the [docs](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/node-api.md) making no mention of this. This remedies that by cleaning up usage of `moduleId` internally as proposed in #2128.

The one place I've left any reference to `moduleId` is within `lib/-private/module-status-cache` because configuration of `ignore` for specific files (non-glob) currently requires specification as a module instead of a file path with extension, e.g.

```js
// .template-lintrc.js
module.exports = {
  ignore: ['path/to/template'],
};
```

I have, though, added the ability to also specify specific files via path with extension, e.g.

```js
// .template-lintrc.js
module.exports = {
  ignore: ['path/to/template.hbs'],
};
```

This was motivated in part with wanting to have consistency with `overrides` (see #2513).

We may want to deprecate specifying by module and remove in next major version.

### DIFF
--- a/bin/ember-template-lint.js
+++ b/bin/ember-template-lint.js
@@ -20,6 +20,7 @@ import { promisify } from 'node:util';
 import { parseArgv, getFilesToLint } from '../lib/helpers/cli.js';
 import printResults from '../lib/helpers/print-results.js';
 import processResults from '../lib/helpers/process-results.js';
+import removeExt from '../lib/helpers/remove-ext.js';
 import Linter from '../lib/linter.js';
 
 const readFile = promisify(fs.readFile);
@@ -31,10 +32,6 @@ const NOOP_CONSOLE = {
   warn: () => {},
   error: () => {},
 };
-
-function removeExt(filePath) {
-  return filePath.slice(0, -path.extname(filePath).length);
-}
 
 async function buildLinterOptions(workingDir, filePath, filename = '', isReadingStdin) {
   if (isReadingStdin) {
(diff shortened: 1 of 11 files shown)


jamescdavis:
This is out of scope for this PR, but should we still be duplicating the ignores in this example
https://github.com/ember-template-lint/ember-template-lint/blob/ba97a8737ad1dcd26fd225e5639f155b59589981/docs/ignore.md?plain=1#L16-L21
now that ember-cli-template-lint is deprecated?

Also
https://github.com/ember-template-lint/ember-template-lint/blob/ba97a8737ad1dcd26fd225e5639f155b59589981/docs/ignore.md?plain=1#L35
should probably be removed at this point, regardless.

jamescdavis:
Also

https://github.com/ember-template-lint/ember-template-lint/blob/ba97a8737ad1dcd26fd225e5639f155b59589981/docs/ignore.md?plain=1#L35

should probably be removed at this point, regardless.

bmish:
@jamescdavis very sorry about the delay on this. A few questions:

1. Can you rebase this on the latest master?
2. Can you explain in a bit more detail if this is an internal-only change, a bug fix, a feature, or a breaking change?
3. Is this ready to merge?

lifeart:
> Can you rebase this on the latest master?
- done
> Can you explain in a bit more detail if this is an internal-only change, a bug fix, a feature, or a breaking change?
- it looks like a breaking thing for programmatic users
> Is this ready to merge?
- yep

NullVoxPopuli:
out of caution, I think I want to hold off on merging this until we're ready to do v8

PR #2516, opened 2022-05-19
Prevent `no-unsupported-role-attributes` rule from failing if the element does not have an implicit role

### LINKED ISSUE #2514: `no-unsupported-role-attributes` throws an error
Namely `error  Cannot read properties of undefined (reading '0')  undefined`. Gives it on a bunch of files, the simplest one being just: `<ItemCheckbox @model={{@model}} @checkable={{@checkable}} />`.

bertdeblock:
Closes #2514.


bmish (approved):
Thanks!

PR #2523, opened 2022-05-23
Add new rule `require-mandatory-role-attributes`

### LINKED ISSUE #2522: Proposal for new a11y check to enforce that ARIA roles have required attributes
Currently, ember-template-lint does not enforce that elements declaring an ARIA role include the required attributes for that role. This requirement ensures that a given element possesses the necessary states and properties to behave consistently with user expectations for other elements with the same ARIA role.

This proposes the authoring of a new rule that enforces that elements with an ARIA role also declare all required ARIA attributes for that role.

Reference: [WAI-ARIA Roles - Accessibility _ MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)

judithhinlung:
This PR adds the new rule role-has-required-attributes which fixes #2522 .

### Background
Elements with ARIA roles must also include all required attributes for that role. This ensures that a given element possesses the necessary states and properties to behave consistently with user expectations for other elements with the same ARIA role.

This rule enforces that elements with an ARIA role also declare all required ARIA attributes for that role. This rule is analogous to JSXs A11y [role-has-required-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-has-required-aria-props.md) rule.

**Allowed**:
```
# examples
<div role="option" aria-selected="false" />
<CustomComponent role="checkbox" aria-required="true" aria-checked="false" />
{{some-component role="heading" aria-boguslevel="2" aria-level="2"}}
```

**Forbidden**:
```
# examples
<div role="option" />
<CustomComponent role="checkbox" aria-required="true" />
{{some-component role="heading" aria-boguslevel="2"}}
```

Reference: [WAI-ARIA Roles - Accessibility _ MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)

###  Testing
Test Suites: 140 passed, 140 total
Tests:       5 skipped, 7728 passed, 7733 total
Snapshots:   1014 passed, 1014 total
Time:        77.767 s
Ran all test suites.
✨  Done in 82.26s.

### DIFF
--- a/docs/rule/role-has-required-attributes.md
+++ b/docs/rule/role-has-required-attributes.md
@@ -1,0 +1,27 @@
+# role-has-required-attributes
+
+Elements with ARIA roles must also include all required attributes for that role. This ensures that a given element possesses the necessary states and properties to behave consistently with user expectations for other elements with the same ARIA role.
+
+This rule enforces that elements with an ARIA role also declare all required ARIA attributes for that role.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<div role="option" />
+<CustomComponent role="checkbox" aria-required="true" />
+{{some-component role="heading" aria-boguslevel="2"}}
+```
+
+This rule **allows** the following:
+
+```hbs
+<div role="option" aria-selected="false" />
+<CustomComponent role="checkbox" aria-required="true" aria-checked="false" />
+{{some-component role="heading" aria-boguslevel="2" aria-level="2"}}
+```
+
+## References
+
+- [WAI-ARIA Roles - Accessibility _ MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)
--- a/test/unit/rules/require-mandatory-role-attributes-test.js
+++ b/test/unit/rules/require-mandatory-role-attributes-test.js
@@ -6,15 +6,23 @@ generateRuleTests({
   config: true,
 
   good: [
+    '<div />',
+    '<div aria-disabled="true" />',
     '<div role="complementary" />',
     '<div role="combobox" aria-expanded="false" aria-controls="ctrlId" />',
     '<div role="option" aria-selected={{false}} />',
+    '<FakeComponent />',
+    '<FakeComponent role="fakerole" />',
     '<CustomComponent role="checkbox" aria-checked="false" />',
     '<SomeComponent role={{this.role}} aria-notreal="bar" />',
     '<OtherComponent @role={{@role}} aria-required={{this.required}} />',
     '<FakeElement aria-disabled="true" />',
+    '{{some-component}}',
+    '{{some-component foo="true"}}',
     '{{some-component role="heading" aria-level="2"}}',
     '{{foo-component role="button"}}',
+    '{{foo-component role="unknown"}}',
+    '{{foo-component role=role}}',
   ],
 
   bad: [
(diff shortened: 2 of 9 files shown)


docs/rule/role-has-required-attributes.md
```diff
@@ -0,0 +1,27 @@
...
+<CustomComponent role="checkbox" aria-required="true" />
+{{some-component role="heading" aria-boguslevel="2"}}
+```
+
+This rule **allows** the following:
+
+```hbs
+<div role="option" aria-selected="false" />
+<CustomComponent role="checkbox" aria-required="true" aria-checked="false" />
+{{some-component role="heading" aria-boguslevel="2" aria-level="2"}}
```
thegilby, line 22:
```suggestion
{{some-component role="heading" aria-level="2"}}
```

While it is _technically_ allowed by definition (and handled by the `no-invalid-aria-attributes` rule), can we clean up this allowed example to remove the `aria-boguslevel` attribute?

thegilby (commented):
Looks great @judithhinlung! Just left a small comment below. Thanks for adding this rule!

bmish:
What do you think of a rule name like one of the following to be more consistent with the naming of our other rules?
* `require-mandatory-role-attributes`
* `no-missing-role-attributes`

test/unit/rules/role-has-required-attributes-test.js
```diff
@@ -0,0 +1,128 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'role-has-required-attributes',
+
+  config: true,
+
+  good: [
```
bmish, line 8: Remember to test on an existing codebase to ensure no crashes.

bertdeblock: FWIW, just did a quick test on our codebase. No crashes here 👍.

judithhinlung: Great, thanks so much @bertdeblock !

judithhinlung:
> What do you think of a rule name like one of the following to be more consistent with the naming of our other rules?
> 
> * `require-mandatory-role-attributes`
> * `no-missing-role-attributes`
> * something else?

Thank you so much for the suggestions @bmish, I have changed it to require-mandatory-role-attributes.

docs/rule/role-has-required-attributes.md
thegilby: Thanks for updating this @judithhinlung! I also dig the updated rule name. 🙌🏽

test/unit/rules/require-mandatory-role-attributes-test.js
```diff
@@ -0,0 +1,128 @@
...
+    '<div role="complementary" />',
+    '<div role="combobox" aria-expanded="false" aria-controls="ctrlId" />',
+    '<div role="option" aria-selected={{false}} />',
+    '<CustomComponent role="checkbox" aria-checked="false" />',
+    '<SomeComponent role={{this.role}} aria-notreal="bar" />',
+    '<OtherComponent @role={{@role}} aria-required={{this.required}} />',
+    '<FakeElement aria-disabled="true" />',
+    '{{some-component role="heading" aria-level="2"}}',
+    '{{foo-component role="button"}}',
+  ],
```
bmish, line 18: We should always have some minimal test cases like the below to ensure we don't break stuff that doesn't use a role at all:

* `<div>`
* `<div foo="true">`
* `<FooComponent>`
* `<FooComponent foo="true">`
* `{{foo-component}}`
* `{{foo-component foo="true"}}`

bmish: One more tip is to look at the code coverage report output by `npm run test:jest` to ensure every line of the new rule is covered by tests.

bmish (approved):
Thanks!

PR #2527, opened 2022-06-03
Add route action name to error message in `no-route-action` rule

tylerbecks:
This PR adds the route-action name to the message. Instead of this:
```
  3:23  todo  Do not use `route-action` as {{route-action ...}}. Instead, use controller actions.  no-route-action
  4:19  todo  Do not use `route-action` as {{route-action ...}}. Instead, use controller actions.  no-route-action
  5:29  todo  Do not use `route-action` as {{route-action ...}}. Instead, use controller actions.  no-route-action
```

The no-route-action rule will produce this:
```
  3:23  todo  Do not use `route-action` as {{route-action "rescheduleInterview"}}. Instead, use controller actions.  no-route-action
  4:19  todo  Do not use `route-action` as {{route-action "cancelInterview"}}. Instead, use controller actions.  no-route-action
  5:29  todo  Do not use `route-action` as {{route-action "sendCandidateConfirmation"}}. Instead, use controller actions.  no-route-action
```


lib/rules/no-route-action.js
```diff
@@ -1,15 +1,16 @@
 import Rule from './_base.js';
 
 const ERROR_MESSAGE = 'Do not use `route-action` as %. Instead, use controller actions.';
```
bmish, line 3: Can we change this into a function? That's how we typically generate dynamic error messages instead of trying to do replacements. 

```
function makeErrorMessage(actionName) { ... }
```

lib/rules/no-route-action.js
```diff
@@ -25,12 +23,26 @@ export default class NoRouteAction extends Rule {
...
 }
+
+function makeErrorMessage(actionName, isSubExpression) {
+  const PLACE_HOLDER = '%';
+  const ERROR_MESSAGE = `Do not use \`route-action\` as ${PLACE_HOLDER}. Instead, use controller actions.`;
+  const usageContext = isSubExpression
+    ? `(route-action '${actionName}')`
+    : `{{route-action '${actionName}'}}`;
+
+  return ERROR_MESSAGE.replace(PLACE_HOLDER, usageContext);
```
bmish, line 47: Can we eliminate this replacement? I don't see why there should be any replacements/placeholders involved.

bmish:
> @bmish is the memory heap issue a known issue? I have a failing check for Node 18.x

Yes that's a known issue which you can ignore.

bmish (approved):
Thanks!

PR #2533, opened 2022-06-13
Add `no-action` rule migration example

angelayanpan:
add example to rule: no-action

### DIFF
--- a/docs/rule/no-action.md
+++ b/docs/rule/no-action.md
@@ -4,7 +4,7 @@
 
 What's wrong with `{{action}}`?
 
-"Action" is an overloaded term in Ember parlance. Actions are:
+"Action" is an overloaded term in Ember idiom. Actions are:
 
 1.) Methods on the `actions` hash
 
@@ -101,6 +101,16 @@ to
 </select>
 ```
 
+
+```hbs
+<myComponent @onValidationChange={{action "onDateValidation"}} />
+```
+
+to 
+
+```hbs
+<myComponent @onValidationChange={{this.onDateValidation}} />
+```
 ## Related Rules
 
 * [no-action-modifiers](no-action-modifiers.md)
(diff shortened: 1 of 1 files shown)


docs/rule/no-action.md
```diff
@@ -4,7 +4,7 @@
 
 What's wrong with `{{action}}`?
 
-"Action" is an overloaded term in Ember parlance. Actions are:
+"Action" is an overloaded term in Ember idiom. Actions are:
```
bmish, line 7: Not a big deal but parlance seems a bit more fitting here?

docs/rule/no-action.md
```diff
@@ -101,6 +101,17 @@ to
...
+```hbs
+<myComponent @onValidationChange={{action "onDateValidation"}} />
+```
+
+to 
+
+```hbs
+<myComponent @onValidationChange={{this.onDateValidation}} />
+```
+
```
tylerbecks, line 114: Components are usually capitalized.

```suggestion
```hbs
<MyComponent @onValidationChange={{action "onDateValidation"}} />
```

to 

```hbs
<MyComponent @onValidationChange={{this.onDateValidation}} />
```

```

File: docs/rule/no-action.md (hunk 1 above)
Comment (tylerbecks) [reply docs/rule/no-action.md:7]: While somewhat unrelated to this PR, I'm in favor of this change.

Maybe a nice alternative would be `diction`, `language`, or `vocabulary`.

Or, simply drop the word all-together.

```
"Action" is an overloaded term in Ember. Actions are:
```

bmish: Any of these new or existing options sound fine to me. Feel free to choose whatever you think will be most-easily understood.

angelayanpan: ok, I proposed the change because I didn't know what parlance is.  But looking at the alternatives, I also feel like we can drop it all together.  since it already says `is an overloaded term`

PR #2535, opened 2022-06-15
Clarify violation message for `no-dynamic-subexpression-invocations` rule

chrisrng:
Suggesting a clarification to the message here since we've had a case where someone thought this meant it will invoke the function with arguments. Not that we can invoke a function with arguments with this helper.


chrisrng:
@bmish will do!

bmish (approved):
Thanks!

PR #2540, opened 2022-06-25
Fix config documentation for `table-groups` rule

nulle:
There seem to be two bugs with one of the examples for `table-groups` rule:
1. misspelled name of the configuration key itself
2. configuration value should be array of strings


bmish (approved):
Thanks!

PR #2541, opened 2022-06-25
Fix helper used by several rules for dasherizing multi-level nested component names

nulle:
Fixing a bug where component that is more than 1 level nested would get dasherized incorrectly.

Example: `<Popup::List::SearchButton>` would get incorrectly dasherized as `popup/list::search-button`. 
Expected result: `popup/list/search-button`.

### DIFF
--- a/lib/helpers/dasherize-component-name.js
+++ b/lib/helpers/dasherize-component-name.js
@@ -14,5 +14,5 @@ export default function (key) {
 
       return `-${char.toLowerCase()}`;
     })
-    .replace('::', '/');
+    .replace(/::/g, '/');
 }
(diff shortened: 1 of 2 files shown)


bmish:
Can you add tests for this? Unit tests for this function would be good. Or at the minimum, it could be tested by a rule that uses this helper.

nulle:
@bmish added a unit test [c44ca4e](https://github.com/ember-template-lint/ember-template-lint/pull/2541/commits/c44ca4ed80dee3bdc8e04c564854b3097c25c653), syntax based on the [existing test for curly-invocation-component](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/helpers/curly-invocation-component-test.js).

bmish (approved):
Nice, thanks!

PR #2554, opened 2022-07-14
Add new rule `no-at-ember-render-modifiers`

NullVoxPopuli:
This library is not recommended.

Should it be a stylistic rule or default?

There are a number of errors around the rule not being found -- any pointers would be great -- then I can add more tests :upside_down_face: 

Here is the work for the equiv gjs/gts no-importing from this library: https://github.com/ember-cli/eslint-plugin-ember/issues/1901


lib/config/recommended.js
```diff
@@ -12,6 +12,8 @@ export default {
     'no-args-paths': 'error',
     'no-arguments-for-html-elements': 'error',
     'no-aria-hidden-body': 'error',
+    // Error on next major
+    'no-at-ember-render-modifiers': 'warn',
```
bmish, line 16: We can add this as a `recommended` rule in the next major release but we don't typically `warn` for any rules so I would avoid adding it to this file.

bmish (commented):
1. You'll need to run `npm run update:indices`.
2. We use npm not pnpm.

test/unit/rules/no-at-ember-render-modifiers-test.js
```diff
@@ -0,0 +1,73 @@
+import { MESSAGES } from '../../../lib/rules/no-at-ember-render-modifiers.js';
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-at-ember-render-modifiers',
+
+  config: true,
+
+  good: ['<div {{this.someModifier}}></div>', '<div {{someModifier}}></div>'],
```
bmish, line 9: Looks like we need a valid test case for:
* `<div {{did-foo}}></div>`
* `{{did-insert}}` (by itself)

docs/rule/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,50 @@
...
+<div {{did-update this.someFunction}}>
+```
+
+```hbs
+<div {{will-destroy this.someFunction}}>
+```
+
+## Migration
+
+The migration path typically depends on what the render-modifier was used for, but if you need a customm modifier, the [`ember-modifier` README](https://github.com/ember-modifier/ember-modifier) covers everything you need to know for making custom modifiers.
```
bmish, line 25:
```suggestion
The migration path typically depends on what the render-modifier was used for, but if you need a custom modifier, the [`ember-modifier` README](https://github.com/ember-modifier/ember-modifier) covers everything you need to know for making custom modifiers.
```

docs/rule/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,50 @@
+# no-at-ember-render-modifiers
+
+`@ember/render-modifiers` were meant as a transitional tool from the pre-Octane era, and not long-term usage.
+
+More information about this can be found on the [`@ember/render-modifiers` README](https://github.com/emberjs/ember-render-modifiers#when-to-use-these-modifiers-and-when-not-to-use-them)
```
bmish, line 5:
```suggestion
More information about this can be found on the [`@ember/render-modifiers` README](https://github.com/emberjs/ember-render-modifiers#when-to-use-these-modifiers-and-when-not-to-use-them).
```

docs/rule/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,50 @@
...
+}
+```
+
+```hbs
+<div {{this.myModifier}}>
+```
+
+## References
+
+- [`@ember/render-modifiers`](https://github.com/emberjs/ember-render-modifiers)
```
bmish, line 49: Perhaps we should indicate that this link is deprecated to make it easier to distinguish from the ember-modifier link below?

```suggestion
- [`@ember/render-modifiers`](https://github.com/emberjs/ember-render-modifiers) (deprecated)
```

test/unit/rules/no-at-ember-render-modifiers-test.js
```diff
@@ -0,0 +1,73 @@
...
+      verifyResults(results) {
+        expect(results).toMatchInlineSnapshot(`
+          [
+            {
+              "column": 5,
+              "endColumn": 19,
+              "endLine": 1,
+              "filePath": "layout.hbs",
+              "line": 1,
+              "message": "${MESSAGES['did-insert']}",
```
bmish, line 23: I'm not sure that snapshots should be using variables like this (presumably you manually added the variable here?). The point of snapshots is that they can be automatically updated by jest whenever anything in this object changes.

lib/rules/no-at-ember-render-modifiers.js
```diff
@@ -0,0 +1,30 @@
...
+      ElementModifierStatement(node, { parentNode }) {
+        let modifierName = node.path.original;
+        if (!MODIFIERS.has(modifierName)) {
+          return;
+        }
+
+        this.log({
+          message: MESSAGES[modifierName],
+          node,
+          source: this.sourceForNode(parentNode),
```
bmish, line 25: Why are we highlighting the source for the parent node instead of just the modifier node itself?

docs/rule/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,50 @@
...
+<div {{will-destroy this.someFunction}}>
+```
+
+## Migration
+
+The migration path typically depends on what the render-modifier was used for, but if you need a customm modifier, the [`ember-modifier` README](https://github.com/ember-modifier/ember-modifier) covers everything you need to know for making custom modifiers.
+
+For example, if render modifiers were used for setup/teardown, the migration to `ember-modifier` could be the following:
+
+```js
```
bmish, line 29: Can we add the import for `Component` as well for completeness of this example?

rob-long:
@NullVoxPopuli I'm interested in getting this PR to the finish line. I might have time coming up if you're ok with me jumping in?

NullVoxPopuli:
Go for it, i forgot this existed <3

NullVoxPopuli:
I am picking this back up 🎉

test/unit/rules/no-at-ember-render-modifiers-test.js
NullVoxPopuli: done! thanks!

lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: I don't know -- I changed it to the modifier, because that makes sense

test/unit/rules/no-at-ember-render-modifiers-test.js
NullVoxPopuli: Updated (thanks to a new lint rule!)

docs/rule/no-at-ember-render-modifiers.md
NullVoxPopuli: yup!

test/unit/rules/no-at-ember-render-modifiers-test.js
```diff
@@ -6,67 +6,133 @@ generateRuleTests({
...
+                filePath: 'layout.hbs',
+                line: 1,
+                message: MESSAGES['did-insert'],
+                rule: 'no-at-ember-render-modifiers',
+                severity: 2,
+                source: '<div {{did-insert}}></div>',
+              },
+            ],
+          },
+          `
```
bmish, line 40: Looks like there's some old leftover snapshots in these tests. There should only be one arg to `toMatchInlineSnapshot`.

docs/rule/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,51 @@
...
+    element.addEventListener('eventName', handleEvent);
+
+    return () => element.removeEventListener('eventName', handelEvent);
+  });
+}
+```
+
+```hbs
+<div {{this.myModifier}}>
+```
```
bmish, line 46: What is this code sample for?

bmish:
Command to update the readme needs to be run.

test/unit/rules/no-at-ember-render-modifiers-test.js
NullVoxPopuli: the lint prevented from using a string, so I had to change to a real object.
then, when jest was running, it updated the string stuff after success -- the string part is the real snapshot persisted by jest, and the js object is the "expected" and helps out with the diff and generated the snapshot if the actual matches the expected.

Is this not how it's supposed to work?

docs/rule/no-at-ember-render-modifiers.md
NullVoxPopuli: it's the template for the above class

NullVoxPopuli:
> Command to update the readme needs to be run.

pushed, thanks!

test/unit/rules/no-at-ember-render-modifiers-test.js
bmish: Jest's auto-updating of inline snapshots is buggy. It left the old snapshot in there. Just clear out the snapshots and let it regenerate without the extra arg.

bmish: There are two arguments passed to `toMatchInlineSnapshot` in your code. There should only be one. One is a duplicate of the other. You need to delete the arguments to `toMatchInlineSnapshot` and re-run jest to regenerate the snapshots.

NullVoxPopuli: but if I do that, jest adds the second argument back -- no diff is seen when doing `git status` / `git diff`.

To prove, I've added two commits:
- first removes the second argument, as you suggest
- second, is what happens after running the tests again

diff:
```
❯ git diff HEAD~2

ember-template-lint on  no-at-ember-render-modifiers [⇡] 
```

NullVoxPopuli: thanks! tho, now I don't know why the lint rule I was seeing isn't triggering :sweat_smile:

bmish (approved):
Thanks!

PR #2568, opened 2022-08-03
Add autofixer to `no-unnecessary-concat` rule

locks:
Also tested on a production codebase.


bmish:
The test indicates the rule doc needs to be updated too.

bmish (approved):
Thank you! Excited to see more autofixers.

PR #2575, opened 2022-08-07
Element attributes order

rob-long:
This adds new rule `element-attributes-order` for issue https://github.com/ember-template-lint/ember-template-lint/issues/684.  It takes inspiration from the previous PR made by @lifeart but makes some large changes to:

a. allow configuration of the attribute order by `attributeOrder`
b. allows config of whether or not to alphabetize within the attribute groupings (defaults false)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Each rule has emojis denoting:
 | [builtin-component-arguments](./docs/rule/builtin-component-arguments.md)                                 | ✅  |     |     |     |
 | [deprecated-inline-view-helper](./docs/rule/deprecated-inline-view-helper.md)                             | ✅  |     |     |     |
 | [deprecated-render-helper](./docs/rule/deprecated-render-helper.md)                                       | ✅  |     |     |     |
+| [element-attributes-order](./docs/rule/element-attributes-order.md)                                       |     |     |     |     |
 | [eol-last](./docs/rule/eol-last.md)                                                                       |     | 💅  |     | 🔧  |
 | [inline-link-to](./docs/rule/inline-link-to.md)                                                           |     |     |     | 🔧  |
 | [linebreak-style](./docs/rule/linebreak-style.md)                                                         |     | 💅  |     |     |
--- a/docs/rule/element-attributes-order.md
+++ b/docs/rule/element-attributes-order.md
@@ -1,0 +1,48 @@
+# element-attributes-order
+
+Element attributes must be ordered, by default the order is:
+
+- arguments
+- attributes
+- element modifiers
+- ...attributes
+- ?attributes
+
+## Configuration
+
+The following values are valid configuration:
+
+- attributeOrder: [`arguments`, `attributes`, `modifiers`, `splattributes`]
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<MyComponent ...attributes @name='Hello' />
+```
+
+```hbs
+<MyComponent {{did-render this.someAction}} @name='Hello' />
+```
+
+```hbs
+<MyComponent ...attributes {{did-render this.someAction}} />
+```
+
+This rule **allows** the following:
+
+```hbs
+<MyComponent @name='1' bar='baz' {{did-render this.someAction}} ...attributes aria-role='button' />
+<MyComponent @name='1' aria-role='button' />
+```
+
+```hbs
+<MyComponent @name='1' ...attributes />
+<MyComponent @name='1' {{did-render this.someAction}} />
+<MyComponent @name='1' bar='baz' />
+```
+
+## References
+
+- [eslint-plugin-vue/rules/attributes-order](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/attributes-order.md)
--- a/docs/rule/element-attributes-order.md
+++ b/docs/rule/element-attributes-order.md
@@ -12,7 +12,8 @@ Element attributes must be ordered, by default the order is:
 
 The following values are valid configuration:
 
-- attributeOrder: [`arguments`, `attributes`, `modifiers`, `splattributes`]
+- attributeOrder: default [`arguments`, `attributes`, `modifiers`, `splattributes`]
+- alphabetize: default `false`, `true` to enforce alphabetically ordered values within attributes
 
 ## Examples
 
--- a/test/unit/rules/element-attributes-order-test.js
+++ b/test/unit/rules/element-attributes-order-test.js
@@ -79,5 +79,17 @@ generateRuleTests({
         column: 38,
       },
     },
+    {
+      config: {
+        attributeOrder: ['attributes', 'arguments', 'modifiers', 'splattributes'],
+      },
+      template: '<div @foo="1" {{did-render this.ok}} ...attributes aria-label="foo"></div>',
+      result: {
+        message: 'Attribute aria-label="foo" must go before arguments, modifiers and splattributes',
+        source: 'aria-label="foo"',
+        line: 1,
+        column: 51,
+      },
+    },
   ],
 });
--- a/test/unit/rules/element-attributes-order-test.js
+++ b/test/unit/rules/element-attributes-order-test.js
@@ -1,0 +1,83 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'element-attributes-order',
+
+  config: true,
+
+  good: [
+    '<div @foo="1" bar="baz" {{did-render this.ok}} ...attributes aria-label="foo"></div>',
+    '<div @foo="1" aria-label="foo"></div>',
+    '<div @foo="1" ...attributes></div>',
+    '<div @foo="1" {{did-render this.ok}}></div>',
+    '<div @foo="1" bar="baz"></div>',
+    '<div bar="baz" {{did-render this.ok}} ...attributes aria-label="foo"></div>',
+    '<div bar="baz" aria-label="foo"></div>',
+    '<div bar="baz" ...attributes></div>',
+    '<div bar="baz" {{did-render this.ok}}></div>',
+    '<div {{did-render this.oks}} ...attributes aria-label="foo"></div>',
+    '<div {{did-render this.ok}} ...attributes></div>',
+    '<div ...attributes aria-label="foo"></div>',
+    '<div aria-label="foo"></div>',
+    '<MyComponent @value="5" @change={{this.foo}} local-class="foo" data-test-foo {{on "click" this.foo}} ...attributes as |sth|>content</MyComponent>',
+  ],
+
+  bad: [
+    {
+      template: '<div ...attributes @a="1"></div>',
+      result: {
+        message: 'Argument @a="1" must go before attributes, modifiers and splattributes',
+        source: '@a="1"',
+        line: 1,
+        column: 19,
+      },
+    },
+    {
+      template: '<div contenteditable @a="1"></div>',
+      result: {
+        message: 'Argument @a="1" must go before attributes, modifiers and splattributes',
+        source: '@a="1"',
+        line: 1,
+        column: 21,
+      },
+    },
+    {
+      template: '<div {{did-render this.someAction}} @a="1"></div>',
+      result: {
+        message: 'Argument @a="1" must go before attributes, modifiers and splattributes',
+        source: '@a="1"',
+        line: 1,
+        column: 36,
+      },
+    },
+    {
+      template: '<div ...attributes {{did-render this.someAction}}></div>',
+      result: {
+        message: 'Splattribute ...attributes must go after modifiers',
+        source: '...attributes',
+        line: 1,
+        column: 5,
+      },
+    },
+    {
+      template: '<div {{did-render this.someAction}} aria-label="button"></div>',
+      result: {
+        message: 'Modifier {{did-render this.someAction}} must go after attributes',
+        source: '{{did-render this.someAction}}',
+        line: 1,
+        column: 5,
+      },
+    },
+    {
+      template:
+        '<MyComponent @value="5" data-test-foo @change={{this.foo}} local-class="foo" {{on "click" this.foo}} ...attributes as |sth|>content</MyComponent>',
+      result: {
+        message:
+          'Argument @change={{this.foo}} must go before attributes, modifiers and splattributes',
+        source: '@change={{this.foo}}',
+        line: 1,
+        column: 38,
+      },
+    },
+  ],
+});
--- a/lib/rules/element-attributes-order.js
+++ b/lib/rules/element-attributes-order.js
@@ -45,7 +45,7 @@ export default class ElementAttributesOrder extends Rule {
     const sourceForNode = this.sourceForNode.bind(this);
     const config = this.config;
 
-    function extractElementTokens(node, nodeText) {
+    function getTokens(node, nodeText) {
       const keys = [];
       for (const attr of node.attributes) {
         const source = sourceForNode(attr);
@@ -63,22 +63,18 @@ export default class ElementAttributesOrder extends Rule {
       return keys;
     }
 
-    function combineElementTokens(tokens) {
-      const result = {
-        modifiers: [],
-        attributes: [],
-        splattributes: [],
-        arguments: [],
-      };
-      // const result = Object.values(AttributeType).reduce((acc, attribute) => {
-      // }
+    function groupTokens(tokens) {
+      const result = Object.values(AttributeType).reduce((acc, attributeType) => {
+        acc[attributeType] = [];
+        return acc;
+      }, {});
       for (let [name, index, source, node] of tokens) {
         result[name].push([index, source, node]);
       }
       return result;
     }
 
-    function addMetaToCombinedTokens(tokens) {
+    function addMetadata(tokens) {
       for (const tokenName of Object.keys(tokens)) {
         const indexes = tokens[tokenName].map(([index]) => Number(index));
         const firstLetters = tokens[tokenName].map(([, names]) =>
@@ -89,13 +85,19 @@ export default class ElementAttributesOrder extends Rule {
           exists: tokens[tokenName].length !== 0,
           last_index: Math.max(...indexes),
           first_index: Math.min(...indexes),
-          ordered: firstLetters.every((b, i, { [i - 1]: a }) => !i || a <= b),
+          unalphabetized_item_pos: firstLetters.findIndex(
+            (e, idx) => idx > 0 && e < firstLetters[idx - 1]
+          ),
         };
       }
       return tokens;
     }
 
-    function findBadItem(attributeType, nodeTokens) {
+    function findUnalphabetizedElement(attributeType, nodeTokens) {
+      return nodeTokens[attributeType].items[nodeTokens[attributeType].unalphabetized_item_pos];
+    }
+
+    function findCurrentElement(attributeType, nodeTokens) {
       return nodeTokens[attributeType].items.find(
         ([index]) => index === nodeTokens[attributeType].last_index
       );
@@ -105,17 +107,18 @@ export default class ElementAttributesOrder extends Rule {
       return string[0].toUpperCase() + string.slice(1, -1);
     }
 
-    function createErrorMessage(attributeType, source) {
+    function createNotAlphabetizedErrorMessage(attributeType, source) {
+      return `${capitalizedAttribute(attributeType)} ${source} is not alphabetized`;
+    }
+
+    function createUnorderedErrorMessage(attributeType, source) {
       const attributeOrder = config.attributeOrder.indexOf(attributeType);
       if (attributeOrder === 0) {
         const otherAttributes = `${config.attributeOrder[1]}, ${config.attributeOrder[2]} and ${config.attributeOrder[3]}`;
 
-        return `${capitalizedAttribute(
-          attributeType,
-          true
-        )} ${source} must go before ${otherAttributes}`;
+        return `${capitalizedAttribute(attributeType)} ${source} must go before ${otherAttributes}`;
       }
-      return `${capitalizedAttribute(attributeType, true)} ${source} must go after ${
+      return `${capitalizedAttribute(attributeType)} ${source} must go after ${
         config.attributeOrder[attributeOrder - 1]
       }`;
     }
@@ -157,15 +160,23 @@ export default class ElementAttributesOrder extends Rule {
       if (!sourceForNode(node)) {
         return;
       }
-      const nodeTokens = addMetaToCombinedTokens(
-        combineElementTokens(extractElementTokens(node, sourceForNode(node)))
-      );
+      const nodeTokens = addMetadata(groupTokens(getTokens(node, sourceForNode(node))));
 
       for (const attributeType of config.attributeOrder) {
-        if (isAttributeUnordered(attributeType, nodeTokens)) {
-          const [, source, item] = findBadItem(attributeType, nodeTokens);
+        if (config.alphabetize && nodeTokens[attributeType].unalphabetized_item_pos >= 0) {
+          const [, source, item] = findUnalphabetizedElement(attributeType, nodeTokens);
           log({
-            message: createErrorMessage(attributeType, source),
+            message: createNotAlphabetizedErrorMessage(attributeType, source),
+            line: item.loc.start.line,
+            column: item.loc.start.column,
+            source,
+            node,
+          });
+        }
+        if (isAttributeUnordered(attributeType, nodeTokens)) {
+          const [, source, item] = findCurrentElement(attributeType, nodeTokens);
+          log({
+            message: createUnorderedErrorMessage(attributeType, source),
             line: item.loc.start.line,
             column: item.loc.start.column,
             source,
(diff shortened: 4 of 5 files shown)


lifeart (approved):

lifeart:
@rob-long good job! Is a chance we also will get fixer for rule :) ?

test/unit/rules/element-attributes-order-test.js
```diff
@@ -0,0 +1,108 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
```
bmish, line 3: Can you add `error` test cases to ensure the rule does not allow invalid configuration? 

Example: https://github.com/ember-template-lint/ember-template-lint/blob/149a28067ec4db022397c63ef489250ece1db825/test/unit/rules/table-groups-test.js#L867

bmish:
1. Would it make sense for this rule to cover the older style mustache statements too in addition to newer style angle brackets? `{{my-component this.foo "bar" arg1="val" ...extra}}`. Generally our rules should still handle both styles so if it makes sense I prefer that it does. If so, we could call this rule `attributes-order`? If not, can you explain why not? Either way, we would still need to add test cases for mustache statements.
2. An autofixer is a definitely critical for this (current PR or quick follow-up PR).
3. Any thoughts on whether this would be a good candidate for our [stylistic](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/stylistic.js) config in the next major release or is it too opinionated even for that?
4. How well does this play with / agree with [ember-template-lint-plugin-prettier](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier)?

docs/rule/element-attributes-order.md
```diff
@@ -0,0 +1,49 @@
...
+- element modifiers
+- ...attributes
+- ?attributes
+
+## Configuration
+
+The following values are valid configuration:
+
+- attributeOrder: default [`arguments`, `attributes`, `modifiers`, `splattributes`]
+- alphabetize: default `false`, `true` to enforce alphabetically ordered values within attributes
```
bmish, line 16: What's the reasoning for default `false` for `alphabetize`? Just too noisy for most people?

test/unit/rules/element-attributes-order-test.js
```diff
@@ -0,0 +1,108 @@
...
+    '<div {{did-render this.ok}} ...attributes></div>',
+    '<div ...attributes aria-label="foo"></div>',
+    '<div aria-label="foo"></div>',
+    '<MyComponent @value="5" @change={{this.foo}} local-class="foo" data-test-foo {{on "click" this.foo}} ...attributes as |sth|>content</MyComponent>',
+  ],
+
+  bad: [
+    {
+      template: '<div ...attributes @a="1"></div>',
+      result: {
```
bmish, line 28: Can you please switch all these bad test cases to use:

```js
verifyResults(results) {
  expect(results).toMatchInlineSnapshot(...);
}
```

This is easier to maintain and we updated all our other rules to use this as well.

Info on how to switch: https://github.com/ember-template-lint/ember-template-lint/pull/2150

docs/rule/element-attributes-order.md
rob-long: That was my original thought but I think I can follow-up this PR with an auto-fixer which should address this concern.

rob-long:
> Thanks so much for working on this!
> 
> 1. Would it make sense for this rule to cover the older style mustache statements too in addition to newer style angle brackets? `{{my-component this.foo "bar" arg1="val" ...extra}}`. Generally our rules should still handle both styles so if it makes sense I prefer that it does. If so, we could call this rule `attributes-order`? If not, can you explain why not? Either way, we would still need to add test cases for mustache statements.
> 2. An autofixer is a definitely critical for this (current PR or quick follow-up PR).
> 3. Any thoughts on whether this would be a good candidate for our [stylistic](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/stylistic.js) config?
> 4. How well does this play with / agree with [ember-template-lint-plugin-prettier](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier)?

1. Updated!
2. I'll do a "quick" follow-up PR
3. Yes, I think so!
4. Didn't see any issues but I'm also not a big user of `ember-template-lint-plugin-prettier` so let me know if there is anything.

docs/rule/element-attributes-order.md
bmish: So you would prefer `true` as the default for `alphabetize` given that the autofixer will handle it? If that's the case, we need to use the ideal default now, as changing the default later is a breaking change.

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,310 @@
...
+              "severity": 2,
+              "source": "a=\\"1\\"",
+            },
+          ]
+        `);
+      },
+    },
+  ],
+
+  error: [
```
bmish, line 272: We should also have error test cases with invalid values for the actual options:

* `attributeOrder: ['arguments', 'NOT_AN_OPTION']`
* `alphabetize: 123`

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,229 @@
...
+      const nodeTokens = addMetadata(groupTokens(getMustacheTokens(node, sourceForNode(node))));
+      const attributeType = AttributeType.ATTRIBUTES;
+      if (config.alphabetize && nodeTokens[attributeType].unalphabetized_item_pos >= 0) {
+        const [, source, item] = findUnalphabetizedElement(attributeType, nodeTokens);
+        log({
+          message: createNotAlphabetizedErrorMessage(attributeType, source),
+          line: item.loc.start.line,
+          column: item.loc.start.column,
+          source,
+          node,
```
bmish, line 185: Shouldn't the `node` be the specific attribute?

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,229 @@
...
+          });
+        }
+        if (isAttributeUnordered(attributeType, nodeTokens)) {
+          const [, source, item] = findCurrentElement(attributeType, nodeTokens);
+          log({
+            message: createUnorderedErrorMessage(attributeType, source),
+            line: item.loc.start.line,
+            column: item.loc.start.column,
+            source,
+            node,
```
bmish, line 214: Shouldn't the `node` be the specific attribute?

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,229 @@
...
+
+      for (const attributeType of config.attributeOrder) {
+        if (config.alphabetize && nodeTokens[attributeType].unalphabetized_item_pos >= 0) {
+          const [, source, item] = findUnalphabetizedElement(attributeType, nodeTokens);
+          log({
+            message: createNotAlphabetizedErrorMessage(attributeType, source),
+            line: item.loc.start.line,
+            column: item.loc.start.column,
+            source,
+            node,
```
bmish, line 204: Shouldn't the `node` be the specific attribute?

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,310 @@
...
+              "severity": 2,
+              "source": "@change={{this.foo}}",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      config: {
+        attributeOrder: ['attributes', 'arguments', 'modifiers', 'splattributes'],
```
bmish, line 202: Need a test case for custom `attributeOrder` for mustache components too.

docs/rule/element-attributes-order.md
rob-long: Forgot to update the docs, `alphabetize` now defaults to `true`!

bmish (commented):
[ember-template-lint-plugin-prettier](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier) is popular so it would be appreciated if you could just verify that  it doesn't enforce anything about the attribute order that would conflict. CC: @dcyriller

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,310 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+const invalidConfigMessage = 'Your config does not match the allowed values';
```
bmish, line 3: Can you export this error message from the rule? Other rules do this with `ERROR_MESSAGE`.

rob-long:
> [ember-template-lint-plugin-prettier](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier) is popular so it would be appreciated if you could just verify that it doesn't enforce anything about the attribute order that would conflict. CC: @dcyriller

Tried on a project and did not run into conflicts.

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,231 @@
...
+            column: item.loc.start.column,
+            source,
+            node: item,
+          });
+        }
+        if (isAttributeUnordered(attributeType, nodeTokens)) {
+          const [, source, item] = findCurrentElement(attributeType, nodeTokens);
+          log({
+            message: createUnorderedErrorMessage(attributeType, source),
+            line: item.loc.start.line,
```
bmish, line 213: For all of these log statements: When you specify the current `node` to display the violation on, I don't think we should need to include `line`, `column`, `source`.

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,231 @@
...
+        keys.push([type, nodeText.indexOf(source), source, attr]);
+      }
+      for (const modifier of node.modifiers) {
+        const source = sourceForNode(modifier);
+        keys.push([AttributeType.MODIFIERS, nodeText.indexOf(source), source, modifier]);
+      }
+      return keys;
+    }
+
+    function groupTokens(tokens) {
```
bmish, line 74: Can you add jsdoc comments to these functions (at least the non-trivial ones) explaining what they are for and details about the parameters / return types?

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,231 @@
...
+            );
+          }
+          break;
+
+        default:
+          break;
+      }
+    }
+
+    function checkMustacheAttributesOrder(node) {
```
bmish, line 174: Can you add some high-level commenting to explain the general strategy for checking the order?

lib/helpers/is-valid-config-object.js
```diff
@@ -0,0 +1,34 @@
+export default function isValidConfigObjectFormat(config, POSSIBLE_CONFIG) {
```
bmish, line 1: Can we add some unit tests for this function (in a corresponding `is-valid-config-object-test.js` file alongside other helper unit tests)? Should be easy enough to test.

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
...
+
+For angle bracket invocation syntax:
+
+- arguments
+- attributes
+- element modifiers
+- ...attributes
+- ?attributes
+
+For mustache/curly invoication syntax:
```
bmish, line 13:
```suggestion
For mustache/curly invocation syntax:
```

bmish, line 13: So there's only one type (`attributes`) for mustache/curly? No other possible types? To make sure I'm understanding this correctly, that means the only affect of this rule on mustache/curly invocation is alphabetizing the attributes? If that's the case, maybe we could say that more clearly like this:

```md
Element attributes must be ordered. By default, the order is:

* ...
* ...
* ...
* ...

Note that this ordering only applies to angle bracket invocation syntax, as mustache/curly invocation syntax only includes a single type `attributes`.
```

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
+# attributes-order
+
+Element attributes must be ordered, by default the order is:
```
bmish, line 3:
```suggestion
Element attributes must be ordered. By default, the order is:
```

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
...
+- attributeOrder: default [`arguments`, `attributes`, `modifiers`, `splattributes`]
+  - [`attributes`] for curly invocation
+- alphabetize: default `true` - `true` to enforce alphabetically ordered values within attributes
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<MyComponent ...attributes @name='Hello' />
```
bmish, line 30: The position of `...attributes` is significant in that the other arguments will override or be overridden by anything in `...attributes` depending on whether it comes before or after, right? Is that the only type of attribute whose ordering is meaningful? If reordering any types of attributes can cause a behavior change, we should call that out in the rule doc.

bmish: One follow-up to this: assuming this is true, then there are two concerns I think we need to address:

* Should we just allow the spread to be in any position so the user can achieve their desired behavior? I.e. don't enforce spread ordering?
* An autofix shouldn't change the code behavior as that could cause breakages, so we need to make sure we don't move anything from one of the spread to the other.

rob-long: Autofix has been implemented in a way that the spread is left in the first position or last position while other attribute types are ordered per configuration.  When not in the first or last position, checking does not occur as ordering becomes exponentially more complicated.  Such ordering should be extremely rare in any case.

lib/helpers/is-valid-config-object.js
rob-long: Tests have been added!

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,59 @@
+# attributes-order
+
+Element attributes must be ordered. By default, the order is:
+
+- arguments
+- attributes
+- element modifiers
+- ...attributes
```
bmish, line 8: Splattributes shouldn't be mentioned in the ordering anymore right? Since we allow the user to place the splattributes wherever they like now from my understanding.

README.md
```diff
@@ -88,6 +88,7 @@ Each rule has emojis denoting:
 | Name                                                                                                      | ✅  | 💅  | ⌨️  | 🔧  |
 | :-------------------------------------------------------------------------------------------------------- | :-- | :-- | :-- | --- |
 | [attribute-indentation](./docs/rule/attribute-indentation.md)                                             |     |     |     |     |
+| [attributes-order](./docs/rule/attributes-order.md)                                                       |     |     |     |     |
```
bmish, line 91: `npm run update:readme`

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,59 @@
+# attributes-order
+
+Element attributes must be ordered. By default, the order is:
```
bmish, line 3: Missing fixable notice, see the documentation test output.

docs/rule/attributes-order.md
bmish: Thanks! As long as spread can be in any position, I'm happy. Can you update these examples? It seems these are outdated now.

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,663 @@
...
+        expect(results).toMatchInlineSnapshot(`
+          [
+            {
+              "column": 37,
+              "endColumn": 53,
+              "endLine": 1,
+              "filePath": "layout.hbs",
+              "isFixable": true,
+              "line": 1,
+              "message": "Attribute aria-label=\\"foo\\" must go before arguments, modifiers and splattributes",
```
bmish, line 53: The error message shouldn't mention splattributes anymore, should it?

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,663 @@
...
+              "severity": 2,
+              "source": "@foo=\\"1\\"",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      template: '<div ...attributes @a="1"></div>',
+      fixedTemplate: '<div @a="1" ...attributes></div>',
```
bmish, line 77: I'm confused as I thought we were allowing spread anywhere?

bmish:
Thanks for the continued work on this! Excited to get it in soon.

lib/helpers/is-valid-config-object.js
```diff
@@ -0,0 +1,34 @@
...
+  const validLiteralValues = [true, false, 'warn', 'error'];
+  if (validLiteralValues.includes(config)) {
+    return true;
+  }
+
+  if (config === null && typeof config !== 'undefined') {
```
seanCodes, line 7: Looks like the second condition might be unnecessary (since if config is `null` it’s guaranteed to not be `undefined` also)
```suggestion
  if (config === null) {
```

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,502 @@
...
+    const errorMessage = createAttributesOrderErrorMessage(config);
+    if (!isValidConfigObjectFormat(config, DEFAULT_CONFIG)) {
+      throw new Error(errorMessage);
+    }
+
+    if (config && typeof config === 'object') {
+      return { ...DEFAULT_CONFIG, ...config };
+    }
+
+    return DEFAULT_CONFIG;
```
seanCodes, line 39: I think the case of `config` being `false` needs to be handled here (like [here](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/attribute-indentation.js#L506)).

Also, the `config && typeof config === 'object'` could just be `typeof config === 'object')`:
```suggestion
    if (config === false) {
      return false;
    }

    if (typeof config === 'object') {
      return { ...DEFAULT_CONFIG, ...config };
    }

    return DEFAULT_CONFIG;
```

seanCodes (commented):
Great work on this! Thanks for carrying the torch and getting it done. 🙏🏽 

Just had some suggestions—feel free to ignore any of them!

rob-long:
> Great work on this! Thanks for carrying the torch and getting it done. 🙏🏽
> 
> Just had some suggestions—feel free to ignore any of them!

Thanks!  The suggestions were all great and have been included.

rob-long:
> @rob-long let us know when you think the PR is in good shape to get in. Note that there are pre-existing memory issues causing CI to fail that you can ignore.

@bmish  I ran the rule on a large code base and was happy with the result.  It's in good shape from my perspective!

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,506 @@
...
+        .filter((i) => Number.isInteger(i))
+    );
+
+    const attributeOrderMinusSplattributes = [
+      ...this.config.attributeOrder.filter(
+        (attribute) => attribute !== AttributeType.SPLATTRIBUTES
+      ),
+    ];
+
+    const splattributeLastIndex = nodeTokens[AttributeType.SPLATTRIBUTES]?.last_index;
```
bmish, line 91: CI is failing because `?.` isn't available yet since we still support Node 12. Can you fix that?

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
...
+The following values are valid configuration:
+
+- boolean -- `true` to enable / `false` to disable
+- object -- An object with the following keys:
+  - `attributeOrder` -- An array of attribute types (defaults to `['arguments', 'attributes', 'modifiers']`
+    - Note that curly component invocations will default to `['attributes']` since they don’t have other attribute types
+- `alphabetize` -- `true` to enforce ordering attributes alphabetically (defaults to `true`)
+
+Note that the position of `...attributes` (“splattributes”) is significant: attributes that appear _before_ `...attributes` can be overridden by values present in `...attributes` but attributes appearing _after_ `...attributes` will always take precedence and cannot be overridden by `...attributes`.
+For this reason, nodes which contain `...attributes` surrounded by other attribute types cannot be safely reordered without causing breakages. If `...attributes` is the first or last attribute of a node however, it will be left as-is and sorting will occur as normal.
```
bmish, line 55: > it will be left as-is and sorting will occur as normal.

This sounds contradictory to me. Does it get sorted or does it get left as-is? Can you clarify this?

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
...
+<MyComponent @name="1" bar="baz" />
+```
+
+## Configuration
+
+The following values are valid configuration:
+
+- boolean -- `true` to enable / `false` to disable
+- object -- An object with the following keys:
+  - `attributeOrder` -- An array of attribute types (defaults to `['arguments', 'attributes', 'modifiers']`
```
bmish, line 50: Should we just call this option `order`? Is there any benefit or need to be specific about what ordering we're talking about it here? Or is the type of ordering implicit from the rule name so we could just call it `order`? What do you think?

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,60 @@
...
+    - Note that curly component invocations will default to `['attributes']` since they don’t have other attribute types
+- `alphabetize` -- `true` to enforce ordering attributes alphabetically (defaults to `true`)
+
+Note that the position of `...attributes` (“splattributes”) is significant: attributes that appear _before_ `...attributes` can be overridden by values present in `...attributes` but attributes appearing _after_ `...attributes` will always take precedence and cannot be overridden by `...attributes`.
+For this reason, nodes which contain `...attributes` surrounded by other attribute types cannot be safely reordered without causing breakages. If `...attributes` is the first or last attribute of a node however, it will be left as-is and sorting will occur as normal.
+
+## References
+
+- [eslint-plugin-vue/rules/attributes-order](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/attributes-order.md)
+- [Ember guides on the `...attributes` syntax](https://guides.emberjs.com/release/components/component-arguments-and-html-attributes/#toc_html-attributes).)
```
bmish, line 60: errant parenthesis

```suggestion
- [Ember guides on the `...attributes` syntax](https://guides.emberjs.com/release/components/component-arguments-and-html-attributes/#toc_html-attributes)
```

test/unit/rules/attributes-order-test.js
```diff
@@ -0,0 +1,785 @@
...
+              "severity": 2,
+              "source": "@change={{this.foo}}",
+            },
+          ]
+        `);
+      },
+    },
+    {
+      config: {
+        attributeOrder: ['attributes'],
```
bmish, line 402: What happens when not all types are ordered? Does it make sense?

bmish (changes requested):
Really close! Just a few minor things.

docs/rule/attributes-order.md
rob-long: Simplifying to `order` makes sense given the rule name already gives the context.

docs/rule/attributes-order.md
rob-long: I added documentation around this that hopefully clarifies things.

test/unit/rules/attributes-order-test.js
rob-long: I've added a check to the configuration parsing to enforce that all types are included as it eliminates the need for guesswork.

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,74 @@
+# attributes-order
+
+Element attributes must be ordered. By default, the order is:
```
bmish, line 3:
```suggestion
# attributes-order

🔧 The `--fix` option on the command line can automatically fix some of the problems reported by this rule.

Element attributes must be ordered. By default, the order is:
```

docs/rule/attributes-order.md
```diff
@@ -0,0 +1,74 @@
...
+```
+Fixed version:
+```hbs
+<Component @value="5" @change={{this.foo}} data-test-foo local-class="foo" {{on "click" this.foo}} ...attributes as |sth|>content</Component>
+```
+
+A different example with splattributes where ordering can not occur automatically:
+```hbs
+<div bar="baz" {{did-render this.ok}} ...attributes label="foo"></div>
+```
```
bmish, line 69:
```suggestion

```hbs
<Component @value="5" data-test-foo @change={{this.foo}} local-class="foo" {{on "click" this.foo}} ...attributes as |sth|>content</Component>
```

Fixed version:

```hbs
<Component @value="5" @change={{this.foo}} data-test-foo local-class="foo" {{on "click" this.foo}} ...attributes as |sth|>content</Component>
```

A different example with splattributes where ordering can not occur automatically:

```hbs
<div bar="baz" {{did-render this.ok}} ...attributes label="foo"></div>
```
```

File: docs/rule/attributes-order.md [hunk 31]
```diff
@@ -0,0 +1,74 @@
...
+<MyComponent @name="1" bar="baz" />
+```
+
+## Configuration
+
+The following values are valid configuration:
+
+- boolean -- `true` to enable / `false` to disable
+- object -- An object with the following keys:
+  - `attributeOrder` -- An array of attribute types (defaults to `['arguments', 'attributes', 'modifiers']`
```
bmish, line 50:
```suggestion
  - `order` -- An array of attribute types (defaults to `['arguments', 'attributes', 'modifiers']`
```

test/unit/rules/attributes-order-test.js
bmish: Nice. Can you add an `error` test case for this?

bmish:
@rob-long can we also rename this rule to `attribute-order`? It sounds more natural to me, and is more consistent with the other rule we have `attribute-indentation`.

Issue #2571, opened 2022-08-04
Rules that need autofixers


bmish:
List of rules where it should be easy / practical / useful to implement an autofixer:

* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/linebreak-style.md (change the linebreak style)
* #2573
* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-quoteless-attributes.md (add quotes?)
* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-trailing-spaces.md (delete the trailing spaces)
* #2574
* #2568
* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-unused-block-params.md (delete the unused param)
* #2569
* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/self-closing-void-elements.md (remove the self-closing slash?)
* https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/simple-unless.md (some autofixer cases might be easier to implement than others and could be in separate PRs)
* Any others?

Many rules don't have autofixers because autofixing capability didn't exist when these rules were created.

This is a good "help wanted" / "first issue" for anyone who wants to help out and claim one or more of these.

PR #2580, opened 2022-08-11
Add autofixer to `no-unused-block-params` rule

locks:
Part of https://github.com/ember-template-lint/ember-template-lint/issues/2571. [no-unused-block-params](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-unused-block-params.md)

I'm having a bit of trouble with this one.

- for nested `{{each}}`, visitor only seems to trigger for innermost one?
- haven't figured out how not to remove block params if the params list would be empty
- haven't figuredd out how to remove block params completely for self-closing component tag


bmish:
Still interested in this!

PR #2582, opened 2022-08-11
Add autofixer to `no-quoteless-attributes` rule

locks:
Part of https://github.com/ember-template-lint/ember-template-lint/issues/2571. [no-quoteless-attributes](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-quoteless-attributes.md)

I switched over from TextNode to AttrNode because it feels cleaner reach into the node than up the parent.


bmish:
Rule doc notice needs to be added.

bmish (approved):
Thanks!

PR #2583, opened 2022-08-12
Add autofixer to `self-closing-void-elements` rule

locks:
Part of https://github.com/ember-template-lint/ember-template-lint/issues/2571. [self-closing-void-elements](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/self-closing-void-elements.md)

### DIFF
--- a/lib/rules/self-closing-void-elements.js
+++ b/lib/rules/self-closing-void-elements.js
@@ -1,3 +1,5 @@
+import { builders } from 'ember-template-recast';
+
 import createErrorMessage from '../helpers/create-error-message.js';
 import Rule from './_base.js';
 
@@ -83,16 +85,30 @@ export default class SelfClosingVoidElements extends Rule {
               ? `${source.slice(0, -1)}/>`
               : `${source.slice(0, -2)}>`;
 
-            this.log({
-              message: isSelfClosingRequired
-                ? 'Self-closing a void element is required'
-                : 'Self-closing a void element is redundant',
-              node,
-              source,
-              fix: {
-                text: expected,
-              },
-            });
+            if (this.mode === 'fix') {
+              let editedNode = builders.element(node.tag, {
+                attrs: node.attributes,
+                modifiers: node.modifiers,
+                children: node.children,
+                comments: node.comments,
+                blockParams: node.blockParams,
+                loc: node.loc,
+              });
+              editedNode.selfClosing = isSelfClosingRequired;
+              return editedNode;
+            } else {
+              this.log({
+                message: isSelfClosingRequired
+                  ? 'Self-closing a void element is required'
+                  : 'Self-closing a void element is redundant',
+                node,
+                source,
+                fix: {
+                  text: expected,
+                },
+                isFixable: true,
+              });
+            }
           }
         }
       },
--- a/test/unit/rules/self-closing-void-elements-test.js
+++ b/test/unit/rules/self-closing-void-elements-test.js
@@ -938,5 +938,32 @@ generateRuleTests({
         `);
       },
     },
+    {
+      template: 'foo<wbr data-custom="50" {{my-modifier true "baz"}} as |paramA paramB|>bar',
+      fixedTemplate: 'foo<wbr data-custom="50" {{my-modifier true "baz"}} as |paramA paramB| />bar',
+      config: 'require',
+
+      verifyResults(results) {
+        expect(results).toMatchInlineSnapshot(`
+          [
+            {
+              "column": 3,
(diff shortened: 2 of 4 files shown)


lib/rules/self-closing-void-elements.js
```diff
@@ -83,16 +85,30 @@ export default class SelfClosingVoidElements extends Rule {
...
-                ? 'Self-closing a void element is required'
-                : 'Self-closing a void element is redundant',
-              node,
-              source,
-              fix: {
-                text: expected,
-              },
-            });
+            if (this.mode === 'fix') {
+              let editedNode = builders.element(node.tag, {
```
bmish, line 89: Are you sure we can't use the existing node instead of creating a new one? Wouldn't that be easier? And then we wouldn't have to worry about remembering to pass all these arguments to it.

locks: I tried setting `selfClosing` but it wasn't doing anything. We could also do `{ ...node, attrs: node.attributes }`, but I didn't want to pass too much, in case the API adds options.

bmish: Hmm, okay, that's weird, I'm not sure how we can trigger it to update after just modifying that boolean property.

So we can stick with your current approach, but we need to make sure all these arguments are tested then. So you'll need to update at least one test case to add all these arguments in the `template` and make sure that they are copied over to the `fixedTemplate`.

locks: So I was working a bit more on this and I'm wondering. Attributes and modifiers are fine, but self-closing tags can't really have block params or children right? They're a leaf node.

bmish: My guess is you don't need to worry about that as only the relevant data will be rendered depending on how `selfClosing` is set. So try just copying everything over to the new node.

bmish: Looks like all we need to finish this PR is to add some test cases that demonstrate everything is copied from the old node to new node correctly.

test/unit/rules/self-closing-void-elements-test.js
```diff
@@ -874,5 +938,32 @@ generateRuleTests({
         `);
       },
     },
+    {
+      template: 'foo<wbr data-custom="50" {{my-modifier true "baz"}} as |paramA paramB|>bar',
```
bmish, line 942: Can we make sure that comments and children are also tested?
```hbs
foo<div data-custom="50" {{my-modifier true "baz"}} as |paramA paramB| {{!comment}}>child</div>bar
```

test/unit/rules/self-closing-void-elements-test.js
```diff
@@ -874,5 +938,32 @@ generateRuleTests({
...
+              "line": 1,
+              "message": "Self-closing a void element is required",
+              "rule": "self-closing-void-elements",
+              "severity": 2,
+              "source": "<wbr data-custom=\\"50\\" {{my-modifier true \\"baz\\"}} as |paramA paramB|>",
+            },
+          ]
+        `);
+      },
+    },
```
bmish, line 967:
```suggestion
    {
      template:
        'foo<wbr data-custom="50" {{my-modifier true "baz"}} {{!comment}} as |paramA paramB| >bar',
      fixedTemplate:
        'foo<wbr data-custom="50" {{my-modifier true "baz"}} {{!comment}} as |paramA paramB| />bar',
      config: 'require',

      verifyResults(results) {
        expect(results).toMatchInlineSnapshot(`
          [
            {
              "column": 3,
              "endColumn": 85,
              "endLine": 1,
              "filePath": "layout.hbs",
              "fix": {
                "text": "<wbr data-custom=\\"50\\" {{my-modifier true \\"baz\\"}} {{!comment}} as |paramA paramB| />",
              },
              "isFixable": true,
              "line": 1,
              "message": "Self-closing a void element is required",
              "rule": "self-closing-void-elements",
              "severity": 2,
              "source": "<wbr data-custom=\\"50\\" {{my-modifier true \\"baz\\"}} {{!comment}} as |paramA paramB| >",
            },
          ]
        `);
      },
    },
```

test/unit/rules/self-closing-void-elements-test.js
bmish: Actually it doesn't make sense to test children inside a void element so I just added the comment part instead: https://github.com/ember-template-lint/ember-template-lint/pull/2583#discussion_r952816841

bmish (approved):
Thanks!

Issue #2588, opened 2022-08-23
<s> and <u> are not obsolete anymore (no-obsolete-elements)


sukima:
As far as I can tell from the [WATWG HTML spec](https://html.spec.whatwg.org/#non-conforming-features)—first link in the [referenced documentation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-obsolete-elements.md)—they explain that “Elements in [non-conforming features] list are entirely obsolete, and must not be used by authors” for which the only reference to `<s>` is the following:

> **strike**
> Use [`del`](https://html.spec.whatwg.org/#the-del-element) instead if the element is marking an edit, otherwise use [`s`](https://html.spec.whatwg.org/#the-s-element) instead.

I have been unable to find any reference to the deprecation of the `<s>` element.

bmish:
It was added here: https://github.com/ember-template-lint/ember-template-lint/pull/1532

The link from that PR seems to be broken. Need to see where the page was moved to so we can check the original source for this.

sukima:
Seems per that PR that it is obsolete but there appears to be no record of that and as noted above the current living spec only references it as an alternative. However, `<u>` is also not there in the spec. So how do we know it is obsolete and what is the alternative.

sukima:
@bmish Best I can tell it was moved here: https://html.spec.whatwg.org/#non-conforming-features but reading through that has no mention to `<b>`, `<u>`, or `<s>`. It does say `<s>` replaced `<strike>` which is even _more_ confusing.

sukima:
I found this [obscure reference](https://www.linkedin.com/pulse/obsolete-deprecated-tags-html5-shadi-namrouti):

> `<s>` Was similar to the deprecated `<strike>`. It is redefined as non-accurate text.

bmish:
I found an archived version of the original source showing `<s>` and `<u>` in the list of `Non-conforming features - Elements in the following list are entirely obsolete, and must not be used by authors - Use CSS instead`.

https://web.archive.org/web/20220407040721/https://dev.w3.org/html5/pf-summary/obsolete.html

> 13.2 Non-conforming features
> Elements in the following list are entirely obsolete, and must not be used by authors:
> 
> [applet](https://web.archive.org/web/20220407040721/https://dev.w3.org/html5/pf-summary/obsolete.html#the-applet-element)
[...]

PR #2589, opened 2022-08-23
Remove <s> from obsolete elements list

sukima:
According to the linked WATWG Specs for HTML5 the <s> element is not obsolete or non-conforming. It instead is referenced as an alternative to <strike> which is non-conforming.

https://html.spec.whatwg.org/#non-conforming-features

Fixes issue #2588


bmish (approved):
Thanks!

PR #2591, opened 2022-08-26
Add Kakoune Formatter and fix `--no-ignore-pattern` cli argument

velrest:
- feat(formatters): add kakoune: Add a formatter for the code editor [Kakoune](https://kakoune.org/)
- fix(cli): fix no-ignore-pattern parameter


lib/helpers/cli.js
```diff
@@ -170,7 +170,11 @@ export function expandFileGlobs(workingDir, filePatterns, ignorePattern, glob =
     let isLiteralPath = !isGlob(pattern) && isFile(path.resolve(workingDir, pattern));
 
     if (isLiteralPath) {
-      let isIgnored = micromatch.isMatch(pattern, ignorePattern);
+      // If `--no-ignore-pattern` is passed, the ignorePatter is `[false]`.
+      let isIgnored =
+        ignorePattern?.length !== 1 &&
+        ignorePattern?.[0] !== false &&
+        micromatch.isMatch(pattern, ignorePattern);
```
bmish, line 177: If this is a bug, can we extract it to a separate PR, and add a test for it?

package.json
```diff
@@ -70,6 +70,7 @@
     "ci-info": "^3.3.2",
     "date-fns": "^2.29.1",
     "ember-template-recast": "^6.1.3",
+    "eslint-formatter-kakoune": "Delapouite/eslint-formatter-kakoune#01f7697",
```
bmish, line 73: What's the fork for?

lib/formatters/kakoune.js
```diff
@@ -0,0 +1,39 @@
...
+      filteredErrors[filePath] = [...errorsFiltered, ...warnings];
+    }
+
+    results = Object.entries(filteredErrors).map(([filePath, messages]) => {
+      return {
+        filePath,
+        messages: [...messages].map((message) => {
+          return {
+            ...message,
+            // For some reason the column is 0 if the error occurs on the first line and this breaks kakoune
```
bmish, line 30: Will it successfully highlight the violation in the correct location or off-by-one?

package.json
velrest: Its not a fork, its the upstream repo.

lib/formatters/kakoune.js
velrest: I`ll test this again.

lib/helpers/cli.js
velrest: Fair point, will do.

package.json
velrest: I dont think it will be published as the repo is 6 years old. I will open an issue and ask if they plan on releasing a package.

bmish: Should we just copy this [file](https://github.com/Delapouite/eslint-formatter-kakoune/blob/master/index.js) in with a comment about where it came from? Normally I wouldn't want to copy in external code like that but it seems like this could be better or safer than using a commit as a dependency?

velrest: Probably, i think in this case it makes more sense to do that. I'll copy the file as well as test it again.

velrest: With eslint, you can pass a repostitory to `--formater ./eslint-formatter-kakoune`. Do you know why template lint uses a different object format than eslint? This patch wouldn't be needed.

lib/helpers/cli.js
velrest: Bugfix pr: https://github.com/ember-template-lint/ember-template-lint/pull/2725

velrest: Sorry for the long wait!

package.json
velrest: Formatter has been released, updated package.json.

lib/formatters/kakoune.js
velrest: This works as expected

velrest:
@bmish i've updated the code according to your reviews

lib/formatters/load-formatter.js
```diff
@@ -4,6 +4,7 @@ import path from 'node:path';
 import JsonFormatter from './json.js';
 // The following disable should be safe. This particular rule does not need to identify
 // cycles that are broken when using dynamic imports. See https://github.com/import-js/eslint-plugin-import/issues/2265
+import KakouneFormatter from './kakoune.js';
```
bmish, line 7: The comment is above the wrong line now.

```suggestion
import KakouneFormatter from './kakoune.js';
// The following disable should be safe. This particular rule does not need to identify
// cycles that are broken when using dynamic imports. See https://github.com/import-js/eslint-plugin-import/issues/2265
```

lib/formatters/kakoune.js
bmish: Can you update or remove the comment then assuming it works properly?

test/acceptance/formatters/kakoune-test.js
```diff
@@ -0,0 +1,217 @@
...
+      app/templates/application.hbs:1:24: Error: Non-translated string used (no-bare-strings) 
+      app/templates/application.hbs:1:53: Warning: HTML comment detected (no-html-comments) 
+
+      3 problems"
+    `);
+    expect(result.stderr).toBeFalsy();
+  });
+
+  describe('with --quiet option', function () {
+    it('should print valid JSON string with errors, omitting warnings', async function () {
```
bmish, line 167: 1. Where is the JSON string?
2. Which rule in this test had warning violations that were silenced?

test/acceptance/formatters/kakoune-test.js
```diff
@@ -0,0 +1,217 @@
...
+      app/templates/application.hbs:1:24: Error: Non-translated string used (no-bare-strings) 
+      app/templates/application.hbs:1:53: Warning: HTML comment detected (no-html-comments) 
+
+      3 problems"
+    `);
+
+    expect(result.stderr).toBeFalsy();
+  });
+
+  it('should include information about available fixes', async function () {
```
bmish, line 74: I don't see any output about fix information?

lib/formatters/kakoune.js
velrest: Done

lib/formatters/load-formatter.js
velrest: Done

test/acceptance/formatters/kakoune-test.js
velrest: The test was originally from the json test. Is fixed now.

test/acceptance/formatters/kakoune-test.js
velrest: Kakoune doesn't support fix information so i removed the test.

bmish (approved):
Thanks!

Issue #2592, opened 2022-08-28
Usage of `unique-id` helper causes rule triggers


geneukum:
In Ember v4.4.0, a [`unique-id` helper](https://api.emberjs.com/ember/4.6/classes/Ember.Templates.helpers/methods/unique-id?anchor=unique-id) was added which generates a unique ID string suitable for use as an ID attribute in the DOM. When this helper is used as: `<p>{{unique-id}}</p>`, both the no-implicit-this and no-curly-component-invocation rules flag it as an issue.

This should be a perfectly legal usage of this helper as far as those two rules are concerned, so I think that we should update the allowed-lists in each.

PR #2593, opened 2022-08-28
Add `unique-id` to allowed built-in helpers in `no-curly-component-invocation` and `no-implicit-this` rules

geneukum:
In Ember v4.4.0, a [`unique-id` helper](https://api.emberjs.com/ember/4.6/classes/Ember.Templates.helpers/methods/unique-id?anchor=unique-id) was added which generates a unique ID string suitable for use as an ID attribute in the DOM. When this helper is used as: `<p>{{unique-id}}</p>`, both the no-implicit-this and no-curly-component-invocation rules flag it as an issue.

Let's update the allowed-lists in both rules such that they no longer flag this usage.

### DIFF
--- a/lib/rules/no-curly-component-invocation.js
+++ b/lib/rules/no-curly-component-invocation.js
@@ -39,6 +39,7 @@ const BUILT_INS = new Set([
   'query-params',
   'textarea',
   'unbound',
+  'unique-id',
   'unless',
   'with',
   '-in-element',
--- a/lib/rules/no-implicit-this.js
+++ b/lib/rules/no-implicit-this.js
@@ -41,6 +41,7 @@ export const ARGLESS_BUILTIN_HELPERS = [
   'query-params',
   'textarea',
   'yield',
+  'unique-id',
 ];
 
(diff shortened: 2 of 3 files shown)


bertdeblock:
Could also add the helper [here](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/rules/no-curly-component-invocation-test.js#L30-L40)? The `no-implicit-this` rule test generates these good cases dynamically.

geneukum:
> Could also add the helper [here](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/rules/no-curly-component-invocation-test.js#L30-L40)? The `no-implicit-this` rule test generates these good cases dynamically.

Great catch, @bertdeblock! I fixed-up the commit which originally handled the rule modification to also include this test modification.

bmish (approved):
Thanks!

bmish:
Remember to include "Fixes https://github.com/ember-template-lint/ember-template-lint/issues/2592" in your PR description so that the linked issue will be resolved with the PR. (I added it for you afterward).

PR #2629, opened 2022-10-09
Add autofixer to `simple-unless` rule for simplest case

godric3:
Part of https://github.com/ember-template-lint/ember-template-lint/issues/2571.

For start auto fix for simplest case of:
```hbs
{{#unless condition}}
  A
{{else}}
  B
{{/unless}}
```
fixed to
```hbs
{{#if condition}}
  B
{{else}}
  A
{{/if}}
```
For other cases I will need to dive more into ast manipulations


bmish (approved):
Awesome, thanks!

PR #2639, opened 2022-10-17
Add autofix for case with helpers to `simple-unless` rule

godric3:
Part of https://github.com/ember-template-lint/ember-template-lint/issues/2571.

Replace `unless` with helpers with `if not` with helpers.
I'm not sure if result is good enough to be auto fixed, any ideas to improve that are welcome.

### DIFF
--- a/test/unit/rules/simple-unless-test.js
+++ b/test/unit/rules/simple-unless-test.js
@@ -71,6 +71,7 @@ generateRuleTests({
         maxHelpers: 2,
       },
       template: "{{unless (if (or true))  'Please no'}}",
+      fixedTemplate: "{{if (not (if (or true)))  'Please no'}}",
 
       verifyResults(results) {
         expect(results).toMatchInlineSnapshot(`
@@ -80,7 +81,7 @@ generateRuleTests({
               "endColumn": 23,
               "endLine": 1,
               "filePath": "layout.hbs",
-              "isFixable": false,
+              "isFixable": true,
               "line": 1,
               "message": "Using {{unless}} in combination with other helpers should be avoided. Allowed helpers: or,eq,not-eq",
               "rule": "simple-unless",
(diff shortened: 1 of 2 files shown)


test/unit/rules/simple-unless-test.js
```diff
@@ -372,6 +380,11 @@ generateRuleTests({
         '  I think I am a brown stick',
         '{{/unless}}',
       ].join('\n'),
+      fixedTemplate: [
+        '{{#if (not (not isBrown isSticky))}}',
```
godric3, line 384: should we detect such cases and simplify them?

bmish:
I was never too big of a fan of how this `simple-unless` rules disallows helpers with `unless` entirely. In many of my codebases at work, we have disabled the helpers check entirely with `'simple-unless': { maxHelpers: -1 }`. Might be worth re-evaluating the default for this.

godric3:
> Let's make sure this rule is compatible with [no-negated-condition](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-negated-condition.md). That means they don't fight with each other, and the end result of autofixing from both of them is desirable.

I believe that those are compatible as simple-unless will create `if not` construct only if `unless` was used with some helper. In such case no-negated-condition won't autofix because it doesn't want to create `unless `+ helper. Maybe in case of some more complex configuration of allowed/restricted helpers 🤔

test/unit/rules/simple-unless-test.js
godric3: Ok `(not (not` chain will be skipped, but if original not have more than one param it will be replaced with `or` helper.
Will also take a look into handling not not in no-negated-conditions

godric3: related to no-negated-conditions, what should happen with such code?
```hbs
{{#if (not c1 c2)}}{{/if}}
```
If I'm looking corrently it may be broken and result in 
```hbs
{{#unless c1}}{{/unless}}
```

bmish: @godric3 good catch, can you file that as a bug? Would this be correct and desirable for that?

```hbs
{{#unless c1 c2}}{{/unless}}
```

If there's no good improvement, we could also just not flag a violation in a situation like this.

bmish (approved):
Thanks!

PR #2648, opened 2022-10-18
Allow negation with multiple parameters in `no-negated-condition` rule

### LINKED ISSUE #2647: `no-negated-condition` rule wrongly marks and autofix case when `not` have multiple parameters
```hbs
{{#if (not c1 c2)}}{{/if}}
```
is autofixed into
```hbs
{{#unlesss c1}}{{/unless}}
```
completely skipping all parameters after first one.
I think autofixing that into simple thing is not possible because it would require `NOR` helper, so this should at least not be marked as violation.

Didn't tested that on real codebase but only in rule tests.

godric3:
Fixes #2647

### DIFF
--- a/lib/rules/no-negated-condition.js
+++ b/lib/rules/no-negated-condition.js
@@ -50,6 +50,11 @@ export default class NoNegatedCondition extends Rule {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


bmish (approved):
Thanks!

PR #2653, opened 2022-10-18
Add `simplifyHelpers` option in `no-negated-condition` rule

### LINKED ISSUE #2644: no-negated-condition rule could detect `(not (not ...))`
> However, I think this should also be a new violation detected and autofixed by `no-negated-condition` (in the next major release since adding new new violations to existing rules is a breaking change). If interested, you could open a PR for that.

_Originally posted by @bmish in https://github.com/ember-template-lint/ember-template-lint/pull/2639#discussion_r997009538_

Would need to be part of a major release (https://github.com/ember-template-lint/ember-template-lint/issues/2319).

### LINKED ISSUE #2655: `no-negated-condition` rule could detect and fix `(not (eq...` and others
Negation connected with some other helpers could be simplified:
- `(not (eq ...))` -> `(not-eq ...)`
- `(not (not-eq ...))` -> `(eq ...)`
- `(not (gt ...))` -> `(lte ...)`
- `(not (gte ...))` -> `(lt ...)`
- `(not (lt ...))` -> `(gte ...)`
- `(not (lte ...))` -> `(gt ...)`

Cases with eq and not-eq looks harmless to autofix, but rest may not always be desirable so it could be enabled with some setting.

godric3:
Fixes #2644 
Fixes #2655 
Add new violation for double negation


test/unit/rules/no-negated-condition-test.js
```diff
@@ -221,6 +266,31 @@ generateRuleTests({
         `);
       },
     },
+    {
+      template:
+        '{{#unless (not (not condition))}}<img>{{else if (not (not condition))}}<input>{{/unless}}',
+      fixedTemplate:
+        '{{#unless condition}}<img>{{else if (not (not condition))}}<input>{{/unless}}',
```
godric3, line 273: Should we keep this rule principle to not touch `{{else if` blocks, or try to fix double negation also in this case?

bmish: I think we should touch `else if` too. Let's fix the double negation here.

lib/rules/no-negated-condition.js
```diff
@@ -59,15 +63,30 @@ export default class NoNegatedCondition extends Rule {
       const isIfElseBlockStatement = node.type === 'BlockStatement' && node.inverse;
       const isIfElseNonBlockStatement = node.type !== 'BlockStatement' && node.params.length === 3;
       const shouldFlipCondition = isIfElseBlockStatement || isIfElseNonBlockStatement;
-      const message = isUnless
+      const message = doubleNegation
```
bmish, line 66: Rename to `hasDoubleNegation` to be consistent with other boolean variable names.

lib/rules/no-negated-condition.js
```diff
@@ -59,15 +63,30 @@ export default class NoNegatedCondition extends Rule {
...
         ? ERROR_MESSAGE_USE_IF
         : shouldFlipCondition
         ? ERROR_MESSAGE_FLIP_IF
         : ERROR_MESSAGE_USE_UNLESS;
 
+      this.log({
+        message,
+        node,
+        isFixable: true,
+      });
```
bmish, line 78: Normally logging should not happen during a fix. Should be written like this:

```js
if (this.mode === 'fix') {
  ...
} else {
  this.log(...)
}
```

lib/rules/no-negated-condition.js
```diff
@@ -5,6 +5,7 @@ const ERROR_MESSAGE_FLIP_IF =
   'Change `{{if (not condition)}} {{prop1}} {{else}} {{prop2}} {{/if}}` to `{{if condition}} {{prop2}} {{else}} {{prop1}} {{/if}}`.';
 const ERROR_MESSAGE_USE_IF = 'Change `unless (not condition)` to `if condition`.';
 const ERROR_MESSAGE_USE_UNLESS = 'Change `if (not condition)` to `unless condition`.';
+const ERROR_MESSAGE_DOUBLE_NEGATION = 'Remove double negation';
```
bmish, line 8:
```suggestion
const ERROR_MESSAGE_DOUBLE_NEGATION = 'Remove unnecessary double negation.';
```

godric3:
I've included helpers mentioned in #2655 and introduced `simplifyHelpers` setting for this rule that specify if we want to have all that simplification of helpers (including double negation), defults to false. Other name for that setting could be `allowNegatedHelpers` (with true as default), not sure which would fit better.

docs/rule/no-negated-condition.md
```diff
@@ -81,6 +81,14 @@ And similar examples with non-block forms like:
 {{input class=(if condition "some-class" "other-class")}}
 ```
 
+## Configuration
+
+
+* boolean -- `true` for enabled / `false` for disabled
+* object --
+  * `simplifyHelpers` -- boolean - simplify negated helpers like `(not (eq ...))` or `(not (gt ...))`, `true` to enable / `false` to disable (default)
```
bmish, line 90:
```suggestion
  * `simplifyHelpers` -- boolean - whether to flag and autofix negated helpers like `(not (eq ...))` or `(not (gt ...))` that can be simplified (default `false`) (TODO: enable by default in next major release)
```

lib/rules/no-negated-condition.js
```diff
@@ -1,24 +1,144 @@
...
+      case 'undefined':
+        return false;
+    }
+
+    let errorMessage = createErrorMessage(
+      this.ruleName,
+      [
+        '  * boolean -- `true` for enabled / `false` for disabled\n' +
+          '  * object --\n' +
+          '    *  `simplifyHelpers` -- boolean - simplify negated helpers like `(not (eq ..))` or `(not (gt ...))`, `true` to enable / `false` to disable (default)',
```
bmish, line 50: Update to match suggestion here: https://github.com/ember-template-lint/ember-template-lint/pull/2653/files#r1007167436

test/unit/rules/no-negated-condition-test.js
```diff
@@ -3,7 +3,7 @@ import generateRuleTests from '../../helpers/rule-test-harness.js';
 generateRuleTests({
   name: 'no-negated-condition',
 
-  config: true,
+  config: { simplifyHelpers: true },
```
bmish, line 6:
```suggestion
  config: true,
```

Let's just test the default behavior here. Then in individual test cases, we can test the explicit option behavior from: `{ simplifyHelpers: true }` and `{ simplifyHelpers: false }`.

bmish (approved):

bmish:
Thanks!

PR #2658, opened 2022-10-25
feat: utilize new contents data from ember-template-imports

gabrielcsapo:
# Summary

in https://github.com/ember-template-imports/ember-template-imports/pull/74 we added the ability to get the contents from the parsed template code so we can utilize that instead of slicing data.

### DIFF
--- a/lib/extract-templates.js
+++ b/lib/extract-templates.js
@@ -65,11 +65,9 @@ export function extractTemplates(moduleSource, relativePath) {
     let { start, end } = templateInfo;
     let templateStart = start.index + start[0].length;
 
-    let template = moduleSource.slice(templateStart, end.index);
-
     return makeTemplateInfo(
       coordinatesOf(moduleSource, templateStart, end.index),
-      template,
+      templateInfo.contents,
       templateInfo,
       true
     );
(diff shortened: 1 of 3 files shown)


gabrielcsapo:
@rwjblue i can update the CI version to the right version for 14 so tests can pass, so you want me to do that in this PR or another one and rebase after that is merged?

rwjblue:
Ya, lets throw a second PR up for bumping the min 14 version

bmish:
Yeah, a separate PR for dropping old Node versions would be good. We have a big major release planned in #2319. Would like to get this out soon.

bmish:
Is this a breaking change? If so, let's include it in the upcoming v5 release (#2319).

rwjblue:
> Is this a breaking change? 

Nope! Just some nice API improvements by recent `ember-template-imports` releases.

bmish:
Okay, well it depends on Node 12/17 being dropped right? So let's re-target to `next` and merge it.

rwjblue:
> well it depends on Node 12/17 being dropped right?

Nope! The comments RE: Node 12 were addressed by https://github.com/ember-template-lint/ember-template-lint/pull/2659 (eslint-plugin-import-helpers added an engines config that excluded Node 12, but since CI was generally red anyways we didn't catch it before merging the original PR updating it).

PR #2694, opened 2022-11-08
Use `npx` in stale todo command suggestion

bertdeblock:
We do the same for `npx @lint-todo/migrator`:
https://github.com/ember-template-lint/ember-template-lint/blob/master/bin/ember-template-lint.js#L127

We also mention in the README that it's intended to install `ember-template-lint` locally within the project:
https://github.com/ember-template-lint/ember-template-lint#usage

This probably means that most users don't have `ember-template-lint` installed globally and that copy-pasting and using the current suggested command will fail with: `command not found: ember-template-lint`.

### DIFF
--- a/docs/todos.md
+++ b/docs/todos.md
@@ -29,7 +29,7 @@ ember-template-lint . --include-todo
 If an error is fixed manually, `ember-template-lint` will let you know that there's a stale todo file. You'll see this error:
 
 ```bash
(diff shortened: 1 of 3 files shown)


bmish (approved):

bmish:
I typically just prefix any command like this with `yarn` before running it. But I guess this works too.

rwjblue (approved):

PR #2695, opened 2022-11-08
Automate extracting/replacing bare strings

netes:


netes:
Sorry, added by mistake.

PR #2696, opened 2022-11-08
Failing test cases for `--fix` with `.gjs`/`.gts` files

rwjblue:
Failing test cases for https://github.com/ember-template-lint/ember-template-lint/issues/2681

PR #2697, opened 2022-11-09
fix: test cases to show that fixer does not work with gjs and gts files

gabrielcsapo:
# Summary

Adds the fix to the issue showcased in https://github.com/ember-template-lint/ember-template-lint/pull/2696

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -221,12 +221,14 @@ export default class Linter {
   async _applyFixes(options, results) {
     let fixableIssues = results.filter((r) => r.isFixable);
 
+    this.console.log(options, fixableIssues);
     // nothing to do, bail out
     if (fixableIssues.length === 0) {
       return options.source;
     }
 
     let currentSource = options.source;
+    this.console.log(options);
     let fileConfig = this._moduleStatusCache.getConfigForFile(options);
 
     let ruleNames = new Set(fixableIssues.map((r) => r.rule));
@@ -245,7 +247,20 @@ export default class Linter {
         let visitor = await rule.getVisitor();
         let { code } = transform(templateInfo.template, () => visitor);
 
-        currentSource = currentSource.replace(templateInfo.template, code);
+        if (code !== templateInfo.template) {
+          currentSource =
+            currentSource.slice(0, templateInfo.start) +
+            code +
+            currentSource.slice(templateInfo.end, templateInfo.template.length);
+
+          return await this._applyFixes(
+            {
+              ...options,
+              source: currentSource,
+            },
+            results
+          );
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


lib/linter.js
```diff
@@ -230,17 +230,23 @@ export default class Linter {
...
+          shouldFix: true,
+          filePath: options.filePath,
+          rawSource: templateInfo.template,
+          fileConfig,
+        });
+
+        let visitor = await rule.getVisitor();
+        let { code } = transform(templateInfo.template, () => visitor);
+
+        currentSource = currentSource.replace(templateInfo.template, code);
```
rwjblue, line 248: Can you add a test case for:

```js
// this one shouldn't be touched by a `--fix`
const bar = <template>{{! <button></button> }}</template>;
// this should get `--fix`'ed
const foo = <template><button></button></template>;
```

I think the current `string.replace` system will replace both `bar` and `foo` there. I believe that we need to manually splice the string contents, something like:

```js
let currentSource = `
const bar = <template>{{! <button></button> }}</template>;
const foo = <template><button></button></template>;`
// -> '\nconst bar = <template>{{! <button></button> }}</template>;\nconst foo = <template><button></button></template>;'

currentSource = currentSource.slice(0, 82) + `<button type="submit"></button>` + currentSource.slice(99)
// -> '\nconst bar = <template>{{! <button></button> }}</template>;\nconst foo = <template><button type="submit"></button></template>;'
```

test/unit/rules/require-button-type-test.js
```diff
@@ -142,18 +140,19 @@ generateRuleTests({
       fixedTemplate: '<form><button type="submit"></button></form>',
     },
     {
-      template: '/** silly example <button> usage */ <button><button>',
+      template: '<template>/** silly example <button/> usage */ <button></button></template>',
```
rwjblue, line 143:
```suggestion
      template: '<template>/** silly example <button> usage */ <button></button></template>',
```

I think the `button` in the comment probably shouldn't self-close

test/unit/rules/require-button-type-test.js
```diff
@@ -142,18 +140,19 @@ generateRuleTests({
...
       meta: {
         filePath: 'layout.gjs',
       },
-      fixedTemplate: '/** silly example <button> usage */ <button type="button"></button>',
+      fixedTemplate:
+        '<template>/** silly example <button type="button" /> usage */ <button type="button"></button></template>',
```
rwjblue, line 148: This is proving my point in the other comment. We can't use `currentSource.replace`. The `button` in the comment should never be touched.

rwjblue (changes requested):

test/unit/rules/require-button-type-test.js
gabrielcsapo: I rebid on your branch upstream, good call.

PR #2699, opened 2022-11-10
chore: cleanup based on feedback from rjackson

gabrielcsapo:


lib/linter.js
```diff
@@ -251,17 +249,18 @@ export default class Linter {
 
         if (code !== templateInfo.template) {
           let prefix;
-          if (templateInfo.templateMatch.type === 'template-tag') {
+          if (templateInfo.templateMatch?.type === 'template-tag') {
```
rwjblue, line 252: We cannot use `?.` just yet (this targets master, which still runs on Node 12)

lib/linter.js
```diff
@@ -251,17 +249,18 @@ export default class Linter {
...
             let { start } = templateMatch;
 
             prefix = currentSource.slice(start.index, start.index + start[0].length);
           } else {
-            prefix = templateMatch.tagName + '`';
+            prefix = `${templateMatch?.tagName}\``;
```
rwjblue, line 257: Same here, also why would `templateMatch` not be defined?

lib/linter.js
```diff
@@ -251,7 +249,7 @@ export default class Linter {
 
         if (code !== templateInfo.template) {
           let prefix;
-          if (templateInfo.templateMatch.type === 'template-tag') {
+          if (templateInfo.templateMatch && templateInfo.templateMatch.type === 'template-tag') {
```
rwjblue, line 252: I do not think that it is possible for `templateInfo.templateMatch` to be undefined, I don't see why we need this initial guard

PR #2701, opened 2022-11-14
Fix autofixer for value-less attributes and comment order in `attribute-order` rule

### LINKED ISSUE #2692: `--fix` for `attribute-order` breaks `@tagName=""`
```hbs
<Foo class="asd" @tagName="" />
```

Running `--fix` on that leads to:

```hbs
<Foo @tagName class="asd" />
```

Which is broken.

### LINKED ISSUE #2693: `--fix` for `attribute-order` moves around comments
```hbs
<Foo
  class="asd asd qwe rty asd zxc asd qwe asd zxc asd qwe asd zxc qwe qwe qwe qwe qwe asd zxc qwe qwe qwe qwe qwe"
  {{! @glint-expect-error }}
  @foo={{1}} />
```

(the long `class` is for Prettier to not make the whole thing on one line)

Running `--fix` on that leads to:

```hbs
<Foo
  @foo={{1}}
  class="asd asd qwe rty asd zxc asd qwe asd zxc asd qwe asd zxc qwe qwe qwe qwe qwe asd zxc qwe qwe qwe qwe qwe"
  {{! @glint-expect-error }}
/>
```

Which is not nice because the comment is supposed to go with the `@foo` argument but it was moved at the end. This is a breaking change for `glint` directives. And just an annoyance for normal comments. :)

Not sure if that's a bug that's fixable though...

rob-long:
Fixes #2693
Fixes #2692 
- add additional test coverage for these fixes
- refactor for clarity
- add more inline docs

### DIFF
--- a/lib/rules/attribute-order.js
+++ b/lib/rules/attribute-order.js
@@ -29,9 +29,11 @@ export function createAttributesOrderErrorMessage(config) {
 // Class representing all of attribute tokens of a given tokenType for a node
 class TokenGroup {
   _items;
+  _context;
 
-  constructor(items) {
+  constructor(items, context) {
     this._items = items;
+    this._context = context;
   }
 
   get items() {
@@ -43,7 +45,13 @@ class TokenGroup {
   }
 
   get indexes() {
-    return this.items.map((attr) => attr.loc.start.line * attr.loc.start.column);
+    return this.items.map((node) => this._context.calculateAttributeIndex(node));
+  }
+
+  get indexesWithSource() {
+    return this.items.map((node) => {
+      return { node, index: this._context.calculateAttributeIndex(node) };
+    });
   }
 
   get lastIndex() {
@@ -54,9 +62,9 @@ class TokenGroup {
     return this.indexes.length ? Math.min(...this.indexes) : undefined;
   }
 
-  unalphabetizedItemIndex(context) {
+  get unalphabetizedItemIndex() {
     const alpha = this.items.map((attr) => {
-      return context.sourceForNode(attr).match(/([^A-Za-z]*)([\w-]*)/)[2];
+      return this._context.sourceForNode(attr).match(/([^A-Za-z]*)([\w-]*)/)[2];
     });
     return alpha.findIndex((e, idx) => idx > 0 && e < alpha[idx - 1]);
   }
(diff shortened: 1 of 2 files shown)


lib/rules/attribute-order.js
```diff
@@ -453,45 +497,56 @@ export default class AttributeOrder extends Rule {
...
         if (orderA === orderB) {
           if (a.loc && b.loc) {
-            return a.loc.start.line * a.loc.start.column > b.loc.start.line * b.loc.start.column;
+            return this.calculateAttributeIndex(a) > this.calculateAttributeIndex(b);
           }
         }
         return orderA > orderB ? 1 : -1;
       });
 
+      for (const comment of node[AttributeType.COMMENTS]) {
```
bmish, line 515: Can you add a high-level sentence ahead of this loop to explain what it does?

lib/rules/attribute-order.js
```diff
@@ -178,6 +193,33 @@ export default class AttributeOrder extends Rule {
       : AttributeType.ATTRIBUTES;
   }
 
+  getAttribute(nodeTokens, { type, name, value }) {
```
bmish, line 196: Can you add a high-level sentence ahead of each function to explain what it does?

test/unit/rules/attribute-order-test.js
```diff
@@ -734,6 +734,79 @@ generateRuleTests({
...
+        `);
+      },
+    },
+    {
+      template: `<Foo
+      {{!-- @glint-expect-error --}}
+      id="op"
+      {{!-- @second --}}
+      @foo={{1}}
+    />`,
```
bmish, line 777: Can we also test with a double comment and with a trailing comment?

```
`<Foo
      {{!-- @glint-expect-error --}}
      id="op"
      {{!-- double comment --}}
      {{!-- @second --}}
      @foo={{1}}
      {{!-- trailing comment --}}
    />`

bmish:
Note that you have to use the syntax `Fixes #123` in the PR description so this PR will automatically link and close the issues.

lib/rules/attribute-order.js
rob-long: Refactored for clarity and added docs to most functions.

bmish (approved):
Thanks!

PR #2722, opened 2022-11-28
Fix autofixer bug with `firstObject` inside a `MustacheStatement` in `no-array-prototype-extensions`

tgvrssanthosh:
## Summary
`no-array-prototype-extensions` auto-fixer is generating an incorrect code when the path expression containing `firstObject` is inside `mustachestatement` as a param.

For example, the following template
```
<div data-test={{eq this.list.firstObject.abc "def"}}>Hello</div>
```

is modified to

```
<div data-test={{eq this.list "0.abc"}}>Hello</div>
```

instead of

```
<div data-test={{eq (get this.list "0.abc") "def"}}>Hello</div>
```

## Fix
The root cause of the issue was that there was a special handling of `firstObject` inside `MustachStatement`. Ideally, that piece of logic should run only when the mustache statement directly contains a path expression containing `firstObject`. In this PR, an additional condition has been added to check for the same.


tgvrssanthosh:
@bmish Can you help review this PR? Thank you!

smilland (approved):

smilland:
Thank you!

PR #2725, opened 2022-12-02
Fix `--no-ignore-pattern`

velrest:
add type check since `--no-ignore-patter` returns an array which `micromath.isMatch` can't handle

### DIFF
--- a/lib/helpers/cli.js
+++ b/lib/helpers/cli.js
@@ -177,7 +177,10 @@ export function expandFileGlobs(workingDir, filePatterns, ignorePattern, glob = 
     let isLiteralPath = !isGlob(pattern) && isFile(path.resolve(workingDir, pattern));
 
     if (isLiteralPath) {
-      let isIgnored = micromatch.isMatch(pattern, ignorePattern);
+      // If `--no-ignore-pattern` is passed, the ignorePatter is `[false]`.
+      let isIgnored =
+        !(ignorePattern?.length === 1 && ignorePattern[0] === false) &&
+        micromatch.isMatch(pattern, ignorePattern);
 
       if (!isIgnored) {
         result.add(pattern);


velrest:
Tests also seem to fail on master

lib/helpers/cli.js
```diff
@@ -177,7 +177,11 @@ export function expandFileGlobs(workingDir, filePatterns, ignorePattern, glob =
     let isLiteralPath = !isGlob(pattern) && isFile(path.resolve(workingDir, pattern));
 
     if (isLiteralPath) {
-      let isIgnored = micromatch.isMatch(pattern, ignorePattern);
+      // If `--no-ignore-pattern` is passed, the ignorePatter is `[false]`.
+      let isIgnored =
+        ignorePattern?.length !== 1 &&
+        ignorePattern?.[0] !== false &&
```
bmish, line 183: I think this is what you want:

```suggestion
        `!(ignorePattern?.length === 1 && ignorePattern[0] === false) &&
```

velrest:
I just tested this:

According to `git bisect`, 1dcfd1b342cc1f95d46c6b8b79f05cbab1fdef8a is the first bad commit.

Test script run with `git bisect` to verify a commit:
```fish
if contains "yarn.lock" (ls)
  yarn --frozen-lockfile
else if contains "package-lock.json" (ls)
  npm i
else
  echo "No matching package manager found"
  exit 1
end
string match -r ".*\b(require-button-type)\b.*" (
  node bin/ember-template-lint.js --config-path  ../.template-lintrc.js --no-ignore-pattern /tmp/topic-list.hbs
) && not string match -r ".*\b(require-button-type)\b.*" (
  node bin/ember-template-lint.js --config-path  ../.template-lintrc.js /tmp/topic-list.hbs
)
```

bmish (approved):
Thanks!

Issue #2700, opened 2022-11-13
no-unknown-arguments-for-builtin-components: Argument "@route" or "@query" is required for <LinkTo /> component


jelhan:
The `no-unknown-arguments-for-builtin-components` complains about `<LinkTo />` component having `@model` (or `@models`) argument only. It requires to have either `@route` or `@query`. But having only `@model` or `@models` should be fine.

`<LinkTo />` having only `@model` or `@models` transitions to other dynamic segments of the same route keeping query parameters untouched. This is covered by public API as far as I know.

Originally the rule even required `@route` to be present always. Addressing #1802 it was extended to either `@route` or `@query`. But the case of `@model` or `@models` only seem to be missed on that one.

geneukum:
Good catch, I think that you're right per https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L291.

I opened https://github.com/ember-template-lint/ember-template-lint/pull/2726 to hopefully resolve this issue.

jelhan:
Thanks a lot for fixing it!

PR #2726, opened 2022-12-04
Allow passing only `@model` or `@models` for `<LinkTo>` in `no-unknown-arguments-for-builtin-components` rule

geneukum:
The no-unknown-arguments-for-builtin-components lint rule is triggered when a <LinkTo /> component only has the `@model` or `@models` argument, stating that it must have either `@route` or `@query` as well.

The <LinkTo /> component supports passing only the `@model` or `@models` argumment. In this case, it will simply reuse the current route / queryParams.

As such, we should update the linter rule to allow this case.

This closes https://github.com/ember-template-lint/ember-template-lint/issues/2700.


bmish (approved):
Thanks!

Issue #1888, opened 2021-04-08
Require-valid-alt-text's object accessible text


frykten:
Hi!

Looking at the documentation, I found info that looked weird to me on this [require-valid-alt-text rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/require-valid-alt-text.md#object):
```
Add alternative text to all embedded <object> elements using either inner text, setting the title prop [...]
[...]
<object width="128" height="256" title="Middle-sized"></object>
```

From what I know and understand, the `title` attribute isn't _fully_ accessible: it gives a tooltip with some text for **visual** users, but **screen readers** do _not_ understand it. It should then not be used solely, but accompanied by an `aria-label` or inner text, right?

MelSumner:
@frykten screen reader support is spotty but it _is_ valid according to the specification. Therefore we can advise against it but can't mark it as invalid.

geneukum:
Hey @MelSumner, came across this and made an attempt to add in that note over in https://github.com/ember-template-lint/ember-template-lint/pull/2727. I hope that's alright!

I'm definitely happy to make any changes if needed! :smile:

PR #2727, opened 2022-12-04
Add note about `title` property limitations to `require-valid-alt-text` rule doc

geneukum:
The `require-valid-alt-text` rule documentation suggests that alt text be added to all embedded `<object>` elements via inner text, `title` prop, or `aria-label`/`aria-labelledby`.

In general, the `title` prop may be a little bit less reliable, but it _is_ still valid according to the spec. As such, we might want to advise developers against using in case they're referencing this doc.

This closes https://github.com/ember-template-lint/ember-template-lint/issues/1888.


bmish (approved):
Thanks!

Issue #2728, opened 2022-12-04
`no-obsolete-elements` should flag usages of param element 


geneukum:
The HTML WhatWG spec has obsoleted the `param` element with the recommendation now being to use the data attribute of the object element to set the URL of the external resource. See: https://github.com/whatwg/html/commit/04a11f278dfafd395a497172face6b5dba082d2c for more details.
    
As such, we should go ahead and add this to our "obsolete list" in the no-obsolete-elements rule so that the linter can flag it as an issue.

Issue #2712, opened 2022-11-19
Plan v6 Release


bmish:
### Planned changes for v6

* TODO

### Future changes not yet ready:

* Match ESLint behavior
   * [Allow any file extension to be matched in globs to match ESLint's behavior](https://github.com/ember-template-lint/ember-template-lint/pull/2366#issuecomment-1035008823)
   * Print out full paths by default (variation of https://github.com/ember-template-lint/ember-template-lint/pull/2491#issuecomment-1115253179)
   * #2366
* Add more rules to `stylistic` config
   * #2575 (this rule might be too noisy/opinionated to include)
* New `recommended` rules
   * @runspired [requested](https://discord.com/channels/480462759797063690/666416704418611233/927660252420259860) to have [no-model-argument-in-route-templates](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-model-argument-in-route-templates.md) considered as a `recommended` rule but the rule and Ember documentation may need tweaks before this would make sense
* Re-evaluate whether we should be publicly-exporting so many [helpers](https://github.com/ember-template-lint/ember-template-lint/blob/c9730dd5b3e07b0b727e3f5a5b2bb4795ca53b97/lib/index.js#L10) in our Node API

For reference, [v5.0.0](https://github.com/ember-template-lint/ember-template-lint/releases/tag/v5.0.0) was released 2022-11-19, discussion in #2319.

bmish:
First alpha is out:
* https://github.com/ember-template-lint/ember-template-lint/releases/tag/v6.0.0-alpha.0

bmish:
@NullVoxPopuli I think it's best for me to release the final v6 now. It's already a large release. #3013 can be saved for the next release whenever it's ready so it's not rushed. That will give time for users to use v6, and v7 can come as soon as a few months later.

NullVoxPopuli:
sounds good!

ef4:
In the future, let's please not add any rules to the recommended set without an actual ember RFC backing them up. 

When upgrading an app, most users don't distinguish "ember is making me do this" from "ember-template-lint is making me do this".  So adding a rule like `builtin-form-components` to the default set is basically a backdoor deprecation of those features. Maybe we *should* deprecate them, but then let's actually deprecate them properly. Right now we're in an awkward place where they're still fully active in the docs and guide and codebase but forbidden by the default lint rules.

bmish:
[...]

PR #2729, opened 2022-12-04
Add `param` to obsolete list in `no-obsolete-elements` rule

geneukum:
The HTML spec has obsoleted the `param` element with the recommendation being to use the data attribute of the object element to set the URL of the external resource.

As such, let's go ahead and add this to our "obsolete list" so that the no-obsolete-elements rule will flag it.

See: https://github.com/whatwg/html/commit/04a11f278dfafd395a497172face6b5dba082d2c for more details.

This closes https://github.com/ember-template-lint/ember-template-lint/issues/2728.


bmish:
Thanks! Will probably consider this a breaking change, similar to what we did with:
* #1532
And include it in the next major release:
* #2712

bmish (approved):
Thanks!

Issue #2628, opened 2022-10-07
Valid menu/sub-menu pattern is reporting errors


KamiKillertO:
The following markup which is from W3C WAI example is not valid according with ember-template-lint.

```hbs
<ul role="menubar" aria-label="functions" id="appmenu">
  <li role="menuitem" aria-haspopup="true">
    File
    <ul role="menu">
      <li role="menuitem">New</li>
      <li role="menuitem">Open</li>
      <li role="menuitem">Print</li>
    </ul>
  </li>
  {{!-- ... --}}
</ul>
````
ember-template-lint is reporting the following error 
> Do not use an element with \`role="menuitem"\` inside an element with \`role="menuitem"\` : ember-template-lint ([no-nested-interactive](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-nested-interactive.md))

After discussing with @MelSumner, it appears that the submenu pattern should be supported by ember-template-lint as it is valid markup

bmish:
Yeah looks like we aren't handling nested menus. Can you include the rule name with the report? PR welcome.

KamiKillertO:
@bmish I've updated the report and include the rule name :+1:

geneukum:
hey @KamiKillertO, did you still want to look into this? If not, I don't mind trying to submit a PR.

bmish:
Any PRs appreciated!

KamiKillertO:
@geneukum no I did not have the time to work on it. Feel free to have a look

PR #2731, opened 2022-12-04
Allow nested `menuitem`s in `no-nested-interactive` rule

geneukum:
Right now, the `no-nested-interactive` rule will flag nested elements with the `menuitem` role. This is a valid pattern used to create submenus, and as such should be allowed by the rule.

This closes https://github.com/ember-template-lint/ember-template-lint/issues/2628.


bmish (approved):
Thanks!

Issue #2286, opened 2021-12-21
Automatically exclude all helpers from `no-curly-component-invocation`


shankarsridhar:
Hey folks, thanks for this wonderful add-on.
I would like to know if there is a way to make this add-on automatically exclude all helpers in the app so that its not needed to manually add the helpers in the app code in the configuration file(.template-lintrc.js, for example).

That is, 
Current way of doing (manual):
```js
rules: {
    'no-curly-component-invocation': {
      allow: ['my-helper', 'my-other-helper'],
    },
  },
```

What I would like to have/proposing
```js
rules: {
// config not needed in this case, since all the exclusions are for helpers 
// which are automatically excluded for the rule 'no-curly-component-invocation'
  },
```

I am using version `3.15.0` of this add-on

lifeart:
Hi @shankarsridhar, not, it's not really possible with current state of things,  but, you could use snippet:

```js
var componentLikeHelpers = Object.keys(require.entries)
    .filter(name=>(name.includes('/helpers/')|| name.includes('/helper')))
    .filter(name=>!name.includes('/-')).map(name=>{
        let path = name.split('/helpers/');
        return path.pop();
    }).filter(name=>!name.includes('/')).uniq();

console.log(JSON.stringify(componentLikeHelpers))
```

to collect all app helpers

bmish:
Maybe the rule doc could be updated to link to this "skipping helpers" part of the codemod doc in case people want to use it for the lint rule too: https://github.com/ember-codemods/ember-angle-brackets-codemod#skipping-helpers

PR #2732, opened 2022-12-04
 Link to "skipping helpers" part of the codemod doc in `no-curly-component-invocation` docs

geneukum:
In order to ignore custom helpers in the `no-curly-component-invocation` linter rule, a developer manually must add these to the config `allow` list. The `ember-angle-brackets-codemod` README has a useful snippet that can be run in the console in order to generate a list of helpers in a given ember app.

Let's add a reference to that here in case folks want to use it to configure this lint rule too.

This closes https://github.com/ember-template-lint/ember-template-lint/issues/2286.


bmish (approved):
Thanks!

PR #2746, opened 2022-12-14
Add autofixer to `no-multiple-empty-lines` rule

Shivareddy-Aluri:
added --fix for no-multiple empty lines

[no-multiple-empty-lines](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-multiple-empty-lines.md)

### DIFF
--- a/test/unit/rules/no-multiple-empty-lines-test.js
+++ b/test/unit/rules/no-multiple-empty-lines-test.js
@@ -26,6 +26,7 @@ generateRuleTests({
   bad: [
     {
       template: '<div>foo</div>\n\n\n<div>bar</div>',
+      fixedTemplate: '<div>foo</div>\n<div>bar</div>',
 
       verifyResults(results) {
         expect(results).toMatchInlineSnapshot(`
@@ -35,6 +36,7 @@ generateRuleTests({
               "endColumn": 14,
               "endLine": 4,
               "filePath": "layout.hbs",
+              "isFixable": true,
               "line": 2,
               "message": "More than 1 blank line not allowed.",
               "rule": "no-multiple-empty-lines",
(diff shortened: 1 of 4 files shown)


test/unit/rules/no-multiple-empty-lines-test.js
```diff
@@ -26,6 +26,7 @@ generateRuleTests({
   bad: [
     {
       template: '<div>foo</div>\n\n\n<div>bar</div>',
+      fixedTemplate: '<div>foo</div>\n<div>bar</div>',
```
bmish, line 29: I'm not sure this is right. Shouldn't we collapse collapse any number of blank lines to just one? This collapses it to no blank lines.

```
foo


bar
```

should be fixed to:

```
foo

bar
```

which is equivalent to having `\n\n`.

Shivareddy-Aluri: oh yeah, done!

test/unit/rules/no-multiple-empty-lines-test.js
```diff
@@ -76,6 +80,7 @@ generateRuleTests({
       config: { max: 3 },
 
       template: '<div>foo</div>\n\n\n\n\n<div>bar</div>',
+      fixedTemplate: '<div>foo</div>\n\n\n\n<div>bar</div>',
```
bmish, line 83: Shouldn't this be collapsed to `\n\n` as well? Always collapse to one blank line, per my other comment.

Shivareddy-Aluri: max is 3 here! so 3 lines as per config!

bmish: Missed that, thanks!

bmish (approved):
Thanks!

PR #2753, opened 2022-12-19
Add autofixer to `no-trailing-spaces` rule

Shivareddy-Aluri:
[no-trailing-spaces](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-trailing-spaces.md)

### DIFF
--- a/lib/rules/no-trailing-spaces.js
+++ b/lib/rules/no-trailing-spaces.js
@@ -9,17 +9,34 @@ export default class NoTrailingSpaces extends Rule {
 
         exit(node) {
           let source = this.sourceForNode(node);
-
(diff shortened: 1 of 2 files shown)


Shivareddy-Aluri:
@bmish tried implementing the trailing spaces, however for BlockStatement even though i have returned the modified node, it still takes the original node. how can i solve this? how do i modify a block statement node?

bmish:
Will re-creating the BlockStatement help?

We create nodes in other parts of the codebase like here https://github.com/ember-template-lint/ember-template-lint/blob/5ce0a419e219ff602d96b51c1d33ecf991e75712/lib/rules/attribute-order.js#L564

Issue #617, opened 2019-01-19
more options to quotes rule


kellyselden:
I implemented the quotes rule the quickest and to satisfy the style I was using. It is now clear that there should be at least one more option for when you enter hbs syntax.

Some people prefer

```hbs
<foo bar="asdf" baz={{concat 'asdf' 'asdf'}} />
```

This is not possible with the current rule, where it is all or nothing double qoutes.

I'm not sure the api, but maybe something like

```js
'quotes': {
  outsideCurlies: 'double',
  insideCurlies: 'single'
}
```

pieter-v:
This becomes more important, since prettier formats templates this way.

robclancy:
Been open since 2019... how do we do this? When I enabled this rule I kinda expected it to just ignore the HTML quotes.

Also could the options be "htmlQuotes" and "hdsQuotes"?

robclancy:
I think it's done https://github.com/ember-template-lint/ember-template-lint/pull/2754

bmish:
Try it out in: https://github.com/ember-template-lint/ember-template-lint/releases/tag/v5.9.0

PR #2754, opened 2022-12-22
Add options for different `html` / `hbs` quote styles in `quotes` rule

robclancy:
I think I did this right?

Fixes https://github.com/ember-template-lint/ember-template-lint/issues/617

Personally I think the default config should be as described in the issue above with double only used for HTML attributes. Also the config forces the need to specify both `hbs` and `html` but it could also just default to `double` for whichever one of those is missing. So for example you could just do `{ hbs: 'single' }` to get the same functionality, essentially still using `double` as a default. 

Some code/messages etc in here could probably use better words. Maybe moustache/curlies instead of handlebars for example. And HTML Attributes instead of just HTML. I'm too tired to deal with naming things.

I think it might need more tests too?

### DIFF
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -69,7 +69,10 @@ export default class Quotes extends Rule {
     } else if (goodChars.hbs === chars.double && goodChars.html === chars.double) {
       message = 'you must use double quotes in templates';
     } else {
-      message = 'you must use double quotes x and single quotes in y in templates';
+      const double = goodChars.hbs === chars.double ? 'Handlebars syntax' : 'HTML attributes';
+      const single = goodChars.hbs === chars.single ? 'Handlebars syntax' : 'HTML attributes';
+
+      message = `you must use double quotes for ${double} and single quotes for ${single} in templates`;
     }
 
     return {
--- a/lib/rules/quotes.js
+++ b/lib/rules/quotes.js
@@ -14,6 +14,18 @@ export default class Quotes extends Rule {
       }
       case 'string': {
         if (['double', 'single'].includes(config)) {
+          return {
+            hbs: config,
+            html: config,
+          };
+        }
+        break;
+      }
+      case 'object': {
+        if (
+          ['double', 'single'].includes(config.hbs) &&
+          ['double', 'single'].includes(config.html)
+        ) {
           return config;
         }
         break;
@@ -28,6 +40,7 @@ export default class Quotes extends Rule {
       [
         '  * "double" - requires the use of double quotes wherever possible',
         '  * "single" - requires the use of single quotes wherever possible',
+        '  * { hbs: `single`|`double`, html: `single`|`double` } - requires different quotes for Handlebars and HTML syntax',
       ],
       config
     );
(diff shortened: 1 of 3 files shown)


lib/rules/quotes.js
```diff
@@ -14,6 +14,18 @@ export default class Quotes extends Rule {
...
+        }
+        break;
+      }
+      case 'object': {
+        if (
+          ['double', 'single'].includes(config.hbs) &&
+          ['double', 'single'].includes(config.html)
+        ) {
           return config;
         }
```
bmish, line 30: Can we throw an error if there are any unknown properties in the object? And add an error test case for this of course.

lib/rules/quotes.js
```diff
@@ -14,6 +14,18 @@ export default class Quotes extends Rule {
...
+            hbs: config,
+            html: config,
+          };
+        }
+        break;
+      }
+      case 'object': {
+        if (
+          ['double', 'single'].includes(config.hbs) &&
+          ['double', 'single'].includes(config.html)
```
bmish, line 27: Wouldn't it make sense to allow specifying one of them, and falling back to the default for the one not specified? And note we can't change defaults now (only in a future breaking change PR).

lib/rules/quotes.js
```diff
@@ -28,6 +40,7 @@ export default class Quotes extends Rule {
       [
         '  * "double" - requires the use of double quotes wherever possible',
         '  * "single" - requires the use of single quotes wherever possible',
+        '  * { hbs: `single`|`double`, html: `single`|`double` } - requires different quotes for Handlebars and HTML syntax',
```
bmish, line 43: We'll need to mention the new options in the rule documentation file.

bmish, line 43: Regarding the names `hbs` and `html`, can you explain a bit more how you came to those names?

bmish (commented):
Thanks for starting on this!

lib/rules/quotes.js
bmish, line 43: Should we allow the user to indicate they don't want to enforce one of these types? E.g. `false` turns off validation for one type so either kind of quotes is allowed.

bmish, line 43: For the record, what do you think the ideal default should be for each type? So we can consider it in the nexst major version.

robclancy: It's just how I see the difference, I didn't really think much about it as I was in a rush and pretty tired. I don't know what they should be called.

robclancy: I think it should be double quotes HTML attributes and single everything else. That's how most code I see is written. The only reason to go with double always is because it is the default prettier does, however most people seem to want it changed to single with this issue being very unclear on weather it will change or not. https://github.com/prettier/prettier/issues/4102

NullVoxPopuli: The hbs file extension doesn't exist in strict mode, so i wonder if this should be 'expressions'?
- expressions
- template expressions
- curlies
- within curlies
- within curly braces
- within expressions 

Ond then:
- within html
- outside expressions
- outside curly braces

?

Idk

bmish: What about `html` and `mustache`? Does that accurately describe the difference?

robclancy: I think `html` and `curlies` would be good. Mainly because `curlies` is a term used a fair bit with ember.

Been pretty busy so haven't gotten back to this but should be good over the weekend or early next week.

bmish: That's fair, there are two ember-template-lint rules with `curlies` in the name, for prior art. Let's proceed with that then and get this merged in the coming days!

lib/rules/quotes.js
robclancy: The problem with this is the default is actually off. The "double" is part of the stylistic preset. I think it's best to just disable the rule for the one not specified, much like your other suggestion of when it is false.

lib/rules/quotes.js
bmish: Looks like this is still needed.

lib/rules/quotes.js
bmish: Looks like an update to the rule doc is still needed. Let's mention the behavior of what happens when one of these is not provided. Let's also provide an example showing what code `hbs` applies to and what code `html` applies to.

bmish: What do you think of this? Would be nice to have a way to explicitly specify that the option is off besides just omitting the option.

robclancy: Yeah I'm still working on your comments, I just rebased to master before and didn't push my changes yet as I ran out of time.

bmish: Great, thanks for continuing to work on this!

bmish:
What's the status of this?

lin-ll:
> What's the status of this?

I noticed that this hadn't been picked up in a while, so I just made some changes to resolve the comments and push it across the finish line. I think it's good to go now - lmk if you have any additional comments. @bmish

bmish (approved):
Thanks!

PR #2778, opened 2023-01-22
Ignore `unique-id` helper in `no-duplicate-id` rule

### LINKED ISSUE #2777: no-duplicate-id rule errors when using ember's built-in unique-id helper
When two elements both assign `id={{unique-id}}` the no-duplicate-ids rule triggers an error.

Since `unique-id` is a built-in helper that always returns a unique id this should not be a violation.

unique-id

lifeart:
Unique-id added to ignore list

Fixes #2777.

### DIFF
--- a/lib/rules/no-duplicate-id.js
+++ b/lib/rules/no-duplicate-id.js
@@ -2,10 +2,11 @@ import AstNodeInfo from '../helpers/ast-node-info.js';
 import { match } from '../helpers/node-matcher.js';
 import Rule from './_base.js';
 
+const IGNORE_IDS = new Set(['{{unique-id}}']); // Built-in helper
+
 class ConditionalScope {
   constructor() {
     this.seenIdStack = [new Set()];
-    this.ignoreIds = new Set(['{{unique-id}}']);
     this.conditionals = [];
   }
 
@@ -35,7 +36,7 @@ class ConditionalScope {
   }
 
   isIgnoredId(id) {
-    return this.ignoreIds.has(id);
+    return IGNORE_IDS.has(id);
   }
 
   isDuplicateId(id) {
--- a/lib/rules/no-duplicate-id.js
+++ b/lib/rules/no-duplicate-id.js
@@ -5,7 +5,7 @@ import Rule from './_base.js';
 class ConditionalScope {
   constructor() {
     this.seenIdStack = [new Set()];
-
+    this.ignoreIds = new Set(['{{unique-id}}']);
     this.conditionals = [];
   }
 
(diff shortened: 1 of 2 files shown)


lib/rules/no-duplicate-id.js
```diff
@@ -5,7 +5,7 @@ import Rule from './_base.js';
 class ConditionalScope {
   constructor() {
     this.seenIdStack = [new Set()];
-
+    this.ignoreIds = new Set(['{{unique-id}}']);
```
bmish, line 8: Since this will never change during runtime and isn't related to the scope class, shall we just define it outside the class as `IGNORE_IDS` or `ID_ALLOWLIST`?

bmish, line 8: Can you add a comment next to this entry in the set: `// Built-in helper`

test/unit/rules/no-duplicate-id-test.js
```diff
@@ -8,6 +8,9 @@ generateRuleTests({
   good: [
     // Unique sibling TextNode IDs
     '<div id="id-00"></div><div id="id-01"></div>',
+    '<div id={{unique-id}}></div><div id={{unique-id}}></div>',
+    '<div id="{{unique-id}}"></div><div id="{{unique-id}}"></div>',
+    "<div id='{{unique-id}}'></div><div id='{{unique-id}}'></div>",
```
bmish, line 13: Should we include these two test cases? It's a bit confusing to me since these are just strings that look like the helper.

test/unit/rules/no-duplicate-id-test.js
```diff
@@ -8,6 +8,9 @@ generateRuleTests({
   good: [
     // Unique sibling TextNode IDs
     '<div id="id-00"></div><div id="id-01"></div>',
+    '<div id={{unique-id}}></div><div id={{unique-id}}></div>',
+    '<div id="{{unique-id}}"></div><div id="{{unique-id}}"></div>',
```
bmish, line 12:
```suggestion

    // Built-in helpers:
    '<div id="{{unique-id}}"></div><div id="{{unique-id}}"></div>',
```

bmish:
`Fixes #x` needs to be in the PR description for linking to work.

test/unit/rules/no-duplicate-id-test.js
lifeart: it's not a strings, it's concat statement, and we check what identificators correctly handled for ` and " inside concats.

image

bmish (approved):
Thanks!

PR #2780, opened 2023-01-22
Ignore `{{(unique-id)}}` helper variation in `no-duplicate-id` rule

lifeart:
fixes https://github.com/ember-template-lint/ember-template-lint/issues/2777

TL DR:
This rule does not normalize id's I have tried to do it, but it seems a quite complex task.

### DIFF
--- a/lib/rules/no-duplicate-id.js
+++ b/lib/rules/no-duplicate-id.js
@@ -2,7 +2,7 @@ import AstNodeInfo from '../helpers/ast-node-info.js';
 import { match } from '../helpers/node-matcher.js';
 import Rule from './_base.js';
 
-const IGNORE_IDS = new Set(['{{unique-id}}']); // Built-in helper
+const IGNORE_IDS = new Set(['{{unique-id}}', '{{(unique-id)}}']); // Built-in helper, with and without parens
 
 class ConditionalScope {
   constructor() {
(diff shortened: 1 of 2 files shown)


lib/rules/no-duplicate-id.js
```diff
@@ -2,7 +2,7 @@ import AstNodeInfo from '../helpers/ast-node-info.js';
 import { match } from '../helpers/node-matcher.js';
 import Rule from './_base.js';
 
-const IGNORE_IDS = new Set(['{{unique-id}}']); // Built-in helper
+const IGNORE_IDS = new Set(['{{unique-id}}', '{{(unique-id)}}']); // Built-in helper, with and without parens
```
bmish, line 5: How about changing this to storing just the helper name? And then the code can check different variations of the usage.
```suggestion
const IGNORED_HELPERS = new Set(['unique-id']); 
```

lifeart: @bmish I gave it a try, and it's overhead for current rule architecture, noted it in TLDR: https://github.com/ember-template-lint/ember-template-lint/pull/2780#issue-1552260436

bmish: Wouldn't the only difference for this change be:

```js
isIgnoredId(id) {
    return IGNORED_HELPERS.has('{{'+id+'}}') || IGNORED_HELPERS.has('{{('+id+')}}');
}
```

lifeart: @bmish problem not with `{`, problem with `id` itself, it's resolved from `sourceForNode`, not from `path` and this  is main architectural problem for this rule.

PR #2792, opened 2023-02-01
Add new rule `no-unnecessary-curly-parens`

raycohen:
`{{( expression )}}` -> `{{ expression }}`


bmish:
I'm glad to see a PR for this new rule. It looks good.

Just thinking of some names that could be a bit clearer. What about `no-redundant-curly-parens`? We use "no-redundant" in several existing rules, and we also use "curlies" in several rules instead of "mustache".

bmish:
Or `no-unnecessary-curly-parens` to align with the name of:
* #2794

raycohen:
updated to use no-unnecessary-curly-parens throughout

lib/rules/no-unnecessary-curly-parens.js
```diff
@@ -0,0 +1,28 @@
...
+              path.parentNode,
+              path.parentKey,
+              b.mustache(node.path.path, node.path.params, node.path.hash)
+            );
+          }
+          this.log({
+            node,
+            isFixable: true,
+            message: 'Unnecessary parentheses enclosing statement',
+          });
```
bmish, line 23: Can you move this into an `else`? Not sure if it's strictly required but all our other rules follow `if (this.mode === 'fix') { } else { this.log() }`.

bmish:
Also run: `npm run update:readme`

bmish (approved):
Thanks!

test/unit/rules/no-unnecessary-curly-parens-test.js
```diff
@@ -0,0 +1,83 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-unnecessary-curly-parens',
+
+  config: true,
+
+  good: ['{{foo}}', '{{concat "a" "b"}}', '{{concat (capitalize "foo") "-bar"}}'],
```
bertdeblock, line 8: We should add and allow cases like `<Foo @bar={{(bar-helper)}} />`.
In the context of an argument or attribute, `{{(bar-helper)}}` and `{{bar-helper}}` are not the same thing.

raycohen: I can open another PR. Seems like this will reduce the cases we can confirm are unnecessary since we can't tell if `{{(this.something)}}` is a helper or not. That said most of the cases I see this actually happening in our codebase, the helper is being passed arguments so those can still be deemed unnecessary, right?

bertdeblock: Yup, that's right I think!

raycohen: https://github.com/ember-template-lint/ember-template-lint/pull/2800

Issue #2776, opened 2023-01-16
[Rule proposal] Disallow redundant curly braces


boris-petrov:
```hbs
<SomeComponent @argument={{"asd"}} />
```

There is no need to wrap the string in curly braces. The same could be written as:

```hbs
<SomeComponent @argument="asd" />
```

Which improves readability and performance.

runspired:
I am aware that https://github.com/ember-template-lint/ember-template-lint/pull/2794 has been merged, and some folks may want this, I would like to push back on this eventually becoming part of the recommended config without more investment in a mechanism for hinting to the rule [ember/no-bare-strings](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md).

Currently, wrapping string literals in curlies is the only way by which to opt-out of this intl rule when needed. This is key because unfortunately that rule cannot be intelligent enough to understand when a string is used as a key or a path vs when it is used in a human legible form. The `elementAttributes` approach to this has in my experience not been a very useful one for many situations and proven much more difficult to maintain.

bmish:
@runspired what about just using `{{! template-lint-disable no-bare-strings}}` for that situation? I'm worried about encouraging these unnecessary curly strings just as a means to workaround another rule.

runspired:
A codebase with thousands of those would be a mess. Also very annoying to type everywhere, and applies to the whole block iirc.

runspired:
I suspect the answer here is to change no-bare-strings to have an allowList for what doesn't need to be translated as opposed to one for what does need to be translated.

bmish:
Definitely interest to make no-bare-strings smarter or more configurable to solve this...

PR #2794, opened 2023-02-05
Add new rule `no-unnecessary-curly-strings`

Shivareddy-Aluri:
closes: https://github.com/ember-template-lint/ember-template-lint/issues/2776


docs/rule/no-unnecessary-curly-braces-for-strings.md
```diff
@@ -0,0 +1,21 @@
+# no-unnecessary-curly-braces-for-strings
```
bmish, line 1: Perhaps just `no-unnecessary-curly-strings` for the rule name? Would align nicely with this other new rule:

* #2792 (likely being renamed to `no-unnecessary-curly-parens`)

docs/rule/no-unnecessary-curly-braces-for-strings.md
```diff
@@ -0,0 +1,21 @@
...
+🔧 The `--fix` option on the command line can automatically fix some of the problems reported by this rule.
+
+string values need not be wrapped in the mustache expressions.
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+class={{"btn"}}
```
bmish, line 12: How about a barebones example and/or a more complete example with the entire element:

```suggestion
{{"btn"}}
```

```suggestion
<FooBar class={{"btn"}}>
```

docs/rule/no-unnecessary-curly-braces-for-strings.md
```diff
@@ -0,0 +1,21 @@
...
+```
+This rule **allows** the following:
+
+```hbs
+class="btn"
```
bmish, line 17: Going along with the other suggestion:

```suggestion
<FooBar class="btn">
```

test/unit/rules/no-unnecessary-curly-braces-for-strings-test.js
```diff
@@ -0,0 +1,93 @@
...
+              "isFixable": true,
+              "line": 1,
+              "message": "Unnecessary curly braces around string",
+              "rule": "no-unnecessary-curly-braces-for-strings",
+              "severity": 2,
+              "source": "{{\\"xyz\\"}}",
+            },
+          ]
+        `);
+      },
```
bmish, line 90: I'm not sure these last 2 invalid test cases actually test anything different? The rule doesn't care about the surrounding code. We could remove them if they aren't useful.

test/unit/rules/no-unnecessary-curly-braces-for-strings-test.js
```diff
@@ -0,0 +1,93 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-unnecessary-curly-braces-for-strings',
+
+  config: true,
+
+  good: ['class="btn"'],
```
bmish, line 8: Need more valid examples:

* `{{foo}}`
* `{{(foo)}}`
* `<FooBar class="btn">`

bmish (changes requested):

docs/rule/no-unnecessary-curly-braces-for-strings.md
Shivareddy-Aluri: not able to change the rule names and run the tests..should i raise a new PR? getting this logs followed by errors
Screenshot 2023-02-06 at 1 11 20 AM

bmish: No need for a new PR. You just have to make sure all the files and references are renamed.

Shivareddy-Aluri: done with the changes @bmish , thank you!

test/unit/rules/no-unnecessary-curly-strings-test.js
```diff
@@ -0,0 +1,58 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-unnecessary-curly-strings',
+
+  config: true,
+
+  good: ['<FooBar class="btn" />', '{{foo}}'],
```
bmish, line 8:
```suggestion
  good: [
    '<FooBar class="btn" />',
    '{{foo}}',
    '{{(foo)}}',
    '{{this.calculate 1 2 op="add"}}',
    '{{get address part}}',
    'foo',
    '"foo"',
  ],
```

bmish:
I realized that there are other types of literals that we likely want to include in this rule.

* BooleanLiteral `{{true}}`
* StringLiteral `{{"Foo"}}`
* NumberLiteral `{{123}}`
* UndefinedLiteral `{{undefined}}`
* NullLiteral `{{null}}`

How about we include all literal nodes and rename rule to `no-unnecessary-curly-literals`?

bmish:
@Shivareddy-Aluri can you resolve the conflict and just disable the autofix for those cases for now? That way, either of us can look into fixing the autofix as a follow-up.

lib/rules/no-unnecessary-curly-literals.js
```diff
@@ -0,0 +1,24 @@
...
+  'UndefinedLiteral',
+  'NullLiteral',
+]);
+
+export default class NoUnnecessaryCurlyLiterals extends Rule {
+  visitor() {
+    return {
+      MustacheStatement(node) {
+        if (LITERALS.has(node.path.type)) {
+          this.log({
```
bmish, line 16: We should keep the autofix for strings. We should only disable the autofix in the specific cases you're having issues with.

Shivareddy-Aluri: even with strings i have the issue, say i have an example like this `< FooBar class={{'btn'}} />` this fix will change it to `< FooBar class="btn"}} />`. single quotes to double quotes. is that okay?

bmish: That's fair. Let's preserve the type of quotes. Can you only enable the autofix when double quotes are present then? You can check the quoteType from the StringLiteral.

Shivareddy-Aluri: sorry looks like i had blind spot, it didn't work for double quotes either(asserted wrong in my past commit where i check auto fix for double quotes). verified now. 😞

bmish: Got it. Okay let's follow-up to autofix it.

bmish (approved):
Thanks!

test/unit/rules/no-unnecessary-curly-literals-test.js
```diff
@@ -0,0 +1,150 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'no-unnecessary-curly-literals',
+
+  config: true,
+
+  good: [
```
bertdeblock, line 8: We should check component arguments as well, because now the rule also flags stuff like `@foo={{true}}`.

bertdeblock: Also `attr={{null}}` and `attr=null`, and `attr={{undefined}}` and `attr=undefined` aren't the same thing.
`attr={{null}}` does not render the attribute and `attr=null` renders `attr="null"`.

boris-petrov:
Thanks for taking the time to do this! But I'm confused. This is valid syntax - `<FooBar value=12345 />`?! Also `<FooBar value=true />`? Never knew that. Is there official Ember documentation about that?

test/unit/rules/no-unnecessary-curly-literals-test.js
bmish: Hmm, was I wrong to suggest this should be expanded to other types of literals? What types of literals should be included in this rule?

lin-ll: Hmm I thought that strings should be the only literals applicable to this rule? I believe other literals (like number or boolean) do need to be wrapped in curly braces

boris-petrov: > Hmm I thought that strings should be the only literals applicable to this rule? I believe other literals (like number or boolean) do need to be wrapped in curly braces

That's what I also think. Or thought. I just tried it with numbers and booleans and... it works. :smile: Never knew that. But I'm not sure it's really recommended by the Ember community/core team. Perhaps it's frowned upon.

Also, note there is the [no-quoteless-attributes rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-quoteless-attributes.md).

bmish: Okay, someone can rename the rule to `no-unnecessary-curly-strings` and we can release it as a bug fix if that's best.

bmish: Can you folks please review the PR to rename the rule to `no-unnecessary-curly-strings` and confirm that it makes sense now? https://github.com/ember-template-lint/ember-template-lint/pull/2802?

PR #2800, opened 2023-02-08
no-unnecessary-curly-parens: allow parens around helper

raycohen:
See https://github.com/ember-template-lint/ember-template-lint/pull/2792#discussion_r1100088097

There is a difference between `{{some-helper}}` and `{{(some-helper)}}`, where the former passes a reference to the helper, while the later invokes the helper and passes a value. If a helper is passed params, it is always invoked.

This PR updates the rule to only consider parens around the statement redundant if it is being passed params.


bmish:
Can you add a note in the rule doc explaining why this is allowed?

PR #2802, opened 2023-02-09
Rename `no-unnecessary-curly-literals` rule to `no-unnecessary-curly-strings`

bmish:
Follow-up to fix my incorrect assumption that the rule could apply to all literals instead of just string literals: https://github.com/ember-template-lint/ember-template-lint/pull/2794#discussion_r1100660805

Releasing as a bug fix since the rule requires this fundamental fix to make sense.

CC: @boris-petrov @Shivareddy-Aluri @bertdeblock @lin-ll


lin-ll (approved):

docs/rule/no-unnecessary-curly-strings.md
```diff
@@ -0,0 +1,29 @@
...
+<FooBar class="btn" />
+```
+
+```hbs
+<FooBar class="btn">Hello</FooBar>
+```
+
+## References
+
+* Ember's [Helper Functions](https://guides.emberjs.com/release/components/helper-functions/) guide
```
bertdeblock, line 29: I think this can go?

docs/rule/no-unnecessary-curly-strings.md
```diff
@@ -0,0 +1,29 @@
+# no-unnecessary-curly-strings
+
+Strings need not be wrapped in the mustache expressions.
```
bertdeblock, line 3:
```suggestion
Strings need not be wrapped in curly braces.
```

Maybe? Since the rule name uses `curly`, and we also use `curly braces` in the error message.

bmish: Updated mention both.

docs/rule/no-unnecessary-curly-strings.md
bmish: Removed.

bertdeblock (approved):

docs/rule/no-unnecessary-curly-strings.md
bmish: It looks like we require one reference so I had to add back one generic link for now.

PR #2812, opened 2023-02-16
Export additional helpers for plugin support

rob-long:
Streamline helper export and add additional exported helpers from module index to support plugin development

  createErrorMessage
  isDasherizedComponentOrHelperName
  isRouteTemplate
  isValidConfigObjectFormat
  replaceNode


rob-long:
@bmish Small PR to export some helpers which are helpful for custom rules...

bmish:
Do you have a specific need for these?

If we're going to export more helpers, I want to make sure we are thoughtful about how they are named and organized, since they become part of our public API then.

PR #2818, opened 2023-02-23
Add `additionalNonSemanticTags` option to `require-presentational-children` rule

Techn1x:
affects `require-presentational-children`
* ~Adds `<i>` and `<svg>` to the list of non-semantic tags - these will be allowed as children of elements like "button" without needing role="presentation" added~ https://github.com/ember-template-lint/ember-template-lint/pull/2821
* Adds config option `nonSemanticTags` to add additional tags that should also be considered non-semantic / presentation

---

This is a very common pattern in my codebase full of tables. I'd like to use the config to add `FaIcon` to the list, to avoid the need to add role="presentation" all the time.
```hbs
<th role="button" {{on "click" this.doSort}}>
  <div class="flex items-center justify-between">
    Some Column
    <FaIcon role="presentation" @icon="sort" />
  </div>
</th>
```

### DIFF
--- a/docs/rule/require-presentational-children.md
+++ b/docs/rule/require-presentational-children.md
@@ -58,7 +58,7 @@ If violations are found, remediation should be planned to either add `role="pres
  The following values are valid configuration:
 
 * object -- An object with the following keys:
-  * `nonSemanticTags` -- An array of additional tags that should be considered presentation
+  * `additionalNonSemanticTags` -- An array of additional tags that should be considered presentation
 
 ## References
 
--- a/docs/rule/require-presentational-children.md
+++ b/docs/rule/require-presentational-children.md
@@ -53,6 +53,13 @@ This rule **allows** the following:
 
 If violations are found, remediation should be planned to either add `role="presentation"` to the descendants as a quickfix. A better fix is to not use semantic descendants.
 
+## Configuration
+
+ The following values are valid configuration:
+
+* object -- An object with the following keys:
+  * `nonSemanticTags` -- An array of additional tags that should be considered presentation
+
 ## References
 
 * [Roles That Automatically Hide Semantics by Making Their Descendants Presentational](https://w3c.github.io/aria-practices/#children_presentational)
--- a/lib/rules/require-presentational-children.js
+++ b/lib/rules/require-presentational-children.js
@@ -72,6 +72,10 @@ const SKIPPED_TAGS = new Set([
 ]);
 
 export default class RequirePresentationalChildren extends Rule {
+  parseConfig(config) {
+    return { additionalNonSemanticTags: config.additionalNonSemanticTags ?? [] };
+  }
+
   visitor() {
     return {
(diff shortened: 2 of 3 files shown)


lib/rules/require-presentational-children.js
```diff
@@ -62,11 +62,13 @@ const NON_SEMANTIC_TAGS = new Set([
   'strike',
   'tt',
   'u',
+  'i',
+  'svg',
```
bmish, line 66: Is updating this list a breaking change? Either way, let's make any changes to the list in a separate PR from adding the configuration option.

docs/rule/require-presentational-children.md
```diff
@@ -53,6 +53,13 @@ This rule **allows** the following:
 
 If violations are found, remediation should be planned to either add `role="presentation"` to the descendants as a quickfix. A better fix is to not use semantic descendants.
 
+## Configuration
+
+ The following values are valid configuration:
+
+* object -- An object with the following keys:
+  * `nonSemanticTags` -- An array of additional tags that should be considered presentation
```
bmish, line 61: Probably we should name this `additionalNonSemanticTags` similar to `additionalInteractiveTags` in https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-invalid-interactive.md#configuration

bmish:
Need tests.

test/unit/rules/require-presentational-children-test.js
```diff
@@ -25,6 +25,10 @@ generateRuleTests({
       <title>Title here</title>
       <circle cx="10" cy="10" r="10"></circle>
     </svg>`,
+    {
+      config: { additionalNonSemanticTags: 'custom-element' },
```
bmish, line 29: Isn't `additionalNonSemanticTags` supposed to be an array?

lib/rules/require-presentational-children.js
```diff
@@ -92,8 +96,11 @@ export default class RequirePresentationalChildren extends Rule {
         const children = getAllElementNodeDescendants(node);
 
         for (const child of children) {
+          // additional non semantic tags
+          const { additionalNonSemanticTags } = this.config;
+          const nonSemantic = new Set([...NON_SEMANTIC_TAGS, ...additionalNonSemanticTags]);
```
bmish, line 101: Will this crash if a future option is passed and `additionalNonSemanticTags` is left as undefined?

test/unit/rules/require-presentational-children-test.js
Techn1x: Oops! I wrote this at 2am in my timezone, silly me. Will fix.

lib/rules/require-presentational-children.js
Techn1x: shouldn't crash, parseConfig handler above should cater for that

bmish (approved):
Thanks!

Issue #2637, opened 2022-10-12
require-context-role incorrectly requires elements to be direct children of their context element


robbytx:
Given the example markup for a [Bootstrap nav](https://getbootstrap.com/docs/5.2/components/navs-tabs/#javascript-behavior):
```html
<ul class="nav nav-tabs" id="myTab" role="tablist">
  <li class="nav-item" role="presentation">
    <button class="nav-link active" id="home-tab" data-bs-toggle="tab" data-bs-target="#home-tab-pane" type="button" role="tab" aria-controls="home-tab-pane" aria-selected="true">Home</button>
  </li>
  <li class="nav-item" role="presentation">
    <button class="nav-link" id="profile-tab" data-bs-toggle="tab" data-bs-target="#profile-tab-pane" type="button" role="tab" aria-controls="profile-tab-pane" aria-selected="false">Profile</button>
  </li>
  <li class="nav-item" role="presentation">
    <button class="nav-link" id="contact-tab" data-bs-toggle="tab" data-bs-target="#contact-tab-pane" type="button" role="tab" aria-controls="contact-tab-pane" aria-selected="false">Contact</button>
  </li>
  <li class="nav-item" role="presentation">
    <button class="nav-link" id="disabled-tab" data-bs-toggle="tab" data-bs-target="#disabled-tab-pane" type="button" role="tab" aria-controls="disabled-tab-pane" aria-selected="false" disabled>Disabled</button>
  </li>
</ul>
<div class="tab-content" id="myTabContent">
  <div class="tab-pane fade show active" id="home-tab-pane" role="tabpanel" aria-labelledby="home-tab" tabindex="0">...</div>
  <div class="tab-pane fade" id="profile-tab-pane" role="tabpanel" aria-labelledby="profile-tab" tabindex="0">...</div>
  <div class="tab-pane fade" id="contact-tab-pane" role="tabpanel" aria-labelledby="contact-tab" tabindex="0">...</div>
  <div class="tab-pane fade" id="disabled-tab-pane" role="tabpanel" aria-labelledby="disabled-tab" tabindex="0">...</div>
</div>
```

`ember-template-lint@4.14.0` issues the errors:
```
  2:2  error  Use of presentation role on <li> detected. Semantic elements should not be used for presentation.  no-invalid-role
  5:2  error  Use of presentation role on <li> detected. Semantic elements should not be used for presentation.  no-invalid-role
[...]

PR #2819, opened 2023-02-23
Allow intermediate presentation role before context in `require-context-role` rule

Techn1x:
affects `require-context-role`
* no longer stops at the immediate parent when searching for context role, if that immediate parent has role "presentation" or "none".

Closes https://github.com/ember-template-lint/ember-template-lint/issues/2637

Various official docs (like MDN) all do this in their examples.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#example_1_navigation_menu

The spec even has an example doing it. 
https://w3c.github.io/aria/#example-14
```html
<ul role="tree">
  <li role="presentation">
	<a role="treeitem" aria-expanded="true">An expanded tree node</a>
  </li>
  …
</ul>
```

My interpretation of the spec is that it can be many levels deep. But I'm happy to be corrected on that

### DIFF
--- a/docs/rule/require-context-role.md
+++ b/docs/rule/require-context-role.md
@@ -4,7 +4,7 @@
 
 ## `<* role><* role /></*>`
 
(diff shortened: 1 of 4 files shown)


bmish (approved):
Thanks!

Issue #2817, opened 2023-02-22
`<li role="presentation">` should not cause `no-invalid-role`


Techn1x:
You can see examples of presentation role being recommended for `li` tags in a few places

MDN docs
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#example_1_navigation_menu
```html
<ul id="menu" role="menu" aria-labelledby="menubutton">
  <li role="presentation">
    <a role="menuitem" href="#description"> Description </a>
  </li>
  ...
```

Bootstrap docs
https://getbootstrap.com/docs/5.2/components/navs-tabs/#javascript-behavior

```html
<ul class="nav nav-tabs" id="myTab" role="tablist">
  <li class="nav-item" role="presentation">
    <button class="nav-link active" id="home-tab" data-bs-toggle="tab" data-bs-target="#home-tab-pane" type="button" role="tab" aria-controls="home-tab-pane" aria-selected="true">Home</button>
  </li>
  ...
```

Techn1x:
Similarly, and perhaps a separate issue - `menuitem` role as below should not violate `require-context-role` - it does not have to be an immediate descendent of group/menu/menubar if the parent role is presentation.

```html
<ul id="menu" role="menu" aria-labelledby="menubutton">
  <li role="presentation">
    <a role="menuitem" href="#description"> Description </a>
  </li>
```

bmish:
Related? #2637

Techn1x:
Yeah looks like that covers the second issue I mentioned in my comment, but not the main issue from the original post

It's definitely all related/adjacent pain points of menus with these rules though

PR #2820, opened 2023-02-23
Allow presentation role for `li` in `no-invalid-role` rule

Techn1x:
affects `no-invalid-role`
* removes `li` from the list of elements that are not allowed to have `role="presentation"`

It's pretty clear from docs and examples that `<li role="presentation">` is not only okay, but recommended in a lot of cases.

Closes https://github.com/ember-template-lint/ember-template-lint/issues/2817

Related PR https://github.com/ember-template-lint/ember-template-lint/pull/2819


bmish (approved):
Thanks!

PR #2821, opened 2023-02-23
Add `svg`/`i` non-semantic tags list in `require-presentational-children` rule

Techn1x:
affects `require-presentational-children`

These tags would almost certainly be used for presentation. I'm not tied to this though so feel free to correct me.

Shouldn't be a breaking change - it'll mean that users can remove `role="presentation"` from their `i` or `svg` tags that appear in roles that cannot support semantic children (eg buttons)

### DIFF
--- a/lib/rules/require-presentational-children.js
+++ b/lib/rules/require-presentational-children.js
@@ -62,6 +62,8 @@ const NON_SEMANTIC_TAGS = new Set([
   'strike',
   'tt',
   'u',
+  'i',
+  'svg',
 ]);
 
 const SKIPPED_TAGS = new Set([


bmish:
Can you find a source/supporting evidence for this change?

Techn1x:
Sadly no. Happy for this to wait for further input from others, or just close it.

bmish:
Okay. Open to accepting if we receive some justification for it. Will leave open for now.

Techn1x:
Going to close this one - I just added them to the list as I felt they seemed like they were missing, but I do not have anything to back that up, nor a need for the change.

PR #2828, opened 2023-03-01
Remove `role` requirement for `img` tags in `require-valid-alt-text`

### LINKED ISSUE #2190: Nonsensical addition of role in require-valid-alt-text
The [require-valid-alt-text rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/require-valid-alt-text.md) states (emphasis mine):

> An `` must have the `alt` attribute. It must have either meaningful text, or be an empty string. If it is an empty string, the `` element tag _must also have `role="presentation"` or `role="none"`_.

That last part makes no sense to me. Why is this rule forcing people to add roles that aren’t necessary? The image is already exposed as decoration/irrelevant, why remove the semantics? That doesn’t add anything I don’t think.

Am I missing something?

---

In addition, `` is invalid HTML according to the [W3C Nu Validator](https://validator.w3.org/nu/#textarea).

[Related twitter thread](https://twitter.com/ZoeBijl/status/1460575553065521159)

### LINKED ISSUE #2579: Should require-valid-alt-text rule require a role="presentation/none" when the alt text is blank like alt=""
I noticed the linter fails for [require-valid-alt-text](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/require-valid-alt-text.md) when adding alt="" without one of these roles. 

Based on W3C: "A null (empty) alt text should be provided (alt="") so that they can be ignored by assistive technologies, such as screen readers.

Screen readers also allow the use of WAI-ARIA to hide elements by using role="presentation". However, currently, this feature is not as widely supported as using a null alt attribute."

And you can see a number of decorative images with the alt="" attribute and no presentation/none role attribute.

Source: https://www.w3.org/WAI/tutorials/images/decorative/

So my question is, should it be changed to allow a simple alt="" without one of these two presentation/none roles added?

yannbertrand:
Affects `require-valid-alt-text`.

`role` should never be needed when using an `img` tag according to the [W3C Nu Validator](https://validator.w3.org/nu/#textarea).

Closes #2190 
Closes #2579.

### DIFF
--- a/lib/rules/require-valid-alt-text.js
+++ b/lib/rules/require-valid-alt-text.js
@@ -55,14 +55,11 @@ export default class RequireValidAltText extends Rule {
             }
           }
           const hasRole = AstNodeInfo.hasAttribute(node, 'role');
-          let roleAttr, roleValue;
           if (hasRole) {
-            roleAttr = AstNodeInfo.findAttribute(node, 'role');
-            if (roleAttr.value.type === 'TextNode') {
-              roleValue = roleAttr.value.chars;
-            } else {
-              roleValue = roleAttr.value;
-            }
+            this.logNode({
+              message: 'The `role` attribute should not be given in `<img>` tags',
+              node,
+            });
           }
 
           let hasSrcAttr = AstNodeInfo.hasAttribute(node, 'src');
@@ -73,20 +70,6 @@ export default class RequireValidAltText extends Rule {
               srcValue = srcAttr.value.chars;
             } else {
               srcValue = srcAttr.value;
-            }
-          }
-
-          // if the role value is a mustache statement we can not validate it
-          if (hasAltAttribute && hasRole && !roleValue.type) {
-            if (
-              ['none', 'presentation'].includes(roleValue.trim().toLowerCase()) &&
-              altValue !== ''
-            ) {
-              this.logNode({
-                message:
-                  'The `alt` attribute should be empty if `<img>` has `role` of `none` or `presentation`',
-                node,
-              });
             }
           }
 
@@ -118,18 +101,6 @@ export default class RequireValidAltText extends Rule {
                 }
               } else {
                 normalizedAltValue = null;
-              }
-
-              if (normalizedAltValue === '') {
-                if (['presentation', 'none'].includes(roleValue)) {
-                  return;
-                }
-                this.logNode({
-                  message:
-                    'If the `alt` attribute is present and the value is an empty string, `role="presentation"` or `role="none"` must be present',
-                  node,
-                });
-                return;
               }
 
               if (normalizedAltValue !== null) {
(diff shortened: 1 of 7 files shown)


MelSumner (changes requested):
Let's make this rule optional instead of removing it; it's weird but also the only way we can currently identify developer intent. It's also on the topic list for the next ARIA face to face.

yannbertrand:
> Let's make this rule optional instead of removing it; it's weird but also the only way we can currently identify developer intent.

Thanks for the review. What do you mean by making the rule optional?

Should I remove [these lines](https://github.com/ember-template-lint/ember-template-lint/pull/2828/files#diff-036baf1a0cc3501594ec83f25c8ff5e0b2f981203127e56adadb5adaed1b7427R57-R62)?

> It's also on the topic list for the next ARIA face to face.

Interesting! Is there a planning of these events we can find somewhere?

yannbertrand:
I've added a commit to accept `role` attributes on `img` tags but do not apply any validation on them, rather than denying them.

jaswilli:
I think the intended changes here have become more urgent with the recent [aria-query](https://github.com/A11yance/aria-query) updates. Now _both_ versions of an `` with an empty alt tag are error conditions:

empty alt with presentation role:
```bash
$ echo '<img src="src" alt="" role="presentation">' | npx ember-template-lint --
undefined
  1:0  error  Use of redundant or invalid role: presentation on <img> detected.  no-redundant-role
```

empty alt without role:
```bash
$ echo '<img src="src" alt="">' | npx ember-template-lint --
undefined
  1:0  error  If the `alt` attribute is present and the value is an empty string, `role="presentation"` or `role="none"` must be present  require-valid-alt-text
```

test/unit/rules/require-valid-alt-text-test.js
```diff
@@ -11,13 +11,16 @@ generateRuleTests({
     '<img alt="blah {{derp}}">',
     '<img aria-hidden="true">',
     '<img hidden>',
+    '<img alt="" src="zoey.jpg">',
     '<img alt="" role="none" src="zoey.jpg">',
     '<img alt="" role="presentation" src="zoey.jpg">',
     '<img alt="a stylized graphic of a female hamster" src="zoey.jpg">',
+    '<img alt="a stylized graphic of a female hamster" role="presentation" src="zoey.jpg">',
```
HeroicEric, line 18: `role="presentation"` and `role="none"` are not valid when `alt` is a non-empty string

lib/rules/require-valid-alt-text.js
```diff
@@ -76,20 +66,6 @@ export default class RequireValidAltText extends Rule {
...
-              ['none', 'presentation'].includes(roleValue.trim().toLowerCase()) &&
-              altValue !== ''
-            ) {
-              this.logNode({
-                message:
-                  'The `alt` attribute should be empty if `<img>` has `role` of `none` or `presentation`',
-                node,
-              });
-            }
-          }
```
HeroicEric, line 91: I don't think this should be removed. It is not valid to have a non-empty alt attribute if `role="presentation"` or `role="none"`

yannbertrand:
I'm closing this in favor of #2954

Issue #2826, opened 2023-03-01
embedded templates indented incorrectly


tylerbecks:
I recently bumped ember-template-lint from v4 to v5.

I ran `yarn ember-template-lint tests/ --fix` and it formatted all embedded templates in our test files.

Before the change, our code looked like this:
```js
module('Integration | Component | recruiting-activities', (hooks) => {
  test('it renders', async (assert) => {
    await render(hbs`<EmberTsProfile$RecruitingActivities @activeContracts={{this.activeContracts}} @inactiveContracts={{this.inactiveContracts}} @profileId={{this.profileId}} @profile={{this.profile}} @selectedContract={{this.selectedContract}} />`);

    assert.dom(recruitingActivitiesSelector).exists('Recruiting Activities component renders');
  });
});
```

After running --fix, it looks like this:
```js
module('Integration | Component | recruiting-activities', (hooks) => {
  test('it renders', async (assert) => {
    await render(hbs`<EmberTsProfile$RecruitingActivities
  @activeContracts={{this.activeContracts}}
  @inactiveContracts={{this.inactiveContracts}}
  @profileId={{this.profileId}}
  @profile={{this.profile}}
  @selectedContract={{this.selectedContract}}
/>`);

    assert.dom(recruitingActivitiesSelector).exists('Recruiting Activities component renders');
  });
});
```

As you can see, the embedded template is formatted as if there is no javascript indentation. 

I expected the output to format like this:
```js
module('Integration | Component | recruiting-activities', (hooks) => {
  test('it renders', async (assert) => {
    await render(hbs`<EmberTsProfile$RecruitingActivities
      @activeContracts={{this.activeContracts}}
      @inactiveContracts={{this.inactiveContracts}}
      @profileId={{this.profileId}}
      @profile={{this.profile}}
      @selectedContract={{this.selectedContract}}
    />`);

    assert.dom(recruitingActivitiesSelector).exists('Recruiting Activities component renders');
  });
});
```

dwickern:
If you're having this problem and using the ember-template-lint-plugin-prettier plugin, see https://github.com/ember-template-lint/ember-template-lint-plugin-prettier/issues/290

lifeart:
Closing issue, but it's still listed here: https://github.com/ember-template-lint/ember-template-lint/issues/3209

PR #2831, opened 2023-03-03
Skip `block-indentation` rule in embedded templates

robinborst95:
The issue with the `block-indentation` rule in embedded templates as reported in https://github.com/ember-template-lint/ember-template-lint/issues/2826 is pretty annoying in existing tests. Turning off the rule entirely is not ideal, as it's really valuable to keep linting this. So, this PR ~adds an option to skip embedded templates entirely, basically keeping the functionality of the v4 version until a proper solution can be found.~ disables the rule in embedded templates for now. I have attempted to fix it fully in https://github.com/ember-template-lint/ember-template-lint/pull/2838, but I can imagine that might need some iterating before it can land. 

I'm not sure if this will be considered a proper way of fixing it, but this is close to being a blocker for me to upgrade to v5, and it's a small change with a great effect :pray: .

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -242,6 +242,7 @@ export default class Linter {
           filePath: options.filePath,
           rawSource: templateInfo.template,
           fileConfig,
+          isStrictMode: templateInfo.isStrictMode,
         });
 
         let visitor = await rule.getVisitor();
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -108,6 +108,10 @@ function removeWhitespaceEnd(str) {
 
 export default class BlockIndentation extends Rule {
   parseConfig(config) {
+    if (!this.isStrictMode) {
+      return false;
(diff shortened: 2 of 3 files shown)


lib/linter.js
```diff
@@ -373,6 +373,10 @@ export default class Linter {
       results.push(...failures);
 
       for (let rule of rules) {
+        if (templateInfo.isEmbedded && rule.config.skipEmbedded) {
```
robinborst95, line 376: I didn't see how `isEmbedded` can be accessed from within a rule, and I didn't know what would be a good way of making that work. It also feels more sensible to not visit the rule at all in this case, because you basically want to skip the entire rule conditionally. Finally, there are probably more rules that are affected by this issue (e.g. `attribute-indentation`), so then such rules can potentially also get this option if wanted.

robinborst95:
Closing this since #2838 has landed instead

Issue #2832, opened 2023-03-03
Embedded templates incorrectly violate `no-trailing-spaces` rule


robinborst95:
Embedded templates with the ending backtick on a newline cause a violation of the `no-trailing-spaces` rule, which feels incorrect. Example:
```js
test('it renders', async (assert) => {
  await render(hbs`
    <div class="parent">
      <div class="child"></div>
    </div>
  `);
);
```

PR #2833, opened 2023-03-03
Fix incorrect violation with embedded templates in `no-trailing-spaces` rule

robinborst95:
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/2832. At first I thought `templateInfo.contents` => `templateInfo.contents.trimEnd()` in `extract-templates` would be the solution (it does fix this added test), but it made other tests fail for good reasons it seemed. So, after fixing another rule, I cherry-picked the `columnOffset` changes from https://github.com/ember-template-lint/ember-template-lint/pull/2838, because this made it quite trivial to fix! I'm curious to hear anyone's thoughts on this (and the other PR) :)


bmish (approved):
Seems reasonable, thanks!

PR #2838, opened 2023-03-08
Fix embedded templates handling in `block-indentation` rule 

robinborst95:
This partially fixes the issue described in https://github.com/ember-template-lint/ember-template-lint/issues/2826. The example in that issue is a combination of `block-indentation` and `attribute-indentation` I think, and this PR only fixes the `block-indentation`, which I guess is more important to fix than attribute indentation. I've added some examples to the tests that work now (including autofix), but I'm happy to get feedback on other scenarios that need to be tested as well. An example of what this PR fixes:

```js
import { hbs } from 'ember-cli-htmlbars';

test('it renders', async (assert) => {
  await render(hbs`
    <div class="parent">
      <div class="child"></div>
    </div>
  `);
);
```

Before this PR, the linter would only be happy if it were formatted like:
```js
import { hbs } from 'ember-cli-htmlbars';

test('it renders', async (assert) => {
  await render(hbs`
<div class="parent">
  <div class="child"></div>
</div>
  `);
);
``` 
which is not very pretty and intuitive.

### DIFF
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -102,8 +102,12 @@ function get(node, path) {
   return value;
 }
 
-function removeWhitespaceEnd(str) {
-  return str.replace(/ +$/, '');
+function addWhitespaceEnd(str, count) {
+  return `${str}${' '.repeat(count)}`;
+}
+
+function removeWhitespaceEnd(str, count) {
+  return str.replace(new RegExp(` {${count}}$`), '');
 }
 
 export default class BlockIndentation extends Rule {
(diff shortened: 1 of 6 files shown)


lib/rules/block-indentation.js
```diff
@@ -477,7 +494,7 @@ export default class BlockIndentation extends Rule {
 
           fixedNode = this.fixLine(
             lines,
-            childStartLine - 1,
+            childStartLine - (path ? 1 : fixedNode.loc.start.line),
```
robinborst95, line 497: This actually seemed to be an existing bug. When I ran the autofixer with `master` on our own embedded templates, then I got `TypeError: Cannot read property 'startsWith' of undefined` because the `childStartLine` was wrong

test/unit/rules/block-indentation-test.js
```diff
@@ -1327,5 +1429,255 @@ generateRuleTests({
...
+        '    </div>',
+        '  `);',
+        ');',
+      ].join('\n'),
+      meta: {
+        filePath: 'layout.js',
+      },
+
+      verifyResults(results) {
+        expect(results).toMatchInlineSnapshot(`
```
robinborst95, line 1460: Creating these snapshots was not trivial to do, is there a command for this?

bmish: `jest --updateSnapshot` ?

bmish (approved):
Seems reasonable, thanks! What part of https://github.com/ember-template-lint/ember-template-lint/issues/2826 is still not fixed? CC: @tylerbecks

robinborst95:
Thanks for merging! The part of #2826 that is not fixed is actually the first example mentioned there. I think it's the `attribute-indentation` rule that does this, and that's not fixed with this PR of course. However, when the `attribute-indentation` is not enabled or not violated, the same issue with indentation would occur, so that's why I described it as partially fixed.

Issue #2798, opened 2023-02-06
`no-invalid-link-text` error with handlebars `aria-label`


Yelinz:
Example
```html
<a aria-label={{this.text}}/>
```
When using an aria-label with the value not set by a plain string it throws this error
```
TypeError: Cannot read properties of undefined (reading 'replace')
```
Which stems from this line
https://github.com/ember-template-lint/ember-template-lint/blob/29f240a7ab30a4ea88652c81c73c9331916b9a2d/lib/rules/no-invalid-link-text.js#L14

The problem is probably trying to use `.chars`
https://github.com/ember-template-lint/ember-template-lint/blob/29f240a7ab30a4ea88652c81c73c9331916b9a2d/lib/rules/no-invalid-link-text.js#L38

Log of `ariaLabelAttr` with  `<a aria-label={{t "translation"}}>`
```js
{
  type: 'AttrNode',
  name: 'aria-label',
  value: {
    type: 'MustacheStatement',
    path: PathExpressionImplV1 {
      original: 't',
      loc: [SourceSpan],
      type: 'PathExpression',
      this: false,
      data: false,
      _head: undefined,
      parts: [Array]
    },
    params: [ [Object] ],
    hash: { type: 'Hash', pairs: [], loc: [SourceSpan] },
    escaped: true,
    trusting: false,
    loc: SourceSpan { data: [HbsSpan], isInvisible: false },
    strip: { open: false, close: false }
  },
  loc: SourceSpan {
    data: HbsSpan {
      source: [Source],
      hbsPositions: [Object],
      kind: 'HbsPosition',
      _charPosSpan: null,
      _providedHbsLoc: null
    },
    isInvisible: false
  },
  isValueless: false,
  quoteType: null
}
```

PR #2843, opened 2023-03-14
Avoid crash with dynamic `aria-label` in `no-invalid-link-text` rule

StephanH90:
Since MustacheStatements can return a valid aria string but don't respond to the regular string checks, we need to exit the rule early.

Fixes https://github.com/ember-template-lint/ember-template-lint/issues/2798


bmish (approved):
Thanks!

I've updated the PR to use the `Fixes #xxx` syntax to correctly link to the issue.

PR #2844, opened 2023-03-14
Disable `no-curly-component-invocation` and `no-implicit-this` rules for `gjs` / `gts` files (#2844)Co-authored-by: Bryan Mishkin <698306+bmish@users.noreply.github.com>

### LINKED ISSUE #2830: Should we start a Polaris config that folks can opt in to when using gjs/gts?
I think we need to change defaults when using gjs/gts, for example

several rules throw false negatives because templates can reference things imported in JS, but not defined in the template. for example:
```js
import { overInvalidatingClock, clock } from './-resources';

const SomeClocks = <template>
  {{clock}}
  {{overInvalidatingClock}}
</template>;
```
this throws errors for the following rules:
 - `no-curly-component-invocation`
 - `no-implicit-this`
 

the behavior of checking if locals are defined or not is handled be eslint / eslint-plugin-ember via transform.

So, for ember-template-lint and a `polaris` config, I think we should start disabling some rules specifically for gjs/gts files -- and we could use this for the polaris config:

```js
'use strict';

module.exports = {
  extends: 'recommended',
  overrides: [
    {
      files: ['**/*.gjs', '**/*.gts'],
      rules: {
        'no-curly-component-invocation': 'off',
        'no-implicit-this': 'off',
      },
    },
  ],
};
```

Related: https://github.com/ember-template-lint/ember-template-lint/issues/2786

NullVoxPopuli:
Fixes #2830

Normally this would be a breaking change under any other circumstances, but gjs/gts folks need to disable these rules anyway, so this is more a bugfix.

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -91,4 +91,13 @@ export default {
     'style-concatenation': 'error',
     'table-groups': 'error',
   },
+  overrides: [
+    {
+      files: ['**/*.gjs', '**/*.gts'],
+      rules: {
+        'no-curly-component-invocation': 'off',
+        'no-implicit-this': 'off',
+      },
+    },
+  ],
 };
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -95,17 +95,17 @@ export default {
     {
       files: ['**/*.gjs', '**/*.gts'],
       rules: {
-        // in gjs/gts, curly-component-invocation, and no-implicit-this were incorrectly triggered, 
+        // in gjs/gts, curly-component-invocation, and no-implicit-this were incorrectly triggered,
         // because {{foo}}, which is ambiguous in loose-mode templates could have "foo" defined in JS scope.
         // For example:
-        // 
+        //
         //  const foo = "hello there";
         //  <template>{{foo}}</template>
-        // 
+        //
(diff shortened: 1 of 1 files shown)


lib/config/recommended.js
```diff
@@ -91,4 +91,13 @@ export default {
     'style-concatenation': 'error',
     'table-groups': 'error',
   },
+  overrides: [
+    {
+      files: ['**/*.gjs', '**/*.gts'],
+      rules: {
+        'no-curly-component-invocation': 'off',
+        'no-implicit-this': 'off',
```
bmish, line 99: Mind adding a comment next to each of these rules explaining why they are disabled for gjs/gts?

NullVoxPopuli: updated!

bmish: Lint failing.

NullVoxPopuli: that's what I get for trying to do this all in-browser :upside_down_face:

ynotdraw:
For some strange reason I'm still having to override this in an app using `gts`/`gjs` https://github.com/CrowdStrike/ember-toucan-core/pull/29/files#diff-47149a39ab889db56d9810a099412dea44af3fa4b519ac990f8841c10c000ad6.  Am I missing something?

Issue #2660, opened 2022-10-26
embedded templates not working properly 


vagarii:
Hi there, I tried using ember-template-lint for linting embedded templates(inside .js not .hbs) but it seems like it didn't work. Wondering if it's ignore js file so I did add `--no-ignore-pattern` but still no luck. The output is just `done in xxx second` but I am pretty sure the files I am working on have lint issue. Can you guys help me with this, does the `embedded templates` unstable since it's quite new feature?

Here is my current version: 
- "ember-template-lint": "^4.15.1"
- "node": "14.17.5"

cc: @tstewart15

tstewart15:
Hi team, to provide further details, the support for linting embedded hbs in .js files was added in #2483 by @rwjblue and @ventuno 

We tried to pull this in as we're eager to get template linting in our GlimmerX applications (learning-guest-frontend and learning-subscription-frontend for you LinkedIn folks), but are unable to get it working. We're thinking it has something to do with the default behavior of [ignoring JS files](https://github.com/ember-template-lint/ember-template-lint/pull/2483/files#diff-11b3352c70fabcab37f5c8836ae8a377af62daa7907dc04f0f0a9a133a5d44d1), so we tried using `--no-ignore-pattern` (which takes forever since it lints EVERYTHING in the project) and then overriding `--ignore-pattern` by doing `yarn ember-template-lint . --ignore-pattern "**/dist/*,*/tmp/**,**/node_modules/**"`.

Are you able to get the .js embedded template linting to work in GlimmerX? If so, how?

ventuno:
@tstewart15, @vagarii: you should be able to pass the list of patterns to ignore by passing the `--ignore-pattern` flag multiple times (notice the `'` around `**/tmp/**`):
```
ember-template-lint --config '{ "rules": { "no-debugger": "error" } }' --ignore-pattern **/dist/** --ignore-pattern **/node_modules/** --ignore-pattern '**/tmp/**' .
```

This is required because [`yargs` (the parser used by process optstrings) runs with the `greedy-arrays` flag set to `false`](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/helpers/cli.js#L148).

tstewart15:
Thanks, will give that a shot! @ventuno Curious why only the "tmp" item has `'` around it? Or are you saying the final item `'` are because of the greedy-arrays flag?

ventuno:
[...]

PR #2845, opened 2023-03-14
Recognize imports from '@glimmerx/component'

tstewart15:
This should resolve https://github.com/ember-template-lint/ember-template-lint/issues/2660


bmish (approved):
Thanks!

PR #2846, opened 2023-03-15
Disable embedded templates handling for `eol-last` rule (#2846

robinborst95:
In our test suite this is the last remaining (enabled) rule that provides incorrect results in embedded templates. Enforcing a newline at the end of an embedded template doesn't make much sense to me, so disabling the rule seems the easiest solution. You could argue to set the config to `"never"` instead of disabling completely, but that requires more changes and in the end it's not what this rule is meant for (which is having a newline or not at the end of a **file**, rather than at the end of a **template**).

### DIFF
--- a/lib/rules/eol-last.js
+++ b/lib/rules/eol-last.js
@@ -5,6 +5,12 @@ import Rule from './_base.js';
 
 export default class EolLast extends Rule {
   parseConfig(config) {
+    // In strict mode (= template is embedded in e.g. a JS file) we want to disable this rule,
+    // because it is not meant for templates specifically, but for template files.
+    if (!this.isStrictMode) {
+      return false;
+    }
+
     let configType = typeof config;
 
     switch (configType) {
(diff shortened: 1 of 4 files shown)


test/unit/rules/eol-last-test.js
```diff
@@ -43,6 +43,19 @@ generateRuleTests({
       config: 'always',
       template: '{{#my-component}}{{/my-component}}\n',
     },
+    {
```
bmish, line 46: Can you add a comment in the test case explaining why the rule is off in this case.

lib/rules/eol-last.js
```diff
@@ -5,6 +5,10 @@ import Rule from './_base.js';
 
 export default class EolLast extends Rule {
   parseConfig(config) {
+    if (!this.isStrictMode) {
```
bmish, line 8: Can you add a comment here explaining why the rule is off here and what strict mode means?

bmish, line 8: Let's also add a note to the rule doc that this rule doesn't apply to embedded templates?

robinborst95: I've adapted the rule doc as well! I figured the rule doc doesn't need an explanation, as it already said `files` specifically, which I made even more explicit by saying `template files` instead.

bmish (approved):
Thanks!

PR #2887, opened 2023-05-18
Add new rule `simple-modifiers`

### LINKED ISSUE #2863: New rule: `simple-modifier`
Modifiers should have the string as the first param to make things less convoluted and prevent issues with other rules.

Not good:

```
<div {{(modifier (unless this.hasBeenClicked "track-interaction") "click" customizeData=this.customizeClickData)}}>
  ...
</div>
```

Good:

```
<div {{(unless this.hasBeenClicked (modifier "track-interaction" "click" customizeData=this.customizeClickData))}}>
  ...
</div>
```

rmonzon:
Fixes #2863.

Modifiers should have the string as the first param to make things less convoluted and prevent issues with other rules.

Not good:

```
<div {{(modifier (unless this.hasBeenClicked "track-interaction") "click" customizeData=this.customizeClickData)}}>
  ...
</div>
```

Good:

```
<div {{(unless this.hasBeenClicked (modifier "track-interaction" "click" customizeData=this.customizeClickData))}}>
  ...
</div>
```


docs/rule/simple-modifiers.md
```diff
@@ -0,0 +1,34 @@
+# simple-modifiers
+
+This rule strongly advises against using the `{{modifier}}` helper with its first argument being either:
+
+- a string literal (.e.g. `{{modifier "tracking-interaction"}}`)
+- a variable in the template's JS backing class context (e.g. `{{modifier this.trackingInteraction}}`)
+
+A common misuse of this rule is to apply the modifier conditionally `{{(modifier (unless this.hasBeenClicked "track-interaction") "click" customizeData=this.customizeClickData)}}`, which technically is not wrong, since the helper `modifier` simply ignores `null` and `undefined`, producing a no-op modifier.
```
chrisrng, line 8:
```suggestion
A common issue this rule will catch is declaring the modifier's name conditionally like `{{(modifier (unless this.hasBeenClicked "track-interaction") "click" customizeData=this.customizeClickData)}}`. This is technically correct since the helper `modifier` simply ignores `null` and `undefined` values producing a no-op modifier but it is confusing. Ideally the we should instead do `{{(unless this.hasBeenClicked (modifier "track-interaction" "click" customizeData=this.customizeClickData))}}` where it's clear the modifier will not get called.
```

Trying to make it easier to understand the motivation here ^

test/unit/rules/simple-modifiers-test.js
```diff
@@ -0,0 +1,38 @@
+import generateRuleTests from '../../helpers/rule-test-harness.js';
+
+generateRuleTests({
+  name: 'simple-modifiers',
+
+  config: true,
+
+  good: [
```
bmish, line 8: Can you add some valid test cases to ensure that the things that are disallowed with modifiers are still allowed with other elements/helpers? We need to make sure this rule limits its affects to modifiers only.

rmonzon: Added a few other examples covering other helpers as well as custom modifiers that do not use the `modifier` helper explicitly.

lin-ll:
I think the checks are still mad about the alphabetical order of the rules in the README. Did you use this command to update it? `npm run update:readme`

docs/rule/simple-modifiers.md
```diff
@@ -0,0 +1,34 @@
+# simple-modifiers
+
+This rule strongly advises against using the `{{modifier}}` helper with its first argument being either:
+
+- a string literal (.e.g. `{{modifier "tracking-interaction"}}`)
+- a variable in the template's JS backing class context (e.g. `{{modifier this.trackingInteraction}}`)
```
bmish, line 6: Shouldn't there be two invalid test cases, one for each of these?

lin-ll: These are the valid use cases - agree that the wording doesn't seem quite right here.

docs/rule/simple-modifiers.md
```diff
@@ -0,0 +1,34 @@
+# simple-modifiers
+
+This rule strongly advises against using the `{{modifier}}` helper with its first argument being either:
```
lin-ll, line 3:
```suggestion
This rule strongly advises against passing complex statements or conditionals to the first argument of the `{{modifier}}` helper. Instead, it recommends using the `{{modifier}}` helper with its first argument being either:
```

lib/rules/simple-modifiers.js
```diff
@@ -0,0 +1,36 @@
...
+export default class SimpleModifiers extends Rule {
+  visitor() {
+    return {
+      SubExpression(node) {
+        if (!this._isModifier(node)) {
+          return;
+        }
+
+        const firstModifierParam = node.params[0];
+        if (firstModifierParam) {
```
bmish, line 13: It looks like this line isn't fully covered by the tests according to `yarn test:jest --coverage`. I think you'll need to test a modifier with no parameters.

docs/rule/simple-modifiers.md
rmonzon: I've updated the wording with your suggestion, @lin-ll.

@bmish these are correct cases. I've updated the wording above to reflect that.

lib/rules/simple-modifiers.js
rmonzon: Added one more test for 100% coverage.

lin-ll (approved):

bmish (approved):
Thanks!

Issue #2785, opened 2023-01-23
`no-implicit-this` + strict mode false positives


Windvis:
I experimented with ember-template-imports again now that the ["Road to stable" issue](https://github.com/ember-template-imports/ember-template-imports/issues/35) has been marked as completed, but I'm running into a lot of no-implicit-this false positives. 

According to [some comments on the original PR](https://github.com/ember-template-lint/ember-template-lint/pull/2483#discussion_r858957036), it seems the intended behaviour is to disable this rule if it's a strict mode template.

However, [this check](https://github.com/ember-template-lint/ember-template-lint/blob/0b4983a496b40452d3747a921c062cc5fa50c336/lib/rules/no-implicit-this.js#L56) seems to be inversed? On quick glance it also seems that the `this.isStrictMode` value [isn't always correctly set either](https://github.com/ember-template-lint/ember-template-lint/blob/0b4983a496b40452d3747a921c062cc5fa50c336/lib/extract-templates.js#L96-L97). Normal .hbs files also seem to have `this.isStrictMode` set to true.

I added a failing test in [this PR](https://github.com/ember-template-lint/ember-template-lint/pull/2784).

Techn1x:
I have also hit this issue today.

With a glint file like below, `Foo` is flagged as an ambiguous reference

```gts
import type { TOC } from '@ember/component/template-only'
import { Foo } from 'my-app/foos/foo'
import SomeComponent from 'my-app/components/some-component'

export interface Signature {}

export const MyComponent: TOC<Signature> = <template>
<div>
  <SomeComponent @foo={{Foo}} />
</div>
```

NullVoxPopuli:
As a temporary work around, disabling the lint is an option.


I think i know how to make it work for gjs tho

NullVoxPopuli:
I had original thought that the build of gjs/gts files would protect us from needing this lint, but gjs/gts files are still compiled fine with missing references, deferring the error to boot-time (when modules start loading).

So this lint _does_ need fixing.

NullVoxPopuli:
also, the `no-implicit-this` lint rule was already disabled in this PR: https://github.com/ember-template-lint/ember-template-lint/pull/2844
released in: 5.7.1: https://github.com/ember-template-lint/ember-template-lint/blob/master/CHANGELOG.md#v571-2023-03-16

So we can close this issue, @bmish

Techn1x:
> As a temporary work around, disabling the lint is an option.

[...]

PR #2891, opened 2023-05-23
Fix override for gjs files in `recommended` config

NullVoxPopuli:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/2785
The perceived effect of this bug is that gjs and gts files do not have these rules disabled: https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/recommended.js#L94

The `recommended` config has `overrides`, but that `overrides` was not making it in to user-configs.

Using the default config,
```js
'use strict';

module.exports = {
  plugins: [],
  extends: 'recommended'
};
```

**Before**
```
❯ pnpm ember-template-lint --print-config app/components/limber/layout/controls/index.gts
...
  "overrides": [],
```

**After**
```
❯ pnpm ember-template-lint --print-config app/components/limber/layout/controls/index.gts
...
  "overrides": [
    {
      "files": [
        "**/*.gjs",
        "**/*.gts"
      ],
      "rules": {
        "no-curly-component-invocation": {
          "config": false,
          "severity": 0
        },
        "no-implicit-this": {
          "config": false,
          "severity": 0
        }
      }
    }
  ],

```

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -310,7 +310,7 @@ function processExtends(config) {
     }
 
     config.rules = Object.assign({}, extendedRules, config.rules);
-    config.overrides = [...extendedOverrides, ...config.overrides];
+    config.overrides = [...extendedOverrides, ...(config.overrides || [])];
   }
 }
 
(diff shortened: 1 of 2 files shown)


bmish (approved):
Looks reasonable. Thanks.

PR #2892, opened 2023-05-23
Adds rule to the recommended config

rmonzon:
Add submitted new rule https://github.com/ember-template-lint/ember-template-lint/pull/2887 to recommended config.


lin-ll:
Since adding a rule to the recommended config is a breaking change, we'll wait for the next major release to do this. Plans for the major release can be tracked/found here: https://github.com/ember-template-lint/ember-template-lint/issues/2712

bmish:
Yeah, thank you, but let's come back to this later as part of #2712.

Issue #2906, opened 2023-06-06
Block-indentation Reports Same Value for Start and End Location


judithhinlung:
Currently, the 'block-indentation' rule is reporting the same value for start and end locations for certain indentation issues.

```
# examples
  {{#each cats as |dog|}}
       {{/each}}
```

```hbs
<LongComponentName>
  <p>Some text goes here</p></LongComponentName>
```

Sample output:
           [
             {
              "column": 8,
               "endColumn": 8,
               "endLine": 2,
               "filePath": "layout.hbs",
               "isFixable": true,
              "line": 2,
               "message": "Incorrect indentation for \`div\` beginning at L1:C0. Expected \`</div>\` ending at L2:C8 to be at an indentation of 0 but was found at 2.",
               "rule": "block-indentation",
               "severity": 2,
  {{#each cats as |dog|}}
       {{/each}}
             }
           ]

This bug could cause static analysis tools to throw an exception when using  Ember-template-lint to scan for issues.

PR #2907, opened 2023-06-06
Fix location reporting in `block-indentation` rule

judithhinlung:
### Background
This PR fixes [#2906](https://github.com/ember-template-lint/ember-template-lint/issues/2906
), a bug in the `block-indentation` rule in which the same value is reported for start and end line locations.

```hbs
  {{#each cats as |dog|}}
       {{/each}}

**Before**:
             {
              "column": 8,
               "endColumn": 8,
               "endLine": 2,
               "filePath": "layout.hbs",
               "isFixable": true,
              "line": 2,
...

**After**:
                "column": 0,
               "endColumn": 8,
               "endLine": 2,
               "filePath": "layout.hbs",
               "isFixable": true,
                "line": 1,
...

### Testing
Test Suites: 148 passed, 148 total
Tests:       5 skipped, 8517 passed, 8522 total
Snapshots:   1106 passed, 1106 total
Time:        139.666 s
Ran all test suites.
✨  Done in 174.15s.

### DIFF
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -393,8 +393,8 @@ export default class BlockIndentation extends Rule {
         this.log({
           message: warning,
           node,
-          line: node.loc.end.line,
-          column: node.loc.end.column,
+          line: node.loc.start.line,
+          column: node.loc.start.column,
           isFixable: true,
         });
       }
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -393,8 +393,6 @@ export default class BlockIndentation extends Rule {
         this.log({
           message: warning,
           node,
-          line: node.loc.start.line,
-          column: node.loc.start.column,
           isFixable: true,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/rules/block-indentation.js [hunk 1]
```diff
@@ -393,8 +393,8 @@ export default class BlockIndentation extends Rule {
         this.log({
           message: warning,
           node,
-          line: node.loc.end.line,
-          column: node.loc.end.column,
+          line: node.loc.start.line,
+          column: node.loc.start.column,
```
lin-ll, line 397: Quick question - could you just remove these two lines? I think by default it should try to take the `line` and `column` from `node.loc.start`

judithhinlung: Thanks so much for calling this out @lin-ll! Yes removing the two lines also resolves the reporting error and in a cleaner way too.

lin-ll (approved):

PR #2913, opened 2023-06-13
Add additional elements to `DOCUMENT_STRUCTURE_ROLES` for `no-invalid-role` rule

mrloop:
See document structure roles here
https://www.w3.org/TR/wai-aria/#document_structure_roles
https://github.com/w3c/aria/blob/main/index.html#L789


lib/rules/no-invalid-role.js
```diff
@@ -141,28 +141,47 @@ const COMPOSITE_WIDGET_ROLES = new Set([
 const DOCUMENT_STRUCTURE_ROLES = new Set([
```
bmish, line 141: Can you add a comment above this array with the URL this list is supposed to come from?

bmish (commented):
Is this a breaking change? If so, we can include it in next major release.

lib/rules/no-invalid-role.js
mrloop: Good idea, added!

MelSumner (approved):

PR #2914, opened 2023-06-14
Fix `aria-query` related crash in `no-unsupported-role-attributes` rule

bmish:
Fixes failing tests with the latest version of aria-query.

Fixes #2912.

PR #2915, opened 2023-06-16
Update incorrect tests

ef4:
This builds off https://github.com/ember-template-lint/ember-template-lint/pull/2914.

I'm pretty sure these tests were incorrect and it was just hidden by the other bug.

The checkbox role *does* support aria-invalid. So I changed that one to button role, which does not.

Also, the menuitem element (not to be confused with the menuitem role) does not appear at all in the set of elementRoles from aria-query. So I changed that test to use `<a role="menuitem">` instead. The menuitem element is [deprecated and non-standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem).

Also note that the "command" role is abstract and not supposed to be used by authors, which is why I replaced it with the "menuitem" role, which is one of the concrete children of "command".

### DIFF
--- a/lib/rules/no-unsupported-role-attributes.js
+++ b/lib/rules/no-unsupported-role-attributes.js
@@ -23,7 +23,8 @@ function getImplicitRole(element, typeAttribute) {
       }
     }
   }
-  let implicitRoles = elementRoles.get(elementRoles.keys().find((key) => key.name === element));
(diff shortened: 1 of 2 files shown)


bmish (approved):
Awesome, thanks! Merging on top of my PR.

PR #2917, opened 2023-06-17
Fix crash in `no-unsupported-role-attributes` rule

### LINKED ISSUE #2916: attributes is not iterable error when using aria-query 5.2.1
This error is similar to https://github.com/ember-template-lint/ember-template-lint/issues/2912 but the the manifestation is different, where this error comes from an input with a type that doesn't match any of aria-query's mappings. A minimal reproduction is simply:

```bash
echo '<input type="password">' | npx ember-template-lint --verbose - --
  -:-  error  attributes is not iterable  undefined
TypeError: attributes is not iterable
    at getImplicitRole (file:///Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/ember-template-lint/lib/rules/no-unsupported-role-attributes.js:18:31)
    at NoUnsupportedRoleAttributes.ElementNode (file:///Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/ember-template-lint/lib/rules/no-unsupported-role-attributes.js:52:18)
    at file:///Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/ember-template-lint/lib/rules/_base.js:323:46
    at visitNode (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/@glimmer/syntax/dist/commonjs/es2017/lib/traversal/traverse.js:86:14)
    at visitArray (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/@glimmer/syntax/dist/commonjs/es2017/lib/traversal/traverse.js:179:18)
    at visitKey (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/@glimmer/syntax/dist/commonjs/es2017/lib/traversal/traverse.js:155:5)
    at visitNode (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/@glimmer/syntax/dist/commonjs/es2017/lib/traversal/traverse.js:107:7)
    at traverse (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/@glimmer/syntax/dist/commonjs/es2017/lib/traversal/traverse.js:220:3)
    at transform (/Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/ember-template-recast/lib/index.js:68:27)
    at Linter.verify (file:///Users/jtw/.npm/_npx/279f7d8bf6868ccd/node_modules/ember-template-lint/lib/linter.js:381:11)
```

ember-template-lint's behavior prior to aria-query@5.2.1 was that inside `getImplicitRole` all of the `key.attributes` were iterable, they just didn't match and the return clause was never hit and we got the default behavior, which was to return a `button`. With how the data has changed in 5.2.1 some of the keys have no attributes defined so we get the error above.

Here's the relevant section of code: https://github.com/ember-template-lint/ember-template-lint/blob/516d5318299af3caa337f1b51d3673aa6eb34f85/lib/rules/no-unsupported-role-attributes.js#L13-L29

I'm happy to work up a PR for a workaround but I'm a little confused about whether or not the current behavior where an `input` with a `typeAttribute` that doesn't match any of the aria-query element role mappings gets "defaulted" to a `button` is correct? Following the "it's gotten us this far" rule I suppose staying the course is fine and we can just keep returning `button` but if that existing behavior is a bug and what we really want is to fall back to something like an input of type text then I'm also happy to make that change.

jaswilli:
Resolves #2916.

Handle the case where aria-query's role mapping data does not have an attributes property defined.

Note: as I mentioned in #2916 I'm a little confused by the current behavior where an `<input>` with a `type` that doesn't match anything sort of defaults to being treated like a `<button>`. If that's correct (or correct enough) then this should work, but if that's also a bug then we'll want a different solution here.

### DIFF
--- a/lib/rules/no-unsupported-role-attributes.js
+++ b/lib/rules/no-unsupported-role-attributes.js
@@ -13,7 +13,7 @@ function createUnsupportedAttributeErrorMessage(attr, role, element) {
 function getImplicitRole(element, typeAttribute) {
   if (element === 'input') {
     for (let key of elementRoles.keys()) {
-      if (key.name === element) {
+      if (key.name === element && key.attributes) {
         let attributes = key.attributes;
         for (let attribute of attributes) {
           if (attribute.name === 'type' && attribute.value === typeAttribute) {
--- a/test/unit/rules/no-unsupported-role-attributes-test.js
+++ b/test/unit/rules/no-unsupported-role-attributes-test.js
@@ -24,6 +24,7 @@ generateRuleTests({
     '{{other-component role=this.role aria-bogus="true"}}',
     '<ItemCheckbox @model={{@model}} @checkable={{@checkable}} />',
     '<some-custom-element />',
+    '<input type="password">',
   ],
 
   bad: [


bmish (approved):
Fixing the crash for now sounds good. Thanks!

CC: @judithhinlung the rule author who perhaps can speak to the desired behavior here.

Issue #2951, opened 2023-08-04
Incorrect no-redundant-role on img


amk221:
Given

```hbs
<img src="thing.svg" alt="" role="presentation" />
```

ember-template-lint outputs:

> Use of redundant or invalid role on `<img>` detected

But this use of `role` is what the [docs](https://www.w3.org/WAI/tutorials/images/decorative/) say to do:

> Decorative images don’t add information to the content of a page. For example, the information provided by the image might already be given using adjacent text, or the image might be included to make the website more visually attractive.
> In these cases, a null (empty) alt text should be provided (alt="")

bertdeblock:
Suddenly seeing this as well.

HeroicEric:
Seeing this as well but with a twist...


```
$ ember-template-lint .
app/components/avatar.gts
  19:2  error  Use of redundant or invalid role: presentation on <img> detected.  no-redundant-role

$ ember-template-lint . --fix
app/components/avatar.gts
  19:2  error  If the `alt` attribute is present and the value is an empty string, `role="presentation"` or `role="none"` must be present  require-valid-alt-text
```

It seems like these rules are in a battle with each other. The initial error about using `role="presentation"` seems to be the one that is not working correctly because, like @amk221 said, setting the `role="presentation"` should be valid.

HeroicEric:
Looked into this a little bit more and it seems that this bug is being caused by the `aria-query` update here https://github.com/ember-template-lint/ember-template-lint/pull/2927.

The issue is not present with `ember-template-lint@5.11.0` but present with `ember-template-lint@5.11.1` so anyone running into this can just downgrade for now.

It seems that the issue was introduced to `aria-query` in version `5.2.0` but that version is totally broken because their lockfile was not updated so `5.2.1` fixed that.

HeroicEric:
AFAICT `<img alt="" />` does [_implicitly_ have the role "presentation"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#technical_summary), so also setting `role="presentation"` would be redundant in that case.

This means that the `require-valid-alt-text` rule is incorrectly requiring setting `role="presentation"` in that case.

void-mAlex:
> The issue is not present with `ember-template-lint@5.11.0` but present with `ember-template-lint@5.11.1` so anyone running into this can just downgrade for now.
> 
seeing the issue with 5.11.0 as well 
[...]

PR #2954, opened 2023-08-09
Don't require redundant role for `<img alt="" />` in `require-valid-alt-text` rule

HeroicEric:
Closes https://github.com/ember-template-lint/ember-template-lint/issues/2951, 
Closes https://github.com/ember-template-lint/ember-template-lint/issues/2190, 
Closes https://github.com/ember-template-lint/ember-template-lint/issues/2579, 
Closes https://github.com/ember-template-lint/ember-template-lint/pull/2828

This updates the `require-valid-alt-text` rule so that it does not *require* setting `role="presentation"` or `role="none"` for an ``.

Context

After bumping `aria-query` in https://github.com/ember-template-lint/ember-template-lint/pull/2927, `ember-template-lint` began reporting an error for the following:

```html
<img alt="" role="presentation" />
```

The error is from the `no-redundant-role`. I believe it is correct to report an error in this case because `` already [implicitly has a `role` of `presentation`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#technical_summary).

However, this new behavior was in direct conflict with part of the `require-valid-alt-text` rule that required setting `role="presentation"` or `role="none"` whenever an image has `alt=""`. It is valid to set one of these roles but it should not be required. Removing this requirement allows these two rules (`no-redundant-role` and `require-valid-alt-text`) to continue to be used together.

### DIFF
--- a/docs/rule/require-valid-alt-text.md
+++ b/docs/rule/require-valid-alt-text.md
@@ -12,7 +12,7 @@ This rule **forbids** the following:
 
 ### `<img>`
 
-An `<img>` must have the `alt` attribute. It must have either meaningful text, or be an empty string. If it is an empty string, the `<img>` element tag must also have `role="presentation"` or `role="none"`.
+An `<img>` must have the `alt` attribute. It must have either meaningful text, or be an empty string.
 
 The content of an `alt` attribute is used to calculate the machine-readable label of an element, whereas the text content is used to produce a label for the element. For this reason, adding a label to an icon can produce a confusing or duplicated label on a control that already has appropriate text content.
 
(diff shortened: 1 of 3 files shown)


HeroicEric:
Just realized someone already started on this in https://github.com/ember-template-lint/ember-template-lint/pull/2828 🙃 

I believe the fix in this branch to be more correct because https://github.com/ember-template-lint/ember-template-lint/pull/2828#discussion_r1289164402

bmish:
Seems reasonable. Should we proceed with this PR? @yannbertrand @jaswilli @MelSumner

bmish (approved):
Thanks!

Issue #2482, opened 2022-04-21
[Todo feature] Missing newline when generating `remove` lines in `.lint-todo`


Windvis:
After fixing some linting errors that were previously added to the `.lint-todo` file I ran `npm run lint:hbs` to check if I didn't create any new issues. To my confusion a totally unrelated linting error was shown.

Upon inspection of the `.lint-todo` file, I noticed that new `remove` prefixed lines were added but the first one didn't start on a new line and instead started right at the end of the previous `add` line which made the error appear again in the linting output:

```
... other lines that were added previously
add|ember-template-lint|simple-unless|134|8|134|8|02b279bc993a1ee6632c7e6edf341a2aabc519cb|1649980800000|||app/templates/subsidy/applications/edit/step/edit.hbsremove|ember-template-lint|no-action|51|20|51|20|9d210688d82ca2d429ebdcf858994aa0c1da7828|1649980800000|||app/components/shared/persoon/create-persoon.hbs
... other remove lines
```

Manually moving it to a new line makes it work as expected again.

angelayanpan:
![Screen Shot 2022-06-06 at 3 08 02 PM](https://user-images.githubusercontent.com/2455168/172231376-9af7b21b-d4fc-4e3c-a321-af43f2eacec5.png)

angelayanpan:
also running into this bug

angelayanpan:
I saw this behavior after running: `yarn ember-template-lint . --update-todo`

Techn1x:
This could have been fixed already, but should be fixed _for sure_ in `@lint-todo/utils` 13.1.1
https://github.com/lint-todo/utils/pull/578

I've posted a PR for ember-template-lint to update to it
https://github.com/ember-template-lint/ember-template-lint/pull/2955

Techn1x:
The issue was that certain commands were not applying a newline to the end of the file, depending on what version you were using. Then, after that, if you did an append operation (like update-todo) it would be joined to the line prior since there was no newline.

13.1.1 should have fixed all the remaining commands to apply a newline after their output - but if you have a `.lint-todo` file from a prior version it still might not have the newline at the end until you run your next command, so it's possible you'll end up with one more doubled todo operation line, _then_ it'll be resolved.

bmish:
Should be fixed by https://github.com/ember-template-lint/ember-template-lint/pull/2955.

PR #2955, opened 2023-08-12
chore: lint todo utils to 13.1.1, fixes compact todo issues

Techn1x:
I recently contributed to `@lint-todo/utils` package and would love to see that fix given to everyone.

This fix ensures `--compact-todo` ends the file with a newline (previously it was always stripping the final newline in the file). This avoids other knock on issues like;
* doubled operation lines after an update-todo or other "append" task, which then cause;
  * lint errors to reappear
  * more merge conflicts
  * errornous lines in the file that are never cleaned up

Above all I think it would be worthwhile just to ensure that reported issues in ETL are as reported and not subtly caused by the above.
https://github.com/ember-template-lint/ember-template-lint/issues/2482

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.11.1",
       "license": "MIT",
       "dependencies": {
-        "@lint-todo/utils": "^13.0.3",
+        "@lint-todo/utils": "^13.1.1",
         "aria-query": "^5.3.0",
         "chalk": "^5.3.0",
         "ci-info": "^3.8.0",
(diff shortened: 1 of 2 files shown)


Techn1x:
Note that while it fixes the commands - users with `.lint-todo` files generated by a prior version of `@lint-todo/utils` where they don't have the newline at the end will still have their _next_ command append and cause a doubled-operation line. But hopefully that just goes away over time.

Techn1x:
Not sure what that failing windows test is, but I assume it's flaky and not something to do with this PR.

bmish:
Since this is just a minor version upgrade (https://github.com/lint-todo/utils/releases/tag/v13.1.1), and we use caret version ranges anyway, I'll go ahead and merge this, thanks. Note that we will get upgrades like this from Dependabot too.

peternowee:
@bmish Title should be 13.1.1.

Techn1x:
thx fixed

PR #2970, opened 2023-09-01
Support auto lookup of `.mjs` and `.cjs` config files

bertdeblock:
`ember-template-lint` is written in ESM, which means plugins must be written in ESM as well.
If you want to `import` a plugin in your `ember-template-lint` config file, your project must be in ESM or your `ember-template-lint` config file must be an `.mjs` file, but it seems `ember-template-lint` doesn't auto lookup `.template-lintrc.mjs` config files. You can do `--config-path=.template-lintrc.mjs`, but that's a little verbose if you have multiple `ember-template-lint` invocations.

### DIFF
--- a/test/unit/get-config-test.js
+++ b/test/unit/get-config-test.js
@@ -970,4 +970,36 @@ describe('getRuleFromString', function () {
       );
     }
   });
+
+  it('supports a `.template-lintrc.mjs` config file', async function () {
+    let project = await Project.defaultSetup();
+
+    project.write({
+      '.template-lintrc.mjs': `export default { extends: 'recommended' };`,
+    });
+
+    try {
+      const config = await resolveProjectConfig(project.baseDir, {});
+
+      expect(config).toEqual({ extends: 'recommended' });
+    } finally {
+      project.dispose();
+    }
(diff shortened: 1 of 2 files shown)


bertdeblock:
@bmish If you have some time, would you mind looking at this PR?

test/unit/get-config-test.js
```diff
@@ -970,4 +970,36 @@ describe('getRuleFromString', function () {
       );
     }
   });
+
+  it('supports a `.template-lintrc.mjs` config file', async function () {
+    let project = await Project.defaultSetup();
+
+    project.write({
+      '.template-lintrc.mjs': `module.exports = { extends: 'recommended' };`,
```
bmish, line 978: should we use ESM code `export default = ...` in this example given that it's mjs?

bmish:
This looks reasonable pending my one comment.

test/unit/get-config-test.js
bertdeblock: Yup, oversight on my part. Updated.

bmish (approved):
Thanks.

PR #2990, opened 2023-11-08
Add rule `no-builtin-form-components`

gilest:
Was surprised to find that no such rule exists.

Combined with `no-mut-helper` this can help prevent two-way binding being present in a codebase.

[Rule docs rendered](https://github.com/ember-template-lint/ember-template-lint/blob/0f684349a4bc11ac41a15913dbac164be62febbf/docs/rule/no-builtin-form-components.md)

### DIFF
--- a/docs/rule/no-builtin-form-components.md
+++ b/docs/rule/no-builtin-form-components.md
@@ -1,6 +1,6 @@
 # no-builtin-form-components
 
-Ember's built-in form components use two-way data binding, where the property as `@value` or `@checked` is mutated by user interaction. This goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
+Ember's built-in form components use two-way data binding, where the property passed as `@value` or `@checked` is mutated by user interaction. This goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
 
 ## Examples
 
@@ -17,6 +17,8 @@ This rule **forbids** the following:
 ## Migration
 
 The migration path typically involves replacing the built-in form component with a native HTML element and binding an event listener to handle user input.
+
+In the following example the initial value of a field is controlled by a local tracked property, which is updated by an event listener.
 
 ```js
 import Component from '@glimmer/component';
@@ -51,6 +53,8 @@ You may consider composing the [set helper](https://github.com/pzuraq/ember-set-
 />
 ```
 
+Depending on your requirements, consider using form management patterns like the "light" [Form component from ember-primitives](https://ember-primitives.pages.dev/7-forms/1-intro) or the "controlled" [ember-headless-form](https://ember-headless-form.pages.dev/).
+
 ## Related Rules
 
 * [no-mut-helper](no-mut-helper.md)
@@ -63,3 +67,5 @@ You may consider composing the [set helper](https://github.com/pzuraq/ember-set-
 * [Native HTML `input`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)
 * [Native HTML `textarea`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)
 * [The `on` modifier](https://guides.emberjs.com/release/components/component-state-and-actions/#toc_html-modifiers-and-actions)
+* [Form component – ember-primitives](https://ember-primitives.pages.dev/7-forms/1-intro)
+* [ember-headless-form](https://ember-headless-form.pages.dev/)
--- a/docs/rule/no-builtin-form-components.md
+++ b/docs/rule/no-builtin-form-components.md
@@ -1,0 +1,65 @@
+# no-builtin-form-components
+
+Ember's built-in form components use two-way data binding, where the property as `@value` or `@checked` is mutated by user interaction. This goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
+
+## Examples
+
+This rule **forbids** the following:
+
+```hbs
+<Input />
+```
+
+```hbs
+<Textarea></Textarea>
+```
+
+## Migration
+
+The migration path typically involves replacing the built-in form component with a native HTML element and binding an event listener to handle user input.
+
+```js
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class MyComponent extends Component {
+  @tracked name;
+
+  @action
+  updateName(event) {
+    this.name = event.target.value;
+  }
+}
+```
+
+```hbs
+<input
+  type="text"
+  value={{this.name}}
+  {{on "input" this.updateName}}
+/>
+```
+
+You may consider composing the [set helper](https://github.com/pzuraq/ember-set-helper) with the [pick helper](https://github.com/DockYard/ember-composable-helpers#pick) to avoid creating an action within a component class.
+
+```hbs
+<input
+  type="text"
+  value={{this.name}}
+  {{on "input" (pick "target.value" (set this "name"))}}
+/>
+```
+
+## Related Rules
+
+* [no-mut-helper](no-mut-helper.md)
+
+## References
+
+* [Built-in components guides](https://guides.emberjs.com/release/components/built-in-components/)
+* [Built-in `Input` component API](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input)
+* [Built-in `Textarea` component API](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Textarea?anchor=Textarea)
+* [Native HTML `input`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)
+* [Native HTML `textarea`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)
+* [The `on` modifier](https://guides.emberjs.com/release/components/component-state-and-actions/#toc_html-modifiers-and-actions)
--- a/docs/rule/no-builtin-form-components.md
+++ b/docs/rule/no-builtin-form-components.md
@@ -16,7 +16,35 @@ This rule **forbids** the following:
 
 ## Migration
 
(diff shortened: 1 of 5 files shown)


docs/rule/no-builtin-form-components.md
```diff
@@ -0,0 +1,65 @@
+# no-builtin-form-components
+
+Ember's built-in form components use two-way data binding, where the property as `@value` or `@checked` is mutated by user interaction. This goes against the Data Down Actions Up principle, goes against Glimmer Components’ intention to have immutable arguments, and is [discouraged by the Ember Core team](https://www.pzuraq.com/on-mut-and-2-way-binding/).
```
gilest, line 3: The pzuraq post is fairly old and mostly concerned with the `mut` helper. Would be nice to link to a better resource here.

docs/rule/no-builtin-form-components.md
```diff
@@ -0,0 +1,65 @@
...
+
+```hbs
+<Input />
+```
+
+```hbs
+<Textarea></Textarea>
+```
+
+## Migration
```
NullVoxPopuli, line 17: There is another migration path, too.

Which is faster, and involves less binding.

See:
- https://tutorial.glimdown.com/7-form-data/1-automatic-binding
- and https://ember-primitives.pages.dev/7-forms/1-intro

NullVoxPopuli:
I'd like to propose this be enabled by default since we're working on the next major anyway

docs/rule/no-builtin-form-components.md
gilest: True. I've seen some of this before but never used these patterns exactly. Not sure I understand them well enough to write concisely about it.

How would you suggest the rule docs are structured?

Should we branch the suggestions based on their existing form type somehow or just present both (light/controlled) as options?

I guess ideally the cognitive load is fairly simple for users with simple forms but  🤷🏻

NullVoxPopuli: Option 2 would be my preference.
You can copy from the above links 🎉

lib/rules/no-builtin-form-components.js
```diff
@@ -0,0 +1,25 @@
...
+  Input: `Do not use the \`Input\` component. ${WHY} ${ACTION}`,
+  Textarea: `Do not use the \`Textarea\` component. ${WHY} ${ACTION}`,
+};
+
+const COMPONENTS = new Set(['Input', 'Textarea']);
+
+export default class NoBuiltinFormComponents extends Rule {
+  visitor() {
+    return {
+      ElementNode(node) {
```
patricklx, line 15: Does this work?
All glimmer Ast should need prefix `Glimmer`.
So `GlimmerElementNode`

bmish:
Just to confirm, should this rule be added to `recommended`? Note that we shouldn't do that in this PR but we can line that up for #2712.

NullVoxPopuli:
@bmish yeah, i think so.
I think long term these two components are going to hopefully move out of core and in to an addon or something

docs/rule/no-builtin-form-components.md
gilest: I've made minor changes to incorporate your recommendations.

To be frank the docs on the ember primitives page are not easy to follow and don't clearly explain the light/controlled concepts so I don't really want to copy that over to this rule doc 😅 

For many cases they will be the ideal upgrade path but I think keeping an example of a low-level migration without addons is valuable for users to understand.

NullVoxPopuli: can you open an issue on ember-primitives? docs should be easy to understand, and it's a bug that they're not https://github.com/universal-ember/ember-primitives/ thanks!!

docs/rule/no-builtin-form-components.md
```diff
@@ -51,6 +53,8 @@ You may consider composing the [set helper](https://github.com/pzuraq/ember-set-
 />
 ```
 
+Depending on your requirements, consider using form management patterns like the "light" [Form component from ember-primitives](https://ember-primitives.pages.dev/7-forms/1-intro) or the "controlled" [ember-headless-form](https://ember-headless-form.pages.dev/).
```
NullVoxPopuli, line 56: this isn't quite what I meant about adding. :sweat_smile: 

[This is docs, for example](https://limber.glimdown.com/edit?c=MQAgYg9gTgtgzgKAcFICqcCWA7A5iAfQBUALAUxAAUAbAQwBcAzaGApANVu02rpAClaAN1oBlAMZRMAB3ogStOCDJCyUAJ70SOfAHcK2MmQAmIehHldj1Cs1ghjDWgBoHZALRx12cSExzGKAgYEAgAVygQOHCocQpHeloQKxBxCF4ycTlaXhAwuDU-bGkwuTDpCGwosIAjGEw4LEqAOiQAeS1C2hq4eihaLMxKuFc4MPESZKVOkAAiCa5cMgL6WdTKuNl17EScZJBqCDkIRjNtJUYwn3oh7Bz-dWTsUwVVZONjf1ucvJvqL%2BW1QmUxAQV4NQGAGtRndpHASEcRiA7JCdK4yPRxK4aqUzgw-EokuZpJhfHYnkdyJFjGQYBBWggAJJVGZ2EI1MiHXSufSpKBkBgUJKcbi8WiAIkIQABtAASRAAsgAZZEsAC6UoATKrXClKtRHrQPnMAKIwDlQNZpGkqyIKZ7-PBnWwsap1BpNKopRiYTmmHAlehwZogABCjxpjFoYWo9FcbNd9UatyUuh41FSUYK%2B2ktCWoM5EENowseySWDwNncOaW7lo0mkPIMRlM5hA0n5qh2eLkEajMZAHNeQygDIwOidIDuNzeAEFKIzXFLILAACJOdUARm1IF54i4IFwDCpE-yhX9pRT5BZ5DyBUi7eWBWmN7GbrkOPo5mwDIABn-cAAVko-yqAgmAwBUUByAA3qEVQAL7IkEIQAOQAAK0uaAD0dKfN6agoQA3GBEHQDBqScumiGBMEIAoZhajuPy0QRHEcBEUgaTYL0USJPQFAALwUbwAAUACUxEIFxPF2tYZDMgGIBCSJKhkDsYlKQAfCA0EICABwYjaMBrokSmTmQujgCwJm0Cpnb0M04gRPyOxELQUBLPQEl6QZPZOGZbQ1ABmQOTRMDGjsUjLCJbI2c0al9D6cDid5Pm9IKjnOQlZn8KIbQAHLNL0Uh4JgjDqCJCQuJO0bUK4GrefBknSXIsk2KItT1HIynKPZGkCdpun6apOzNA%2B9krmQkbRvQ4nEfpbXycUpR2QljWSQAPPxEF0Pxmk%2BRtbI%2Bfp0HQZUdHnvQKGWPaS0BvB8HHTpZ1VChr5dddi0dW%2BD1PeIdCNAJsy4FIpiHtI7garMT29OoNhAzAtAAB7uKmxhaAAXCAADMAAMuPSEj0P6ft%2Bn6RtdActQ2lgJgUA8fltAwGQT3k5dk5M2QAkod69P0IzzMoaTZMbVhlOcvtT0U90EvgMI0D%2BBQJks2TIvs-Q6jSFzKEJGQ113Mz3ORkICv8crQtS2LMvU0gqsbR%2BX5mJr2vvf4QvfV1osO5Uwui2ykvkzUUCafbwe21Zq5OBjB0Pppp3pfxmVQC59APaLsfhxtsM2ML%2Bns7BNTQDSUBYxuhNROkmDGIRICPeTWHZ2Q%2B2i9t0i7U3CB-j%2BSAbSQYfIKgsw0AwbJrAASlNahqWxSDuNKP6HSwmk-nBIDyiu%2BXqlqCBz1KP7LsZTgr%2Bd6%2Bb1KW5ICgoCmuaIAT4wU8%2BMss-SjMP6VCvuFlT6UDqtjqpIHPqqLGJBPxwgxlhLCNJVCHC1iOOkAAvNMtBmjQFwFhNS7g0CiCgRAcQcAsIAHUyA1CwnORkWED42QQJqYB8gwFwAgVA1SsC1DNEQcg1BHkMHYCwTg4weCCHENIXKJUWFjQ2GZjsLCR0pT-xAQwphuAwhV2WPFM0aggKOWCFhfkNhFBkCwmkUihgdgEOMRUUx9BPB8Q8FYWsgxhhYWAOYcQBBQEwGoO4L%2B%2BF6a1meA4m4wwEBAA&format=glimdown)

I think we should show an "uncontrolled" example before the controlled example. I think it's important that folks see how to _not_ write JavaScript before they reach for javascript.

gilest: Ahh yes. That makes a lot more sense 😅 

FWIW I appreciate the resources you're creating and the modern improvements.

But we should have empathy for the frustrated engineer who is on the receiving end of lint errors, and try to give them a very clear explanation of the problem and how to solve it.

That's why I'm not using the new component authoring format in the examples, too.

docs/rule/no-builtin-form-components.md
gilest: You got it – https://github.com/universal-ember/ember-primitives/issues/165

docs/rule/no-builtin-form-components.md
NullVoxPopuli: > FWIW I appreciate the resources you're creating and the modern improvements.

<3 


> other new ideas.

controlling every aspect of a form is actually the new idea -- the uncontrolled way is the very old way to do forms, it builds on the foundation of the platform and allows the developer to do less work.

Additionally, React has been trying to teach uncontrolled vs controlled forms for 5+ years! :muscle: 

> That's why I'm not using cell or the new component authoring format in the examples.

that's totally fine! the _details_ are a means to the concept, and not actually important to the concept -- I think that's probably one thing people get too hung up on when learning new things.  It's important to be able to adapt a concept in to your existing code, as not all documentation can match what your existing code is doing.

> Anyway I had another run at it. Let me know your thoughts

will do, thanks a ton for working on this and being willing to have the back and forth!!

docs/rule/no-builtin-form-components.md
```diff
@@ -0,0 +1,97 @@
...
+
+```hbs
+<Textarea></Textarea>
+```
+
+## Migration
+
+Many forms may be simplified by switching to a light one-way data approach.
+
+For example – vanilla JavaScript has everything we need to handle form data, de-sync it from our source data and collect all user input in a single object.
```
NullVoxPopuli, line 21: this looks great! nice work!

NullVoxPopuli (approved):

docs/rule/no-builtin-form-components.md
NullVoxPopuli: > a means to the concept, and not actually important to the concept
> can match what your existing code is doing.

which you've don't exactly! yay!

gilest:
From my perspective this is ready @bmish

PR #3021, opened 2024-02-14
Remove template-imports, add content-tag, drop support for the hbs template literals from ember-template-imports (keeping support for ember-cli-htmlbars)

NullVoxPopuli:
This is a breaking change, all stemming from swapping out the template extraction tool.

Removing ember-template-imports, and adding content-tag means:
- no linting templates made with hbs\`\` from ember-template-imports
- js / ts files will be parsed with babel 
- we fix ~ 40 bugs with gjs/gts files

Also, we're adding a new option to disable checking of js/ts files in general -- this will slightly improve template-lint perf.

We're keeping js/ts for the next major, but we could flip the default to false (don't check js/ts) in the following major.


bmish:
I assume you'll want to get this into v6?

* https://github.com/ember-template-lint/ember-template-lint/issues/2712

NullVoxPopuli:
aye, will be important as template-tag are now in the guides: https://guides.emberjs.com/release/components/template-tag-format/

NullVoxPopuli:
missed v6. v7 here we go!

ef4:
From tooling discussion: the major question here is whether we're OK with losing support for linting inline hbs by landing this PR.

Switching to the content-tag implementation is important, it fixes a lot of bugs, so this is a valuable PR. Reimplementing inline-hbs support would be a whole new separate feature space for code that we plan to codemod people off of anyway, so it's not super attractive to block on.

But we don't want to make migrations harder by letting lint errors creep into inline hbs (primarily tests) only to later re-add linting when those tests port to template tag.

test/unit/temp-babel-backtick-extract-test.js
```diff
@@ -0,0 +1,133 @@
...
+            "template": "
+          Hello, <Name @name={{@name}} />!
+        ",
+          },
+        ]
+      `);
+    });
+  });
+
+  // it("noop", () => {
```
wagenet, line 79: Why the disabled tests?

wagenet (approved):
I'm leaning heavily on tests for this approval.

test/unit/temp-babel-backtick-extract-test.js
NullVoxPopuli: ah these aren't relevant -- they're copy-paste from content-tag-utils -- I'll delete

NullVoxPopuli:
> I'm leaning heavily on tests for this approval.

so am i!

the thing that helps me have any confidence is how the tests that are relevant (most of them) don't need any changes -- all the indexes remain the same

PR #2956, opened 2023-08-12
Prettier 3 (and friends), min node 16

Techn1x:
Updates;
* prettier 3.0.1
  * added trailingComma es5 to avoid big diff in this PR - a follow up PR should remove this and autofix
* eslint & plugins that had updates
  * autofixed for the new unicorn rules, there weren't too many
* switches from eslint-plugin-node (not maintained) to eslint-plugin-n (an eslint community plugin)

eslint-plugin-unicorn and eslint-plugin-n both require node 16. Node 14 was EOL a long while ago, even 16 is just 4 weeks away from EOL. Expanded the node versions to 16, 18 and 20+

PR #3025, opened 2024-02-15
Update to prettier v3

Techn1x:
Updates to prettier 3, eslint-plugin-prettier 5, and eslint-config-prettier 9

Split out of https://github.com/ember-template-lint/ember-template-lint/pull/2956


.prettierrc.cjs
```diff
@@ -1,4 +1,5 @@
 module.exports = {
   "singleQuote": true,
   "printWidth": 100,
+  "trailingComma": "es5"
```
Techn1x, line 4: avoids big diff with prettier v3 new defaults

I would recommend a later PR to remove this custom config item and run eslint fix over everything to do commas the standard way

bmish: Sounds good.

bmish (approved):
Thanks!

PR #3089, opened 2024-10-04
Enhancement: Update require-presentational-children rule

MelSumner:
## Summary

If merged, this PR would update the `require-presentational-children` rule to allow child elements marked with the `aria-hidden` attribute. 

## Details

- Added check for `aria-hidden`
- Updated test for rule with things that should be allowed


lib/rules/require-presentational-children.js
```diff
@@ -93,6 +98,11 @@ export default class RequirePresentationalChildren extends Rule {
           return;
         }
 
+        // If node tag has an attribute in this list, we return
```
MelSumner, line 101: Added a test that is failing so this is not the correct approach, will refactor.

judithhinlung:
I think it will be good to update the docs for `require-presentational-children` with a note regarding these exceptions. The docs examples can also be updated.

lifeart:
Hey @MelSumner! Do you have time to solve conflicts here? If not - I think I could help here.

PR #3101, opened 2025-01-17
Fix strict mode detection

simonihmig:
https://github.com/ember-template-lint/ember-template-lint/pull/2483 introduced detection of embedded and strictMode templates. However, it seems it got the logic wrong. All `*.hbs` files were considered strict mode for example! The wrong logic is quite visible in the tests, see the inline comments.

Seems there were some issues coming up due to this, like for `no-implicit-this` in https://github.com/ember-template-lint/ember-template-lint/issues/2785, which was fixed by disabling the rule in config (https://github.com/ember-template-lint/ember-template-lint/pull/2891), but the rule was still getting things wrong (same PR).


lib/rules/eol-last.js
```diff
@@ -5,9 +5,9 @@ import Rule from './_base.js';
 
 export default class EolLast extends Rule {
   parseConfig(config) {
-    // In strict mode (= template is embedded in e.g. a JS file) we want to disable this rule,
```
simonihmig, line 8: strict mode is not the same as embedded! `hbs` tagged template is embedded, but not strict!

As strictMode was fixed here, which now behaves correctly but differently, I needed to pass `isEmbedded` to rules to get this right here!

lib/rules/no-implicit-this.js
```diff
@@ -53,7 +53,7 @@ const ARGLESS_DEFAULT_BLUEPRINT = [
 
 export default class NoImplicitThis extends Rule {
   parseConfig(config) {
-    if (config === false || config === undefined || !this.isStrictMode) {
```
simonihmig, line 56: Evidence that `no-implicit-this` was also wrong: it should skip _if_ the template is strict mode. It just worked accidentally because `isStrictMode` was also completely wrong! 🤪

test/unit/extract-templates-test.js
```diff
@@ -28,7 +28,7 @@ describe('extractTemplates', function () {
             "columnOffset": 0,
             "end": 0,
             "isEmbedded": undefined,
-            "isStrictMode": true,
```
simonihmig, line 31: This test was asserting that a plain `handlebarsTemplate` is strict mode! 🙃

test/unit/rules/no-implicit-this-test.js
```diff
@@ -29,42 +29,11 @@ let good = [
...
+    template: '<template>{{book}}</template>',
     meta: {
       filePath: 'layout.gjs',
     },
   },
-  {
-    template: `import { hbs } from 'ember-cli-htmlbars';
-      import { setComponentTemplate } from '@ember/component';
-      import templateOnly from '@ember/component/template-only';
-      export const SomeComponent = setComponentTemplate(hbs\`{{book}}\`, templateOnly());`,
```
simonihmig, line 44: Those tests were also asserting the completely wrong thing: the are not strict mode, so _should_ trigger a linting error. Moved them to `bad`

NullVoxPopuli (approved):

PR #3130, opened 2025-02-26
Report unused disables

wagenet:
Can either be set in config or via command line.

### DIFF
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -246,7 +246,7 @@ export default class {
           commentNode,
           parentNode: pNode,
           parentKey,
-          errors: 0
+          errors: 0,
         });
       }
 
@@ -275,7 +275,7 @@ export default class {
             commentNode,
             parentNode: elementNode,
             parentKey: 'comments',
-            errors: 0
+            errors: 0,
           });
         }
 
--- a/lib/rules/_base.js
+++ b/lib/rules/_base.js
@@ -45,9 +45,14 @@ export default class {
     this._ruleNames = options.ruleNames;
     this._allowInlineConfig =
       typeof options.allowInlineConfig === 'boolean' ? options.allowInlineConfig : true;
+    this._reportUnusedDisableDirectives =
+      typeof options.reportUnusedDisableDirectives === 'boolean'
+        ? options.reportUnusedDisableDirectives
+        : options.fileConfig?.reportUnusedDisableDirectives ?? false;
 
     this.severity = options.defaultSeverity;
     this.results = [];
+    this.disableComments = new Map();
     this.mode = options.shouldFix ? 'fix' : 'report';
 
     this.isStrictMode = options.isStrictMode;
@@ -62,7 +67,7 @@ export default class {
     // this.config so the rule implementations don't need to worry about the
     // fact that there is a stack.
     this.config = this.parseConfig(options.config);
-    this._configStack = [this.config];
+    this._configStack = [{ config: this.config }];
 
     this[MODULE_NAME] = options.moduleName;
     this._rawSource = options.rawSource;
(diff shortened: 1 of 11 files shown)


lib/rules/_base.js
```diff
@@ -372,18 +389,47 @@ export default class {
...
+      const node = this._configStack.at(-1)?.commentNode;
+      if (node) {
+        const source = this.sourceForNode(node);
+        const value = node.value.trim();
+        const parts = value.split(reWhitespace).slice(1);
+        if (parts.length > 0) {
+          const updatedParts = parts.filter((part) => part !== this.ruleName);
+          let fix;
+          if (updatedParts.length > 0) {
+            fix = {
```
NullVoxPopuli, line 413: nice

vitest.config.ts
```diff
@@ -5,15 +5,15 @@ export default defineConfig({
...
-      provider: 'v8',
-      enabled: true,
-      thresholds: {
-        lines: 86,
-        functions: 91,
-        branches: 82,
-        statements: 86,
-      },
-    },
+    // coverage: {
```
NullVoxPopuli, line 8: what happened here?

wagenet: I got annoyed by the output and forgot to re-enable it :D

lib/rules/_base.js
```diff
@@ -372,18 +389,47 @@ export default class {
...
+    this.disabledErrors = 0;
   }
 
   _popConfig(config) {
-    if (last(this._configStack) !== config) {
+    if (last(this._configStack).config !== config) {
       throw new Error('Configuration stack out of sync with AST!');
     }
 
+    if (this._configStack.at(-1).disabled && this.disabledErrors === 0) {
```
wagenet, line 403: It might make sense to split this into its own method, but I do think `_popConfig` is probably still the place to call it.

lib/rules/_base.js
```diff
@@ -320,8 +325,20 @@ export default class {
         }
       }
 
-      if (ruleHandler && !plugin.isDisabled()) {
-        maybeRuleHandlerResult = ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
+      if (
+        ruleHandler &&
+        plugin.config &&
+        (!plugin.isDisabled() || plugin._reportUnusedDisableDirectives)
```
wagenet, line 331: If we're going to report on unused, we need to still run the lint so we can find out if it would have failed.

NullVoxPopuli (approved):

lib/rules/_base.js
```diff
@@ -320,8 +325,20 @@ export default class {
...
+      ) {
+        const result = ruleHandler.apply(plugin, arguments); // eslint-disable-line prefer-rest-params
+        if (plugin.isDisabled()) {
+          const frame = plugin.scope.frames.at(-1);
+          if (frame) {
+            frame.hasDisabledErrors = true;
```
wagenet, line 337: I need to remove this. This was from an earlier approach.

lib/rules/_base.js
```diff
@@ -440,15 +486,16 @@ export default class {
       }
 
       if (m[1] === 'disable') {
-        config.value = false;
+        config.value = this._configStack[0]?.config;
+        config.disabled = true;
```
wagenet, line 490: Instead of not returning the config, we keep it but return that it is disabled.

lib/rules/_base.js
```diff
@@ -510,10 +557,15 @@ export default class {
   }
 
   isDisabled() {
-    return !this.config;
+    return this._configStack.at(-1)?.disabled || !this.config;
   }
 
   log(result) {
+    if (this.isDisabled() && result.rule !== 'global') {
```
wagenet, line 564: I don't _love_ how this works, but it does what I want.

NullVoxPopuli (approved):
seems good

NullVoxPopuli:
@ab-abbot approve-testing

NullVoxPopuli:
this isn't the full fix, as the ember blueprint is now on eslint9 + flat config

wagenet:
For reference, times on a large project:
```
**6.0.0**: 53.349s
**master (no unused reporting)**: 52.403s
**master (with unused reporting)**: 51.547s
```

PR #3131, opened 2025-02-26
Don't process ignored files

wagenet:
Also adds more logging on run.

### DIFF
--- a/lib/helpers/cli.js
+++ b/lib/helpers/cli.js
@@ -224,7 +224,8 @@ export function getFilesToLint(
   filePatterns,
   ignorePattern,
   errorOnUnmatchedPattern,
-  config
+  config,
+  _console
 ) {
   let files;
 
@@ -240,6 +241,20 @@ export function getFilesToLint(
       errorOnUnmatchedPattern
     );
   }
+
+  const counts = new Map();
+  for (const file of files) {
+    const ext = path.extname(file);
+    const count = counts.get(ext) || 0;
+    counts.set(ext, count + 1);
+  }
+
+  let fileCounts = '';
+  for (const [ext, count] of counts.entries()) {
+    fileCounts += `\t${ext}: ${count}\n`;
+  }
+
+  _console.log(`Linting ${files.size} Total Files with TemplateLint\n${fileCounts}`);
 
   return files;
 }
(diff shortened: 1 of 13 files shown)


lib/helpers/cli.js
```diff
@@ -201,33 +206,56 @@ export function expandFileGlobs(
...
+      if (!shouldIgnoreFileDueToConfig(filePath, config)) {
+        result.add(filePath);
+      }
     }
   }
 
   return result;
 }
 
+export function filePatternsShouldIncludeSTDIN(filePatterns) {
```
NullVoxPopuli, line 218: thanks for doing this -- chained conditions can become a headache 💯

test/acceptance/cli-test.js
```diff
@@ -959,17 +1010,23 @@ describe('ember-template-lint executable', function () {
         let result = await runBin('app/**/*', '--no-ignore-pattern');
 
         expect(result.exitCode).toEqual(1);
-        expect(result.stdout).toEqual(
-          `app/other.gjs
-  1:39  error  Non-translated string used  no-bare-strings
+        expect(result.stdout).toMatchInlineSnapshot(
```
NullVoxPopuli, line 1013: I like this output a lot! nice !

NullVoxPopuli (approved):

boris-petrov:
I believe this commit (or some other from the new release) started linting `.ts` files (and especially `.d.ts`) which breaks:

```
types/htmlbars-inline-precompile/index.d.ts
  1:0  error  Non-translated string used  no-bare-strings
  1:0  error  template cannot end with newline  eol-last

types/tracked-toolbox/index.d.ts
  9:28  error  attribute name cannot start with equals sign: (error occurred in 'an unknown module' @ line 9 : column 28)  undefined
```

I have a bunch of those. Also:

```
Linting 1113 Total Files with TemplateLint
        .js: 21
        .ts: 733
        .hbs: 359
```

I don't believe template-lint should be linting `.ts` and `.js` files.

NullVoxPopuli:
Template-lint was already previously linting js and ts files. 

This behavior can now be disabled with: 
```
checkHbsTemplateLiterals: false
```

I believe any other suble changes like what you are observing would be in here: https://github.com/ember-template-lint/ember-template-lint/pull/3021/

It's possible i messed up some filtering or something on the file extensions.
( And that existing tests were insufficient)

PR #3133, opened 2025-03-03
Correctly handled file extensions like .d.ts and .css.d.ts

ijlee2:
## Background

`ember-template-lint@7.0.0` lints `*.d.ts` files, leading to unexpected errors depending on the project setup (e.g. with `ember-template-lint-plugin-prettier`, with `*.css.d.ts`). Here are just 2 examples:

<details>

<summary><code>app/styles/app.css.d.ts</code></summary>

File:

```ts
declare const styles: Record<string, never>;

export default styles;
```

Error:

```sh
app/assets/app.css.d.ts
  1:28  error  Unclosed element `string,`: 

|
|  <string, never>
|

(error occurred in 'an unknown module' @ line 1 : column 28)  undefined
```

</details>

<details>

<summary><code>unpublished-development-types/index.d.ts</code></summary>

File:

```ts
// Add any types here that you need for local development only.
// These will *not* be published as part of your addon, so be careful that your published code does not rely on them!

import '@glint/environment-ember-loose';
import '@glint/environment-ember-template-imports';

import type EmbroiderCssModulesRegistry from 'embroider-css-modules/template-registry';

import type MyV2AddonRegistry from '../src/template-registry.ts';

declare module '@glint/environment-ember-loose/registry' {
  export default interface Registry
    extends MyV2AddonRegistry,
      EmbroiderCssModulesRegistry {
    // Add any registry entries from other addons here that your addon itself uses (in non-strict mode templates)
    // See https://typed-ember.gitbook.io/glint/using-glint/ember/using-addons
  }
}
```

Error:

```sh
unpublished-development-types/index.d.ts
  1:63  error  Replace `⏎//·These·will·*not*·be·published·as·part·of·your·addon,·so·be·careful·that·your·published·code·does·not·rely·on·them!⏎⏎import·'@glint/environment-ember-loose';⏎import·'@glint/environment-ember-template-imports';⏎⏎import·type·` with `·//·These·will⏎*not*·be·published·as·part·of·your·addon,·so·be·careful·that·your·published·code⏎does·not·rely·on·them!·import·'@glint/environment-ember-loose';·import⏎'@glint/environment-ember-template-imports';·import·type⏎`  prettier
  7:88  error  Replace `⏎import·type·MyV2AddonRegistry·from·'../src/template-registry.ts';⏎⏎declare·module·'@glint/environment-ember-loose/registry'·{⏎·` with `import·type·MyV2AddonRegistry·from·'../src/template-registry.ts';·declare·module⏎'@glint/environment-ember-loose/registry'·{`  prettier
  12:36  error  Replace `····extends·MyV2AddonRegistry,⏎······EmbroiderCssModulesRegistry·{⏎····//·Add·any·registry·entries·from·other·addons·here·that·your·addon·itself·uses·(in·non-strict·mode·templates)⏎····//·See·https://typed-ember.gitbook.io/glint/using-glint/ember/using-addons⏎··}⏎}⏎` with `extends·MyV2AddonRegistry,·EmbroiderCssModulesRegistry·{·//·Add·any·registry⏎entries·from·other·addons·here·that·your·addon·itself·uses·(in·non-strict·mode⏎templates)·//·See⏎https://typed-ember.gitbook.io/glint/using-glint/ember/using-addons·}·}`  prettier
```

</details>

## Solution

The file utilities in `7.0.0` are prone to bugs, because they make [weak assertions via `.endsWith()`](https://github.com/ember-template-lint/ember-template-lint/blob/v7.0.0-ember-template-lint/lib/file-types.js#L13-L31). We shouldn't continue to [use `isSupported()`](https://github.com/ember-template-lint/ember-template-lint/blob/v7.0.0-ember-template-lint/lib/linter.js#L487), because `*.d.ts` and `*.css.d.ts` files end up being considered as `*.hbs`, leading to errors like the one for `app/styles/app.css.d.ts`.

After writing a failing test, I created utilities that make stronger assertions through `===` and `.includes()` (which is an extension of `===`). Then, I refactored a couple of files to use the new ones and checked that all tests pass.


test/acceptance/cli-test.js
```diff
@@ -699,7 +699,7 @@ describe('ember-template-lint executable', function () {
     });
 
     describe('given - (stdin) path', function () {
-      it('should print errors', async function () {
+      it('should not print errors', async function () {
```
ijlee2, line 702: Not sure what this test is about and why `result.exitCode` is now `0`.

test/acceptance/cli-test.js
```diff
@@ -725,14 +725,14 @@ describe('ember-template-lint executable', function () {
...
-        expect(result.exitCode).toEqual(1);
+        expect(result.exitCode).toEqual(0);
         expect(result.stdout).toBeTruthy();
         expect(result.stderr).toBeFalsy();
       });
     });
 
     describe('given /dev/stdin path', function () {
-      it('should print errors', async function () {
+      it('should not print errors', async function () {
```
ijlee2, line 735: Not sure what this test is about and why `result.exitCode` is now `0`.

test/acceptance/cli-test.js
```diff
@@ -725,14 +725,14 @@ describe('ember-template-lint executable', function () {
           shell: true,
         });
 
-        expect(result.exitCode).toEqual(1);
+        expect(result.exitCode).toEqual(0);
```
NullVoxPopuli, line 728: Why should this be 0?

We enabled a lint, And the template violates that lint, ya?

test/acceptance/cli-test.js
NullVoxPopuli: The rule used for this test isn't fixable, why should the exit code change?

test/acceptance/public-api-test.js
```diff
@@ -480,6 +480,29 @@ describe('public api', function () {
       project.dispose();
     });
 
+    it('ignores files that are not supported', async function () {
```
NullVoxPopuli, line 483: Thanks for adding a test!

test/acceptance/cli-test.js
```diff
@@ -351,8 +351,8 @@ describe('ember-template-lint executable', function () {
       });
     });
 
-    describe('given path to single file with custom extension with errors', function () {
-      it('should print errors', async function () {
+    describe('given path to single file with custom extension', function () {
+      it('should not print errors', async function () {
```
NullVoxPopuli, line 355: This isn't a mistake, we do support passing arbitrary files.
Can you revert the changes to this test?

NullVoxPopuli (changes requested):
Thanks for addressing this! 

We need to revert one little thing, and then i left some questions. 

Thank you!

lib/-private/file-path.js
```diff
@@ -0,0 +1,73 @@
...
+  if (!filePath) {
+    return false;
+  }
+
+  const { ext } = parseFilePath(filePath);
+
+  return isDTS(ext) || isGlimmerScript(ext) || isHBS(ext) || isHTML(ext) || isScript(ext);
+}
+
+export function parseFilePath(filePath) {
```
ijlee2, line 39: Code tested in [`@codemod-utils/files`](https://github.com/ijlee2/codemod-utils/blob/2.3.2/packages/files/src/files/parse-file-path.ts). `filePath` is assumed to be well-defined (has the type of `string`).

lib/-private/file-path.js
```diff
@@ -0,0 +1,73 @@
...
+  return ext === HBS;
+}
+
+function isHTML(ext) {
+  return ext === HTML;
+}
+
+function isScript(ext) {
+  return SOON_DEPRECATED.includes(ext);
+}
```
ijlee2, line 27: To avoid unintended errors (especially in the absence of TypeScript), I'm performing exact matches instead of partial (what had been done before via `.endsWith()`). It's okay to use `.endsWith()` to match file extensions `*.d.ts` and `*.css.d.ts`.

lib/linter.js
```diff
@@ -405,26 +405,28 @@ export default class Linter {
...
-    }
-
-    if (isHBS(options.filePath) || isHTML(options.filePath)) {
-      return this.#byType.template.fix(ruleNames, options, fileConfig);
-    }
-
-    if (isHbsTemplateInScriptCheckingEnabled(options) && isScript(options.filePath)) {
-      return await this.#byType.script.fix(ruleNames, options, fileConfig);
-    }
+    return await processFile(filePath, {
```
ijlee2, line 412: The two sets of `if`-statements (before) has the risk of getting out-of-sync. The utility `processFile()` helps us apply the same order (while hiding details).

test/acceptance/cli-test.js
ijlee2: I reverted commits and found a different solution.

ijlee2:
@NullVoxPopuli Ready for a review again.

lib/-private/file-path.js
NullVoxPopuli, line 39: If we're going to parse the file path, let's use the built in utility instead:

https://nodejs.org/api/path.html#pathparsepath

lib/helpers/cli.js
```diff
@@ -287,15 +287,17 @@ export function getPossibleOptionNames(specifiedOptions) {
 }
 
 function executeGlobby(workingDir, pattern, ignore) {
-  let supportedExtensions = new Set(['.hbs', '.handlebars', ...GJS, ...SOON_DEPRECATED]);
```
NullVoxPopuli, line 290: Why this change?

lib/linter.js
NullVoxPopuli: Thank you!

lib/-private/file-path.js
ijlee2: I briefly alluded to a problem in https://github.com/ember-template-lint/ember-template-lint/pull/3133#discussion_r1981170408. When a file extension has more than 1 `.`, Node's `parse()` returns just the last part (e.g. `.ts` instead of `.d.ts` or `.css.d.ts`), which makes it hard for us to distinguish files.

lib/helpers/cli.js
ijlee2: I wanted to export as few things from the (now private) utility file as possible, so that they don't get misused (imported to a place that hadn't been intended).

Here, I thought duplication would be better. I kept `SOON_DEPRECATED`, though, because it seemed to have been created to indicate places where code may be later removed.

PR #3148, opened 2025-03-31
Support decorator usage in source files

Windvis:


mkszepp:
@Windvis thank you! works fine 🎉

Issue #3134, opened 2025-03-04
Bug in 7.0.0 in gjs files


karl-sjogren:
After upgrading to 7.0.0 in our project I'm getting the following error when linting `.gjs` files.

> Incorrect indentation for `<div>` beginning at L2:C4. Expected `<div>` to be at an indentation of NaN, but was found at 4. (fixable)ember-template-lint[block-indentation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/block-indentation.md)

The indentation is correct in the file. If I break the indentation in a `.hbs` file it reports as expected.

> Incorrect indentation for `section` beginning at L179:C4. Expected `</section>` ending at L183:C16 to be at an indentation of 4 but was found at 6. (fixable)ember-template-lint[block-indentation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/block-indentation.md)

This is my `.template-lintrc.js` file.

```js
'use strict';

module.exports = {
  extends: 'recommended',
  rules: {
    'no-bare-strings': false,
    'quotes': 'double',
    'require-button-type': false, // 2021-07-30 KJ Maybe add this back later
    'require-valid-alt-text': false,
    'require-input-label': false,
    'block-indentation': true,
    'require-media-caption': false,
    'require-each-key': true,
    'no-builtin-form-components': false,
    'no-at-ember-render-modifiers': false
  }
};
```

These are my installed packages.

```js
{
  "name": "am",
  "version": "0.0.0",
  "packageManager": "yarn@4.6.0",
  "private": true,
  "devDependencies": {
    "@babel/core": "^7.26.9",
    "@babel/eslint-parser": "^7.26.8",
    "@babel/plugin-proposal-decorators": "^7.25.9",
    "@babel/preset-env": "^7.26.9",
    "@ember/optional-features": "^2.2.0",
    "@ember/render-modifiers": "^3.0.0",
    "@ember/string": "^4.0.1",
    "@ember/test-helpers": "^5.1.0",
    "@eslint/js": "^9.21.0",
    "@glimmer/component": "~1.1.2",
    "@glimmer/tracking": "~1.1.2",
    "@microsoft/signalr": "^8.0.7",
    "@stylistic/eslint-plugin-js": "^4.2.0",
    "@stylistic/stylelint-plugin": "^3.1.2",
    "@zxing/library": "^0.21.3",
    "babel-plugin-module-extension": "^0.1.3",
[...]

PR #3151, opened 2025-03-31
Fix block-indentation not working correctly for GJS files

robinborst95:
# Reason
Fixes https://github.com/ember-template-lint/ember-template-lint/issues/3134. Related to https://github.com/ember-template-lint/ember-template-lint/pull/2838, where I used `columnOffset > 0` as check for whether a template is embedded or not, but this is not necessarily the case.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -294,6 +294,7 @@ export default class Linter {
             {
               line: coordinates.line,
               column: coordinates.column,
+              columnOffset: coordinates.columnOffset,
               template,
               isEmbedded: true,
               isStrictMode: true,
--- a/lib/rules/block-indentation.js
+++ b/lib/rules/block-indentation.js
@@ -310,8 +310,11 @@ export default class BlockIndentation extends Rule {
     const hasLeadingContent =
       !isDirectChildOfTemplate || this.hasLeadingContent(node, this.template.body);
     const startColumn = node.loc.start.column;
+
+    // Embedded templates should be indented 1 level past the starting offset (unless the first line)
+    const isStandalone = !this.isEmbedded || !this.isStrictMode;
     const expectedStartColumn =
-      this.columnOffset === 0 || node.loc.start.line === 1
+      (this.columnOffset === 0 && isStandalone) || node.loc.start.line === 1
         ? 0
         : this.columnOffset + this.config.indentation;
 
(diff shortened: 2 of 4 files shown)


lib/linter.js
```diff
@@ -294,6 +294,7 @@ export default class Linter {
             {
               line: coordinates.line,
               column: coordinates.column,
+              columnOffset: coordinates.columnOffset,
```
robinborst95, line 297: This is what caused the `NaN` in the error message in the linked issue.

lib/rules/block-indentation.js
```diff
@@ -387,7 +390,7 @@ export default class BlockIndentation extends Rule {
 
         let warning =
           `Incorrect indentation for \`${displayName}\` beginning at ${startLocation}` +
-          `. Expected \`${display}\` ending at ${endLocation} to be at an indentation of ${expectedEndColumn} but ` +
+          `. Expected \`${display}\` ending at ${endLocation} to be at an indentation of ${expectedEndColumn}, but ` +
```
robinborst95, line 393: I noticed some inconsistency with the error messages, so I sneaked that in as well :grimacing:

lib/rules/block-indentation.js
```diff
@@ -552,7 +555,7 @@ export default class BlockIndentation extends Rule {
         const elseLine = node.program.loc.end.line;
         const fixedNode = this.fixLine(
           lines,
-          elseLine - 1,
+          elseLine - (path ? 1 : node.loc.start.line),
```
robinborst95, line 558: Took me some time to find, but this is basically similar to for example line 498 above. At first I didn't have the `path ? 1` part in there, but I added an example test that failed without this :+1:

test/unit/rules/block-indentation-test.js
```diff
@@ -334,6 +334,110 @@ generateRuleTests({
...
+      ].join('\n'),
+      meta: {
+        filePath: 'layout.gjs',
+      },
+    },
+    {
+      template: [
+        'function fetchComponent() {',
+        '  return <template>',
+        '    <div class="parent">',
```
robinborst95, line 375: This is example of the `columnOffset` (= indentation of the line that contains `<template`, in this case 2 spaces before `return`) being necessary to indent this at level 2 + 2 = 4.

NullVoxPopuli (approved):
Thank you for all the tests!!

PR #3165, opened 2025-04-08
Support compound extensions ending in .hbs

JuiHsuan-Kuo:
## Issue
The linter currently only recognizes files with exactly `.hbs` extension, but fails to process files with compound extensions like `.component.hbs`. This affects users who need to use compound extensions for their Handlebars templates.
Resolves #3162 

## Current Behavior
When running `ember-template-lint --fix` on files with compound extensions (e.g., `*.component.hbs`), the linter fails to process them because the file extension detection is too strict.

## Expected Behavior
The linter should process any file that ends with `.hbs`, regardless of additional extensions before it (e.g., `.test.hbs`, `.component.hbs`, etc.).

## Technical Details
The issue is in `lib/-private/file-path.js` where the `isHBS` function performs an exact match:

```javascript
function isHBS(ext) {
  return ext === HBS;  // HBS is '.hbs'
}
```

This should be changed to:

```javascript
function isHBS(ext) {
  return ext.endsWith(HBS);
}
```

## Impact
This affects users who:
1. Use compound extensions for their Handlebars templates
2. Follow naming conventions that require additional extensions

## Environment
- ember-template-lint version: 7.0.2
- Node.js version: ^18.18.0 || >= 20.9.0

PR #3167, opened 2025-04-09
fix: support compound extensions when globbing

### LINKED ISSUE #3166: Globbing should support compound extensions ending in .hbs
## Issue
The linter globbing currently only recognizes files with exactly `.hbs` extension, but fails to process files with compound extensions like `.component.hbs`. This affects users who need to use compound extensions for their Handlebars templates.

See also #3162 #3165 - related issue & PR for `--fix`

## Current Behavior
When running `ember-template-lint .` or `ember-template-lint **/*` on files with compound extensions (e.g., `*.component.hbs`), the linter fails to process them because the file extension detection is too strict.

## Expected Behavior
The linter should process any file that ends with `.hbs`, regardless of additional extensions before it (e.g., `.test.hbs`, `.component.hbs`, etc.).

## Technical Details

The issue is in `lib/helpers/cli.js` where globby results are filtered:

https://github.com/ember-template-lint/ember-template-lint/blob/2db07504fac1bba9fabbab0e61a6d326f3848a45/lib/helpers/cli.js#L289-L301

Note that `parseFilePath` from `lib/-private/file-path.js` causes the inequality by looping over and collecting all extension segments in `ext`.

https://github.com/ember-template-lint/ember-template-lint/blob/2db07504fac1bba9fabbab0e61a6d326f3848a45/lib/-private/file-path.js#L64-L80

## Impact
This affects users who:

1. Use compound extensions for their Handlebars templates
2. Follow naming conventions that require additional extensions
3. Attempt to use globs or directory expansion in the CLI to find Handlebars templates

## Environment
* ember-template-lint version: 7.0.2
* Node.js version: ^18.18.0 || >= 20.9.0

simshanith:
Fixes #3166 

- add `shortExt` to `parseFilePath`
- use `shortExt` in `executeGlobby`
- export & add tests for `isDTS` 
- exclude `.d.ts` in `executeGlobby`
- add tests for templates with compound file extensions in `test/acceptance/cli-test.js`
  - these new tests fail when using `ext` as current behavior does in `executeGlobby`
- ~~`package-lock.json` updated via `npm install`; looks like version number only~~

### DIFF
--- a/lib/-private/file-path.js
+++ b/lib/-private/file-path.js
@@ -6,7 +6,7 @@ const HBS = '.hbs';
 const HTML = '.html';
 export const SOON_DEPRECATED = ['.js', '.ts'];
 
-function isDTS(ext) {
+export function isDTS(ext) {
   return ext.endsWith(DTS);
 }
 
@@ -47,7 +47,7 @@ export function canProcessFile(filePath) {
  *
  * @return
  *
- * An object with `base`, `dir`, `ext`, and `name`.
+ * An object with `base`, `dir`, `ext`, `shortExt`, and `name`.
  *
  * @example
  *
@@ -55,14 +55,17 @@ export function canProcessFile(filePath) {
  * const filePath = 'src/components/navigation-menu.d.ts';
  * const { base, dir, ext, name } = parseFilePath(filePath);
  *
- * // base -> 'navigation-menu.d.ts'
- * // dir  -> 'src/components'
- * // ext  -> '.d.ts'
- * // name -> 'navigation-menu'
+ * // base     -> 'navigation-menu.d.ts'
+ * // dir      -> 'src/components'
+ * // ext      -> '.d.ts'
+ * // shortExt -> '.ts'
+ * // name     -> 'navigation-menu'
  * ```
  */
 export function parseFilePath(filePath) {
   let { base, dir, ext, name } = parse(filePath);
+
+  const shortExt = ext;
 
   // eslint-disable-next-line no-constant-condition
   while (true) {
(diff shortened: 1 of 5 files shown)


lifeart:
Hi @simshanith! Thank you for your contribution!
I think short extensions may break `.d.ts` case - https://github.com/ember-template-lint/ember-template-lint/blob/e6bb6958942796eb472d98018f5a6aa33893a6b4/lib/-private/file-path.js#L3

https://github.com/ember-template-lint/ember-template-lint/pull/3165 seems less invasive here

NullVoxPopuli:
I went with #3165 -- do we still need this PR?

simshanith:
#3165 does not address the globbing; the test cases added in `cli-test.js` here will still fail.

re: @lifeart :

> I think short extensions may break .d.ts case

I think you're right, as `SOON_DEPRECATED` includes `.ts` and now we're matching `shortExt` `.d.ts` will be globbed when previously not. Need further test cases and logic to exclude here if not desired. I'll take a stab at this soon.

simshanith:
https://github.com/ember-template-lint/ember-template-lint/commit/271fbc0fd0c286212c4d12c86f168b46bf451443#r155145549 comment on package-lock.json got lost in force-push; I've since restored the `package-lock.json`

lifeart (approved):

NullVoxPopuli (approved):

simonihmig:
I haven't looked closely at this, but if this was the only change in 7.0.6 (which according to the release notes it is), then this has caused a regression: https://github.com/ember-template-lint/ember-template-lint/issues/3171

simshanith:
The `7.0.6` release changed behavior in that the `image.helper.ts` was previously ignored by the `ember-template-lint .` directory globbing.

Could add an `ignore: ['**/*.ts']` property to your `.template-lintrc.js` to exclude explicitly now that the compound extension no longer implicitly ignores.

Issue #3122, opened 2025-01-29
`no-bare-strings`: no configuration to ignore more HTML elements


BlueCutOfficial:
The `no-bare-strings` rule knows a hardcoded list of `IGNORED_ELEMENTS` like `pre` and `script`, so their content is not interpreted as a UI text. There's currently no configuration to modify this list, that would be a nice addition. 

**Usecase**: 

ember-template-lint is great at linting templates, and I am running into a case where we decided to use ember-template-lint to lint some templates that are not `hbs` but [`mjml`](https://mjml.io/) files. It doesn't make a lot of difference, it's mostly Handlebars syntax except there are mjml tags everywhere. I have a problem with the `mj-style` element though. This one contains CSS styles that can be interpreted as bare strings by `no-bare-strings` rule, which is incorrect.

If mjml templates sound too specific to get `mj-style` included in `IGNORED_ELEMENTS`, then at least I would like to have the possibility to add it to my config. I would appreciate feedback about the best approach, I can eventually open a PR.

PR #3178, opened 2025-04-16
fix: #3122 extend no-bare-strings rule config

lifeart:
Resolves: https://github.com/ember-template-lint/ember-template-lint/issues/3122


NullVoxPopuli (approved):
oh mjml

PR #3183, opened 2025-04-16
fix #2969 allow no-invalid-link-text to be configured for custom link…

### LINKED ISSUE #2969: allow no-invalid-link-text to be configured for custom link component names
Right now, the `no-invalid-link-text` rule checks for invalid link text on `<LinkTo>` and `<a>` tags. It would be great if we could allow a custom option to the config for tag names, so that developers using custom/wrapper components could specify additional component invocations to check.

For example, say a developer has an `ExternalLink` component that they use to consistently style links leading away from the application. We would still want those links to have more meaningful text than "learn more"/etc.

Right now, the rule passes for the following:

```handlebars
<ExternalLink @href={{this.externalLink}}>Learn more</ExternalLink>
```

We would want the above code to fail the `no-invalid-link-text` check.

The config could look something like this:

```js
// .template-lintrc.js

module.exports = {
  extends: 'recommended',
  rules: {
    'no-invalid-link-text': {
      additionalTags: ['ExternalLink']
    }
  }
};

### PR DESCRIPTION
**Pull request by lifeart**
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/2969

### DIFF
--- a/lib/rules/no-invalid-link-text.js
+++ b/lib/rules/no-invalid-link-text.js
@@ -1,7 +1,7 @@
 import AstNodeInfo from '../helpers/ast-node-info.js';
 import Rule from './_base.js';
 
-const DEFAULT_CONFIG = { allowEmptyLinks: false };
+const DEFAULT_CONFIG = { allowEmptyLinks: false, linkComponents: [] };
 const DISALLOWED_LINK_TEXT = new Set([
   // WCAG F84-Provided Examples
   'click here',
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/rules/no-invalid-link-text.js [hunk 1]
```diff
@@ -99,7 +102,11 @@ export default class NoInvalidLinkText extends Rule {
         // In strict mode, LinkTo is ignored even with invalid text
         // This is because we can't tell if the component is actually a built-in LinkTo
         // or just has the same name. It's better to risk false negatives than false positives.
-        if (node.tag === 'a' || (node.tag === 'LinkTo' && !this.isStrictMode)) {
+        if (
+          node.tag === 'a' ||
+          (node.tag === 'LinkTo' && !this.isStrictMode) ||
+          this.config.linkComponents.includes(node.tag)
```
NullVoxPopuli, line 108: shouldn't this also be exclusive to loose mode?

lifeart: not really, because we only extending functionality (non breaking) - user needs to manually specify component name.

NullVoxPopuli (approved):

PR #3187, opened 2025-04-16
Fix #2654 require-presentational-children fails for named blocks

### LINKED ISSUE #2654: `require-presentational-children` fails for named blocks
We have a button component that yields several named blocks for various reasons (mostly used by other re-usable components). The rule `require-presentational-children` fails whenever we add e.g. `role="tab"` to one of these buttons:

```
<MyButton role="tab">
  <:default>Button text</:default>
</MyButton>
```

Fails with the error

> <MyButton> has a role of tab, it cannot have semantic descendants like <:default>

I think named blocks should not be flagged like this, rather the contents should be evaluated instead (if anything). The result is now that people are encouraged to *remove* the role attribute or suppress the lint rule, as those are the only ways to fix the error.

lifeart:
Resolves: https://github.com/ember-template-lint/ember-template-lint/issues/2654

### DIFF
--- a/lib/rules/require-presentational-children.js
+++ b/lib/rules/require-presentational-children.js
@@ -12,6 +12,9 @@ function createErrorMessage(ancestorElement, role, descendantElement) {
  */
 function getAllElementNodeDescendants(node) {
   return AstNodeInfo.childrenFor(node).flatMap((childNode) => {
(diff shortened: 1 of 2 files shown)


joukevandermaas (approved):
Amazing!

NullVoxPopuli (approved):

PR #3188, opened 2025-04-17
[wip] experiment - worker pool

lifeart:
trying to implement poc for https://github.com/ember-template-lint/ember-template-lint/issues/1860

PR #3189, opened 2025-04-17
peformance: pool file processing

lifeart:
trying to maximize CPU usage here
In current state of things we do async lint/verify file-by-file.
Problem with this approach is that `verify` step contain list of  async `io` operations and we basically waiting until promise is resolved.

To address this issue we creating a `linting` pool to process multiple files in same time (for now in one thread)

We should see improvement on projects with large amount of templates or on projects with slow IO (windows)

this is initial steps to https://github.com/ember-template-lint/ember-template-lint/pull/3188


lifeart:
closing prior to https://github.com/ember-template-lint/ember-template-lint/pull/3190

PR #3196, opened 2025-04-28
No Spread Attributes with Class Rule

zfir:
# Add no-splattributes-with-class rule

## Problem

When using `...attributes` (spread attributes) in Ember components, it's possible to also use a `class` attribute. This can lead to confusion about class priority and unexpected styling behavior. For example:

```hbs
{{!-- Parent component --}}
<MyComponent class="background-blue-class" />

{{!-- MyComponent template --}}
<div class="background-red-class"  ...attributes>
  {{!-- Which class takes priority? --}}
</div>
```

In this example, the `background-red-class` will take priority over `background-blue-class` because it comes later in the HTML. This can be confusing and lead to unexpected styling results, especially when components are reused in different contexts.

This issue became apparent when migrating from `ember-css-modules` to `embroider-css-modules`, where class handling and specificity can behave differently. During this migration, we encountered several cases where components had both `...attributes` and `class` attributes, the order of the css classes changed, as we have the same specificity for two classes, we broke the rendering and making the migration process more difficult. To solve this, we had to refactor components to split the classes into wrapper elements:

```hbs
{{!-- Before --}}
<div ...attributes class="background-red-class">
  content
</div>

{{!-- After --}}
<div ...attributes>
  <div class="background-red-class">
    content
  </div>
</div>
```

## Solution

This PR adds a new rule `no-splattributes-with-class` that prevents using both `...attributes` and `class` attributes together. Instead, developers should:

1. Use `...attributes` alone to allow class merging from the parent component
2. Use `class` alone if they want to enforce specific classes

The rule will report an error/warning when it finds both attributes being used together, helping to prevent potential styling issues and making component behavior more predictable.

## Implementation Details

- Added new rule `no-splattributes-with-class`
- Added test cases
- Added docs with examples

### DIFF
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -79,6 +79,7 @@ import norestrictedinvocations from './no-restricted-invocations.js';
 import norouteaction from './no-route-action.js';
 import noscopeoutsidetableheadings from './no-scope-outside-table-headings.js';
 import noshadowedelements from './no-shadowed-elements.js';
+import nosplattributeswithclass from './no-splattributes-with-class.js';
 import nothisintemplateonlycomponents from './no-this-in-template-only-components.js';
 import notrailingspaces from './no-trailing-spaces.js';
 import notriplecurlies from './no-triple-curlies.js';
@@ -202,6 +203,7 @@ export default {
   'no-route-action': norouteaction,
   'no-scope-outside-table-headings': noscopeoutsidetableheadings,
   'no-shadowed-elements': noshadowedelements,
+  'no-splattributes-with-class': nosplattributeswithclass,
   'no-this-in-template-only-components': nothisintemplateonlycomponents,
   'no-trailing-spaces': notrailingspaces,
   'no-triple-curlies': notriplecurlies,
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ Each rule has emojis denoting:
 | [no-route-action](./docs/rule/no-route-action.md)                                                         | ✅  |     |     |     |
 | [no-scope-outside-table-headings](./docs/rule/no-scope-outside-table-headings.md)                         | ✅  |     | ⌨️  |     |
 | [no-shadowed-elements](./docs/rule/no-shadowed-elements.md)                                               | ✅  |     |     |     |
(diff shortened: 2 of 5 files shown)


zfir:
Hello @evoactivity,

We have changed your examples a bit to demonstrate what is the problem.

Example1: https://codepen.io/zfir/pen/NPPwWPX

Example 2: https://share.glimdown.com/n9vSgZCVTe635oVY.07mrw

In the first example, I moved the `make-me-blue-border` above and priority of the css changes. This means that the specificity of the css changes based on the order of the css in the file.

In the second example, I added a parent component and child component, irrespective of the positioning of the `...atttributes`, the css that is appicable in this case is the one from the child, the one that we want to apply is the one from the parent.

evoactivity:
I might have misunderstood your original post as meaning the order of the classes on the class attribute rather than the order they appear in the CSS (or are injected to the page).

> In the first example, I moved the make-me-blue-border above and priority of the css changes. This means that the specificity of the css changes based on the order of the css in the file.

Yes, that is expected.

> In the second example, I added a parent component and child component, irrespective of the positioning of the ...atttributes, the css that is applicable in this case is the one from the child(blue colour), the one that we want to apply is the one from the parent(red colour).

This is more annoying to deal with I'll grant you that. But I think the answer is to use CSS features to manage specifcity. In this case CSS layers would be ideal (it's what they're designed for).

An example with layers - https://share.glimdown.com/Exx04u6CSfylUmJTrr2HBg

I don't think the answer is to recommened through linting against useful base functionality of splattributes.

lifeart:
Here is more details about `...attributes` behaviour - https://guides.emberjs.com/v5.5.0/components/component-arguments-and-html-attributes/

image

lib/rules/index.js
```diff
@@ -202,6 +203,7 @@ export default {
   'no-route-action': norouteaction,
   'no-scope-outside-table-headings': noscopeoutsidetableheadings,
   'no-shadowed-elements': noshadowedelements,
+  'no-splattributes-with-class': nosplattributeswithclass,
```
lifeart, line 206: we need to run `npm run update` to adjust readme.

image

lifeart: @zfir nope

NullVoxPopuli (approved):

PR #3201, opened 2025-05-03
fix: #2534 & #1799 - improve no-duplicate-landmark-elements

### LINKED ISSUE #1799: `no-duplicate-landmark-elements` should not flag elements as duplicate across different conditional branches
`<header>` is flagged as a violation of [no-duplicate-landmark-elements](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-duplicate-landmark-elements.md) in this example:

```hbs
{{#if @someVariable}}
  <header>
    foo
  </header>
{{else}}
  <header>
    bar
  </header>
{{/if}}
```

I know that `<header>` could be moved outside the if statement in this simplified example, but that fix is not always possible. We should ignore duplicates across conditional branches.

The `no-duplicate-id` rule already takes this same problem into consideration.

ember-template-lint: `3.0.0-beta.5`

### LINKED ISSUE #2534: ignore no-duplicate-landmark-elements lint rule for elements displayed conditionally
Given an hbs file with two forms but only one exists at any given moment, ember-template-lint reports no-duplicate-landmark-elements. Is this intentional or a bug?

```
{{#if this.isCreateProjectFromSavedSearchEnabled}}
   <form></form>
{{else}}
   <form></form>
{{/if}}
```

The form elements here are version1 and version 2 of same feature and cannot have different aria-labels.

lifeart:
resolves #2534
resolves #1799

### DIFF
--- a/lib/rules/no-duplicate-landmark-elements.js
+++ b/lib/rules/no-duplicate-landmark-elements.js
@@ -46,6 +46,19 @@ export default class NoDuplicateLandmarkElements extends Rule {
 
   visitor() {
     return {
+      Block: {
+        enter() {
+          const parent = this._landmarksScope.at(-1);
+          const newScope = new Map();
+          for (const [role, labelsForRole] of parent) {
+            newScope.set(role, labelsForRole);
+          }
+          this._landmarksScope.push(newScope);
+        },
+        exit() {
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):
thank you!!!!!

PR #3203, opened 2025-05-03
fix #2386 - list summary in interactive elements

### LINKED ISSUE #2386: Enhancement: Add summary to list of things that should trigger no-nested-interactive rule.
What I expect to happen: 
The linter should throw a `no-nested-interactive` error if I put an interactive element inside of a `summary` element. 

What happens: 
No error is thrown

To fix: 
Add `summary` to the list of interactive elements that cannot contain another interactive to the [`no-nested-interactive`](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-nested-interactive.md) rule.

lifeart:
resolves https://github.com/ember-template-lint/ember-template-lint/issues/2386
resolves https://github.com/ember-template-lint/ember-template-lint/issues/1625

### DIFF
--- a/test/unit/rules/no-nested-interactive-test.js
+++ b/test/unit/rules/no-nested-interactive-test.js
@@ -73,6 +73,27 @@ generateRuleTests({
 
   bad: [
     {
+      template: '<summary><a href="/">button</a></summary>',
+
+      verifyResults(results) {
+        expect(results).toMatchInlineSnapshot(`
+          [
+            {
+              "column": 9,
+              "endColumn": 31,
+              "endLine": 1,
+              "filePath": "layout.hbs",
+              "line": 1,
+              "message": "Do not use an <a> element with the \`href\` attribute inside <summary>",
+              "rule": "no-nested-interactive",
+              "severity": 2,
+              "source": "<a href="/">button</a>",
+            },
+          ]
+        `);
+      },
+    },
+    {
       template: '<a href="/">button<a href="/">!</a></a>',
 
       verifyResults(results) {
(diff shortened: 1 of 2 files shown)


test/unit/rules/no-nested-interactive-test.js
```diff
@@ -178,21 +199,21 @@ generateRuleTests({
       },
     },
     {
-      template: '<button><details><summary>Some details</summary><p>!</p></details></button>',
```
NullVoxPopuli, line 181: why this change?

lifeart: @NullVoxPopuli because here we have 2 different errors, and added summary overshaded details case. Seems in this rule we may have one error from interactive element in one time. And here we have button->details(interactive)->summary(interactive)

NullVoxPopuli (approved):

test/unit/rules/no-nested-interactive-test.js
rrenno: @lifeart is it not the case that `<summary>` is allowed a the first child element of `<details>` despite both being classified as "interactive"? E.g.

```javascript
<details>
  <summary>Details</summary>
  Something small enough to escape casual notice.
</details>
```

I'm seeing lots of seemingly reputable sources documenting this use, but no offering of an alternative.
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/summary
https://html.spec.whatwg.org/multipage/interactive-elements.html#interactive-elements

Any advice on how to provide a "label" to the `details` without using `summary`? Otherwise I suspect this might cause a regression bug. While it is true on their own, interactives should not be nested under `details`, nor `summary`, they have a special case where `summary` can be nested under `details` as the first child.

Thanks in advance for your help!

NullVoxPopuli: aye -- this is a bug!

lifeart: @rrenno thank you for your report! You absolutely right! Here is fix - https://github.com/ember-template-lint/ember-template-lint/pull/3249, in short - we allowing summary inside details if it's first node

PR #3204, opened 2025-05-03
fix #2305 support "if" helper in no-bare-strings

### LINKED ISSUE #2305: `no-bare-strings` false-negative
```hbs
<div>
  {{if true "Yes" "No"}}
</div>
```

This should lead to a `no-bare-strings` warning but it doesn't.

lifeart:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/2305

### DIFF
--- a/lib/rules/no-bare-strings.js
+++ b/lib/rules/no-bare-strings.js
@@ -105,6 +105,10 @@ function isPageTitleHelper(node) {
   return match(node.path, { original: 'page-title', type: 'PathExpression' });
 }
 
+function isIfHelper(node) {
+  return match(node.path, { original: 'if', type: 'PathExpression' });
+}
+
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

boris-petrov:
Thanks for the fix! However, it broke a lot of other things:
```hbs
<div class={{if true "disabled"}}></div>
```
That obviously shouldn't issue a `no-bare-strings` warning but it does. Do you want me to open a new issue for this @lifeart?

lifeart:
@boris-petrov, thank you for report! Happy "allow" tests added, logic fixed in https://github.com/ember-template-lint/ember-template-lint/pull/3215. As bonus - added support for string only `concat` and `unless`.

PR #3205, opened 2025-05-03
fix #1728 support absolute paths in no-this-in-template-only-components

### LINKED ISSUE #1728: no-this-in-template-only-components - does not support absolute filePath's

lifeart:
Resolves https://github.com/ember-template-lint/ember-template-lint/issues/1728

### DIFF
--- a/lib/rules/no-this-in-template-only-components.js
+++ b/lib/rules/no-this-in-template-only-components.js
@@ -24,6 +24,15 @@ function isTemplateOnlyComponent(hbsFilePath, validComponentExtensions) {
     return validComponentExtensions.some((ext) => fs.existsSync(pathWithoutExtension + ext));
   }
   const paths = path.normalize(hbsFilePath).split(path.sep);
+
+  while (paths.length > 0 && paths[0] !== 'app' && paths[0] !== 'addon') {
+    paths.shift();
+  }
+
+  if (paths.length === 0) {
+    // no app or addon directory found
+    return false;
+  }
   if (paths[0] === 'app' || paths[0] === 'addon') {
     if (paths[1] === 'templates') {
       if (paths[2] === 'components') {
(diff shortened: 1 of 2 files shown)


lib/rules/no-this-in-template-only-components.js
```diff
@@ -24,6 +24,15 @@ function isTemplateOnlyComponent(hbsFilePath, validComponentExtensions) {
     return validComponentExtensions.some((ext) => fs.existsSync(pathWithoutExtension + ext));
   }
   const paths = path.normalize(hbsFilePath).split(path.sep);
+
+  while (paths.length > 0 && paths[0] !== 'app' && paths[0] !== 'addon') {
```
NullVoxPopuli, line 28: some addons with templates may not have an `addon` directory (v2 addons and v2 apps may use `src`)

Why check for specific path segments?

lifeart: @NullVoxPopuli you absolutely right, but this rule is kinda "happy-path" rule, and works only for classic layouts. In scope of this PR we don't change rule logic, we only adding support for absolute paths (before - only relative works)

PR #3232, opened 2025-05-05
[performance] bundle bin, worker and index files

lifeart:
Metrics:

Test suite duration (master):
```
Test Files  156 passed (156)
      Tests  8860 passed | 5 skipped (8865)
   Start at  13:06:30
   Duration  55.81s (transform 1.29s, setup 0ms, collect 77.10s, tests 120.25s, environment 24ms, prepare 11.04s)
```

~Test suite duration (bundled (esbuild)):~
```
 Test Files  159 passed (159)
      Tests  8979 passed | 5 skipped (8984)
   Start at  17:00:47
   Duration  41.58s (transform 1.27s, setup 0ms, collect 74.95s, tests 85.65s, environment 21ms, prepare 11.44s)
```

Tes suite duration (bundled (rolldown))
```
 Test Files  159 passed (159)
      Tests  8979 passed | 5 skipped (8984)
   Start at  19:29:11
   Duration  33.75s (transform 1.21s, setup 0ms, collect 70.59s, tests 69.77s, environment 24ms, prepare 10.45s)
```

Improvement: `120.25s` -> `69.77s`

Note, it's likely an startup improvement, not overall linting x2 improvement.

Minified bin bundle size: `1.6 mb`

Huuuge dependency: `language-tags`, to save `800kb` of bundle size we have builtin country code download & resolution logic, it's not 100% spec accurate, but, 800kb is 800kb.

Some dependencies is not really bundlabe, because of error like this: https://github.com/evanw/esbuild/issues/3324, https://github.com/evanw/esbuild/issues/1921

Note:
~We need to pin dependencies to exact versions to have same behaviour as we have if import package directly.~ (now we bundling public api scope, so it's should be safe)


lifeart:
@SergeAstapov, sorry to bothering you here, do you mind to test out this PR (note, you may need to run `npm run build:node` to bundle things)

NullVoxPopuli:
@SergeAstapov how'd it go?

SergeAstapov:
@lifeart @NullVoxPopuli sorry for delay, on my project I see following numbers:

```
~/Code/[my app]/apps/[my app]
time /Users/sastapov/Code/tmp/ember-template-lint/bin/ember-template-lint.js .
Linting 2431 Total Files with TemplateLint
        .js: 1345
        .ts: 579
        .hbs: 284
        .gts: 203
        .gjs: 2
        .scenario.js: 18


/Users/sastapov/Code/tmp/ember-template-lint/bin/ember-template-lint.js .  25.94s user 1.98s system 413% cpu 6.756 total


~/Code/[my app]/apps/[my app]
time pnpm run lint:hbs

> [my app]@1.0.0 lint:hbs /Users/sastapov/Code/[my app]/apps/[my app]
> ember-template-lint .

Linting 2431 Total Files with TemplateLint
        .js: 1345
        .ts: 579
        .hbs: 284
        .gts: 203
        .gjs: 2
        .scenario.js: 18


pnpm run lint:hbs  47.10s user 2.80s system 461% cpu 10.811 total
```

so I do see about 2x speed improvement! 🤯

lifeart:
@SergeAstapov wow! Impressive numbers! 
Wondering, if results is better/worse if we change `MIN_FILES_TO_USE_WORKERS` from `100` to `150` or `200`.

lifeart:
@NullVoxPopuli wondering if we could land it to see if anything suppose to be broken...

NullVoxPopuli:
sounds good to me

lib/helpers/country-codes.js
NullVoxPopuli, line 1: what's this file for?

lib/rules/require-lang-attribute.js
```diff
@@ -1,7 +1,6 @@
-import tags from 'language-tags';
```
NullVoxPopuli, line 1: ah I see -- why don't we want to leave this as is? is it not maintained? :see_no_evil:

build.js
```diff
@@ -0,0 +1,125 @@
+import * as rolldown from 'rolldown';
+
+const externalDeps = [
```
NullVoxPopuli, line 3: why not automate this by reading package.json devdeps vs deps? (and then keep the node stuff, I suppose

e.g.:


external: (x) => {
  return dependencies.includes(x) 
         || x.startsWith('node:') 
         || isKownNodeBuiltIn(x);
}

lib/rules/require-lang-attribute.js
lifeart: It's adding 800kb of `json` to bundle (

lib/helpers/country-codes.js
lifeart: This file to generate country-codes list, for `language-tags` replacement.

build.js
lifeart: not all deps bundlable(, I'm using array to be able simply turn on/off dependency from bundling (simplify debugging)

NullVoxPopuli (approved):

Issue #3086, opened 2024-08-12
Pattern grouping support for `attribute-order`


TSenter:
I really like the idea of the [`attribute-order`](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/attribute-order.md) rule, but I feel like it doesn't quite meet our use case. My company follows a pattern where the names of all of our actions start with the prefix `on` - we also have a pattern of placing all of the actions as the last arguments/attributes on a component. However, the current `attribute-order` rule might mix actions with other properties that come after actions, like `readonly`.

I know this would increase the complexity of this rule, but I think the following configuration would be incredibly helpful:
```js
{
  order: [
    'attributes',
    {
      type: 'arguments',
      pattern: /^on/,
      invert: true,
    },
    {
      type: 'arguments',
      pattern: /^on/,
    },
    'modifiers',
  ],
}
```

This would let more granular groupings for attributes to be applied, like in our use case. Also (and this would only apply if a change for this was implemented), it would be nice to mark different ordering rules as "equal" in terms of ordering, similar to how the [`order-in-components`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/order-in-components.md) rule and the [`order#groups`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#groups-array) rule in [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import).

wozny1989:
We're using the same pattern on our company according to this:
```hbs
<TooltipOnComponent
                                                  ┓
  class="tooltip-base risktag-tooltip z-idx-3001" | ==> Native attributes
                                                  ┛
                                                  ┓
  @event="none"                                   |
  @side="bottom"                                  | ==> Static attributes
  @enableLazyRendering={{true}}                   |
  @spacing={{10}}                                 |
                                                  ┛
                                                  ┓
  @isShown={{this.showToolTip}}                   |
  @target={{concat '#risktag-' @tag.id}}          | ==> Dynamic attributes
                                                  ┛
                                                  ┓
  @onClick={{this.example}}                       | ==> Actions like "onClick" etc.
                                                  ┛
>
  <div>
    <span>example content</span>
  </div>
</TooltipOnComponent>
```
It will be nice to keep order `onSomething` attributes at the end 👍

Issue #3054, opened 2024-04-10
Bug: autofix of `attribute-order` removes empty string attribute/argument values


IgnaceMaes:
## Description

The auto fix of `attribute-order` removes empty string values for attributes and arguments.

## Example

Before:

```hbs
<SomeComponent @b="" @a="foo" />
```

After:

```hbs
<SomeComponent @a="foo" @b />
```

Expected:

```hbs
<SomeComponent @a="foo" @b="" />
```

## Versions

Seeing this behaviour both on `ember-template-lint@5.11.2` and latest (`ember-template-lint@6.0.0`)

IgnaceMaes:
Seems like this was reported before: https://github.com/ember-template-lint/ember-template-lint/issues/2692

And mentioned as fixed in https://github.com/ember-template-lint/ember-template-lint/pull/2701

Why am I still seeing this issue then? 🤔

lifeart:
That's correct issue still exists, and likely related to https://github.com/ember-template-lint/ember-template-recast

Issue #3053, opened 2024-04-10
Running `--fix` when enabling `attribute-order` rule hangs process


IgnaceMaes:
When enabling the `attribute-order` rule like so:

```json5
{
    // ...
	rules: {
	  'attribute-order': true,
      // ...
    },
}
```

Running `pnpm ember-template-lint "**/*.hbs"` succeeds (as in, completes):

```
✖ 1105 problems (1105 errors, 0 warnings)
  1090 errors and 0 warnings potentially fixable with the `--fix` option.
 ELIFECYCLE  Command failed with exit code 1.
```

Running `pnpm ember-template-lint "**/*.hbs" --fix` results in a Node process running at 100% CPU which hangs.

I've yet to pinpoint the exact issue by running on a subset of files. But opening the issue already regardless.

**Versions**

- node@18.18.2
- ember-template-lint@5.11.2

Using Node v20.11.0 gives the same result.

IgnaceMaes:
I was able to create the following minimal reproduction.

Running autofix hangs on the following:

```hbs
<Foo::Bar
  @ca={{this.justARandomName}}
  @cb={{this.anotherRandomName}}
  @t={{this.another.random.prop}}
>
  <div>
    <div>
      <div>
        <LinkTo
          @route="some.route"
          @model={{this.someLongerGetterName}}
          @query={{hash
            l=this.someProp.anotherthing
            mP=this.anotherProp.anotherthing
            p=this.foobar
          }}
          target="_blank"
        >
          test
        </LinkTo>
      </div>
    </div>
  </div>
</Foo::Bar>
```

Running autofix works when either of the following actions is done:
- Removing one level of `div` nesting
- Removing the `ca` argument in the `Foo::Bar` component
- Replacing `someLongerGetterName` with `x`

This all seems random. But as making the line length shorter generally seems to solve this, it might indicate a conflict with a line length rule?

ijlee2:
I enabled `attribute-order` as shown in the OP's description, and encountered this issue when I let `pnpm` run `lint:hbs:fix` on multiple packages. One of the packages had a component that caused the autofix for `attribute-order` to run indefinitely (there may be an infinite loop).

Here are my findings after trying to debug the issue: For all packages, `lint:hbs:fix` does finish when I patch the rule in one of these ways:

[...]

PR #3237, opened 2025-05-05
New rule: sort-invocations

ijlee2:
## Background

Addresses the issues #3053, #3054, and #3086 still present in `ember-template-lint@v7` by introducing a sorting rule, one that is easy to understand and maintain. Source code and tests were copied from [`ember-codemod-sort-invocations`](https://github.com/ijlee2/ember-codemod-sort-invocations).

 I suggest removing `attribute-order` for `v8.0.0`, since it can cause the process to hang (infinite loop).


docs/rule/sort-invocations.md
```diff
@@ -0,0 +1,113 @@
...
+### Modifiers
+
+Similarly to helpers, list the named arguments in alphabetical order.
+
+
+## Limitations
+
+It's intended that there are no options for sorting. Alphabetical sort is the simplest for everyone to understand and to apply across different projects. It's also the easiest to maintain.
+
+To better meet your needs, consider creating a plugin for `ember-template-lint`.
```
ijlee2, line 77: The same approach (limit the API) is taken by [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort/tree/v12.1.1?tab=readme-ov-file#not-for-everyone).

docs/rule/sort-invocations.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Known issues
+
+1\. If you passed an empty string as an argument's value, it has been replaced with the placeholder text `<EMPTY STRING>`. Do a find-and-replace-all to restore the value.
+
+```diff
+- <MyComponent @description="<EMPTY STRING>" />
++ <MyComponent @description="" />
+```
```
ijlee2, line 72: This is due to a bug in `ember-template-recast` (possibly, `@glimmer/syntax`).

docs/rule/sort-invocations.md
```diff
@@ -0,0 +1,97 @@
...
+
+The order clearly shows how the component is customized more and more. Things are alphabetized within each group.
+
+```hbs
+<Ui::Button
+  @label="Submit form"
+  @type="submit"
+  data-test-button
+  {{on "click" this.doSomething}}
+  ...attributes
```
lifeart, line 38: splattributes position may be meaningful - https://guides.emberjs.com/v5.5.0/components/component-arguments-and-html-attributes/

image

ijlee2: Yes, changing the order of `...attributes` may break an app. This shouldn't occur often, though.

In general, I think it's good to recommend writing components in a way that (1) the order doesn't matter and (2) an end-developer's preference can take precedence over the component's.

I'll add the link to the Ember Guides as a warning, appreciate your feedback!

docs/rule/sort-invocations.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Limitations
+
+It's intended that there are no options for sorting. Alphabetical sort is the simplest for everyone to understand and to apply across different projects. It's also the easiest to maintain.
+
+To better meet your needs, consider creating a plugin for `ember-template-lint`.
+
+## Known issues
+
+1\. If you passed an empty string as an argument's value, it has been replaced with the placeholder text `<EMPTY STRING>`. Do a find-and-replace-all to restore the value.
```
lifeart, line 67: I'm suggesting to use `@decription={{""}}` as safe fallback to avoid `not-reading-documentation` error.

docs/rule/sort-invocations.md
```diff
@@ -0,0 +1,97 @@
...
+
+```diff
+<MyComponent
+  @isOpen={{this.isOpen}}
+-   @parentContainerId="{{@parentId}}"
++   @parentContainerId="#{{@parentId}}"
+/>
+```
+
+Better yet, use the `{{concat}}` helper instead (this helps codemods).
```
lifeart, line 86: I think we could manually do it, applying transform, because `concat` is builtin helper - https://api.emberjs.com/ember/3.2/classes/Ember.Templates.helpers/methods/get?anchor=concat

docs/rule/sort-invocations.md
ijlee2: Ah, interesting.

I checked that the syntax `{{""}}` will cause [`no-unnecessary-curly-strings`](https://github.com/ember-template-lint/ember-template-lint/blob/v7.5.0-ember-template-lint/docs/rule/no-unnecessary-curly-strings.md) to fail. I can update the code to insert `{{""}}` instead, but we'll need to rely on the autofix from `no-unnecessary-curly-strings` (i.e. we create a dependency between two rules).

docs/rule/sort-invocations.md
ijlee2: When writing my codemod, I found that I'd need to recreate the `ConcatStatement`. Because we can pass more complex expressions to a concatenated string, I don't think we can easily recreate it without increasing the maintenance cost.

Here, it's better to ask end-developers to refactor code.

docs/rule/sort-invocations.md
lifeart: @ijlee2 how about `{{concat ""}}` ?

docs/rule/sort-invocations.md
lifeart: We could have something like this - https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/no-curly-component-invocation.js#L60

lifeart: `Here, it's better to ask end-developers to refactor code.` - I agree, but, we should produce kinda "safe" code in fixes, in worst case, I would prefer to not sort empty attribute and print meaningful message (saying where it to place and that we can't automatically adjust it)

docs/rule/sort-invocations.md
ijlee2: Code that will run into this issue originally looks something like,

```hbs
<MyComponent @description="" />
```

So I think the better option was to let this rule change the syntax to 

```hbs
<MyComponent @description={{""}} />
```

then allow `no-unnecessary-curly-strings`' autofix to change the syntax back to

```hbs
<MyComponent @description="" />
```

If we add `{{concat ""}}`, this would change the code more drastically. Other developers, who didn't run `ember-template-lint` may be confused as to why the `{{concat}}` helper is used to add an empty string.

docs/rule/sort-invocations.md
ijlee2: Code that will run into this issue should occur rarely (it happened twice in a large production project, and both were caught thanks to tests). So I think this rule shouldn't handle this rare case.

A more maintainable approach would be to write a rule in `ember-template-lint` that encourages people to use the `{{concat}}` helper instead.

lifeart: wondering if we skip this element/component node at all, if issue is rare, it's better to just emit error, fix others and disable autofix only for this case

README.md
```diff
@@ -206,6 +206,7 @@ Each rule has emojis denoting:
 | [self-closing-void-elements](./docs/rule/self-closing-void-elements.md)                                   |     | 💅  |     | 🔧  |
 | [simple-modifiers](./docs/rule/simple-modifiers.md)                                                       | ✅  |     |     |     |
 | [simple-unless](./docs/rule/simple-unless.md)                                                             | ✅  |     |     | 🔧  |
+| [sort-invocations](./docs/rule/sort-invocations.md)                                                       |     | 💅  |     | 🔧  |
```
lifeart, line 209: this thing should be auto-updated (with npm run update)

docs/rule/sort-invocations.md
ijlee2: I came up with a fix in https://github.com/ijlee2/ember-codemod-sort-invocations/pull/13/files. I'll downstream it soon.

lib/rules/sort-invocations.js
```diff
@@ -226,18 +226,42 @@ export default class SortInvocations extends Rule {
...
         }
 
         if (!canSkipSortModifiers(modifiers)) {
-          node.modifiers = sortModifiers(modifiers);
+          if (this.mode === 'fix') {
+            node.modifiers = sortModifiers(modifiers);
+          } else {
+            this.log({
+              isFixable: true,
+              message: 'Modifiers must be sorted',
```
lifeart, line 261: let's print correct order of modifiers (by name) to improve DX, kinda - `sortModifiers([...modifiers]).map((m => m.path.original)).join(', ');

lib/rules/sort-invocations.js
```diff
@@ -248,7 +272,15 @@ export default class SortInvocations extends Rule {
           return;
         }
 
-        node.hash = sortHash(hash);
+        if (this.mode === 'fix') {
+          node.hash = sortHash(hash);
+        } else {
+          this.log({
+            isFixable: true,
+            message: 'Hash keys must be sorted',
```
lifeart, line 280: let's print correct order of keys (by name) to improve DX

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,305 @@
...
+}
+
+function getAttributePosition(node) {
+  const name = getAttributeName(node);
+
+  if (name.startsWith('@')) {
+    return 1;
+  }
+
+  if (name === '...attributes') {
```
lifeart, line 24: wondering if we need specific order for `data-test-` attributes

ijlee2: I don't think so. `data-test-` attributes, used for testing, are just a subset of [(data) attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes). We can let them be alphabetized as usual.

lib/rules/sort-invocations.js
ijlee2: Done.

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,315 @@
...
+function sortModifiers(modifiers) {
+  return modifiers.sort(compareModifiers).map((modifier) => {
+    const { hash, params, path } = modifier;
+
+    return b.elementModifier(path, params, hash);
+  });
+}
+
+export default class SortInvocations extends Rule {
+  visitor() {
```
lifeart, line 220: @ijlee2 do you mind adding jsdoc type here (reference - https://github.com/ember-template-lint/ember-template-lint/pull/3242)?

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,315 @@
...
+
+        const sorted = sortHash(hash);
+
+        if (this.mode === 'fix') {
+          node = b.block(path, params, sorted, b.blockItself());
+        } else {
+          this.log({
+            isFixable: true,
+            message: `Hash keys must be sorted: ${sorted.pairs.map(getHashPairName).join(', ')}`,
+            node,
```
lifeart, line 237: I think we may scary user with whole `node` in RED_UNDERLINE (in editor), I'm proposing to log fist incorrectly sorted node here.

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,315 @@
...
+        if (!canSkipSortAttributes(attributes)) {
+          const sorted = sortAttributes(attributes);
+
+          if (this.mode === 'fix') {
+            node.attributes = sorted;
+          } else {
+            this.log({
+              isFixable: true,
+              message: `Arguments and attributes must be sorted: ${sorted.map(getAttributeName).join(', ')}`,
+              node,
```
lifeart, line 254: I think we may scary user with whole `node` in RED_UNDERLINE (in editor), I'm proposing to log fist incorrectly sorted node here.

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,315 @@
...
+
+        const sorted = sortHash(hash);
+
+        if (this.mode === 'fix') {
+          node.hash = sorted;
+        } else {
+          this.log({
+            isFixable: true,
+            message: `Hash keys must be sorted: ${sorted.pairs.map(getHashPairName).join(', ')}`,
+            node,
```
lifeart, line 289: I think we may scary user with whole `node` in RED_UNDERLINE (in editor), I'm proposing to log fist incorrectly sorted hash-node here.

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,315 @@
...
+
+        const sorted = sortHash(hash);
+
+        if (this.mode === 'fix') {
+          node = b.sexpr(path, params, sorted);
+        } else {
+          this.log({
+            isFixable: true,
+            message: `Hash keys must be sorted: ${sorted.pairs.map(getHashPairName).join(', ')}`,
+            node,
```
lifeart, line 309: I think we may scary user with whole `node` in RED_UNDERLINE (in editor), I'm proposing to log fist incorrectly sorted node here.

docs/rule/sort-invocations.md
NullVoxPopuli: what if most of my work has attributes in the middle of a component?

(such as when I'm working on a design system, and require unoverridable things)

NullVoxPopuli: https://github.com/ember-template-lint/ember-template-lint/pull/3237/files#diff-582f08438c153065f141a545311dac9c0372a68c19db8f209f49ceb33295cba3R46

This feels unsatisfactory :-\

lifeart: I think we could remove `Limitations` section here, to allow contributors extend rule if needed (as now, by default), without explicitly mentioning opinion.

kategengler: I've also seen order of `...attributes` matter, especially with design systems. I believe this would break a significant number of apps.

docs/rule/sort-invocations.md
kategengler: The counter to that approach is https://github.com/import-js/eslint-plugin-import

lib/rules/sort-invocations.js
ijlee2: Done.

lib/rules/sort-invocations.js
ijlee2: Done.

docs/rule/sort-invocations.md
ijlee2: To address @NullVoxPopuli and @kategengler's concerns:

I don't think it's possible for a lint rule to know when exactly it's dangerous to place `...attributes` at the end. This would require analyzing other components, both within the app as well as from outside (addons), and possibly knowing the internal details of Ember. People can always come up with a way to make something that should be order-independent (e.g. [object keys](https://eslint.org/docs/latest/rules/sort-keys)) order-dependent, but this should be a rare occurrence.

I'd say that this rule's targeting [the 80% case](https://en.wikipedia.org/wiki/Pareto_principle) is good enough, since it's a stylistic rule and not a recommended one in `v7`. For the other 20%, the code is simple enough (when compared to that for [`attribute-order`](https://github.com/ember-template-lint/ember-template-lint/blob/v7.6.0-ember-template-lint/lib/rules/attribute-order.js)), so people have the option to write an `ember-template-lint` plugin or a codemod based on mine.

I ran the codemod on several productions apps that, in total, have almost a thousand components. I saw that hundreds of Percy snapshots passed without an issue. In other words, probabilistically speaking, it's all right to place `...attributes` at the end for many "randomly written" templates. I think it'd be good to publish this rule as is, so that people can try it out and report errors with actual examples. Would this be all right with you?

ijlee2: > we could have 2 options (safe config - by default), where we sort only GROUPS of attributes (and not attributes itself inside group), and "stylistic" option - to sort items inside group too

I think that's a possibility. I need to limit my time this week on implementing more features, though. If it's okay, can we implement the safe config later, if many people report a bug with the current code?

lifeart: @ijlee2 sure, we could land it as-is, the limitation only that we can't make rule `recommended` if there is no `safe` fix by default.

test/unit/rules/sort-invocations-test.js
```diff
@@ -0,0 +1,2027 @@
...
+                  />",
+            },
+            {
+              "column": 8,
+              "endColumn": 10,
+              "endLine": 16,
+              "filePath": "layout.hbs",
+              "isFixable": true,
+              "line": 2,
+              "message": "\`on\` must appear later",
```
lifeart, line 648: `must appear later` is not a valuable message, we should point after that it should appear (in my understanding)
Like: 
`"on" modifier should appear after data-test-name attribute` or
`{{on}} modifier should appear before "...attributes"` or
`"id" attribute should appear after "ab" attribute" or
`@name argument should appear after "id" attribute`

docs/rule/sort-invocations.md
ijlee2: @lifeart Understood, I appreciate your help and great suggestions!

kategengler: I think publishing the rule is absolutely fine. I think to make it recommended we would need something like lifeart suggested.

I also believe there are cases where modifier order can matter.

NullVoxPopuli: > where modifier order can matter.

confirm!

test/unit/rules/sort-invocations-test.js
ijlee2: I don't think we can easily implement such precise before and after's. This would require a more drastic change to the code that I wrote. The existing `sort-invocations` also [doesn't specify where exactly something should go](https://github.com/ember-template-lint/ember-template-lint/blob/v7.6.0-ember-template-lint/test/unit/rules/attribute-order-test.js#L106-L118).

`sort-invocations` is really meant to be followed using autofix, so I think the log message can remain brief as is. I'll will surround the modifier names with `{{}}` to more clearly indicate things that are modifiers.

ijlee2: Nvm. We can easily write the after. :)

lib/rules/sort-invocations.js
```diff
@@ -0,0 +1,345 @@
...
+        let index = getUnsortedAttributeIndex(attributes);
+
+        if (index !== -1) {
+          if (this.mode === 'fix') {
+            node.attributes = sortAttributes(attributes);
+            isSorted = true;
+          } else {
+            this.log({
+              isFixable: true,
+              message: `\`${getAttributeName(attributes[index])}\` must appear after \`${getAttributeName(attributes[index + 1])}\``,
```
ijlee2, line 273: Due to how `getUnsortedAttributeIndex()` works, `index + 1` is guaranteed to be at most `attributes.length - 1`.

lifeart (approved):

NullVoxPopuli (approved):

NullVoxPopuli:
this just spins forever: 
image


@ijlee2 want to rebase? maybe that'll fix it

lifeart:
@NullVoxPopuli I see changes already in master 
image

NullVoxPopuli:
🤔 
image

ijlee2:
Since the `master` branch has the new rule, maybe we can close this pull request instead of a rebase?

If I rebase my branch, then I think only @NullVoxPopuli's commit might be shown in this PR, i.e. we lose the step-by-step solution. I'm also unsure if GitHub would then note a merge conflict.

ijlee2:
@NullVoxPopuli `release-plan` failed to include this PR in the release notes and `CHANGELOG`. Can you manually add it to them?

Issue #3217, opened 2025-05-04
`no-this-in-template-only-components` is triggered in class backed gts components


johanrd:
### Steps to reproduce:
1. Have a class backed component in template tag format:

```glimmer-ts
import { on } from '@ember/modifier';
import Component from '@glimmer/component';

export default class Example extends Component {

  handleClick = (event: Event) => {
    console.log('Button clicked!', event);
  };

  <template>
    <button {{on 'click' this.handleClick}} type='button'>+</button>
  </template>
}
```

2. Enable `no-this-in-template-only-components` in `.template-lintrc.js` (as this is proposed as default true in https://github.com/ember-template-lint/ember-template-lint/issues/3210)

```diff
module.exports = {
  extends: ["recommended"],
  rules: {
+    'no-this-in-template-only-components': true
  },
}
```

3. Experience unexpected error when running `pnpm glint`:
```
app/components/a.gts
  11:25  error  Usage of 'this' in path 'this.handleClick' is not allowed in a template-only component. Use '@handleClick' if it is a named argument or create a component.js for this component.  no-this-in-template-only-components
```

### Expected behavior:
The rule `no-this-in-template-only-components` should not be triggered from class backed components

lifeart:
Hi @johanrd, thank yo for report, issue should be fixed here - https://github.com/ember-template-lint/ember-template-lint/pull/3240

PR #3240, opened 2025-05-05
fix: no-this-in-template-only-components should early exit in gjs/gts (skipping this lint)

lifeart:
Resolves: https://github.com/ember-template-lint/ember-template-lint/issues/3217

### DIFF
--- a/lib/rules/no-this-in-template-only-components.js
+++ b/lib/rules/no-this-in-template-only-components.js
@@ -10,6 +10,8 @@ function message(original, fixed) {
   );
 }
 
+const glimmerScriptExtensions = ['.gjs', '.gts'];
+
 /**
  * Tests whether the template has a corresponding component class file
  * @param hbsFilePath - path to the template file
@@ -17,6 +19,13 @@ function message(original, fixed) {
  * @returns {boolean} true if a component class was found
  */
 function isTemplateOnlyComponent(hbsFilePath, validComponentExtensions) {
+  if (
+    glimmerScriptExtensions.some((ext) => {
+      return hbsFilePath.endsWith(ext);
+    })
+  ) {
+    return false;
+  }
   /**
    * Tests whether the component class exists as `component.js`, `component.ts`, etc.
    */
--- a/test/unit/rules/no-this-in-template-only-components-test.js
+++ b/test/unit/rules/no-this-in-template-only-components-test.js
@@ -17,6 +17,24 @@ generateRuleTests({
       },
     },
     {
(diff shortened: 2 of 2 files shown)


test/unit/rules/no-this-in-template-only-components-test.js
```diff
@@ -16,6 +16,24 @@ generateRuleTests({
...
+      },
+    },
+    {
+      template: '{{this.data}}',
+      meta: {
+        filePath: 'addon/templates/components/foo-bar.gjs',
+      },
+    },
+    {
+      template: '{{this.bar}}',
```
NullVoxPopuli, line 32: these aren't gjs 🙈

lifeart: Ohh, @NullVoxPopuli you are right! Changes pushed

lib/rules/no-this-in-template-only-components.js
```diff
@@ -10,13 +10,22 @@ function message(original, fixed) {
...
 /**
  * Tests whether the template has a corresponding component class file
  * @param hbsFilePath - path to the template file
  * @param validComponentExtensions - an array of valid component class extensions
  * @returns {boolean} true if a component class was found
  */
 function isTemplateOnlyComponent(hbsFilePath, validComponentExtensions) {
+  if (
+    glimmerScriptExtensions.some((ext) => {
+      return hbsFilePath.endsWith(ext);
```
johanrd, line 24: Won't this exclude all `gts`/`gjs` files whether they are template only or not?

Maybe not a problem, as my experience is that Glint warns about `this` references in gts/gjs anyways?

lifeart: in this rule don't have proper context for template-only gts/gjs (we need to resolve it from babel), we could add this functionality later

NullVoxPopuli: wouldn't it make more sense for this lint to be disabled in gjs/gts, and then implemented instead in eslint-plugin-ember?

lifeart: @NullVoxPopuli yep, technically, we disabling rule (in this pr) if `filePath` is a gts/gjs.

Note, we have a problems with gjs/gts rules overrides, at the moment - so, it may be tricky to disable rule in config.
Folks may have `gts/gjs` and `.hbs` in same codebase.

NullVoxPopuli: ah! yes, the return false is clear now

PR #3291, opened 2025-06-09
Ensured that ...attributes is listed last

ijlee2:
Downstreams [ember-codemod-sort-invocations#28](https://github.com/ijlee2/ember-codemod-sort-invocations/pull/28) and [#29](https://github.com/ijlee2/ember-codemod-sort-invocations/pull/29).

The check for when we can skip listing `...attributes` last missed a couple of cases.

```hbs
{{! Skipped when it shouldn't be, because attributes and modifiers are sorted correctly }}
<iframe
  ...attributes
  {{did-insert this.doSomething1}}
  {{on "load" this.doSomething2}}
></iframe>

{{! Skipped when it shouldn't be, because attributes and modifiers are sorted correctly }}
<iframe
  {{did-insert this.doSomething1}}
  ...attributes
  {{on "load" this.doSomething2}}
></iframe>

{{! Skipped correctly }}
<iframe
  {{did-insert this.doSomething1}}
  {{on "load" this.doSomething2}}
  ...attributes
></iframe>
```

In addition, the calculations for the new locations of `...attributes` and modifiers didn't work always.

### DIFF
--- a/test/unit/rules/sort-invocations-test.js
+++ b/test/unit/rules/sort-invocations-test.js
@@ -567,7 +567,7 @@ generateRuleTests({
             this.styles
             "button"
             (unless this.enableSubmit "disabled")
-          }} data-cucumber-button="Submit form" data-test-button ...attributes {{autofocus}} {{on "click" @onSubmit}}
+          }} data-cucumber-button="Submit form" data-test-button {{autofocus}} {{on "click" @onSubmit}} ...attributes
         />
 
         <Ui::Button
@@ -575,7 +575,7 @@ generateRuleTests({
             this.styles
             "button"
             (unless this.enableSubmit "disabled")
-          }} data-cucumber-button="Submit form" data-test-button={{""}} ...attributes {{autofocus}} {{on "click" @onSubmit}}
+          }} data-cucumber-button="Submit form" data-test-button={{""}} {{autofocus}} {{on "click" @onSubmit}} ...attributes
         >
           Submit form
         </Ui::Button>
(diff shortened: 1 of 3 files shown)


test/unit/rules/sort-invocations-test.js
```diff
@@ -1746,6 +1746,152 @@ generateRuleTests({
         `);
       },
     },
```
NullVoxPopuli, line 1748: These should all be different tests to help with the debug output size when things go wrong

ijlee2:
Hmm, I see the code change as a bug fix and needing only a patch release. When the rule had been introduced in `v7.7.0`, the documentation explained that [`...attributes` will be listed last](https://github.com/ember-template-lint/ember-template-lint/blob/v7.7.0-ember-template-lint/docs/rule/sort-invocations.md#components). The current code fulfills this expectation in some cases, but not all.

NullVoxPopuli (approved):

test/unit/rules/sort-invocations-test.js
ijlee2: The formatting in this file is due to two things:

- I used the same groupings that I had used in [`ember-codemod-sort-invocations`](https://github.com/ijlee2/ember-codemod-sort-invocations/blob/1.0.5/tests/fixtures/my-app/input/app/components/example-15.hbs).
- `ember-template-lint` project doesn't have a history of separating a large file into multiple (e.g. utilities, test files) to help with readability, so I didn't want to introduce a new pattern in my pull request.

PR #3396, opened 2025-11-07
Exclude script files when `checkHbsTemplateLiterals` is set to `false`

bertdeblock:


bin/source-ember-template-lint.js
```diff
@@ -165,7 +165,6 @@ async function run() {
       config,
       rule: options.rule,
       allowInlineConfig: !options.noInlineConfig,
-      checkHbsTemplateLiterals: options.checkHbsTemplateLiterals,
```
bertdeblock, line 168: `options` here is an object with the CLI args, not the user's config.
The user's config is only loaded below via `await linter.getConfig();`.

test/unit/bin/get-files-to-lint-test.js
```diff
@@ -14,7 +14,6 @@ describe('getFilesToLint', function () {
   beforeEach(async function () {
     project = await setupProject();
     await project.chdir();
-    await project.write({ 'application.hbs': 'almost empty', 'other.hbs': 'ZOMG' });
```
bertdeblock, line 17: We already write these files with the same content below.

NullVoxPopuli (approved):

bertdeblock:
@NullVoxPopuli I forgot that it's still possible to define hbs`` templates in gjs/gts files, so in theory this PR isn't completely correct I assume?

NullVoxPopuli:
Sure, but technical possibility does not mean we should support it, ya? Esp since hbs`` is loose mode, and not a component - it's a silly paradigm

bertdeblock:
Yeah, I'm fine with this change, just wanted to point that out.

bertdeblock:
Should we temp pin `rolldown` (dev dep) to have CI pass again?

